PR #57, opened 2013-08-21
IPython SVG visualizer for Simulator

jaberg:
This PR is developing code to visualize the signal flow within the Simulator.  It isn't quite right yet, but it's a start.

The current visualization method is based on a grid whose columns are signals, and whose rows are time points within the simulator loop (e.g. "just-post-decoder calculations").

Now that I've seen it, I think I'd like to do it a bit differently, so that there are three signal rows corresponding instead to the three buffer sets created within the simulator.  In this view, there would be some lateral connections that would need curved arcs and arrowheads, but the picture would actually match the reference simulator algorithm.

There is *also* an even stronger need to develop a visualization technique for Models, in which nodes are ensembles (i.e. the things that you can "connect" together), and there is one arc per connection. I'm cautiously optimistic that such a model visualizer should be relatively easy to do once the Simulator one is done, because it seems simpler.

The reason this has anything to do with IPython is that these visualizers can be configured to be the default rendering mechanisms for Nengo objects, so that `display(model.sim_obj)` is this visualization.

TODO:
- [ ] Define a "marker" in the output SVG so that the lines have arrowheads on them
- [ ] Use three rows for the three internal buffers
- [ ] Use Bezier curves ('paths') to make it pretty
- [ ] Only draw nodes that are the endpoint of an arrow
- [ ] Use colour to denote constant vs. signal, remove that part of their name

### DIFF
--- a/nengo/__init__.py
+++ b/nengo/__init__.py
@@ -4,3 +4,19 @@ log.addHandler(logging.NullHandler())  # Prevent output if no handler set
 
 from .model import Model
 from .objects import LIF
+
+
+try:
+    import IPython.core.getipython
+    ipy = IPython.core.getipython.get_ipython()
+except ImportError:
+    ipy = None
+
+# -- register an automatic visualizer for IPython
+if ipy:
+    import simulator
+    import simulator_svg
+    simulator.Simulator._repr_svg_ = simulator_svg.simulator_to_svg
+    print 'registering handler for simulator.Simulator'
+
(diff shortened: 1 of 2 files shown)


tbekolay:
With `nengo_gui` now released I think this can be closed. The notion of visualizing something to do with signals and ops would definitely still be cool, but probably needs to start from scratch since lots has changed.

Issue #135, opened 2013-09-12
Have an option to force the simulator to run in real time


bjkomer:
For any real time control, the simulator should not run faster than real time. If for whatever reason it cannot run fast enough to be real time, it should try to be as close as possible.
Any thoughts/ideas on this? Discuss!

tbekolay:
The consensus from the meeting yesterday was that this is a useful thing, as evidenced by the fact that [Nengo GUI now does this](https://github.com/nengo/nengo_gui/pull/661). However, it's likely not something that we would want to put directly into the reference simulator. It could go in `nengo/utilities`, but to keep things simple I'm going to put the Nengo GUI implementation in `nengo_extras` for now, and if it sees a lot of use we can migrate it to this repo in the future.

hunse:
Done in https://github.com/nengo/nengo-extras/pull/85

PR #151, opened 2013-09-17
Fixed-Time Simulation

bjkomer:
Partial solution for issue #135 
Adding a flag to the simulator allowing the user to force the simulation to run in real-time

sim = model.simulator( fixed_time=True )

Currently works for when the simulator can run faster than real-time, but not yet for the simulator running slower than real-time

### DIFF
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -10,6 +10,7 @@ from collections import Mapping
 import logging
 
 import numpy as np
+import time
 
 import nengo.utils.numpy as npext
 from nengo.builder import Model
@@ -187,7 +188,7 @@ class Simulator(object):
 
         self._probe()
 
-    def run(self, time_in_seconds, progress_bar=True):
+    def run(self, time_in_seconds, progress_bar=True, real_time=False):
         """Simulate for the given length of time.
 
         Parameters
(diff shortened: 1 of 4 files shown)


drasmuss:
Reviving this one from the dead :skull:

This feels like a `nengo_extras` thing to me, now that we have that. Basically just a subclass of Simulator with a modified `run_steps` function, and then people could do `nengo_extras.RealtimeSimulator` if they want to use it.  The fact that this has been sitting around since 2013 I think tells us that it's a niche enough feature that I'd be reluctant to give it a special place in the core reference Simulator.

tbekolay:
I'm +1 on moving to `nengo_extras`! It's worth pointing out also that SpiNNaker and Nengo GUI also offer ways to run in real-time or pretty close to realtime, so it is probably desired, though yeah, not by people using the reference simulator. It would be nice to consolidate the Nengo GUI throttling with this implementation, but that's not critical.

@bjkomer do you want to move this to `nengo_extras` or do you mind if I do it?

drasmuss:
I migrated this to nengo-extras (https://github.com/nengo/nengo-extras/pull/85), so closing this one now.  The end of a 6 year journey!

PR #219, opened 2013-12-05
Probes on a diet

youssefzaky:
Made probes that can take a maximum length parameter based on Eric's changes in probes-awesome branch. This makes it fast to probe something when you just need the current value, or last n values, of the thing you're probing. Added tests to test_probe.py.


drasmuss:
Going to close this, since the discussion kind of ended on "not sure if this will be a common use case", and I think our experience since then has shown that other ways of doing this (e.g., using Nodes) have done the trick.  It's also probably been superseded by things like #654.

PR #372, opened 2014-06-02
RebuildNode

otrujill:
Added a rebuild_node() function to the simulator class, a test for it. rebuild_node() allows rebuilding of individual input nodes, specifying a new output value for them without rebuilding the whole model.

Usage is something like:
m = nengo.Network()
with m:
n = nengo.Node(output=x)
sim.run(...)
sim.rebuild_node(n, output=y)
sim.reset()
sim.run(...)

### DIFF
--- a/nengo/builder.py
+++ b/nengo/builder.py
@@ -1108,12 +1108,14 @@ def build_node(node, model, config):
     elif not is_callable(node.output):
         model.sig[node]['out'] = Signal(node.output, name=node.label)
     else:
-        sig_in, sig_out = build_pyfunc(fn=node.output,
-                                       t_in=True,
-                                       n_in=node.size_in,
-                                       n_out=node.size_out,
-                                       label="%s.pyfn" % node.label,
-                                       model=model)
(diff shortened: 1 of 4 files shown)


jgosmann:
This PR is over two years old and marked as "needs changes". It is unclear whether we actually want a rebuild node function (apparently no one needed it in two years) and it seems to be something that needs to be well thought out (are other backends supposed to support it too? What are potential cases where this fails? Should this go in hand with a reset of the simulation?).

I'd propose to close this PR.

hunse:
I'd be okay with closing it. To me, this has always seemed like something that might be nice to have, but also a bit of a hack. Ultimately, it would be nice to have a way that users can just make changes directly to a model, ask to rebuild it, and we'll only rebuild the parts that we have to (and re-use the rest). But this is not easy.

So at the time, this hack seemed like something that might be nice to have since Nodes are something that people might want to change while keeping the rest of the model the same. But you're right, if there hasn't been demand for it in two years, it's probably not high priority.

PR #391, opened 2014-06-06
Made common signals readonly

tbekolay:
I looked over all the signals we make to see if any made sense to be readonly. Right now, it seems like the only critical ones are the `'common'` signals, so this ensures that they are not writeable. I added `test_builder.test_commonsig_readonly` to test that they can't be modified.


hunse:
I rebased this. I propose that we merge it, since it is an improvement. We can revisit making other signals read-only at a later date if we wish.

PR #408, opened 2014-06-13
WIP: Fix for nengo.utils.functions.whitenoise

drasmuss:
Creating a node with `nengo.utils.functions.whitenoise` as the output function crashes the javaviz, because it passes time as a float rather than an nparray.  This fixes that, but javaviz now crashes after 2 timesteps, for some unknown reason.  No error message.  Still trying to figure out why; any javaviz people have some clever ideas (or know where the error message is disappearing to)?

### DIFF
--- a/nengo/utils/functions.py
+++ b/nengo/utils/functions.py
@@ -144,11 +144,11 @@ def whitenoise(step, high, rms=0.5, seed=None, dimensions=None):
     if dimensions is not None:
         signals = [whitenoise(
             step, high, rms=rms, seed=rng.randint(0x7ffffff))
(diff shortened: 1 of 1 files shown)


hunse:
Is this still a problem? The `whitenoise` function has been redone to be a process, so the javaviz problem may have been fixed.

drasmuss:
Just gave it a try, and it looks like it has (pretty much) resolved itself.  It doesn't crash, but the node output doesn't display properly (although it appears to be properly communicated to connected populations).  Given the state of flux javaviz is in though, I would probably just wait and see if that fixes itself as well :wink:

PR #429, opened 2014-06-23
WIP: Eventprobes

youssefzaky:
This is a first pass at implementing 'event probes', i.e. probes that record the time when certain things happen. I implemented it by having probes take in two other arguments: a function that returns a binary vector, and a parameter 'track_changes' that modifies how the times are recorded. Here's an example of how it works:

```python
def func(x):
      #1D
      return x[0] > 0

ensemble = nengo.Ensemble(neurons=10, dimension=2)

#will record all the times when the first dimension of the value represented is > 0
event_probe = Probe(ensemble, 'decoded_output', function=func)
  
#will only record the times when the first dimension of the value represented becomes > 0 
event_probe = Probe(ensemble, 'decoded_output', track_changes=True, function=func)
```

The function that you pass must return a binary vector, and the times are recorded independently for
each dimension of that vector.

```python
def func2(x):
       #2D for the following example
       return x > 0

#will record all the times for each dimension when they are > 0
event_probe = Probe(ensemble, 'decoded_output', function=func)
```
        
This can be used to record spike times too:

```python
event_probe = Probe(ensemble, 'spikes', function=func)
```

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -32,3 +32,4 @@ Please keep this list sorted alphabetically.
 - Travis DeWolf <travis.dewolf@gmail.com>
 - Trevor Bekolay <tbekolay@gmail.com>
 - Xuan Choo <xchoo.mainframe@gmail.com>
+- Youssef Zaky <youssefzaky@gmail.com>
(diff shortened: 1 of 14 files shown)


drasmuss:
Been a WIP without activity for 2 years, so going to mark it "revise and resubmit" and close for now.

Issue #414, opened 2014-06-17
Include Aaron's Semantic Pointer element distribution in utils.dist.


e2crawfo:
Useful for setting evaluation points.

jgosmann:
With #430 merged, is this still an issue?

tbekolay:
I think this can be closed, yeah.

PR #430, opened 2014-06-25
Add SqrtBeta probability distribution

jgosmann:
This addresses #414.

I uploaded the derivation of the probability distribution to the Google drive.

I am unsure whether to add the PDF/CDF. I think, those are useful, but we're not providing those functions for the other probability distributions.


hunse:
I think this is good to go. It's just adding a distribution, so it should be straightforward (i.e. it can't break anything).

PR #441, opened 2014-07-11
Implemented UDP socket utility function for Nodes

xchoo:
Initial working prototype of udp socket functions for nengo Nodes. 
- Sockets can be easily inserted into nengo models by creating the socket class with appropriate parameters, and then creating a nengo.Node and passing the socket class' run function as the output parameter.
- Socket class is designed to be not require any special tweaking on the simulator side, and also designed to close itself after a predefined period of inactivity (or by the user if desired).


drasmuss:
Rebased to master and added a commit that lets you specify multiple destination addresses/ports.

Seanny123:
@tcstewar does this need to create a Node subclass or can it be reviewed as-is?

xchoo:
The intention was to be able to use the UDP socket both within a node and as standalone code.

tcstewar:
> @tcstewar does this need to create a Node subclass or can it be reviewed as-is?

I don't think it needs to be a Node subclass, so should be fine to review as-is.

nengo/utils/tests/test_socket.py
```diff
@@ -0,0 +1,151 @@
...
+            self.lock.set()
+        self.sim.run(self.sim_time)
+
+
+def test_send_recv_chain(Simulator, plt, seed, rng):
+    # Model that sends data
+    udp_send = sockets.UDPSocket(dest_port=54321, send_dim=1)
+    m_send = nengo.Network(label='Send', seed=seed)
+    with m_send:
+        input = nengo.Node(output=lambda t: math.sin(10 * t))
```
Seanny123: This might be crazy, but for code simplicity, why not use `np.sin` and thus not need to `import math` in the first place?

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+        self.dest_addr = (dest_addr if isinstance(dest_addr, list)
+                          else [dest_addr])
+        self.dest_port = (dest_port if isinstance(dest_port, list)
+                          else [dest_port])
+        self.timeout = timeout
+        self.byte_order = '!'
+
+        self.last_t = 0.0
+        self.dt = 0.0                               # Local simulation dt
+        self.dt_remote = max(dt_remote, self.dt)    # dt btw each packet sent
```
Seanny123: Having `dt_remote != dt` is never tested.

nengo/utils/tests/test_socket.py
```diff
@@ -0,0 +1,151 @@
...
+class SimThread(threading.Thread):
+    def __init__(self, simulator, sim_time, lock=None, wait=False):
+        threading.Thread.__init__(self)
+        self.sim = simulator
+        self.sim_time = sim_time
+        self.lock = lock
+        self.wait = wait
+
+    def run(self):
+        if self.lock is not None and self.wait:
```
Seanny123: Why would `self.lock == None` anyways? If that's something you want to check for, shouldn't you check for it in `__init__()`?

nengo/utils/tests/test_socket.py
```diff
@@ -0,0 +1,151 @@
...
+from nengo.utils import sockets
+
+
+class SimThread(threading.Thread):
+    def __init__(self, simulator, sim_time, lock=None, wait=False):
+        threading.Thread.__init__(self)
+        self.sim = simulator
+        self.sim_time = sim_time
+        self.lock = lock
+        self.wait = wait
```
Seanny123: Why is `self.wait` necessary? Does it ever get changed by any outside process?

Seanny123: Ah nvm. I see where it's used now in the test.

nengo/utils/tests/test_socket.py
```diff
@@ -0,0 +1,151 @@
+import math
+import threading
+import numpy as np
+
+import nengo
+from nengo.utils import sockets
+
+
+class SimThread(threading.Thread):
```
Seanny123: I'm just generally confused about how this class is supposed to work in general. I can't match it to any of the threading patterns I've seen before. I understand that because of the GIL, the threads will switch eventually after sending a UDP message, since that's an IO process, but what's with all the locking? What shared state is there to protect?

xchoo: There is no shared state to protect. The threads are using the lock to wait for a threading event signal. This is used in the `test_send_recv_chain` test where it is necessary for the threads to be started in a certain order (the sender thread needs to start sending information before the receiving threads start running).

Seanny123: Comments explaining this in the tests would be greatly appreciated.

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+            # Time to send next packet if time between last packet and current
+            # t + half of dt is >= remote dt
+            if (t - self.last_packet_t + self.dt / 2.0 >= self.dt_remote):
+                for addr in self.dest_addr:
+                    for port in self.dest_port:
+                        self.send_socket.sendto(self.pack_packet(t * 1.0, x),
+                                                (addr, port))
+                self.last_packet_t = t * 1.0   # Copy t (which is an np.scalar)
+
+        if (self.is_receiver):
```
Seanny123: no parentheses needed

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+            found_item = False
+            if (not self.buffer.empty()):
```
Seanny123: No parentheses needed

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+                    data = self.buffer.get()
+                    self.value = data[1]
+                    found_item = True
+                elif (t_peek >= t + self.dt):
+                    # Timestamp of first item in buffer is > t+dt (i.e. all
+                    # items in the buffer are future packets). Assume packet
+                    # for current timestep has been lost.
+                    found_item = True
+
+            while (not found_item):
```
Seanny123: not parentheses needed

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+    def set_byte_order(self, byte_order):
+        if (byte_order.lower() == "little"):
+            self.byte_order = '<'
+        elif (byte_order.lower() == "big"):
+            self.byte_order = '>'
+        else:
+            self.byte_order = byte_order
+
+    def pack_packet(self, t, x):
+        # pack_packet takes a timestamp and data (x) and makes a socket packet
```
Seanny123: If you're going to make a multiline comment explaining how a function works, mark it with triple quotes.

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+        # Default packet data type: float
+        # Default packet structure: [t, x[0], x[1], x[2], ... , x[d]]
+        send_data = [float(t + self.dt_remote / 2.0)] + \
+                    [x[i] for i in range(self.send_dim)]
+        packet = struct.pack(self.byte_order + 'f' * (self.send_dim + 1),
+                             *send_data)
+        return packet
+
+    def unpack_packet(self, packet):
+        # unpack_packet takes a packet and extracts a timestamp and data (x)
```
Seanny123: Triple quotes again.

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+
+        # Empty the buffer
+        while (not self.buffer.empty()):
+            self.buffer.get()
+
+    def _open_socket(self):
+        # Close socket, terminate alive check thread
+        self.close()
+
+        if (self.is_sender):
```
Seanny123: No paren

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+            try:
+                self.send_socket = socket.socket(socket.AF_INET,
+                                                 socket.SOCK_DGRAM)
+                self.send_socket.bind((self.local_addr, 0))
+            except socket.error as error:
+                raise RuntimeError("UDPSocket: Error str: " + str(error))
+
+        if (self.is_receiver):
```
Seanny123: No paren

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+                               "simulation has been run before, wait for "
+                               "previous UDPSocket to release the port. "
+                               "(See 'max_idle_time' argument in class "
+                               "constructor, currently set to %f seconds)" %
+                               (self.local_addr, self.local_port,
+                                self.max_idle_time_current))
+
+    def _retry_connection(self):
+        self._close_recv_socket()
+        while (self.recv_socket is None):
```
Seanny123: No paren

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+        self.is_sender = False
+        self.is_receiver = False
+        self.ignore_timestamp = False
+
+        self.close()
+        self._initialize()
+
+    def config_send_only(self, dest_addr, dest_port):
+        self._config_wipe()
+        if (dest_port > 0):
```
Seanny123: No paren

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+            self.is_sender = True
+            self.dest_addr = dest_addr
+            self.dest_port = dest_port
+        else:
+            raise ValueError("UDPSocket: Invalid send only configuration."
+                             "Destination port should be > 0")
+
+    def config_recv_only(self, local_port, timeout=5, ignore_timestamp=False):
+        self._config_wipe()
+        if (local_port > 0):
```
Seanny123: No paren

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+        # Initialize socket if t > 0, and it has not been initialized
+        if t > 0 and ((self.recv_socket is None and self.is_receiver) or
+           (self.send_socket is None and self.is_sender)):
+            self._open_socket()
+
+        # Calculate dt
+        self.dt = t - self.last_t
+        self.dt_remote = max(self.dt_remote, self.dt)
+        self.last_t = t * 1.0
+        self.last_active = time.time()
```
Seanny123: No paren

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+
+        if (self.is_sender):
```
Seanny123: No paren

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+        if (local_port > 0):
+            self.is_receiver = True
+            self.local_port = local_port
+            self.timeout = timeout
+            self.ignore_timestamp = ignore_timestamp
+        else:
+            raise ValueError("UDPSocket: Invalid recv only configuration."
+                             "Local port should be > 0")
+
+    def config_send_recv(self, local_port, dest_addr, dest_port, timeout=5,
```
Seanny123: These functions are never used or tested. What's their intended use?

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+    def __del__(self):
+        self.close()
+
+    def _initialize(self):
+        self.value = [0.0] * self.recv_dim
+        self.last_t = 0.0
+        self.last_packet_t = 0.0
+
+        # Empty the buffer
+        while (not self.buffer.empty()):
```
Seanny123: No paren

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+        # Keep checking if the socket class is still being used.
+        while (time.time() - self.socket_class.last_active <
+               self.socket_class.max_idle_time_current):
+            time.sleep(self.socket_class.alive_thread_sleep_time)
+        # If the socket class is idle, terminate the sockets
+        self.socket_class._close_recv_socket()
+        self.socket_class._close_send_socket()
+
+
+class UDPSocket:
```
Seanny123: Documentation (in the form of docstrings) for the internal workings of this class (or at least the `run` function) would be appreciated.

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+        self.last_t = 0.0
+        self.last_packet_t = 0.0
+
+        # Empty the buffer
+        while (not self.buffer.empty()):
+            self.buffer.get()
+
+    def _open_socket(self):
+        # Close socket, terminate alive check thread
+        self.close()
```
Seanny123: Why do this at every time step? It seems kind of inefficient?

Seanny123:
I tried my best at reviewing in a helpful manner and will revisit once my style and documentation notes have been resolved. I understand that this may not be for a long while since @xchoo is busy with Spaun and other stuff.

nengo/utils/tests/test_socket.py
```diff
@@ -0,0 +1,151 @@
...
+                       atol=0.0001, rtol=0.0001)
+
+
+def test_time_sync(Simulator, plt, seed, rng):
+    udp1 = sockets.UDPSocket(dest_port=54322, local_port=54321,
+                             send_dim=1, recv_dim=2, timeout=1)
+    m1 = nengo.Network(label='One', seed=seed)
+    with m1:
+        input = nengo.Node(output=lambda t: math.sin(10 * t))
+        socket_node = nengo.Node(size_in=1, output=udp1.run)
```
hunse: I don't like having to pass `.run`. Can these be done as processes so that this isn't necessary? Or would that be confusing because with Processes it's typically fine to have more than one instance (and they're typically independent), whereas with sockets that's not necessarily the case?

nengo/utils/sockets.py
tbekolay: This should also be made into a new-style class by subclassing `object` (i.e., `class UDPSocket(object):`).

nengo/utils/sockets.py
```diff
@@ -0,0 +1,322 @@
...
+                self.last_packet_t = t * 1.0   # Copy t (which is an np.scalar)
+
+        if (self.is_receiver):
+            found_item = False
+            if (not self.buffer.empty()):
+                # There are items (packets with future timestamps) in the
+                # buffer. Therefore, check the buffer for appropriate
+                # information
+                t_peek = self.buffer.queue[0][0]
+                if (t_peek >= t and t_peek < t + self.dt) or \
```
hunse: Can this be a subroutine? Similar code is used below.

xchoo:
Hmmm. There are quite a few substantial changes to the code needed... I won't have time to work on this for months, anyone want to volunteer? :stuck_out_tongue:

Seanny123:
I can take it over. What was the original intent of this pull request? What
were you using it for?

xchoo:
The original intent? To implement a way to add UDP sockets to nengo. It was used to implement the communication between nengo 1.4 and nengo 2.0 for the biospaun stuff.

tcstewar:
> The original intent? To implement a way to add UDP socket communication to nengo. It was used to implement the communication between nengo 1.4 and nengo 2.0 for the biospaun stuff.

Do we need something like this in core Nengo?  Or would it work better as an example somewhere or how to do this?  I think there are lots of situations where people might want to use UDP to send data into and out of Nengo, but in most of those situations they'd want to be able to customize the data format, so they may as well write their own Node, customizing it from an example....

tbekolay:
Might I suggest https://github.com/nengo/nengo_extras?

xchoo:
Not really, but this code standardizes the UDP communication protocol such that you can use it to synchronize multiple simulations together (i.e. time-lock). Also, it standardizes the protocol for sending vector information, so anything that communicates to / from nengo using vector values can use this code.

xchoo:
And yeah, putting it in nengo_extras would be a good thing to do (this code was written way before that was ever created).

Seanny123:
If this is being moved to Nengo extras, should I polish this pull request
up here and merge it there once it's done?

tbekolay:
> should I polish this pull request up here and merge it there once it's done?

I think for the sake of history it's best to polish it up and review here... when it's looking good I'll port the history over to nengo_extras.

nengo/utils/tests/test_socket.py
Seanny123: My opinion is that this is a special enough case that a process shouldn't be used. Another option that Terry was throwing around was making the UDP socket thing a subclass of a Node. What do you think about that option?

hunse: > My opinion is that this is a special enough case that a process shouldn't be used

What do you mean by that? If there are complexities that make this not fit into the Process interface, then I understand that. But I don't understand why you say *shouldn't* be used.

Making a Node subclass might work, but it would be the first time we subclass one of our core NengoObjects, so I'd want to make sure everything works fine.

nengo/utils/tests/test_socket.py
studywolf: possibly the lock might be handling some weird race conditions on startup? if i'm reading it right it's only ever catching here once, then the lock is set and never cleared

nengo/utils/tests/test_socket.py
studywolf: similarly, i think it's being just used to handle some startup issues, making sure that all three threads in the `test_send_recv_chain` are ready to go before any start simulating

nengo/utils/tests/test_socket.py
studywolf: is this ensuring that the threads are started in a certain order, or just that they're all ready to start simulating at the same time?

nengo/utils/sockets.py
studywolf: this is only called once as threads are instantiated

nengo/utils/sockets.py
studywolf: a very good question! took them out and everything still passes, seems fair to assume it's legacy code from some previous iteration...

studywolf:
OK, went through and added some more comments, removed some functions that weren't being used, and made a couple changes for bugs that popped up when I was using over an ethernet connection (instead of local threads)... should be ready for another review...?

xchoo:
Oopa. Just realized the option to set the `ignore_timestamp` flag was removed with the removal of the `config` functions. That should be put back into the class constructor.

studywolf:
@tbekolay how do we go about the move over to extras?

tbekolay:
Hrm... I'm not sure what I was talking about a while back about keeping the history, as probably the history is not critical to keep. So I would probably do it manually, copy over the new files from this PR and put them somewhere appropriate in `nengo_extras`... in the commit message in `nengo_extras`, make sure you note that the PR is co-authored by Xuan.

studywolf:
aight, i moved this over to https://github.com/nengo/nengo_extras/pull/53 ! so closing the PR here!

PR #453, opened 2014-07-17
WIP: Hierarchical builder

tbekolay:
Still a work in progress, but this is an attempt to make the builder even *more* pluggable by making it sort of hierarchical. So, for example, if you're building a learning rule, and it does filtering, you should be able to use the same build function that builds synapses in connections or probes.

Here's what that looks like now, as a sneak preview of the exciting code within

```python
# In builder/synapses.py
@Builder.register(Connection, Lowpass)
@Builder.register(Probe, Lowpass)
def build_lowpass(model, obj, synapse):
    ...

# In builder/connections.py
@Builder.register(Connection)
def build_connection(model, conn):
    ...
    model.build(conn, conn.synapse)
    ....
```

This much is done and working, but I'm hoping there are more places where doing something like this will help. But, I wanted to get this out now though to see what people think of it before doing more changes / refactoring.

Note that this branch is based on the `frontend-refactor` branch, so the first two commits are from #444.

Other changes:

- Split `builder.py` into a subdirectory with a separate file for each build function.
- `Builder.register_builder` is now a decorator (see deea231 and 993c354)
- Split up `NetworkMember` and `NengoObject`; the intention here is to make working with `ObjView`s and `Neurons` more like other Nengo objects, but I haven't made much progress on this.


hunse:
This looks a lot more chaotic than it is, since it involves moving all the builder files, which we've already done.

drasmuss:
Marking this "revise and resubmit", since it's been inactive for a few years (and I think has been superseded/merged into other PRs).

hunse:
I thought that this was on the list for 2.2.0? @tbekolay, have you been working on this? Or is it given up?

tbekolay:
I haven't touched this since the last commit, though now that I'm looking through this again it does look useful. But I think it makes sense to revise this and make another PR as lots has changed.

PR #475, opened 2014-09-09
WIP: Added Gauss' method for convolution.

hunse:
This branch implements Gauss' method for complex number multiplication in the circular convolution network, which allows complex multiplication to be done with 3 multiplies instead of four, thus reducing the number of products needed for circular convolution by 25%.

I still need to document in code exactly how the method works (beyond just pointing people [here](http://en.wikipedia.org/wiki/Multiplication_algorithm#Gauss.27s_complex_multiplication_algorithm)), so I've marked it WIP.

### DIFF
--- a/nengo/networks/circularconvolution.py
+++ b/nengo/networks/circularconvolution.py
@@ -120,22 +120,25 @@ class CircularConvolution(nengo.Network):
     def transform_out(self):
         dims = self.dimensions
         dims2 = (dims // 2 + 1)
-        tr = np.zeros((dims2, 4, dims))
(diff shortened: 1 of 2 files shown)


drasmuss:
This still seems like a pretty cool idea.  @hunse did you ever get a chance to quantify the accuracy (mentioned a WIP tech report)?

hunse:
I could never get a noticeable increase in accuracy, so I kind of gave up on this.

drasmuss:
Sounds good.  I'll mark it revise and resubmit just in case anyone wants to come back and look at this in the future.

Issue #496, opened 2014-10-10
Network graph in and out of IPython notebook


tbekolay:
Make an SVG file

Hook it up to notebook, if in notebook.

tbekolay:
Done in https://github.com/nengo/nengo_gui!

PR #524, opened 2014-10-29
Network graph visualization

jgosmann:
I'm opening this PR for feedback and discussion; there is still quite some work to do. Help is welcome.

This PR addresses #496 (network graph visualization). It is based on the [nengo_gui](https://github.com/ctn-waterloo/nengo_gui) code and I pulled that code with its history into this branch. Thus, the large number of commits in this PR (I recommend just looking at files changed). At a later stage I will filter all of that history so that only relevant commits will be kept.

To tidy up the representation of different kind of data I settled on the following general structure for network graph visualizations:

1. We start with a `model` we want to visualize.
2. The `model` gets converted to a different graph representation `ModelGraph` which allows for easy manipulations. It provides properties to navigate incoming and outgoing edges as well as parents, ancestors, children, and descendants. The `ModelGraph` is intended to store only structural information, but no information about where to place certain vertices in the visualization or styles like color. Before visualizing the graph it might be manipulated by transforms (`transform.py`), e.g. to prune empty networks (`spa.Cortical` seems to produce such empty networks) or to collapse networks. The later is also important for performance reasons as large ensemble arrays can make the visualization in nengo_gui unusable.
3. Where graph vertices get positioned is specified with the config system. The positions can be set automatically with layout algorithms. Currently, I only ported the `feedforward_layout` algorithm. It seems that nengo_gui is using some different algorithm per default. It might be worthwhile to write a layout algorithm using [graphviz](http://www.graphviz.org/).
4. Given the `ModelGraph` and a `config` the graph can be converted to the final visualization by a `Renderer`. Different `Renderer`s for different output formats can be implemented. Currently, there is only the `D3DataRenderer` for producing input data for an interactive IPython notebook visualization with D3.js because it allowed to use a great deal of existing code and the `feedforward_layout` is not good enough to be usable without manual adjustments. But it should be pretty straight forward to implement rendering directly to SVG (without the loop through the notebook) or to the graphviz input format.

Improvements
===========

* Previously, two different graph representations were used for `feedforward_layout` and the rendering of the D3 data. This has been unified to one graph.
* With the new structure different aspects (conversion to the final output format, layouting, etc) are not as intermingled before.
* The new structure allows unit testing of some parts

Further changes
=============

* Wrapped the `main.js` code into a class to prevent pollution of the global namespace and to allow extensions (like in nengo_gui).
* The top level network was previously not shown, now it is (i.e. all elements are wrapped in a gray box). In theory `ModelGraph` supports multiple top level networks, but I am not sure whether there is a use case for that. Going back to the old behavior should be possible and even now the first network added will be saved as `ModelGraph.top` and is a bit special in so far.

Relationship with nengo_gui
======================

To avoid code duplication this PR is accompanied by PR ctn-waterloo/nengo_gui#140 in nengo_gui. It removes the code used in Nengo from nengo_gui and relies on the code provided by Nengo. Right now this leads to a level of coupling between Nengo and nengo_gui, I am not comfortable with (a change in the Nengo implementation is likely to break nengo_gui). With further refactoring Nengo should ideally provide a clearly specified, public interface for nengo_gui (and maybe other visualization programs).

Points for discussion
================

At the moment the config for positioning graph vertices acts on the usual nengo objects. I am not sure whether that is the right approach. It makes it obvious in nengo_gui which options correspond to which element in the code block that gets appended by nengo_gui. But there is not necessarily a 1:1 relationship between nengo objects and the modified `ModelGraph`. The elements of the later are what actually gets displayed so that it might be better to use those for the config.

Partial to do list
============

* [ ] Refactor JavaScript code. I think, it's in a horrible state right now. I even made it a bit more horrible with wrapping it into a class as I do not fully understand (at the moment) what the right way to implement inheritance is in JS. Also, `this` behaves quite strangely in JS. Despite that I believe that an object oriented approach is the way to go.
* [ ] Refactor `feedforward_layout.py`
* [ ] More and better layout algorithms
* [ ] Ensure that multiple visualizations in one notebook work (at the moment this is probably not the case)
* [ ] Direct rendering to SVG
* [ ] Share SVG code for direct rendering and interactive visualization
* [ ] Allow to save SVG from interactive visualization. I already implemented a prove of concept for this, but it's missing the CSS styles and is not really intuitive.
* [ ] Allow to update config from interactive visualization. (This might get a little bit hackish as long as no widgets will be used, but those require at least IPython 2.)
* [ ] More tests, a few places are marked with TODO comments where tests are missing

This is all I can think of right now. I might have forgotten a few points. Feedback is welcome.


jgosmann:
As nengo/nengo_gui#221 is now merged, I propose to close this PR. The nengo_gui implementation provides essentially the same functionality less the functionality to save an SVG (which probably will be introduce at some point?). I think, the approach taken in nengo/nengo_gui#221 is the right one because it has essentially no code duplication and give access to the complete nengo_gui functionality within the notebook.

If there are no objections, I will close the PR and clean out the nengo_gui_2014 history from the nengo repository (which I imported to work on this PR). I would propose to keep the cleaned-up branch for reference (it contains some code which might be useful later on like dynamically generating a download link for the rendered SVG).

tcstewar:
> If there are no objections, I will close the PR and clean out the nengo_gui_2014 history from the nengo repository (which I imported to work on this PR). I would propose to keep the cleaned-up branch for reference (it contains some code which might be useful later on like dynamically generating a download link for the rendered SVG).

Sounds good to me.  :)

tbekolay:
Sounds good to me! Feel free to close this issue once you push the cleaned up branch.

PR #529, opened 2014-11-06
Explain the config system

tbekolay:
There's a lot of confusion about the `config` system, so this PR is an attempt to explain it through an example notebook.

It also contains a few changes to `config.py` to make things more clear and/or correct. Once we're agreed upon some of the details of the code, I'll split those changes into a separate commit.

For the time being, please check out [the rendered notebook](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/6182929/config.ipynb) and give me some feedback on what makes sense and what's confusing.

Additionally, this PR should figure out the appropriate `__str__` methods for `Config`, `ClassParams`, and `InstanceParams`. You can see in the notebook how I envision these being used, but I think there's also an argument to be made for what @hunse implemented (see `config.py`, line 80 and 179) and making what I printed out a different attribute of these objects (e.g., `config.all` or something).


tbekolay:
OK! Went over this again, implemented the requested str/reprs for the config stuff, and cleaned up the history; should be good to merge, I think. [Rendered notebook is here](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/6182929/config.ipynb).

hunse:
The only thing that's still bugging me is that even if you print a config object in context, it's defaults do not take context into account. So the only way to actually check what something will be set to is to call `nengo.Config.default` as you do in the notebook, and that has to be done separately for each parameter you want to inquire about. It would be nice to be able to print the current "state" in a particular context, to see what everything will actually be set to if an object is created right then and there.

tbekolay:
Sure, can do.

tbekolay:
OK, made the change, see b28d1fc09d79fa. If it looks good, LMK and I'll do the squashing (want to split the last commit up so the `config.py` changes are squashed in 38bf55c.

tbekolay:
OK, changes implemented! More deletions than additions in 4810f6ff595ee6a2a5e118b8a4947ae62b1dcf71, which is a good sign; I think the [notebook](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/6182929/config.ipynb) reads better too.

hunse:
A test is failing. Once that is fixed up I think this is good to merge.

PR #535, opened 2014-11-10
Explain network design

tbekolay:
This is based on the `explain-config` branch, which still needs significant cleanup.

For now, I'm just pushing this to get some feedback on the notebooks. In separating out the network stuff from the config system notebook, I figured I'd go whole hog and write some network design tutorial notebooks. See the rendered [basic](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/6182929/network_design.ipynb) and [advanced](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/6182929/network_design_advanced.ipynb) notebooks.

I'm not sure about the advanced stuff; it's super useful for huge, interwoven networks, but it's hard to justify the added complexity for networks small enough that we can use them in an example.


tbekolay:
OK, I rebased this on #529 (which should be merged first), pulled out a `network.py` change into a separate commit, and made sure the notebooks still look alright. I think they're good enough! Can always revisit later, but I think for now good to merge!

PR #552, opened 2014-11-18
Utils for doing function space stuff in nengo

youssefzaky:


youssefzaky:
Finally got around to making this network!

tcstewar:
Hey everyone, I finally got around to taking a look at this PR.  Thank you for putting it together, @youssefzaky !  

I think there's actually two separate questions here.  First, how can we make a useful helper API for people doing function representation, and second, how can we wrap that in a nice Network interface.  This maps nicely to the two separate files in this PR.  But for now I'm just going to focus on the first half, since getting that straight will inform the second.

For the core ```nengo.utils.function_space``` helper API, I think there's a bit of reorganizing needed.  Basically, right now you have to specify the same information multiple times in multiple places.  For example, here's the current code for encoding a nice gaussian bump:

```python
import nengo.utils.function_space as fs

def gaussian(points, mean, sd):
    return np.exp(-(points-mean)**2/(2*sd**2))

n_neurons = 100
domain = fs.uniform_cube(1, d=0.01)
n_eval_points = 2000

fns = fs.gen_funcs(gaussian, 
                   fs.gen_args(n_neurons, nengo.dists.Uniform(-1,1), nengo.dists.Uniform(0.1,0.7)),
                   domain)
svd_fs = fs.SVD_FS(fns, d=0.01, domain_dim=1, n_basis=20)

eval_points = fs.sample_eval_points(gaussian, svd_fs, n_eval_points, 4, domain, 
                                    nengo.dists.Uniform(-1,1), nengo.dists.Uniform(0.1, 0.2))

input_func = fs.sample_comb(gaussian, 4, domain, nengo.dists.Uniform(-1, 1), nengo.dists.Uniform(0.1, 0.1))
signal_coeffs = svd_fs.project(input_func)

model = nengo.Network()
with model:
    ens = nengo.Ensemble(n_neurons=n_neurons, dimensions=svd_fs.n_basis,
                         encoders = svd_fs.encoders(), radius=np.linalg.norm(signal_coeffs),
                         eval_points=eval_points)
    stim = nengo.Node(signal_coeffs)
    nengo.Connection(stim, ens)
```

So, for example, the ```domain``` parameter needs to get passed in to three different places, the ```d``` parameter shows up twice, and even ```n_neurons``` is used twice.  This seems mostly, to me, to be due to the fact that ```fs.sample_eval_points``` and ```fs.sample_comb``` are module-level functions, rather than methods off of the Func_Space object (i.e. if they were member functions like that, then they wouldn't need that extra information passed in to them).

I'm also a bit worried that the current API makes some pretty strong assumptions about how the function space stuff is being used.  For example, the ```encoders()``` function always uses the functions that were used to define the space.  I've often used the eval_points values to define the space, and used a different distribution to define the encoders.  

Here's a first attempt at a slightly different approach to the API that I think is a bit more general and a bit more succinct:

```python
def gaussian(points, mean, sd):
    return np.exp(-(points-mean)**2/(2*sd**2))

fs = FunctionSpace(gaussian, pts=np.linspace(-1, 1, 200), 
                   n_samples=100, n_basis=20,
                   mean=nengo.dists.Uniform(-1, 1), 
                   sd=nengo.dists.Uniform(0.1, 0.7))

model = nengo.Network()
with model:
    ens = nengo.Ensemble(n_neurons=100, dimensions=fs.n_basis)
    fs.set_encoders(ens,
                       mean=nengo.dists.Uniform(-1, 1), 
                       sd=nengo.dists.Uniform(0.1, 0.7))
    fs.set_eval_points(ens, n_eval_points=2000,
                       mean=nengo.dists.Uniform(-1, 1),
                       sd=nengo.dists.Uniform(0.1, 0.2))
    
    stim = fs.sample_fn(n_samples=1,
                    mean=nengo.dists.Uniform(-1, 1), 
                    sd=0.1, 
                    superimpose=4)

    stimulus = nengo.Node(fs.encode(stim)[0])
    nengo.Connection(stimulus, ens)
```

My attempt here is just to encapsulate things a bit into the FunctionSpace object, and to provide the sort of functionality that I've used in my models.  I also made a slight change to how the random distribution parameters are specified (keyword arguments, rather than a list), although I still want to think on a better syntax for that.  Notice also that the encoders and the eval points don't have to be chosen from the same distribution that defines the function space.

Thoughts on this?  I can't think of anything that would be harder to do with this alternate syntax, but a fair number of things that are a bit shorter and more readable (to me, in any case), but there's probably a lot of use cases I'm not thinking of.

tcstewar:
Thinking about this more, here's a quick list of functions that I think need to be on the FunctionSpace object:

 - ```sample```  (takes a function and distributions and returns an array of values at each point)
 - ```project``` (takes an array of values at each point and projects it onto the basis)
 - ```reconstruct``` (takes the projected/represented value and expands it back out to an array of values at each point)
 - ```project_fn``` (helper function for ```sample``` followed by ```project```
 - ```set_encoders``` (takes a function, its distribution, and an ensemble, and sets the ensemble's encoders accordingly by doing ```project_fn```)
 - ```set_eval_points``` (takes a function, its distribution, and an ensemble, and sets the ensemble's eval_points accordingly by doing ```project_fn```)

I'm not happy with the names of those...   Part of me would prefer ```encode``` and ```decode``` rather than ```project``` and ```reconstruct```, but that's some horrible overloading of names.  I'm also not at all sure about the ```project``` and ```project_fn``` pair....  I think it's pretty common to want to both sample and project, but maybe not common enough to have a helper function for it with such a confusing name.

tcstewar:
Okay, this is pretty awesome.  I just took a stab at making a quick nengo gui display node using the function representation stuff.

image

You can use the sliders to adjust the input, and the black line is the decoded representation and the blue line is the ideal projected representation.   And the complete code for the gui stuff is there on the screen.  I'll roll that in as another helper function in FunctionSpace, I think....  :)

tcstewar:
Here's an example showing the proposed alternate syntax in action

image

youssefzaky:
I think this API is definitely is more straightforward and avoids the duplication you pointed out. When I first wrote the class I actually had it organized that way, with more functions being encapsulated in the class. I changed it when I came across the case where I had made functions my own way without sampling from distributions, and then wanted to use them. So I just made the class take these functions and moved methods out as helpers, instead of trying to make the class handle all the possible ways of defining functions.

I'm not sure how common this use case is though, so maybe it's not worth the added complexity.

tcstewar:
> I changed it when I came across the case where I had made functions my own way without sampling from distributions, and then wanted to use them. So I just made the class take these functions and moved methods out as helpers, instead of trying to make the class handle all the possible ways of defining functions.
> 
> I'm not sure how common this use case is though, so maybe it's not worth the added complexity.

Could you give an example of your alternate use case?  I'd intended the version here to also support people defining their own functions and mixing and matching between using functions and just large sets of data, but I'm not sure that I've got in mind the same use cases you're thinking of....

studywolf:
Hey @tcstewar I was just going to play around with this syntax and all that for a thing i'm working on, is function-space3 the most recent version of things?

tcstewar:
> Hey @tcstewar I was just going to play around with this syntax and all that for a thing i'm working on, is function-space3 the most recent version of things?

Yup, and there's also a function-space-3 branch in nengo_gui that has a bunch of examples of using it.

tcstewar:
(note: I'll try to get something written up about the function-space-3 syntax for the meeting on Tuesday)

studywolf:
Just worked through the notebook, and very excited this is super cool! One thought I had was that at the moment the basis functions are autogenerated based on the kinds of functions that I want to model, but it would also be nice to be able to instead just say "OK I want my basis functions to just be a set of Gaussians across state space". What I'm thinking of basically is for the next step where we add a method for learning functions in function space.

tcstewar:
Here's the notebook @studywolf was referring to:

https://github.com/tcstewar/testing_notebooks/blob/master/Function%20Space%20description.ipynb

>  but it would also be nice to be able to instead just say "OK I want my basis functions to just be a set of Gaussians across state space". What I'm thinking of basically is for the next step where we add a method for learning functions in function space.

The problem with that is that's not a valid set of basis functions.  Basis functions have to be orthonormal.  

As for learning functions in the function space, I'm not quite sure what you mean.  As far as I can tell, both types of learning that we currently do should work fine with the system as it is (both types being decoder learning and encoder learning).  Neither of those need to directly interact with the compressed function space, and neither should want to change the compressed function space.

studywolf:
Hmmm. OK maybe it's just that there are a few distinct use cases? 

I would like to have a set of Gaussians that I use the same way I use neural tuning curves (which are not orthonormal), where the weights are represented in the ensemble vector space. And then I would like to have a learning rule over the weights on each of those Gaussians (denoted w) that's like ```delta_w = -alpha_w * error * activities_basisfunctions```, and then modify the decoders for the population with ```delta_d = -alpha_d * delta_w * activities_neurons```. 

There are a bunch of reasons why it would be useful to have this, one of them is that in the case of representing the forcing function in DMPs, where the function represented in function space is the forcing function, you can then easily tell similar trajectories by taking the dot product of the vector of weights over the Gaussians for each of the movements. Which is kind of cool. 

Does that make sense? I think that i'm understanding the function space stuff, but priors tend to suggest otherwise.

tcstewar:
> Does that make sense? I think that i'm understanding the function space stuff, but priors tend to suggest otherwise.

Hmm...   I'm not quite sure I'm understanding, but it sounds to me like what you're talking about is a learning rule on the encoders (delta_w = -alpha_w * error * activities_neurons) and a learning rule on the decoders (a_d = -alpha_d * delta_w * activities_neurons).  I think that's what you mean by "activities_basisfunctions" -- the activity of a neuron that's sensitive to some particular gaussian.  That's what's considered an encoder in this approach.  There's no such thing as the activity of a basis function.  

But if you want "a set of Gaussians that I use the same way I use neural tuning curves", that's exactly what we're doing here by saying here's a bunch of gaussians, use them for the encoders.  The only thing the "basis functions" are doing is being a more efficient way of specifying those gaussians (instead of specifying them with a big giant long vector giving the value of the gaussian at many points).  But the math's all still the same.

tcstewar:
(to hopefully clarify what I said above, I'm pretty sure that yes, you can do exactly what you're wanting to do with the system as it stands, but it doesn't involve changing the basis functions at all.)

studywolf:
I rebased the nengo and nengo_gui function space branches to master in function-space-4 branch, in case anyone's looking, there's also a learning example and a couple other things i'm playing around with in the nengo_gui examples list.

studywolf:
was talking with @tcstewar about plotting 2D functions and got a simple heat map going, it just converts a matrix elements into a values from white to black and sends it over to nengo_gui as an svg image. up on the function-space-4 branch of nengo and nengo_gui, in nengo_gui/examples/function_space/08-2D_function_example. still lost to work out but kind of neat!

tcstewar:
@studywolf did you check in that example?  it doesn't seem to be there for me....

studywolf:
whoops! there now!

tcstewar:
That is *awesome*!   Nicely done!

studywolf:
:smile: function space stuff is gonna be siiiicccckkk, i'm really enjoying playing with it all!

jgosmann:
Can someone bring me up to date on the state of this PR? It is marked as "needs changes", but also as "needs review". What are the required changes? Are this minor enough that one can review this anyways?

studywolf:
Ah yeah Terry rewrote all these, function space 4 is the newest i believe
... it needs a fair bit of love still. I can talk with Trevor about trying
to bring it up to speed?

jgosmann:
It's not urgent (for me), just trying to figure out what can and needs to be reviewed. I'll remove the "needs review" label until it's actually ready for that.

studywolf:
This PR can be closed, things have been reworked in the function-space-4 branch. 
Working on that now, i'll make a PR from there once it's a bit closer to ready!

Issue #220, opened 2013-12-06
alternate syntax for connect?


jaberg:
Quick chat with someone just now reminded me that it's not apparent from the word "connect" that we are numerically _adding_ additional current to an ensemble's input. I remember being confused about that myself when first learning Nengo.

I think in both cases different syntax could have avoided the misunderstanding.

In the spirit of maintaining a distinction between ensembles (representing vectors) and neurons, we might want to allow the vector-implying operators to connect ensembles, since these are meant to mean vector manipulations.

E.g.

``` python
A = Ensemble(dimensions=2)
B = Ensemble(dimensions=2)
C = Ensemble(dimensions=2) 
B.increment_by(A.decode(func=np.sin)) # -- returns connection object
B.increment_by(C.decode(func=np.cos))
B.increment_by(B.decode())
```

Alternatively, if you don't want to store the connection in a local variable, we could abuse the "+=" operator (which sort of looks like "<-" if you use your imagination...):

``` python
B.input += A.decode(func=np.sin)
B.input += C.decode(func=np.cos)
B.input += B.decode()     # or B.input += B ?
# this brings us back to wanting to access the connections by e.g. B.inputs
```

I know we've talked about the API a lot, but FWIW this is not technically inconsistent with what we have. It's just an alternate syntax for connecting stuff to ensembles.  Something like this would have made it clearer for me when I was learning.

PR #553, opened 2014-11-19
Using generic python function to specify learning rule

drasmuss:
The idea is that this allows someone to specify an arbitrary Python function for the learning rule.  You need to define a function that accepts the previous decoders and a "data" vector, and outputs a change in the decoders.  The data vector is the output of some node; the idea is that users connect up whatever inputs they want to a passthrough node (although it doesn't have to be a passthrough, could be something more complex), and then we pass the output of that node to the learning rule.

See `test_learning_rule.py` for an example of implementing the PES rule using this syntax.

### DIFF
--- a/nengo/learning_rules.py
+++ b/nengo/learning_rules.py
@@ -1,9 +1,11 @@
 import warnings
 import weakref
 
+from nengo.base import ObjView
 from nengo.ensemble import Neurons
 from nengo.exceptions import ValidationError
-from nengo.params import FrozenObject, NumberParam, Parameter, IntParam
+from nengo.params import (FrozenObject, NumberParam, Parameter, IntParam,
+                          BoolParam)
 from nengo.utils.compat import is_iterable, itervalues
 
 
@@ -29,6 +31,9 @@ class LearningRule(object):
             self.size_in = (self.connection.post_obj.ensemble.size_in
                             if isinstance(self.connection.post_obj, Neurons)
                             else self.connection.size_out)
+
+    def __getitem__(self, key):
+        return ObjView(self, key)
 
     def __repr__(self):
         return "<LearningRule at 0x%x modifying %r with type %r>" % (
@@ -313,6 +318,43 @@ class Voja(LearningRuleType):
     def __init__(self, post_tau=0.005, learning_rate=1e-2):
         self.post_tau = post_tau
         super(Voja, self).__init__(learning_rate, size_in=1)
+
+
+class GenericRule(LearningRuleType):
+    """Learning rule implemented by generic Python function.
+
+    Parameters
+    ----------
+    function : callable
+        Accepts previous values of learned array and a data vector
+        as input, outputs a delta for the learned array.
+
+        Optionally the function may also accept a `params` argument, in which
+        case `model.params` will be passed to the function as well (can be used
+        to look up things like encoder values).
+    learning_rate : float
+        A scalar used to weight the output of the function
+    size_in : int
+        The dimensionality of the data vector input to the function
+    modifies : 'weights' or 'decoders' or 'encoders'
+        The signal targeted by the learning rule
+    pass_model_params : bool
+        Selects whether or not the compiled model parameters will be passed
+        to the learning rule function
+    """
+
+    probeable = ('target', 'data', 'delta')
+
+    pass_model_params = BoolParam('pass_model_params')
+
+    def __init__(self, function, learning_rate=1.0, size_in=0,
+                 modifies="decoders", pass_model_params=False):
+        self.modifies = modifies
+        self.function = function
+        self.pass_model_params = pass_model_params
+
+        super(GenericRule, self).__init__(learning_rate=learning_rate,
+                                          size_in=size_in)
 
 
 class LearningRuleTypeParam(Parameter):
--- a/nengo/learning_rules.py
+++ b/nengo/learning_rules.py
@@ -5,7 +5,7 @@ from nengo.base import ObjView
 from nengo.ensemble import Neurons
 from nengo.exceptions import ValidationError
 from nengo.params import (FrozenObject, NumberParam, Parameter, IntParam,
-                          BoolParam)
+                          TupleParam)
 from nengo.utils.compat import is_iterable, itervalues
 
 
@@ -343,18 +343,18 @@ class GenericRule(LearningRuleType):
         to the learning rule function
     """
 
-    probeable = ('target', 'data', 'delta')
+    probeable = ('delta',)
 
-    pass_model_params = BoolParam('pass_model_params')
+    data_sources = TupleParam("data_sources")
 
-    def __init__(self, function, learning_rate=1.0, size_in=0,
-                 modifies="decoders", pass_model_params=False):
+    def __init__(self, function, data_sources=None, learning_rate=1.0,
+                 modifies="decoders"):
         self.modifies = modifies
         self.function = function
-        self.pass_model_params = pass_model_params
+        self.data_sources = data_sources
 
         super(GenericRule, self).__init__(learning_rate=learning_rate,
-                                          size_in=size_in)
+                                          size_in=0)
 
 
 class LearningRuleTypeParam(Parameter):
--- a/nengo/learning_rules.py
+++ b/nengo/learning_rules.py
@@ -1,18 +1,8 @@
 import warnings
 
-from nengo.base import NengoObjectParam
 from nengo.exceptions import ValidationError
-from nengo.params import FrozenObject, NumberParam, Parameter
+from nengo.params import FrozenObject, NumberParam, Parameter, IntParam
 from nengo.utils.compat import is_iterable, itervalues
-
-
-class ConnectionParam(NengoObjectParam):
-    def validate(self, instance, conn):
-        from nengo.connection import Connection
-        if not isinstance(conn, Connection):
-            raise ValidationError("'%s' is not a Connection" % conn,
-                                  attr=self.name, obj=instance)
-        super(ConnectionParam, self).validate(instance, conn)
 
 
 class LearningRuleType(FrozenObject):
@@ -54,15 +44,16 @@ class LearningRuleType(FrozenObject):
         The signal targeted by the learning rule.
     """
 
-    error_type = 'none'
     modifies = None
     probeable = ()
 
     learning_rate = NumberParam('learning_rate', low=0, low_open=True)
+    size_in = IntParam('size_in', low=0, optional=True)
 
-    def __init__(self, learning_rate=1e-6):
+    def __init__(self, learning_rate=1e-6, size_in=None):
         super(LearningRuleType, self).__init__()
         self.learning_rate = learning_rate
+        self.size_in = size_in
 
     def __repr__(self):
         return '%s(%s)' % (type(self).__name__, ", ".join(self._argreprs))
@@ -94,7 +85,6 @@ class PES(LearningRuleType):
         Filter constant on activities of neurons in pre population.
     """
 
-    error_type = 'decoded'
     modifies = 'decoders'
     probeable = ('error', 'correction', 'activities', 'delta')
 
@@ -105,7 +95,7 @@ class PES(LearningRuleType):
             warnings.warn("This learning rate is very high, and can result "
                           "in floating point errors from too much current.")
         self.pre_tau = pre_tau
-        super(PES, self).__init__(learning_rate)
+        super(PES, self).__init__(learning_rate, size_in=None)
 
     @property
     def _argreprs(self):
@@ -148,7 +138,6 @@ class BCM(LearningRuleType):
         A scalar indicating the time constant for theta integration.
     """
 
-    error_type = 'none'
     modifies = 'weights'
     probeable = ('theta', 'pre_filtered', 'post_filtered', 'delta')
 
@@ -161,7 +150,7 @@ class BCM(LearningRuleType):
         self.theta_tau = theta_tau
         self.pre_tau = pre_tau
         self.post_tau = post_tau if post_tau is not None else pre_tau
-        super(BCM, self).__init__(learning_rate)
+        super(BCM, self).__init__(learning_rate, size_in=0)
 
     @property
     def _argreprs(self):
@@ -209,7 +198,6 @@ class Oja(LearningRuleType):
         Filter constant on activities of neurons in pre population.
     """
 
-    error_type = 'none'
     modifies = 'weights'
     probeable = ('pre_filtered', 'post_filtered', 'delta')
 
@@ -222,7 +210,7 @@ class Oja(LearningRuleType):
         self.pre_tau = pre_tau
         self.post_tau = post_tau if post_tau is not None else pre_tau
         self.beta = beta
-        super(Oja, self).__init__(learning_rate)
+        super(Oja, self).__init__(learning_rate, size_in=0)
 
     @property
     def _argreprs(self):
@@ -262,7 +250,6 @@ class Voja(LearningRuleType):
         Filter constant on activities of neurons in post population.
     """
 
-    error_type = 'scalar'
     modifies = 'encoders'
     probeable = ('post_filtered', 'scaled_encoders', 'delta')
 
@@ -270,7 +257,7 @@ class Voja(LearningRuleType):
 
     def __init__(self, post_tau=0.005, learning_rate=1e-2):
         self.post_tau = post_tau
-        super(Voja, self).__init__(learning_rate)
+        super(Voja, self).__init__(learning_rate, size_in=1)
 
 
 class LearningRuleTypeParam(Parameter):
@@ -287,10 +274,6 @@ class LearningRuleTypeParam(Parameter):
             raise ValidationError(
                 "'%s' must be a learning rule type or a dict or "
                 "list of such types." % rule, attr=self.name, obj=instance)
-        if rule.error_type not in ('none', 'scalar', 'decoded', 'neuron'):
-            raise ValidationError(
-                "Unrecognized error type %r" % rule.error_type,
-                attr=self.name, obj=instance)
         if rule.modifies not in ('encoders', 'decoders', 'weights'):
             raise ValidationError("Unrecognized target %r" % rule.modifies,
                                   attr=self.name, obj=instance)
--- a/nengo/learning_rules.py
+++ b/nengo/learning_rules.py
@@ -1,8 +1,63 @@
 import warnings
+import weakref
 
+from nengo.ensemble import Neurons
 from nengo.exceptions import ValidationError
 from nengo.params import FrozenObject, NumberParam, Parameter, IntParam
 from nengo.utils.compat import is_iterable, itervalues
+
+
+class LearningRule(object):
+    """An interface for making connections to a learning rule.
+
+    Connections to a learning rule are to allow elements of the network to
+    affect the learning rule. For example, learning rules that use error
+    information can obtain that information through a connection.
+
+    Learning rule objects should only ever be accessed through the
+    ``learning_rule`` attribute of a connection.
+    """
+
+    def __init__(self, connection, learning_rule_type):
+        self._connection = weakref.ref(connection)
+        self.learning_rule_type = learning_rule_type
+
+        if learning_rule_type.size_in is not None:
+            self.size_in = learning_rule_type.size_in
+        else:
+            # infer size_in from post
+            self.size_in = (self.connection.post_obj.ensemble.size_in
+                            if isinstance(self.connection.post_obj, Neurons)
+                            else self.connection.size_out)
+
+    def __repr__(self):
+        return "<LearningRule at 0x%x modifying %r with type %r>" % (
+            id(self), self.connection, self.learning_rule_type)
+
+    def __str__(self):
+        return "<LearningRule modifying %s with type %s>" % (
+            self.connection, self.learning_rule_type)
+
+    @property
+    def connection(self):
+        """(Connection) The connection modified by the learning rule."""
+        return self._connection()
+
+    @property
+    def modifies(self):
+        """(str) The variable modified by the learning rule."""
+        return self.learning_rule_type.modifies
+
+    @property
+    def probeable(self):
+        """(tuple) Signals that can be probed in the learning rule."""
+        return self.learning_rule_type.probeable
+
+    @property
+    def size_out(self):
+        """(int) Cannot connect from learning rules, so always 0."""
+        return 0  # since a learning rule can't connect to anything
+        # TODO: allow probing individual learning rules
 
 
 class LearningRuleType(FrozenObject):
--- a/nengo/tests/test_learning_rules.py
+++ b/nengo/tests/test_learning_rules.py
@@ -503,12 +503,14 @@ def test_generic_rule(Simulator, nl_nodirect, seed, modifies):
                                            params.gain, activities) -
                         target * reg)
 
-        probes = [
-            nengo.Probe(conn, attr="weights"),
-            nengo.Probe(error),
-            nengo.Probe(a.neurons, synapse=0.005)]
-        if modifies == "weights":
-            probes += [nengo.Probe(b, attr='params')]
+        with nengo.Config(nengo.Probe) as conf:
+            conf[nengo.Probe].keep_history = False
+            probes = [
+                nengo.Probe(conn, attr="weights"),
+                nengo.Probe(error),
+                nengo.Probe(a.neurons, synapse=0.005)]
+            if modifies == "weights":
+                probes += [nengo.Probe(b, attr='params')]
 
         conn.learning_rule_type = GenericRule(
             regularized_pes, probes, learning_rate=1e-3, modifies=modifies)
@@ -555,14 +557,16 @@ def test_generic_encoders(Simulator, nl_nodirect, rng, seed):
             return (scale * np.outer(activities, input) -
                     activities[:, None] * enc)
 
-        nengo.Connection(
-            u, x, synapse=None,
-            learning_rule_type=GenericRule(
-                generic_voja, [nengo.Probe(x, 'encoders'),
-                               nengo.Probe(u),
-                               nengo.Probe(x.neurons, synapse=0.005),
-                               nengo.Probe(x, 'params')],
-                learning_rate=1e-1 * dt, modifies="encoders"))
+        with nengo.Config(nengo.Probe) as conf:
+            conf[nengo.Probe].keep_history = False
+            nengo.Connection(
+                u, x, synapse=None,
+                learning_rule_type=GenericRule(
+                    generic_voja,
+                    [nengo.Probe(x, 'encoders'), nengo.Probe(u),
+                     nengo.Probe(x.neurons, synapse=0.005),
+                     nengo.Probe(x, 'params')],
+                    learning_rate=1e-1 * dt, modifies="encoders"))
 
     with Simulator(m, dt=dt) as sim:
         sim.run(1.0)
--- a/nengo/tests/test_learning_rules.py
+++ b/nengo/tests/test_learning_rules.py
@@ -3,7 +3,8 @@ import pytest
 
 import nengo
 from nengo.dists import UniformHypersphere
-from nengo.learning_rules import LearningRuleTypeParam, PES, BCM, Oja, Voja
+from nengo.learning_rules import (LearningRuleTypeParam, PES, BCM, Oja, Voja,
+                                  GenericRule)
 from nengo.processes import WhiteSignal
 
 
@@ -466,3 +467,113 @@ def test_frozen():
 
     with pytest.raises((ValueError, RuntimeError)):
         a.learning_rate = 1e-1
+
+
+@pytest.mark.parametrize('modifies', ["decoders", "weights"])
+def test_generic_rule(Simulator, nl_nodirect, seed, modifies):
+    n = 100
+    input_vector = np.asarray([0.5, -0.5])
+    dt = 0.001
+    reg = 0.1
+
+    m = nengo.Network(seed=seed)
+    with m:
+        m.config[nengo.Ensemble].neuron_type = nl_nodirect()
+        u = nengo.Node(output=input_vector)
+        a = nengo.Ensemble(n, dimensions=2)
+        b = nengo.Ensemble(n, dimensions=2)
+
+        nengo.Connection(u, a)
+
+        if modifies == "decoders":
+            def regularized_pes(target, data):
+                return -dt / n * np.outer(data[:2], data[2:]) - target * reg
+        else:
+            def regularized_pes(target, data, params):
+                encoders = params[b].encoders
+                gains = params[b].gain
+                return (-dt / n *
+                        np.outer(np.dot(encoders, data[:2]) * gains,
+                                 data[2:]) - target * reg)
+
+        rule = GenericRule(regularized_pes, learning_rate=1e-3,
+                           size_in=2 + n, modifies=modifies,
+                           pass_model_params=modifies == "weights")
+
+        conn = nengo.Connection(
+            a if modifies == "decoders" else a.neurons,
+            b if modifies == "decoders" else b.neurons,
+            transform=1 if modifies == "decoders" else
+            nengo.dists.Gaussian(0, 0.001),
+            learning_rule_type=rule)
+        nengo.Connection(b, conn.learning_rule[:2])
+        nengo.Connection(u, conn.learning_rule[:2])
+        nengo.Connection(a.neurons, conn.learning_rule[2:], synapse=0.005)
+
+        b_p = nengo.Probe(b, synapse=0.1)
+        weights_p = nengo.Probe(conn, attr="weights")
+
+    with Simulator(m, dt=dt) as sim:
+        sim.run(1.)
+
+    tmask = sim.trange() > .9
+    assert (np.sum(np.abs(sim.data[weights_p][:10])) >
+            np.sum(np.abs(sim.data[weights_p][-10:])))
+    assert np.allclose(sim.data[b_p][tmask], -input_vector, atol=0.05)
+
+
+def test_generic_encoders(Simulator, nl_nodirect, rng, seed):
+    """Voja test implemented via generic rule."""
+
+    n = 200
+    learned_vector = np.asarray([0.3, -0.4, 0.6])
+    learned_vector /= np.linalg.norm(learned_vector)
+    n_change = n // 2  # modify first half of the encoders
+    dt = 1e-3
+
+    # Set the first half to always fire with random encoders, and the
+    # remainder to never fire due to their encoder's dot product with the input
+    intercepts = np.asarray([-1] * n_change + [0.99] * (n - n_change))
+    rand_encoders = UniformHypersphere(surface=True).sample(
+        n_change, len(learned_vector), rng=rng)
+    encoders = np.append(
+        rand_encoders, [-learned_vector] * (n - n_change), axis=0)
+
+    m = nengo.Network(seed=seed)
+    with m:
+        m.config[nengo.Ensemble].neuron_type = nl_nodirect()
+        u = nengo.Node(output=learned_vector)
+        x = nengo.Ensemble(n, dimensions=len(learned_vector),
+                           intercepts=intercepts, encoders=encoders,
+                           radius=2.0)  # to test encoder scaling
+
+        def generic_voja(enc, data, params):
+            scale = (params[x].gain / x.radius)[:, None]
+            return (scale * np.outer(data[:n], data[n:]) -
+                    data[:n, None] * enc)
+
+        conn = nengo.Connection(
+            u, x, synapse=None,
+            learning_rule_type=GenericRule(
+                generic_voja, learning_rate=1e-1 * dt, modifies="encoders",
+                size_in=len(learned_vector) + n, pass_model_params=True))
+        nengo.Connection(u, conn.learning_rule[n:], synapse=None)
+        nengo.Connection(x.neurons, conn.learning_rule[:n], synapse=0.005)
+
+    with Simulator(m, dt=dt) as sim:
+        sim.run(1.0)
+
+        scaled_encoders = sim.signals[sim.model.sig[x]['encoders']]
+
+    encoder_scale = (sim.model.params[x].gain / x.radius)[:, np.newaxis]
+
+    # Check that the last half kept the same encoders throughout the simulation
+    assert np.allclose(sim.model.params[x].scaled_encoders[n_change:],
+                       scaled_encoders[n_change:])
+    # and that they are also equal to their originally assigned value
+    assert np.allclose(sim.model.params[x].scaled_encoders[n_change:] /
+                       encoder_scale[n_change:], -learned_vector)
+
+    # Check that the first half converged to the input
+    assert np.allclose(scaled_encoders[:n_change] / encoder_scale[:n_change],
+                       learned_vector, atol=0.01)
--- a/nengo/tests/test_learning_rules.py
+++ b/nengo/tests/test_learning_rules.py
@@ -12,7 +12,6 @@ def _test_pes(Simulator, nl, plt, seed,
               pre_neurons=False, post_neurons=False, weight_solver=False,
               vin=np.array([0.5, -0.5]), vout=None, n=200,
               function=None, transform=np.array(1.), rate=1e-3):
-
     vout = np.array(vin) if vout is None else vout
 
     model = nengo.Network(seed=seed)
@@ -83,7 +82,7 @@ def test_pes_weight_solver(Simulator, plt, seed):
 
 def test_pes_ens_slice(Simulator, plt, seed):
     vin = [0.5, -0.5]
-    vout = [vin[0]**2 + vin[1]**2]
+    vout = [vin[0] ** 2 + vin[1] ** 2]
     function = lambda x: [x[0] - x[1]]
     _test_pes(Simulator, nengo.LIF, plt, seed, vin=vin, vout=vout,
               function=function)
@@ -186,7 +185,7 @@ def test_unsupervised(Simulator, rule_type, solver, seed, rng, plt):
     with m:
         u = nengo.Node(WhiteSignal(0.5, high=10), size_out=2)
         a = nengo.Ensemble(n, dimensions=2)
-        b = nengo.Ensemble(n+1, dimensions=2)
+        b = nengo.Ensemble(n + 1, dimensions=2)
         nengo.Connection(u, a)
 
         if solver:
@@ -205,7 +204,7 @@ def test_unsupervised(Simulator, rule_type, solver, seed, rng, plt):
         ap = nengo.Probe(a, synapse=0.03)
         up = nengo.Probe(b, synapse=0.03)
 
-    with Simulator(m, seed=seed+1) as sim:
+    with Simulator(m, seed=seed + 1) as sim:
         sim.run(0.5)
     t = sim.trange()
 
@@ -313,6 +312,7 @@ def test_reset(Simulator, learning_rule, plt, seed, rng):
 
 def test_learningruletypeparam():
     """LearningRuleTypeParam must be one or many learning rules."""
+
     class Test(object):
         lrp = LearningRuleTypeParam('lrp', default=None)
 
@@ -333,6 +333,7 @@ def test_learningruletypeparam():
 
 def test_learningrule_attr(seed):
     """Test learning_rule attribute on Connection"""
+
     def check_rule(rule, conn, rule_type):
         assert rule.connection is conn and rule.learning_rule_type is rule_type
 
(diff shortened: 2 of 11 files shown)


drasmuss:
**Description:**


I'll be honest, I forgot I had this PR.  But now it's back and better than ever!

Same idea as before, it allows you to define a python function implementing some learning rule update, and it'll just call that function every timestep to compute a delta on your weights/encoders/decoders.  So that makes it super easy to play around with new learning rules, or drop learning rules from other people into your code.

The syntax is nicer now, thanks to the new ability to connect directly to learning rules.  Now whatever you connect up to the learning rule will be passed to the function you defined.  So that's how you pass error signals or neural activities from the model to your learning update.

I also made two more wide-reaching changes, separated into different commits:

a0143e319bd44b85acf8a92df67e988d0c54fff7 changes the implementation of Learning Rules slightly.  Before, each learning rule would specify `error_type='none'/'scalar'/'decoded'`.  In practice, all this did was set the `size_in` of the learning rule (i.e., the dimensionality we expect on incoming connections).  I changed it so that  you just directly specify `size_in`.  This has a one-to-one mapping with the old implementation; `none->0`, `scalar->1`, `decoded->None` (`None` because we have to infer this after the learning rule is created based on the dimensionality of post).  The motivation for this change is that learning rules can have incoming connections that are not necessarily an error signal, so we need a more generic way of specifying `size_in`.  If people really liked the old syntax, we could leave it in, and just add the option to set `error_type=<an int>`, in which case we'd use that as `size_in`.  My thinking is that if we were going to have that option, we might as well just use it everywhere and get rid of the `none`/`scalar`/`decoded` string stuff.

268eabae67b092b904af25335152c8d2e9621a11 moves the `LearningRule` class from `connection.py` to `learning_rule.py`.  This is just based on my experience re-acquainting myself with the learning rule system, during which I found it confusing having the learning rule functionality split between these two files.  I realize that the reason we created `LearningRule` was to handle connections to learning rules, which is why it was put in `connection.py`.  But to me it makes a lot more sense from a user's perspective to have all the things related to the function of a learning rule in `learning_rule.py`, not divided up according to how the learning rule is used.  For example, the function of `ensemble.Neurons` is very similar to `LearningRule`; it provides the wrapper code that allows users to make a connection to `my_ensemble.neurons`, just as they can connect to `my_connection.learning_rule`.  But we put `Neurons` in `ensemble.py`, not `connection.py`.  Obviously none of this makes any difference from a functional perspective, it's just a subjective readability/discoverability thing.  So if there are objections it's easy to take out, this way just made more sense to me.

**Motivation and context:**



We seem to frequently get the question "does Nengo have <learning rule x>?" Often the first thing a totally new user wants to do is take some other model and implement it in Nengo, and a lot of those models are based around some specific learning rule implementation in a paper.  The idea here is to make it as simple as possible to implement new learning rules, so the answer is basically "if you can write your learning rule in a python callable, then you can put it in Nengo".  This is also handy for experienced users who just want to try out some different ideas, without having to implement a custom learning rule each time.

Note that this is not a replacement for nengo-specific learning rule implementations.  The main downside here is that the internal processing of the learning rule is hidden from the backend, so it can't be optimized, probed, etc.  So I think of this as a prototyping method, which would then be moved into a proper implementation if it seems worthwhile.


**How has this been tested?**


#1095 (I don't think there's any actual conflicts here, they just touch a lot of the same files so might require a little hand-merging).

I'd go through commit-by-commit, as they're relatively separable changes.

Check out the new test in `test_learning_rules.py` to see how the syntax for the new learning rule works in practice.

**Types of changes:**



**Checklist:**





- [x ] I have read the **CONTRIBUTING.rst** document.
- [x ] I have updated the documentation accordingly.
- [x ] I have included a changelog entry.
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.

hunse:
The fact that we put `Neurons` in `ensemble.py` seems to me to be an argument to keep `LearningRule` in `connection.py`. `Neurons` is used for the `Ensemble.neurons` attribute, so it goes with `Ensemble`. `LearningRule` is used with the `Connection.learning_rule` attribute, so if we want it to be analogous, it should go with `Connection`. Or if you want to make an argument for moving it, you should recommend moving `Neurons` to `neurons.py` as well. I could see an argument for moving them both.

nengo/learning_rules.py
```diff
@@ -262,15 +308,49 @@ class Voja(LearningRuleType):
...
+
+        Optionally the function may also accept a `params` argument, in which
+        case `model.params` will be passed to the function as well (can be used
+        to look up things like encoder values).
+    learning_rate : float
+        A scalar used to weight the output of the function
+    size_in : int
+        The dimensionality of the data vector input to the function
+    modifies : str
+        The signal targeted by the learning rule ('weights' or 'decoders')
```
hunse: Can be 'encoders' too, right?

nengo/learning_rules.py
```diff
@@ -262,15 +308,49 @@ class Voja(LearningRuleType):
...
+        to the learning rule function
+    """
+
+    probeable = ('target', 'data', 'delta')
+
+    def __init__(self, function, learning_rate=1.0, size_in=0,
+                 modifies="decoders", pass_model_params=False):
+        self.modifies = modifies
+        self.function = function
+        self.pass_model_params = pass_model_params
```
hunse: These should all be `Parameter`s. Use `EnumParam` for modifies.

nengo/learning_rules.py
```diff
@@ -262,15 +308,49 @@ class Voja(LearningRuleType):
...
+        as input, outputs a delta for the learned array.
+
+        Optionally the function may also accept a `params` argument, in which
+        case `model.params` will be passed to the function as well (can be used
+        to look up things like encoder values).
+    learning_rate : float
+        A scalar used to weight the output of the function
+    size_in : int
+        The dimensionality of the data vector input to the function
+    modifies : str
```
hunse: Put the alternatives here: `modifies : 'weights' or 'decoders' or 'encoders'`

nengo/tests/test_learning_rules.py
```diff
@@ -466,3 +467,61 @@ def test_frozen():
 
     with pytest.raises((ValueError, RuntimeError)):
         a.learning_rate = 1e-1
+
+
+@pytest.mark.parametrize('modifies', ["decoders", "weights"])
```
hunse: We have a test for encoders here, either part of this test or as a second test.

nengo/tests/test_learning_rules.py
```diff
@@ -466,3 +467,61 @@ def test_frozen():
...
+def test_generic_rule(Simulator, nl_nodirect, seed, modifies):
+    if nl_nodirect == nengo.RectifiedLinear and modifies == "weights":
+        # this is a sloppy way of avoiding this one case without having
+        # to make a big custom parameterization
+        # TODO: why doesn't it work for this case?
```
hunse: That's really strange that this doesn't work. Have you played around with this much, @drasmuss?

hunse:
I added some suggestions. I'm pretty busy over the next few days, but I can make them next week if no one else gets to them before that.

drasmuss:
>The fact that we put Neurons in ensemble.py seems to me to be an argument to keep LearningRule in connection.py. Neurons is used for the Ensemble.neurons attribute, so it goes with Ensemble. LearningRule is used with the Connection.learning_rule attribute, so if we want it to be analogous, it should go with Connection. Or if you want to make an argument for moving it, you should recommend moving Neurons to neurons.py as well. I could see an argument for moving them both.

To me it is not the fact that `Ensemble` and `Neurons` are syntactically related (`a.b`) that puts them in the same file, it is that a) they are closely related from a naive user's perspective (ensemble "isa" group of neurons) and b) `Neurons` is never used other than in the `ensemble.neurons` attribute.  `LearningRule`, on the other hand, is not just used for the `connection.learning_rule` attribute, it has a functional role distinct from connections.

As a completely hypothetical example to illustrate in more detail what prompted this change, lets say someone wants to know how the PES rule works, so they go look up `build_pes` in `test_learning_rules.py`.  The definition of that function takes in two arguments, `pes` and `rule`, and the first line in that function is `conn = rule.connection`.  So now they want to figure out what that `LearningRule` object is, where would they go to look?  I submit that this very reasonable, hypothetical person would expect that to be in `learning_rules.py`.  I don't think there is anything about the definition of a `LearningRule` object that would lead you to expect it in `connection.py`.  Obviously that is nothing that a quick search wouldn't resolve, it just feels a bit awkward.

So yeah, I'd leave `Neurons` where it is and just move `LearningRule`.  But it's all just subjective, so I'd get one or two more opinions and then go with whatever feels right to the most people.

>EDIT: Definitely break that all out into another commit, though, because none of that is necessary for the generic learning rule.

It is in a separate commit at the moment.  Or did you mean make the `Neurons -> neurons.py` swap a different commit from the `LearningRule -> learning_rules.py` swap (if we decide to do that)?

nengo/learning_rules.py
drasmuss: Ah yeah, good call (I was just being lazy and copy-pasting the documentation :astonished:)

nengo/learning_rules.py
drasmuss: Do you think the `modifies` parameter in the other learning rules should be a `Parameter` as well?  It's a bit odd, as in that in those other cases it is not something we expect to be set by the user while in this class it is.  But if only this one is a parameter, then if you print all the parameters of a learning rule (via e.g. `Config.all_defaults`) you'd get the impression that it's a unique attribute of `GenericRule` rather than a property of the `LearningRuleType` base class.

nengo/tests/test_learning_rules.py
drasmuss: I didn't work too hard on it.  My guess is that it's because you end up with much larger gains on `RectifiedLinear` neurons, and that (perhaps combined with the lack of saturation) makes things more unstable.

drasmuss:
Don't worry about making the changes, I can get to it later today or tomorrow.

nengo/learning_rules.py
drasmuss: I ended up leaving `modifies` the way it is, because I noticed we're already checking the values in the `LearningRuleTypeParam` validation.

drasmuss:
Pushed the changes and cleaned up the history.

hunse:
But isn't your hypothetical example also true for `Neurons`? If a user looks at `build_lif`, for example, it gets `neurons`, which is a `Neurons` object. So doesn't the same thing apply?

The only place we create `LearningRule` objects is for `Connection.learning_rule`, just as the only place we create `Neurons` objects is for `Ensemble.neurons`. They're both used to be able to connect to a specific aspect of the larger object. To me, they seem very analogous, and we should do the same for both.

drasmuss:
Yes, but `build_lif` is called from `build_ensemble`.  So if a user is tracing "what is this `neurons` object?", they go to `build_ensemble` in `builder/ensemble.py`, which I think pretty naturally leads them to `nengo/ensemble.py`.  When the user is tracing "what is this `rule` object?", they're lead to `build_learning_rule` in `builder/learning_rules.py`; nothing connects to `connection.py`.  More generally, I think what this tells us is that `Neurons` is more closely related to `Ensemble` than `LearningRule` is to `Connection`.

Another way of saying the same thing: it seems to make sense that `build_learning_rule` belongs with `build_bcm`, `build_pes`, etc., (rather than with `build_connection`) because if you want to understand how `PES` works you really need to understand how `LearningRule` works (e.g., if you want to know how it is getting an error signal).  On the other hand, if you want to understand how `LIF` works, you don't really need to understand how `Neurons` works.  Which is why I don't think anyone found it contradictory to put `build_learning_rule` in `learning_rules.py`.

Again this is all subjective and others may feel `LearningRule` is more closely related to `Connection` than it is to `PES`.  But I don't think there is such a strong analogy between `Neuron` and `LearningRule` that we should feel bound to treat and organize them the same.

hunse:
But `build_lif` appears in `builder/neurons.py`, so if the person in question is looking in that file, how would they know to go to `builder/ensemble.py`?

Also, I think you're slightly misunderstanding what I'm arguing here. I'm not trying to say that we shouldn't move `LearningRule`. I think that's fine, and maybe even somewhat better (clearly, there are arguments for and against). I'm just arguing that `Neurons` should also move. To me, the analogy seems very clear. What I don't understand is why you're opposed to `Neurons` moving?

It is true that `Neurons` has fewer parameters than `LearningRule`, but it still has things like a `size_in` that's used in the functions in `builder/neurons.py`, and to understand those functions, it would help if it's easy to find `Neurons` (in `neurons.py`). So why not move it too?

drasmuss:
I would be totally fine with moving `Neurons`, all I was meaning to say was that I can see people wanting to move one but not the other.  The only reason I was thinking I'd leave it where it was is just because it would be less of a change to the code base (since it didn't seem as necessary as moving `LearningRule`).  But yeah I have no objection to moving `Neurons` as well.

nengo/tests/test_learning_rules.py
```diff
@@ -466,3 +467,118 @@ def test_frozen():
...
+            transform=1 if modifies == "decoders" else np.eye(n),
+            learning_rule_type=rule)
+        nengo.Connection(b, conn.learning_rule[:2], transform=1)
+        nengo.Connection(u, conn.learning_rule[:2], transform=1)
+        nengo.Connection(a.neurons, conn.learning_rule[2:], synapse=0.005)
+
+        b_p = nengo.Probe(b, synapse=0.1)
+        weights_p = nengo.Probe(conn, attr="weights")
+
+    sim = Simulator(m, dt=dt)
```
jgosmann: This should use the `with` syntax.

nengo/learning_rules.py
```diff
@@ -315,6 +320,43 @@ def __init__(self, post_tau=0.005, learning_rate=1e-2):
...
+
+    Parameters
+    ----------
+    function : callable
+        Accepts previous values of learned array and a data vector
+        as input, outputs a delta for the learned array.
+
+        Optionally the function may also accept a `params` argument, in which
+        case `model.params` will be passed to the function as well (can be used
+        to look up things like encoder values).
```
jgosmann: Isn't this specific to the reference backend?
Maybe we should finally specify how backends in general are supposed to provide access to things like encoder etc.?

jgosmann:
> I changed it so that you just directly specify size_in. This has a one-to-one mapping with the old implementation; none->0, scalar->1, decoded->None

Seems like a good thing to do, the only thing that I slightly dislike is that `size_in=None` doesn't really express that `size_in` is inferred from the post population.

nengo/tests/test_learning_rules.py
```diff
@@ -466,3 +467,118 @@ def test_frozen():
...
+                                 data[2:]) - target * reg)
+
+        rule = GenericRule(regularized_pes, learning_rate=1e-3,
+                           size_in=2 + n, modifies=modifies,
+                           pass_model_params=modifies == "weights")
+
+        conn = nengo.Connection(
+            a if modifies == "decoders" else a.neurons,
+            b if modifies == "decoders" else b.neurons,
+            transform=1 if modifies == "decoders" else np.eye(n),
```
jgosmann: Is this line necessary?

nengo/tests/test_learning_rules.py
```diff
@@ -466,3 +467,118 @@ def test_frozen():
...
+            learning_rule_type=rule)
+        nengo.Connection(b, conn.learning_rule[:2], transform=1)
+        nengo.Connection(u, conn.learning_rule[:2], transform=1)
```
jgosmann: `transform=1` can be skipped, I believe?

nengo/tests/test_learning_rules.py
```diff
@@ -466,3 +467,118 @@ def test_frozen():
...
+
+        b_p = nengo.Probe(b, synapse=0.1)
+        weights_p = nengo.Probe(conn, attr="weights")
+
+    sim = Simulator(m, dt=dt)
+    sim.run(1.)
+
+    tmask = sim.trange() > .9
+    assert (np.sum(np.abs(sim.data[weights_p][:10])) >
+            np.sum(np.abs(sim.data[weights_p][-10:])))
```
jgosmann: Why is this the case?

nengo/tests/test_learning_rules.py
```diff
@@ -466,3 +467,118 @@ def test_frozen():
...
+
+        def generic_voja(enc, data, params):
+            scale = (params[x].gain / x.radius)[:, None]
+            return (scale * np.outer(data[:n], data[n:]) -
+                    data[:n, None] * enc)
+
+        conn = nengo.Connection(
+            u, x, synapse=None,
+            learning_rule_type=GenericRule(
+                generic_voja, learning_rate=1e-1 * dt, modifies="encoders",
```
jgosmann: Why is the learning rate scaled by `dt` in this test, but not the other test?

nengo/tests/test_learning_rules.py
drasmuss: Yes. You have to specify the full transform matrix on a neuron--neuron connection or it won't be automatically generated (we just use a `1xn` matrix to apply the gains), which means the learning rule doesn't have the correct sized signal to target.

nengo/tests/test_learning_rules.py
drasmuss: Ah yeah, I was playing around with different transforms just to make sure things worked as expected, but they aren't needed anymore.

nengo/tests/test_learning_rules.py
drasmuss: Because our learning rule is applying an l2 regularization term (see the `regularized_pes` function above)

nengo/learning_rules.py
drasmuss: Yeah, it's awaiting a more permanent solution for how to access properties of the built model.  That's why I have it kind of hidden (optional, and turned off by default).  So it's there if someone needs it, but presumably if they understand what `model.params` is and how to use it, they will also understand the limitations.

nengo/tests/test_learning_rules.py
jgosmann: So the weights are sorted by magnitude? I don't see how that follows from the regularization. I also don't see why is this calculated over exactly 10 weights on each end and not 5 or 20?

nengo/tests/test_learning_rules.py
drasmuss: It's scaled by `dt` in both cases.  In the PES case the `dt` is incorporated inside the `regularized_pes` function.

jgosmann:
I think the first test would be slightly easier to understand if it would learn a communication channel instead of a sign flip.

I don't like it that in both tests the `data` vector is split into two parts because `data[:n]` and `data[n:]` are not very expressive what data each part represents. It also seems to be error prone to set up the connections accordingly if `len(:n) == len(n:)`. Wouldn't it be possible to allow a variable number of named data arguments? Also, I wonder if there is a better way to access data from the pre and post ensembles than additional connections to data. These are already given in the connection the learning rule is applied to, so the extra connections feel redundant.

nengo/tests/test_learning_rules.py
jgosmann: Indeed. I think it would be good to unify because the tests are also somewhat an example of what we recommend. So if we recommend to include the `dt` in the learning rule we should do that in both cases or if we don't that we shouldn't. (To me it seems the former case would be better as the `dt` is defined by the simulator and can change.) This should probably also agree with the specialized implementations of these learning rules.

Maybe the `dt` should be passed as argument to the generic learning rule function from the simulator (to avoid defining it independent of the simulator)?

nengo/tests/test_learning_rules.py
drasmuss: This is an index into probe data, so the `:10` is referencing the first 10 timesteps, not the first 10 weights.

jgosmann: I see. But if we started with very small weights they could still increase despite regularization, couldn't they?

nengo/tests/test_learning_rules.py
drasmuss: I did it the two different ways intentionally as a way to illustrate the different possibilities.  I don't really feel strongly that one way is "better" than the other, but if it were up to me I'd probably get rid of the `learning_rate` parameter altogether and just have the user build that into their learning rule function like they do all the other variables.  But it'd be confusing to have an unused `learning_rate` parameter, so I left the option to do it both ways.

No strong opinion on whether to include `dt` or not.  My inclination is to keep the syntax for the user function as simple as possible, but maybe the convenience of automatically passing `dt` would be worth it.

nengo/tests/test_learning_rules.py
drasmuss: It's not really how small the weights are to start with that matters, it's to what extent increasing the magnitude of the weights can reduce the error.  But yes, a regularization term is just a penalty applied to large weights, it is not a guarantee that weights must decrease in magnitude.

drasmuss:
>Seems like a good thing to do, the only thing that I slightly dislike is that size_in=None doesn't really express that size_in is inferred from the post population.

Yeah, I thought about using a string instead, `"post"` or something like that.  Could go with that if you think it'd be better.

> I don't like it that in both tests the data vector is split into two parts because data[:n] and data[n:] are not very expressive what data each part represents.

I could add `activities = data[n:]` and `output = data[:n]` and then reference those variables instead of the `data` vector if you think that'd be more clear.  

>  It also seems to be error prone to set up the connections accordingly if len(:n) == len(n:). Wouldn't it be possible to allow a variable number of named data arguments? 

It'd be possible, but it would require some entirely new user-facing connection syntax (since the `GenericRule` object is hidden behind the `connection.LearningRule` object, we can't just add new connection logic to the `GenericRule` class).  Which is something I'd rather avoid.

> Also, I wonder if there is a better way to access data from the pre and post ensembles than additional connections to data. These are already given in the connection the learning rule is applied to, so the extra connections feel redundant.

I thought about doing that, but although it is common to use that data in a learning rule, it isn't universal.  It's part of the price you pay for making a generic rule, that it isn't as clean and concise as a more specialized implementation.  I also like the simplicity of just saying "all the inputs to the learning rule need to go into the `data` vector", rather than having a large function definition that novice users need to parse in order to figure out what is being passed into this function.  So yeah, I would just leave this implementation as generic as possible, and then we can build specific implementations if we want a particular rule to be easier for people to use.

nengo/tests/test_learning_rules.py
```diff
@@ -466,3 +467,113 @@ def test_frozen():
...
+
+        nengo.Connection(u, a)
+
+        if modifies == "decoders":
+            def regularized_pes(target, data):
+                return -dt / n * np.outer(data[:2], data[2:]) - target * reg
+        else:
+            def regularized_pes(target, data, params):
+                encoders = params[b].encoders
+                gains = params[b].gain
```
jgosmann: Another thing I don't like is that this function needs to explicitly reference the ensemble, so that I can't reuse it on other connections.

jgosmann:
I feel that there are two main things that could be improved with this PR:
* Access to encoders, gains, radii of pre and post ensemble. I tried to resolve this within this PR, but then realized that it is not nicely possible until #1145 is resolved.
* Non-descriptiveness of split data signal (`data[:2]` and `data[2:]`). It doesn't seem to be very complicated to allow for multiple named data signals. But the base class `LearningRuleType` assumes a single signal because it defines a single `size_in`. So changing that would require some slight changes to the other learning rules too.

I think, we should resolve #1145 before going ahead with this PR, but if you want to get this as fast as possible into master, I would be ok with it given a few minor fixes if it were marked as experimental and made clear that the API is likely to change.

I would also like to allow learning rules to have more than one data/error signal. Mostly because it would make `GenericRule` implementations much easier to understand, but also because maybe some day wants to implement a learning rule that gets two different error signals (who knows?). However, I would like other people's opinion on this first before I invest time in that. I could imagine that such a change could be somewhat controversial.

drasmuss:
I don't think the `data` signal will be that confusing to people.  It's exactly the same way `Nodes` work; if you want multiple input signals to your node, you feed them all into the node's single input signal, and then split them up however you like within your node function.  And while allowing users to connect to different named variables within a GenericRule would not be complicated programmatically, conceptually it adds an entirely new behaviour to the connection logic (connecting to specific signals within an object).    I don't think the minor gain in convenience would justify that added complexity (especially considering that the single-signal system seems to have met all our needs fairly well for nodes).

As for waiting on some future PR that adds a different way of connecting to `model.params`, no strong feelings either way.  There's no particular urgency to merging this in, other than the general point that it would be nice to move this out of the open PR queue.

nengo/tests/test_learning_rules.py
drasmuss: That's entirely up to the person writing the update function.  You could make a callable class that takes the ensemble as a parameter if you want a more general solution.

jgosmann:
> I don't think the data signal will be that confusing to people.

To me it was pretty confusing when I was trying to understand the tests. It also requires that people realize that they can append the different signals into a single vector.

> It's exactly the same way Nodes work; if you want multiple input signals to your node, you feed them all into the node's single input signal, and then split them up however you like within your node function.

I never handled multiple input signals this way (except maybe if all of them were one-dimensional). I usually use a class with multiple functions that provide input with different nodes (though part of the reason might also be that I need some persistent state).

> And while allowing users to connect to different named variables within a GenericRule would not be complicated programmatically, conceptually it adds an entirely new behaviour to the connection logic (connecting to specific signals within an object).

That's not different from connecting to or from ensemble neurons, so I don't think it's conceptually new.

> I don't think the minor gain in convenience would justify that added complexity

To me it's not so much about convenience, but an API that makes it easy to write easily comprehendable and correct code. To me that outweighs the slight increase in complexity in the Nengo code (because many Nengo users are probably less experienced programmers). The current implementations of learning rules in the tests raise quite a few warning flags (e.g. easy to swap `2:` and `:2`, using the same rule in two places might silently fail, ...). This is fine a quick and dirty test implementation, but I don't like to promote such coding practices as part of the API.

jgosmann:
Somehow missed you're comment from 2 days ago ...

> I could add activities = data[n:] and output = data[:n] and then reference those variables instead of the data vector if you think that'd be more clear.

I think that would be the minimum. Still doesn't clear it up in the place where the connection is created, but at least it is stated somewhere what is what and one doesn't have to resort to figure it out purely from context.

>> Also, I wonder if there is a better way to access data from the pre and post ensembles than additional connections to data. These are already given in the connection the learning rule is applied to, so the extra connections feel redundant.
> 
> I thought about doing that, but although it is common to use that data in a learning rule, it isn't universal. It's part of the price you pay for making a generic rule, that it isn't as clean and concise as a more specialized implementation.

Every learning rule is applied to the connection between two ensembles (or Nengo objects?), so that part is universal. Because of that,  I would sort of expect that I am able to access the pre and post objects including their signals. If that would be possible in general it would also solve the problem that the data vector needs to be split into two things. (In most cases at least. With two error signals it would still be an issue, but if that's ever required, it's probably rare enough that I would be ok with concatenating it into a single vector.)

drasmuss:
I'm just going to use this as a gentle reminder of our [new review policy](https://github.com/nengo/nengo/blob/master/docs/contributing.rst).  This isn't particularly aimed at you @jgosmann, I think it's easy for all of us to slip back into old habits (I've done it myself), and the only way to break ourselves of that is to be reminded.

> Please be explicit about your expectations of what will happen in response to your comment. If you're asking a question, then it is clear that the PR author should respond. If you're pointing out an issue that you plan to fix later with a commit, say that in your comment so that the PR author doesn't make that change in the meantime.
>A bad use of an inline comment is to ask for a significant change to the PR. These comments tend to be frustrating for PR authors to respond to and end up prolonging the review process. For significant changes, instead make a commit.
>...
>Instead of asking the PR author for changes, we prefer reviewers make commits to propose changes to a PR. Commits allow the reviewer to propose explicit changes that the PR author can say yes or no to, rather than placing the burden on the PR author and allowing for miscommunication.

jgosmann:
Yeah, I would be happy to make the required changes, but I don't want to waste time on changes that everyone else would be opposed to in the first place (which could be the case with multiple error signals for one learning rule) and (hopefully) less controversial changes are blocked by #1145 right now.

drasmuss:
I pushed a commit that changes the syntax so that you use a list of `Probes` to indicate the input to the learning rule rather than the `data` vector. 

The main benefit is that it allows the learning rule function to define a list of named arguments (as @jgosmann requested), which are matched up with the list of probes.  And then if we implement "static probes" (as in https://github.com/nengo/nengo/pull/1146), we can get rid of the `params` argument and just treat that as another Probe.  

The main downside is that it's kind of shoehorning Probes into a role they weren't really designed for.  The main awkwardness is that we need to build the probes before building this learning rule (normally probes are built last).  So the build function for this learning rule manually calls the builder for its probes.  Nothing technically wrong with that, it just doesn't feel great.

jgosmann:
Thank you for trying that out!

Somehow I can't figure out how the probes are converted into signals.

> So the build function for this learning rule manually calls the builder for its probes. Nothing technically wrong with that, it just doesn't feel great.

Um, this is only guaranteed to work with the reference builder. So it doesn't solve the problem that `params` is specific to the reference backend.


I just had an idea how I might solve the backend dependence, make the implementation independent of the decisions in #1145, and doesn't require probes (no manual call to the `build_probe` required). Hopefully, I'll have some time to implement it this week to see how that looks like.

drasmuss:
> Um, this is only guaranteed to work with the reference builder. 

Not sure what you mean.  This is the backend implementation we're talking about, so it'll definitely only work for that backend.  The general requirement is that when you pass it a list of probes, somehow the data that would normally be probed also get fed into that function, which would be implemented very differently in different backends.  

> So it doesn't solve the problem that params is specific to the reference backend.

I don't think there's any way around the fact that `params` is always going to be backend dependent, because it is a reference into the built parameters of the object (it's basically a way for the frontend to get some access to the backend).  I think all we can hope to accomplish is that there is a standardized interface for accessing those properties (as is the case here).

drasmuss:
> Somehow I can't figure out how the probes are converted into signals.

Sorry, missed this one.  It's this line here https://github.com/nengo/nengo/pull/553/commits/f73de77d8341b119b8ec1f26dd84d5ba493f3ce9#diff-d4a1ea7c740725a930af968ec8143735R659

jgosmann:
>> Um, this is only guaranteed to work with the reference builder.
> 
> Not sure what you mean.

Not every backend might have a `build_probe` function. But I realize now that it is only called in the `build_generic_rule` function that already is backend specific code.

> I think all we can hope to accomplish is that there is a standardized interface for accessing those properties (as is the case here).

Yeah, that's why I opened #1145.

jgosmann:
I implemented another proposal in #1149 and provide reasons why I think it's better in that PR. Note that it is not polished enough to be merged, but good enough to be discussed (keeping in mind that names can easily be changed).

drasmuss:
For comparison, here is the same implementation under 
Option 1:
``` python
def regularized_pes(target, data, params):
    error = data[:2]
    activities = data[2:]
    return (-dt / n * np.outer(
        np.dot(params[b].encoders, error) * params[b].gain, 
        activities) - target * reg)

rule = GenericRule(regularized_pes, learning_rate=1e-3,
    size_in=2 + n, modifies='weights',
    pass_model_params=True)

conn = nengo.Connection(
    a.neurons, b.neurons, transform=nengo.dists.Gaussian(0, 0.001),
    learning_rule_type=rule)
nengo.Connection(b, conn.learning_rule[:2])
nengo.Connection(u, conn.learning_rule[:2])
nengo.Connection(a.neurons, conn.learning_rule[2:], synapse=0.005)
```
Option 2:
``` python
def regularized_pes(target, error, activities, params):
    return (-dt / n * np.outer(
        np.dot(params.encoders, error) * params.gain,
        activities) - target * reg)

conn = nengo.Connection(
    a.neurons, b.neurons,
    transform=nengo.dists.Gaussian(0, 0.001))
nengo.Connection(b, error)
nengo.Connection(u, error)

probes = [
    nengo.Probe(conn, attr="weights"),
    nengo.Probe(error),
    nengo.Probe(a.neurons, synapse=0.005)]
    nengo.Probe(b, attr='params')]
conn.learning_rule_type = GenericRule(
    regularized_pes, probes, learning_rate=1e-3, modifies='weights')
```
Option 3:
``` python
def regularized_pes(pre, post, dt, target, data):
    encoders = post.built_params.encoders
    gains = post.built_params.gain
    return (-dt / n * np.outer(
        np.dot(encoders, data) * gains,
        pre.get_input('neurons.out')) - target * reg)

pes = GenericRule(
    regularized_pes, learning_rate=1e-3, size_in=None, modifies='weights',
    filters={'pre.neurons.out': 0.005})

conn = nengo.Connection(
    a.neurons, b.neurons,
    transform=nengo.dists.Gaussian(0, 0.001),
    learning_rule_type=rule)
nengo.Connection(b, conn.learning_rule)
nengo.Connection(u, conn.learning_rule)
```

tbekolay:
I haven't been following this issue very closely, but since @drasmuss made it convenient for me to give my opinion anyhow, here's my two cents!

Of the three options I prefer Option 1 (or perhaps an option 1.5, sort of in between options 1 and 2)... not necessarily because it's the most clear or the most perfect API, but if you assume that people know how to use Nengo in general, Option 1 requires the least amount of additional learning to use.

Option 2 seems kind of clunky to me because 1) probes don't feel like they should be used that way (though I could probably be swayed on this) and 2) probes are the least well supported aspect of our API in non-reference Nengo tools (e.g. the GUI, which encourages you to not add probes to your model at all).

Option 3 could be sweet, but it feels like a specific instance of a more general idea about being more explicit about an API for frontend-backend coupling. I think this sort of API will be needed eventually, and as we build more backends it will hopefully grow organically, but I would hope that that API would be devised in concert with people that have written the backends needing that kind of API. In part, I have been sort of waiting for people writing backends to start posting issues with things they need/want, but that hasn't happened much even though we've got quite a few backends at this point.

I'm also not super convinced that slicing with `[:2]` and `[2:]` is all that error prone... those are already issues that you have to deal with in other parts of the API (e.g., writing functions across ensemble-to-ensemble connections, writing functions that are used as node outputs) so it feels weird to draw the line at this particular need for slicing. In my eyes, these are the costs of using NumPy very explicitly and giving people low-level access to what are essentially NumPy arrays (even though different backends might not represent them that way). I think that in Nengo we are making the assumption that users know NumPy, or will learn about NumPy through using Nengo, and to me that's a good thing. It reminds me of #227 and #220, which are super old issues that aim to abstract away parts of how Nengo and (to some extent) NumPy work. I think that these things could definitely be developed and be useful, but I'm not sure if core Nengo is the right place for them... it's fine with me if a higher-level API is developed that gets translated down to Nengo calls, that seems like historically how a lot of these types of advancements have been done (e.g., x86 assembler is tedious, so write C instead; C is tedious so write Python instead; etc etc).

Since I'm already ranting, there's a similar argument to be made for plotting. Aside from raster plots (which is a specific type of plot that IMO should be better supported in Matplotlib), we don't implement any plotting functions. You could make the argument that we should implement a `nengo.plot` function that takes in a list of probes and gives you a nicely formatted figure (`pandas` data frames has a plot function like this, for example). However, I would be very resistant to adding such a function because 1) everyone will want that `plot` function to work differently, meaning it will become a gigantic function with 20 optional arguments that's awful to maintain and 2) I want people to learn how to use Matplotlib since it's a super useful tool. Instead, we try to make sure that Nengo exposes the right things to make using Matplotlib easy (which I would argue we do in `Simulator.trange` and `Simulator.data`). Then if there's a need for something more friendly to beginners who aren't ready to learn lots of NumPy and Matplotlib details, it becomes a separate project that depends on Nengo (as happened with `nengo_gui`).

Sooo yeah tl;dr I like Option 1, though I'm not a big fan of passing in the params like that. I'd be inclined (for the time being) to only make it possible to use quantities in the simulation and not things like gains and biases... not very many learning rules need those things -- most people are looking to do hebbian type rules on neural activities, and PES on decoders doesn't need the gain and bias. If people really really need those things, then as a workaround they can generate those quantities themselves, set them on the appropriate objects, and use them in the learning rule function manually.

drasmuss:
Yeah I'd be fine with taking out the `params` passing thing, and we can just add that functionality back in if/when we come up with a more permanent solution for #1145.

jgosmann:
> not necessarily because it's the most clear or the most perfect API, but if you assume that people know how to use Nengo in general, Option 1 requires the least amount of additional learning to use.

I don't think ease of learning justifies the need or potential to write worse code (at least not always).

> Option 3 could be sweet, but it feels like a specific instance of a more general idea about being more explicit about an API for frontend-backend coupling. I think this sort of API will be needed eventually,

I agree, that this is a specific instance of a more general idea and that we will need such an API eventually. Especially, because related things are coming up all the time right now: generic learning rules, spaopt/deferrables, Parisien transform, ... (I'm sure I could find more going through the issues.) But it seems each time we are saying, this is just a special case, we need more examples and delay the development of such an API.

> and as we build more backends it will hopefully grow organically, but I would hope that that API would be devised in concert with people that have written the backends needing that kind of API. I have been sort of waiting for people writing backends to start posting issues with things they need/want, but that hasn't happened much even though we've got quite a few backends at this point.

I don't think, this is going to happen. The backend doesn't need the interaction with the frontend (apart from the model to build that's passed in). It's the frontend that in many instances needs information of the backend. So it is the frontend developer that have to request things they need/want from the backend developer.

> I'm also not super convinced that slicing with [:2] and [2:] is all that error prone...

Yeah, I think what annoys me more is: To understand the learning rule function, I might have to look at a different place to know what connects into `[:2]` and `[2:]` (and remember whether I'm looking for `:2` or `2:` which are visually very similar). It gets a bit better by naming those slices, but that discipline on the programmers side and I still don't know whether `activities` is pre-synaptic or post-synaptic without looking a a completely different place. Also the other way around, the connections don't give away what they are connecting to and what sort of information is supposed to be transmitted and I have to look into the learning function to figure that out (hoping the slices are named).

To me there is also a line that can be drawn to slicing with ensembles because there I explicitly want to represent two types of information in a single ensemble and it makes sense that it has to be sliced. But the error and activities for the learning rule here are pretty independent. And as mentioned earlier, I pretty much never slice node inputs/outputs (I probably wasn't even super aware that I could do that before this discussion).

> I'd be inclined (for the time being) to only make it possible to use quantities in the simulation and not things like gains and biases... not very many learning rules need those things

2 of the 3 implemented rules in the test use `params`. I'm not familiar enough with the learning rules to know whether an implementation without `params` is necessary, but right now it seems to contradict your statement.

> as a workaround they can generate those quantities themselves, set them on the appropriate objects, and use them in the learning rule function manually

Which requires calls to builder functions and it is still not clear to me are users supposed to call these? Are they part of what we consider the public Nengo API? Are they even documented?

jgosmann:
> I think this sort of API will be needed eventually, and as we build more backends it will hopefully grow organically

One more thought: This sounds to me like we're waiting for this sort of API to come into existence magically, because it is hard to design and make the right decisions. But as I said, I don't think this will just happen. I think this is an issue that needs some careful thinking and explicit decisions. Otherwise we might wait forever for this to happen.

drasmuss:
> I don't think ease of learning justifies the need or potential to write worse code (at least not always).

Of course it isn't always justified, but here the primary use-case is for novice users/rapid prototyping, where ease of use really is the primary factor.  This is not a "power user" feature--power users would just build their own learning rule using the pluggable builder system, giving them access to all the back-end features they might want.

(I would also note that "worse code" is a highly subjective statement.  You see `data[:2]` as worse code, Trevor and I don't.)

I'll restate a similar point to Trevor from a different angle, to make my perspective more clear.  The reason I prefer option 1/2 (and the motivation for why they were implemented the way they were) is that they don't introduce new "parts" into Nengo--they re-use existing parts in new ways.  Option 1 makes use of the existing system for connecting into learning rules, and Option 2 makes use of the existing probe system (albeit in a somewhat unusual way, which is why I also lean towards Option 1).  Option 3 adds entirely new systems (`post.build_params`, `pre.get_input('neurons.out')`, `filters={'pre.neurons.out': 0.005}`), that the user has never seen before and will never see elsewhere in the code.  The advantage of adding a new system is that you can customize it to work exactly how you would like for the application, but the disadvantage is that it adds extra cognitive load to anyone learning/maintaining that system.  It will always be a balancing act between those two depending on where you place more weight, but in this case I think the envisioned use-case should sway us towards the more lightweight implementation.

jgosmann:
> I would also note that "worse code" is a highly subjective statement. You see data[:2] as worse code, Trevor and I don't.

To some degree, but I suppose we can agree that everything else equal¹ self-documenting code is better than non-self-documenting code and `data[:2]` is certainly not self-documenting.

¹: This is of course not the case here.

> here the primary use-case is for novice users/rapid prototyping, where ease of use really is the primary factor.

Ease of use is equally subjective and I'm not convinced that option 1 is easier to use.

> Option 3 adds entirely new systems (post.build_params, pre.get_input('neurons.out'), filters={'pre.neurons.out': 0.005}),

`post.build_params` is also added by option 1 as `params` if we don't scrap it.

drasmuss:
Let's have a quick vote on this so we can decide which way to go forward.  We'll say

:smile: = option 1
:tada: = option 2
:heart: = option 3

jgosmann:
:+1: = option 4 (variable number of data arguments):
```python
def regularized_pes(target, error, activities, params):
    encoders = params[b].encoders
    gains = params[b].gain
    return ...

rule = GenericRule(
    regularized_pes, size_in={'error': 2, 'activities': n},
    pass_model_params=True, ...)

conn = nengo.Connection(..., learning_rule_type=rule)
nengo.Connection(b, conn.learning_rule.error)
nengo.Connection(u, conn.learning_rule.error)
nengo.Connection(a.neurons, conn.learning_rule.activities, synapse=0.005)
```

hunse:
I like Option 1, but I'm wondering if we should have any other default arguments. Namely, the last delta? This would make momentum really easy to implement. Users could store that last delta themselves, but as soon as you do you get into the same issues we had with processes, making it hard to use one learning rule function in more than one place. If we add `delta`, then I'm good with option 1 I think.

With regards to option 4, if we allow that type of syntax for `size_in` here, then I think we'd also have to add it for `Node`s. Not that that would be bad, just some more work.

Looking at option 4, it actually looks really nice. And I assume that you'd still be able to do option 1 (passing `size_in=2+n`) and split up the argument yourself. I'm trying to think if it would make things any more difficult for backends, but I can't think so. One minor concern is that the user arguments (`error` and `activities` in the example) will have to be passed as keyword arguments. This means they have to have the same names (shouldn't really be a problem), but also that with our current parameter order, `params` would also have to be a keyword argument. I might flip them around (having `delta`, `targets`, `params`, then user arguments) so that we can pass the first three positionally and don't have to worry if the user switches the name.

drasmuss:
Re: option 4, I'd make the same argument that I don't think it's worth adding new syntax to nengo just for this feature.  However, I could see a larger new feature being "allow users to specify named inputs to Nodes" (as @hunse suggests), and if we added that we could update `GenericRule` to support that as well.  But yeah I'd say that'd be a separate PR than this one.

tbekolay:
> as soon as you do you get into the same issues we had with processes, making it hard to use one learning rule function in more than one place

Perhaps we could instead let users pass in a process for the first argument if they're doing something like that? Since, like you mention, there are very similar considerations at play.

hunse:
> > as soon as you do you get into the same issues we had with processes, making it hard to use one learning rule function in more than one place
> 
> Perhaps we could instead let users pass in a process for the first argument if they're doing something like that? Since, like you mention, there are very similar considerations at play.

I don't follow. This would be so we don't have to add `delta` as another standard argument? How would this work?

tbekolay:
> This would be so we don't have to add delta as another standard argument? How would this work?

Yeah, my first thought was that if users want that they can manage the last delta state by themselves. But you're right that, in that case, things like using the rule in multiple simulators and resetting will not work as expected. So, instead of adding `delta` as a required argument for the callable, we instead make it such that the callable can be a process instance instead, like we do for nodes. So, keeping track of the delta would look something like:

```python
class UsesDelta(nengo.Process):
    def __init__(self, func):
        self.func = func  # func accepts target, data, last_delta

    def make_step(self, shape_in, shape_out, dt, rng):
        last_delta = np.zeros(shape_out)

        def call_func(target, data):
            delta = self.func(target, data, last_delta)
            last_delta[...] = delta
            return delta
```

Only thing that isn't clear to me (but I imagine is clear to @drasmuss) is what `shape_in` and `shape_out` should be up there, and if I'm right to assume that `shape_out` would be the shape of `delta`.

jgosmann:
> Users could store that last delta themselves, but as soon as you do you get into the same issues we had with processes, making it hard to use one learning rule function in more than one place.

I think options except 3 already make it hard to use one learning rule in more than one place because they explicitely reference specific ensembles (`b`). Also, having to specify the input size makes it harder to reuse because it will only work where the same dimensionalities of error and input activities are used.

> I might flip them around (having delta, targets, params, then user arguments) so that we can pass the first three positionally and don't have to worry if the user switches the name.

I'm fine with that, just want to highlight that we might not pass `params` at all? @tbekolay said only few learning rules require it, though I'm still missing support for that statement (as two of three implemented tests need it).

drasmuss:
The process idea sounds interesting, I'll try that out next week to see what it looks like

drasmuss:
Just an update on the status of this (since I just remembered it was sitting around).  I probably won't have time/motivation to do significant work on this any time soon.  So we can just mark it as "revise and resubmit" and close it since it didn't pass review, unless someone else wants to take it over.

Issue #562, opened 2014-11-27
Passing in a list to add_output of network arrays


studywolf:
It would be cool to be able to pass in a list of functions to the network array, so that a bunch of different signals could be calculated and accessed from the same output connection. 

This came up because I wanted something like `output = (x[0] + constant_i) * x[1])` where `constant_i` is different for each population `i`, but I could also see doing things like `output = [x[0]*x[1], x[0]**2*x[1],...]` or whatever. Might be nice!

tbekolay:
Done in #580.

PR #580, opened 2014-12-09
Added function list support for EA.add_output

studywolf:
Allows a list of functions to be passed in to the ensemble_array
add_output function, so that different functions can be calculated
off of the different ensembles.

Also added a test.

Addresses #562.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,7 +24,12 @@ Release History
 
 **Improvements**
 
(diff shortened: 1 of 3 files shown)


hunse:
I reduced some of the code duplication between the `if` blocks. LGTM.

tbekolay:
LGTM too. I made a few changes to the test (just cosmetic), wrote a changelog entry, and squashed the PEP8 fix (will keep @hunse's commit though); merging!

Issue #564, opened 2014-11-27
Scaling of evaluation points


jgosmann:
(not sure if this has been discussed before)

It would be convenient to be able to pass a distribution to the `eval_points` argument of `Ensemble` and `Connection`. But it would make the Nengo code a bit more complicated (but also the user's code less complicated).

My use case right now is to improve the accuracy of a multiplication with a specifically manufactured distribution of evaluation points. But it should be noted that the improvements are not enormous. At most I get an improvement of the RMSE of 16% (all other parameters except radius with their defaults), but with diagonal encoders it's just 3%. Changing the distribution of evaluation points might also improve the approximation of other non-linear functions. But so far I don't have data on that.

tbekolay:
Everything in this thread is addressed now, I think, except for @jgosmann's third suggestion: "Implement boolean parameter to disable scaling." It would be an easy thing to implement, but it would add another parameter to both Ensemble and Connection. Is this heavily sought after, or can it wait?

jgosmann:
I think, it can wait. Maybe we should think about a way to cut down constructor parameters at some point ...

jgosmann:
Just came across this issue again and I don't think we need a parameter to disable scaling. It is a pretty rare use case and there are viable workarounds. So I'm closing this and if we can create a new issue if this should come up again.

PR #595, opened 2014-12-17
Always scale ensemble evaluation points.

jgosmann:
Addresses #564.

### DIFF
--- a/nengo/tests/test_ensemble.py
+++ b/nengo/tests/test_ensemble.py
@@ -5,7 +5,7 @@ import pytest
 
 import nengo
 import nengo.utils.numpy as npext
-from nengo.dists import Choice
+from nengo.dists import Choice, UniformHypersphere
 from nengo.processes import StochasticProcess
 from nengo.utils.testing import warns, allclose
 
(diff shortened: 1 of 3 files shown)


nengo/tests/test_ensemble.py
```diff
@@ -233,6 +233,24 @@ def heuristic(neurons, dims):
     assert points.shape == (heuristic(neurons, dims), dims)
 
 
+@pytest.mark.parametrize('eval_points, radius', [
+    (UniformHypersphere(), 0.5),
+    (UniformHypersphere(), 1.),
+    (UniformHypersphere(), 1.5),
+    (UniformHypersphere().sample(500, 1), 0.5),
+    (UniformHypersphere().sample(500, 1), 1.),
+    (UniformHypersphere().sample(500, 1), 1.5)])
```
jgosmann: Yeah, that would be better.

hunse:
Okay, got this rebased and cleaned up that test (it's also deterministic, now). @tbekolay, want to give it a quick check and merge?

Also, this kind of pre-empts #596 now, since Connection uses the same eval point generation function as ensemble so it has to have the same scaling.

tbekolay:
You beat me to it! Giving it a look-see now.

Re #596, the only additional stuff in that PR is the test for connection eval_points. Should we still include it? I could just squash it into this PR so it's one commit. Thoughts?

hunse:
Yeah, maybe just put it into this PR. I think the test is less necessary now, since we're using the same function for both Ensemble and Connection eval points, so it's hard to have different scaling as we did before, but we might as well keep the test.

PR #596, opened 2014-12-17
Scale connection eval points by ensemble radius.

jgosmann:
Addresses #564. The discussion in that issue and whether a consensus was reached before merging.

### DIFF
--- a/nengo/tests/test_connection.py
+++ b/nengo/tests/test_connection.py
@@ -5,6 +5,7 @@ import pytest
 
 import nengo
 from nengo.connection import ConnectionSolverParam
+from nengo.dists import UniformHypersphere
 from nengo.solvers import LstsqL2
 from nengo.utils.functions import piecewise
 from nengo.utils.testing import allclose
@@ -609,6 +610,23 @@ def test_set_eval_points(Simulator):
             nengo.Connection(a.neurons, b, eval_points=[[0, 0], [0.5, 1]])
 
     Simulator(model)  # Builds fine
+
+
+@pytest.mark.parametrize('eval_points, radius', [
+    (UniformHypersphere().sample(500, 1), 0.5),
+    (UniformHypersphere().sample(500, 1), 1.),
+    (UniformHypersphere().sample(500, 1), 1.5)])
+def test_eval_points_scaling(Simulator, eval_points, radius, seed):
+    model = nengo.Network(seed=seed)
+    with model:
+        a = nengo.Ensemble(1, 1, radius=radius)
+        b = nengo.Ensemble(1, 1)
+        con = nengo.Connection(a, b, eval_points=eval_points)
+
+    sim = Simulator(model)
+    dists = np.linalg.norm(sim.data[con].eval_points, axis=1)
+    assert np.all(dists <= radius)
+    assert np.any(dists >= 0.9 * radius)
 
 
 def test_solverparam():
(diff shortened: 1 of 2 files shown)


nengo/tests/test_connection.py
```diff
@@ -611,6 +612,23 @@ def test_set_eval_points(Simulator):
...
+    model = nengo.Network(seed=seed)
+    with model:
+        a = nengo.Ensemble(1, 1, radius=radius)
+        b = nengo.Ensemble(1, 1)
+        con = nengo.Connection(a, b, eval_points=eval_points)
+
+    sim = Simulator(model)
+    dists = np.linalg.norm(sim.data[con].eval_points, axis=1)
+    assert np.all(dists <= radius)
+    assert np.any(dists >= 0.9 * radius)
```
hunse: I'm not sure if we need this test, since it's the same as for Ensemble in #595. I realize that you put it in because there was a difference between Ensemble and Connection that shouldn't have been there, but sometimes I feel like we get into the habit of finding a bug, fixing the bug, then adding a test that tests whether that specific bug is fixed. This means we end up with a lot of tests, and I'm not really sure how useful it is since there's no way we could test every possible bug, and we really want our tests to highlight bugs for us, not the other way around.

But all that is a general comment on how we do tests sometimes; sorry to take it all out on this test!

jgosmann: I usually try to write the test before fixing the bug to have some automated method to reproduce the bug. Only in the second step I fix the bug. The test will also ensure that a bug is not reintroduced (it was overlooked once, so it might be overlooked again).

Of course it is impossible to test everything, but I'm not sure if it is a good idea to throw away tests that have already been written for some reason. As long as the tests do not hinder refactoring, having many tests doesn't seem to be a problem to me. Maybe we should think about a better organization instead of deleting tests?

tbekolay: My 2 cents: regression tests are useful, as long as they don't add undue length to the test suite. At the same time, we have a lot of overlapping tests at this point, and it would be good to reduce its length, or separate them into those tests that we should run all the time, and those that we only need to run before releases, etc. Issues for another PR though for sure.

tbekolay:
I added a commit to shorten the test a bit, and make it deterministic (as the UniformHypersphere calls weren't seeded before). @jgosmann, if it looks good to you I'll squash and merge.

nengo/tests/test_connection.py
```diff
@@ -611,6 +612,21 @@ def test_set_eval_points(Simulator):
     Simulator(model)  # Builds fine
 
 
+@pytest.mark.parametrize('radius', [0.5, 1., 1.5])
+def test_eval_points_scaling(Simulator, radius, seed, rng):
+    eval_points = UniformHypersphere().sample(500, 1, rng=rng)
```
hunse: Can Connection not take a Distribution for `eval_points`. Just glancing at the builder code, it looked like maybe it can't. We should probably fix that. And if and when it can, we should probably have this test give both a distribution and a bunch of points, just like the test for Ensemble.

tbekolay: It can't, but this has been brought up about 5 times so far so let me fix that.

tbekolay: Done in #612; will change this test when that's merged.

tbekolay:
Incorporated into #595.

PR #598, opened 2014-12-17
Vagrant Windows box

jgosmann:
I set up a Vagrant Windows 7 box for easy testing. [I wrote a wiki page on it.](https://github.com/nengo/nengo/wiki/Running-unit-tests) (Not sure if that's the right place; we might move that documentation somewhere else. It probably needs some editing, too.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,8 @@ log.txt
 .ipynb_checkpoints/
 .cache
 .tox
+.vagrant
+wintest.sh
+Vagrantfile
 
 *.class


tbekolay:
This is awesome! I haven't tested it yet, but will do so soon.

I'm wondering if this is the right place for it, or if it should be in [ctn-waterloo/devscripts](https://github.com/ctn-waterloo/devscripts). Mostly, my concern is that (with the password especially) it's only accessible by those in the CTN anyway, so probably it shouldn't live in a (relatively) high-profile repo. Also, it seems like the VM would be very nice for testing all sorts of models in Windows, especially if it has Nengo 1.4 and some other simulators installed; tying it to Nengo 2 doesn't quite do it justice!

hunse:
I agree, devscripts sounds like a good place for this.

jgosmann:
Yeah, devscripts sounds like a good place. Might make syncing of the data a bit more complicated.

tbekolay:
True... Is there no way to configure the folders to sync? Other than editing the Vagrantfile that is.

jgosmann:
I think the only way is to edit the Vagrantfile. We could put it into devscripts and have everyone who wants to use it copy (or symlink) it their nengo project directory.

jgosmann:
Moved the `Vagrantfile` and `wintest.sh` to ctn-waterloo/devscripts#3. Changed this PR to only add those files to `.gitignore` to prevent them from accidentally being committed.

tbekolay:
Makes sense! I'll merge this in now.

PR #601, opened 2014-12-17
allows empty rasters

celiasmith:
this is useful for generating animations

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,16 @@ Release History
   (`#626 <https://github.com/nengo/nengo/issues/626>`_,
   `#628 <https://github.com/nengo/nengo/pull/628>`_)
 
+**Improvements**
+
+- ``rasterplot`` works with 0 neurons, and generates much smaller PDFs.
+  (`#601 <https://github.com/nengo/nengo/pull/601>`_)
+
+**Bug fixes**
+
+- Fix compatibility with NumPy 1.6.
+  (`#627 <https://github.com/nengo/nengo/pull/627>`_)
+
 2.0.0 (January 15, 2015)
 ========================
 
--- a/nengo/utils/matplotlib.py
+++ b/nengo/utils/matplotlib.py
@@ -42,7 +42,7 @@ def implot(plt, x, y, Z, ax=None, colorbar=True, **kwargs):
         plt.colorbar(image, ax=ax)
 
 
-def rasterplot(time, spikes, ax=None, **kwargs):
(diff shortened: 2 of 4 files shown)


tbekolay:
This looks great! I'm curious if we shouldn't bother with `eventplot`? My reasoning here is that running `test_rasterplot` generates a 15KB PDF without eventplot, and 3.1MB with eventplot. My understanding of why this happens is because `plot` uses [BLITting](https://en.wikipedia.org/wiki/Bit_blit), while `eventplot` does not.

Alternatively, we could use `eventplot` but rasterize it, which reduces the PDF size to 84KB, but (since it's rasterized) doesn't look as nice as the `plot` version when you zoom in closely.

hunse:
The difficulty with plot is choosing a good marker size. This isn't too big of a problem if you know your figure size and number of neurons in advance (such as when making figures for a paper), since you can just play around with the marker size until it looks good. The problem occurs when you want to be able to resize the figure, since there's no way to tie the marker size to the axes (data), so for example if you make the figure larger then the markers will be too small by comparison.

But I didn't realize that the difference in file sizes was so large, and taking out eventplot would make the code easier. Another option is to turn eventplot off by default, but still have it there if people really want it.

tbekolay:
I like the idea of plot by default, and eventplot still available if desired. I'll try implementing a heuristic marker size based on the figure size and the number of neurons at the point when `rasterplot` is called, but yeah, if people change the figure size after the fact we can't do much (though, it is possible to implement a figure resize handler, it'd just be Matplotlib backend-specific).

hunse:
I added a heuristic for the marker size that seems to work pretty well.

tbekolay:
Great! Let me test it out a bit then I'll merge.

tbekolay:
Added some commits; take a look @hunse and if it looks good I'll merge.

hunse:
LGTM, but this is getting tripped up on the Mock plot objects used to skip plotting in the examples.

tbekolay:
Oh right! Forgot about that. Will fix tomorrow.

tbekolay:
Hacked together a fix in c9ae1ef. Look good @hunse or too much of a hack?

hunse:
I was trying to think if there's a decent comment you could put there to explain it, but I can't think of anything.

I think that hack will be fine for the time being. It would be nicer if we could somehow completely skip things like `rasterplot` when we're using the mock plotter, but I don't know an easy way to do that, so we can just patch these kinds of problems as they crop up.

Issue #594, opened 2014-12-17
utils.neurons.rates_isi test failing


tbekolay:
![utils test_neurons test_rates rates_isi](https://cloud.githubusercontent.com/assets/203709/5474861/726001f6-85e2-11e4-9668-a3852a8b65cf.png)

It appears as though it has an issue calculating the rates based on ISI at the very end of the simulation -- I didn't look at the `rates_isi` code in detail, but perhaps this is because those neurons aren't spiking for quite a while at the end?

tbekolay:
Fixed in #602

PR #602, opened 2014-12-17
Fixed bug in `rates_isi`

hunse:
As shown in #594. The function wasn't using the time of
the last spike, causing the rate before the last spike to be
stretched.


tbekolay:
`test_rates_isi` now passes for me!

PR #604, opened 2014-12-18
Fix SPA syntax warnings. Fixes #472.

jgosmann:
This is the minimal change set to fix the SPA syntax warnings by using a config.

I had to add a `similarity` function to `SPA` because the config object is only known to the `SPA` object. Thus, we have now two similarity functions (the other one in `spa.utils`). Not sure if we want to keep it that way or unify them somehow.

Also, we might add some syntactic sugar for accessing the vocabulary. Right now one has to do something like `model.config[probe.target].vocab` which does not seem very straight forward. Before it was just `probe.target.vocab`.

PR #605, opened 2014-12-19
Missing spa examples

jgosmann:
Depends on PR #604 (branch `fix-spa-warnings`).


hunse:
I took a look at these, and I think they need a bit of extra discussion, especially of the results. Right now, they're pretty much all code. I've started adding some discussion for `convolution.ipynb`.

jgosmann:
Also added a few lines to the other two examples. Probably not perfect, but a start.

tbekolay:
What's the status of this PR ATM? I think this and the docs is all that's left.

hunse:
I think it's pretty much good to go. The only thing that's not done is adding labels to plot axes, but that's probably not essential.

tbekolay:
Labels schmabels! I'll review / merge this.

tbekolay:
Added a quick blurb to the start of `spa_parser.ipynb`, which was the only example without a lot of exposition, and added a bunch of these to the optional set of notebooks, as these take a while to simulate. Aside from that looks good; merging!

PR #610, opened 2015-01-07
Handle `tau_ref == 0` in LIF neurons

### LINKED ISSUE #609: LIF neuron with tau_ref=0
I was playing around with tuning curve shapes and found that the LIF neuron model does not handle tau_ref=0 well.

    import nengo
    net = nengo.Network()
    with net:
        ens = nengo.Ensemble(10, 1, neuron_type=nengo.LIF(tau_ref=0))
    sim = nengo.Simulator(net)
    sim.run(1)

Throws

    ZeroDivisionError: float division by zero

I suppose the model should either be able to handle this case or it should be documented and caught earlier.

hunse:
Fixes #609. The issue was minor; it was only because we weren't careful when validating `max_rates` relative to `tau_ref`.

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -175,7 +175,7 @@ class LIFRate(NeuronType):
         intercepts : list of floats
             X-intercepts of neurons.
         """
-        inv_tau_ref = 1. / self.tau_ref
+        inv_tau_ref = 1. / self.tau_ref if self.tau_ref > 0 else np.inf
         if (max_rates > inv_tau_ref).any():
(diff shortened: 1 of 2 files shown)


tbekolay:
Rebased and made a quick test for the zero tau_ref. I know, it goes against the other thread, but I think for bugs it's good to ensure we don't regress back to unfixing the bug. It can be merged into another test later on.

@hunse, if it looks good I'll squash and merge.

hunse:
Looks good.

PR #611, opened 2015-01-07
Parameters outside the config system

hunse:
This build fixes up Parameters so they can be used outside the config system, and puts them to work in NeuronType and LearningRuleType so that we can validate the parameters on these objects.

The central part of this PR is the `Unconfigurable` type, which is used as a value for the Parameter default to indicate that the Parameter has no default and cannot be set with the config system. If you try to use the parameter before it's been set, you get an error. The second commit makes `Unconfigurable` the default value for Parameter defaults, which doesn't break any existing code since any parameter that is configurable needs a default.

### DIFF
--- a/nengo/learning_rules.py
+++ b/nengo/learning_rules.py
@@ -1,7 +1,16 @@
 import warnings
 
-from nengo.params import Parameter
+from nengo.base import NengoObjectParam
+from nengo.params import Parameter, NumberParam, Unconfigurable
 from nengo.utils.compat import is_iterable, itervalues
+
+
+class ConnectionParam(NengoObjectParam):
+    def validate(self, instance, conn):
+        from nengo.connection import Connection
+        if not isinstance(conn, Connection):
+            raise ValueError("'%s' is not a Connection" % conn)
+        super(ConnectionParam, self).validate(instance, conn)
 
 
 class LearningRuleType(object):
@@ -11,6 +20,7 @@ class LearningRuleType(object):
     the Connection on which you want to do learning.
     """
 
+    learning_rate = NumberParam(default=Unconfigurable, low=0, low_open=True)
     probeable = []
 
     def __init__(self, learning_rate=1e-6):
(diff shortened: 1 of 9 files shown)


nengo/learning_rules.py
```diff
@@ -11,6 +20,7 @@ class LearningRuleType(object):
     the Connection on which you want to do learning.
     """
 
+    tau_ref = NumberParam(low=0, low_open=True)
```
tbekolay: Is this supposed to be here?

tbekolay: Or should it be `learning_rate`?

hunse: Should be learning_rate.

nengo/learning_rules.py
```diff
@@ -76,6 +87,10 @@ class BCM(LearningRuleType):
     TODO
     """
 
+    pre_tau = NumberParam(low=0, low_open=True)
+    post_tau = NumberParam(low=0, low_open=True)
+    theta_tau = NumberParam(low=0, low_open=True)
+    learning_rate = NumberParam(low=0, low_open=True)
```
tbekolay: Should be on superclass?

nengo/learning_rules.py
```diff
@@ -109,6 +124,10 @@ class Oja(LearningRuleType):
     TODO
     """
 
+    pre_tau = NumberParam(low=0, low_open=True)
+    post_tau = NumberParam(low=0, low_open=True)
+    beta = NumberParam(low=0)
+    learning_rate = NumberParam(low=0, low_open=True)
```
tbekolay: Should be on superclass?

hunse:
Hmm, this isn't playing nicely with the new `Config.all_defaults`. @tbekolay, go ahead and fix this if you want; I've got meetings until at least 4pm, but I'll try to fix it then if you haven't gotten to it.

tbekolay:
I'll give it a look!

nengo/learning_rules.py
hunse: Yes

nengo/learning_rules.py
hunse: Yes

tbekolay:
'Twas a quick fix; just explicitly skipping unconfigurables in `all_defaults` (which makes sense anyway, as their default can't be set). Also fixed the learning rule stuff. Merging!

PR #612, opened 2015-01-08
Connections can take dists for eval_points

tbekolay:
As has been requested a number of times in the past while :rainbow: I only tested that it builds (in `test_connection.test_set_eval_points`); other PRs are in the queue that will test it functionally.

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -4,6 +4,7 @@ import numpy as np
 
 import nengo.utils.numpy as npext
 from nengo.builder.builder import Builder
+from nengo.builder.ensemble import gen_eval_points
 from nengo.builder.node import build_pyfunc
 from nengo.builder.operator import DotInc, ElementwiseInc, PreserveValue, Reset
 from nengo.builder.signal import Signal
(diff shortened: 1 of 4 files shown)


nengo/builder/ensemble.py
```diff
@@ -24,25 +24,29 @@ def sample(dist, n_samples, rng):
...
 
 
-@Builder.register(Ensemble)  # noqa: C901
-def build_ensemble(model, ens):
-    # Create random number generator
-    rng = np.random.RandomState(model.seeds[ens])
-
-    # Generate eval points
-    if isinstance(ens.eval_points, Distribution):
+def gen_eval_points(ens, eval_points, rng):
```
hunse: Since the only thing we use from `ens` is `n_eval_points`, I'd just pass that straight in.

tbekolay: We use `dimensions` and `radius` too.

nengo/builder/connection.py
```diff
@@ -19,17 +20,16 @@
...
     bias = model.params[conn.pre_obj].bias
 
-    eval_points = conn.eval_points
-    if eval_points is None:
+    if conn.eval_points is None:
         eval_points = npext.array(
             model.params[conn.pre_obj].eval_points, min_dims=2)
     else:
-        eval_points = npext.array(eval_points, min_dims=2)
+        eval_points = gen_eval_points(conn.pre_obj, conn.eval_points, rng)
```
hunse: If `conn.pre_obj` is not an ensemble, this will give a weird error when we try to get `n_eval_points` from it. Since we'll never need eval points if we don't have an ensemble, I'd put in something like `conn.pre_obj.n_eval_points if isinstance(conn.pre_obj, Ensemble) else 0'

nengo/builder/ensemble.py
hunse: That we do! nm

nengo/builder/connection.py
hunse: Oh, never mind on this one, too, since it looks like we just call this whole function if `conn.pre_obj` is an ensemble.

hunse:
LGTM

nengo/builder/connection.py
tbekolay: Hmm... `Connection.EvalPointsParam` already checks that `pre` must be an `Ensemble`, so it won't allow it to be anything but `None`, I think, unless you try really really hard.

Issue #623, opened 2015-01-15
radius breaks target_function


fragapanagos:
Setting the radius to something other than 1.0 breaks target_function.

    import nengo
    from nengo.utils.connection import target_function
    import numpy as np

    x = np.linspace(-1, 1, 10)
    model = nengo.Network()
    with model:
        # a = nengo.Ensemble(10, 1, radius=1.0)  # will work
        a = nengo.Ensemble(10, 1, radius=2.0)  # will break
        b = nengo.Ensemble(10, 1)
        nengo.Connection(a, b, **target_function(x, x))

    sim = nengo.Simulator(model)

produces

    KeyError: (-2.0,)

because the eval_points passed into the connection are scaled by the radius [here](https://github.com/nengo/nengo/blob/master/nengo/builder/ensemble.py#L40). When the scaled eval_points are passed to the function generated by target_function, they are not necessarily in its func_dict.

tcstewar:
Yikes!  That's a pretty major bug to find.  Thank you for posting it!

tcstewar:
Any ideas what to do about this?  This is an unexpected sideeffect of scaling the radius...  for now we might need to just pass in the radius to ```target_function(x, x, radius=2.0)``` but that's rather awkward-looking.  I don't see any other way of dealing with it though.

tcstewar:
Here's another option: we add the scale_eval_point_by_radius parameter that was discussed in #564 and have that be one of the things returned from target_function (along with eval_points and function).

Note that passing in the radius the way I suggested above is problematic for floating point rounding reasons:  we can't just divide by the radius in target_function and allow the builder to multiply by the radius and expect that we'll get exactly the same result.

tcstewar:
I just tried a minimal fix for this by adding a scale_eval_points parameter to Connection.  Not sure at all that this is the right way to go, but I don't see any other options.

tcstewar:
(Btw, @fragapanagos , if you want to try out the fixed version, switch to the ```scaling_taget_function``` branch.  Let me know if that doesn't work for you!)

fragapanagos:
@tcstewar The fixed version seems to be working.

tcstewar:
> The fixed version seems to be working.

Cool.  No idea whether that'll be the eventual fix, but it should let you work with things for now...

PR #624, opened 2015-01-15
Fixing scaling target function

tcstewar:
Here's a quick fix for #623.  This adds a ```scale_eval_points``` to Connection, as discussed in #564 but at the time we didn't think there was a pressing need to add it, so we were holding off.  However, without this feature I don't see a nice way to use ```nengo.utils.target_function``` on an pre-Ensemble with a radius that isn't 1.0. (issue #623).

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -29,7 +29,8 @@ def build_linear_system(model, conn, rng):
         eval_points = npext.array(
             model.params[conn.pre_obj].eval_points, min_dims=2)
     else:
-        eval_points = gen_eval_points(conn.pre_obj, conn.eval_points, rng)
+        eval_points = gen_eval_points(
+            conn.pre_obj, conn.eval_points, rng, conn.scale_eval_points)
 
     x = np.dot(eval_points, encoders.T / conn.pre_obj.radius)
     activities = conn.pre_obj.neuron_type.rates(x, gain, bias)
--- a/nengo/builder/ensemble.py
+++ b/nengo/builder/ensemble.py
@@ -24,7 +24,7 @@ def sample(dist, n_samples, rng):
     return np.array(dist)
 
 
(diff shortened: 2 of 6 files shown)


tcstewar:
(btw, I didn't meant to rush this PR without discussion -- I just tossed together a quick temporary fix for the person who posted the original issue, and then realized that, hmm, actually, that might be the correct fix after all, so I posted it as a PR.  If anyone has any nagging doubts about it, please say so!  )

tcstewar:
> I think this file was included accidentally; I'll strip it out during the merge.

Argh, damnit, yes, that has nothing to do with anything other than me screwing up my gitting.  Again.  Sorry about that...

tbekolay:
> I didn't meant to rush this PR without discussion

Nah don't worry about it; I was hesitant when reading the discussion, but looking at the changes it seems pretty reasonable. I'll give it another once over and merge after #605!

drasmuss:
I just hate it whenever I see another parameter being added to our core objects, especially a pretty niche one like this.  In #564 we decided to scale the eval points because it seemed like it'd be easier, now we have to introduce a new parameter to fix the issues that introduced.  It just feels :crying_cat_face:.  But that being said, we have this PR here, and no PR for the other solution, so I guess that's the most relevant argument.

tcstewar:
> I just hate it whenever I see another parameter being added to our core objects, especially a pretty niche one like this. In #564 we decided to scale the eval points because it seemed like it'd be easier, now we have to introduce a new parameter to fix the issues that introduced. It just feels . But that being said, we have this PR here, and no PR for the other solution, so I guess that's the most relevant argument.

I am so feeling exactly that way right now.  It feels to me like it's indicating there's something slightly wrong in the underlying design, but I don't see what it is.  Still, that's what 2.1 is for....  :)

PR #627, opened 2015-01-20
Fix usage of `astype` for NumPy 1.6

tbekolay:
The `copy` kwarg was added to NumPy in 1.7. Since we depend on `numpy>=1.6`, we shouldn't use it. This uses `asarray` instead.

### DIFF
--- a/nengo/builder/ensemble.py
+++ b/nengo/builder/ensemble.py
@@ -58,9 +58,8 @@ def build_ensemble(model, ens):
     if isinstance(ens.neuron_type, Direct):
         encoders = np.identity(ens.dimensions)
     elif isinstance(ens.encoders, Distribution):
-        encoders = ens.encoders.sample(
-            ens.n_neurons, ens.dimensions, rng=rng).astype(
-            np.float64, copy=False)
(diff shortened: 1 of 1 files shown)


tbekolay:
Confirmed that this allows Nengo to work on @cchan987's machine, which uses NumPy 1.6.1 :+1:

jgosmann:
LGTM, but we should fix it probably in two other places too: [here](https://github.com/nengo/nengo/blob/04e53cc56f7b1852f2bdddbd8a9ee7fc809b4038/nengo/tests/test_solvers.py#L346) and [here](https://github.com/nengo/nengo/blob/04e53cc56f7b1852f2bdddbd8a9ee7fc809b4038/nengo/tests/test_solvers.py#L409)

hunse:
@jgosmann: I'm not too worried about those two places, because they don't use `copy=False` so they won't break on Numpy < 1.7, and they're in tests where I'm not worried about the extra copy of a small array.

tbekolay:
These don't use the `copy` kwarg; the function existed in 1.6, it just didn't have a `copy` kwarg until in 1.7.

jgosmann:
Oh, right. I somewhat missed the part that it is only the `copy` arg. :eyeglasses:

Issue #626, opened 2015-01-19
Potentially confusing Node behaviour


drasmuss:
Pulling a Travis here and submitting bugs in my code as Nengo bugs, but just spent a little while tracking this down and thought I should document it.  Here's a minimal example:
```python
with nengo.Network() as net:
    class Test:
        tmp = 0.0
        
        def __call__(self, t):
            if t == 0.5:
                self.tmp = t
            print self.tmp
            
    a = nengo.Node(output=Test())
    
sim = nengo.Simulator(net, dt=0.25)
sim.run(1.0)
```
I think most people would expect that to print out "0.0, 0.0, 0.5, 0.5, 0.5, ...", but in reality it prints out "0.0, 0.0, 0.5, 0.75, 1.0, ...".  Clearly (in hindsight) the problem is that `t` is an ndarray, not a float, so `tmp` and `t` get aliased.  But the problem is that `t` looks and acts, in basically all other respects, as if it is a float (e.g., `t == 0.5` passes, `t[0]` fails, etc.).

Potential solutions
1) cast `t` to a float before passing it to the `Node` function (cleanest solution, but pay the cost of type cast)
2) document this clearly in `Node` (hope people read it)
3) this is desired behaviour, make fun of Daniel for making CS 101 errors

jgosmann:
4) Pass in a copy of `t` as ndarray. (1 might be faster, but not sure)

tcstewar:
Cute.  I've run into the same problem with the ```x``` value being passed in.  This also give initially unexpected behaviour for the same reason:

```python
data = []
def node_function(t, x):
    data.append(x)
node = nengo.Node(node_function, size_in=1)
```

I've gotten into the habit of doing this instead:

```python
data = []
def node_function(t, x):
    data.append(np.copy(x))
node = nengo.Node(node_function, size_in=1)
```

Not sure what to do about it in general.  I still forget and get annoyed by this now and then.  My instinct is that the overhead involved in doing (1) or (4) is pretty high, but it's something that probably should be benchmarked and checked.   If it is actually a significant cost, then my vote is (2).

hunse:
I would probably have an option on the node whether to copy `t` and `x` or not (we could have one option for each, but I think one for both is fine). This would be settable with the config system, and default to copying. That way, people really concerned about speed can turn it off. I'm guessing the time cost will be relatively small in the grand scheme of things.

tcstewar:
[...]

PR #628, opened 2015-01-21
Node functions get t as float, x as readonly

hunse:
Node functions now receive `t` as a float (as opposed to a
zero-dimensional array), and `x` as a readonly view onto the
simulator data. This addresses #626.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,20 +7,27 @@ Release History
    version (release date)
    ======================
 
-   ** section **
+   **section**
 
    - One-line description of change (link to Github issue/PR)
 
 .. Changes should be organized in one of several sections:
 
-   - API Changes
+   - API changes
    - Improvements
-   - Behavioural Changes
+   - Behavioural changes
    - Bugfixes
    - Documentation
 
 2.0.1 (unreleased)
 ==================
+
+**Behavioural changes**
+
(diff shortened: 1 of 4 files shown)


jgosmann:
LGTM :cake:

tbekolay:
Me too; merging!

tbekolay:
Amended in a changelog entry.

My thought process (which I want to make sure matches other people's) is that every PR should have a corresponding changelog entry; there's a bit of worry that we're duplicating info that is in the commit messages, but changelog entries should always be one line, with a pointer to the issue / PR that introduced the fix / feature. That way the bulk of the info (if people care) is in the commit message / issue / PR discussion, while the changelog makes sure that we don't have to scramble to say what's new in each version. Any objections to that?

hunse:
No, that sounds like a good idea. I'll try to remember to put those in. I think it's also not necessarily a bad thing to do them at the end of a PR (when it's clear what the actual changes are) as opposed to the beginning, but the best is probably to put something in at the beginning and then edit it as need be.

tbekolay:
Yeah, actually this is perhaps more appropriately a job for the merger rather than the author (though the author should feel free to add one if they choose). But just wanted to make sure tracking these is OK!

hunse:
Yeah, it's nice to have it all in one place. An additional benefit is that a `git blame` on the changelog file should always be able to get you to (near) where a particular change was made.

Issue #603, opened 2014-12-18
Optional decoder solver tests fail without scikit learn


jgosmann:
I think, they should be skipped automatically instead of failing. Maybe with some kind of message if that is possible.

```
____________________________________________________________ test_decoder_solver_extra[Solver0] _____________________________________________________________

Solver = <class 'nengo.solvers.LstsqL1'>, plt = <nengo.utils.testing.Mock object at 0x355f0110>, rng = <mtrand.RandomState object at 0x355f0ad0>

    @pytest.mark.optional
    @pytest.mark.parametrize('Solver', [LstsqL1])
    def test_decoder_solver_extra(Solver, plt, rng):
>       test_decoder_solver(Solver, plt, rng)

nengo/tests/test_solvers.py:126: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
nengo/tests/test_solvers.py:88: in test_decoder_solver
    D, _ = Solver()(Atrain, train, rng=rng)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <nengo.solvers.LstsqL1 object at 0x355f0a90>, weights = False, l1 = 0.0001, l2 = 1e-06

    def __init__(self, weights=False, l1=1e-4, l2=1e-6):
        """
            weights : boolean, optional
                If false solve for decoders (default), otherwise solve for weights.
            l1 : float, optional
                Amount of L1 regularization.
            l2 : float, optional
                Amount of L2 regularization.
            """
>       import sklearn.linear_model  # noqa F401, import to check existence
E       ImportError: No module named sklearn.linear_model

nengo/solvers.py:394: ImportError
____________________________________________________________________ test_subsolvers_L1 _____________________________________________________________________

rng = <mtrand.RandomState object at 0x355e0090>

    @pytest.mark.benchmark
[...]

PR #629, opened 2015-01-22
Pytest decorator improvements

jgosmann:
This addresses #603. With this PR there shouldn't be any test failures because of uninstalled dependencies anymore, instead those tests are automatically skipped. The reason for the skip can be displayed (including the missing module name) with the `-rs` option to pytest. I think, this approach is preferrable as a failing test should always indicate "something is wrong" and not "maybe something is wrong, but we don't know because of a missing dependency". Because skipped tests are indicated in it's own category, it is still clear that not all tests ran and some of those might fail.

With this the `optional` mark should not be used for tests with special dependencies anymore. I think, at the moment it is only used for slow tests and it seems the same is true for `benchmark`. Maybe we should get rid of one of those? Maybe the remaining mark should be renamed to `slow` as it is more descriptive? (I might want to use `benchmark` for some other benchmarking related pytest functionality I am planning to try to implement.)

Note: I assume `matplotlib` to be a depency for running the test and did not mark all the tests which fail without it.

### DIFF
--- a/nengo/spa/tests/test_thalamus.py
+++ b/nengo/spa/tests/test_thalamus.py
@@ -6,7 +6,7 @@ from nengo import spa
 import numpy as np
 
 
-@pytest.mark.optional  # Too slow
+@pytest.mark.slow
 def test_thalamus(Simulator, plt, seed):
     model = spa.SPA(seed=seed)
 
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -113,17 +113,20 @@ def pytest_generate_tests(metafunc):
 def pytest_addoption(parser):
     parser.addoption('--benchmarks', action='store_true', default=False,
                      help='Also run benchmarking tests')
+    parser.addoption('--plots', action='store_true', default=False,
+                     help='Also run plotting tests')
     parser.addoption('--noexamples', action='store_false', default=True,
                      help='Do not run examples')
     parser.addoption(
-        '--optional', action='store_true', default=False,
-        help='Also run optional tests that may use optional packages')
+        '--slow', action='store_true', default=False,
+        help='Also run slow tests.')
 
 
 def pytest_runtest_setup(item):
     for mark, option, message in [
             ('benchmark', 'benchmarks', "benchmarks not requested"),
             ('example', 'noexamples', "examples not requested"),
-            ('optional', 'optional', "optional tests not requested")]:
+            ('plot', 'plots', "plots not requested"),
+            ('slow', 'slow', "slow tests not requested")]:
         if getattr(item.obj, mark, None) and not item.config.getvalue(option):
             pytest.skip(message)
--- a/nengo/tests/test_examples.py
+++ b/nengo/tests/test_examples.py
@@ -21,8 +21,8 @@ from nengo.utils.stdlib import execfile
 def pytest_generate_tests(metafunc):
     examples = glob('%s/*.ipynb' % examples_dir)
 
-    # if `--optional` is not set, filter out time-consuming notebooks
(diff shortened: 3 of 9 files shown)


tbekolay:
Notes from dev meeting:

- `--optional` should be changed to `--slow` (for tests that are actually slow) and tests marked `benchmark` should be reviewed to ensure that they're actually benchmarks (i.e., tests that aren't primarily for asserting things, but for testing performance of something and recording it)

jgosmann:
There where exactly two tests marked as optional. One of those is now marked as slow, the other one just produces plots and I marked that one as plot instead. This should play well with the changes in PR #647.

Two of the tests marked as benchmark are indeed a benchmark (one for the cache, one for firing rates of neurons). All other tests marked as benchmark are solver tests which seem to be indeed benchmarks (though some of them have assertions and some might be better marked as `plot`?)

jgosmann:
I would leave the solver marks as they are for now and we might revisit that issue after merging #647 too. After that is merged we might want to rewrite some of the benchmarks anyway.

hunse:
> the other one just produces plots and I marked that one as plot instead

It was my impression from the dev meeting that we were going to use the `--plot` argument on `py.test` to replace `NENGO_TEST_PLOT`. We could also have a `plot` mark, but I think this would get confusing. I would mark the rasterplot test as a benchmark.

hunse:
Though I guess it might be annoying to have to type `--plot --benchmark` to run that test. I guess we could assign "double functionality" to `--plot`, since the intent is similar. There could be some benchmarks which only output plots, though, and I was thinking it would be nice to have one mark for all tests that don't assert anything. But maybe we actually want multiple categories of tests that don't assert anything. Thoughts?

jgosmann:
Not sure if the following is possible, but if it is it might be a good way to handle things:

* One mark for slow tests (will be skipped per default)
* One mark for tests without assertions
  * If the `--plot` flag is set and the test uses the `plt` fixture, the test will run.
  * If the `--benchmark` flag is set and the test uses the `analytics` fixture, the test will run.
  * If some other `--flag` is set, the test will run. This allows to ensure no exceptions are raised. But I would be fine not adding such a flag.
  * Otherwise the test won't run.

That would keep the number of marks to a minimum and tests will only run if they assert something or have an effect (creating a plot, producing some data).

Maybe there should be another flag to skip the normal tests so that it is possible to only create the plots or benchmark data?

jgosmann:
Fixed some stuff and squashed the commits.

tbekolay:
I'll take a look at this shortly! One quick clarification from that last post: all tests will still run even if `--plot` isn't specified right? I think that only the `rasterplot` test is solely a test of plotting, every other test that plots asserts stuff and does plotting mainly as a way to debug failures.

jgosmann:
Most of the changes I proposed in the comment above are not in this pull request (but will be in the #647 PR). To summarize this PR:

* Use `importorskip`, `slow`, `plot` instead of `optional` where appropriate.
* Check for `jinja2` before running examples as it is required.

The only test marked as `plot` is for the rasterplot. All other tests will run without the `--plots` flag depending on the `importorskip` result and the `--slow` flag. There is also the `--benchmarks` flag with an unchanged effect.

The rational of this PR is to get rid of `optional` in favor of more expressive marks. Further changes to this are in PR #647.

tbekolay:
Solid! Reviewing now.

tbekolay:
Added a commit to ignore some new PEP8 errors (new as in newly added to the `pep8` project, not new as in things introduced in this PR). Can address these errors more thoroughly later if desired, but not worth stopping this PR. Merging!

PR #633, opened 2015-01-26
NeuronType functions handle array_like

hunse:
- Allow `NeuronType.rates` and `NeuronType.gain_bias` to handle
  all array_like arguments, including single numbers, lists, and
  arrays.
- To facilitate this, I added the `array_like` decorator that marks
  function arguments to be passed through `np.array` (with
  `copy=False` by default).

### DIFF
--- a/nengo/utils/numpy.py
+++ b/nengo/utils/numpy.py
@@ -7,6 +7,7 @@ import numpy as np
 
 from .compat import PY2, is_integer, is_iterable
 from ..exceptions import ValidationError
+from .magic import FunctionWrapper
 
 maxint = np.iinfo(np.int32).max
 
@@ -201,3 +202,53 @@ if hasattr(np.fft, 'rfftfreq'):
 else:
     def rfftfreq(n, d=1.0):
         return np.abs(np.fft.fftfreq(n=n, d=d)[:n // 2 + 1])
+
+
+class array_like(object):
+    """Decorator to apply `np.asarray` to function arguments
+
+    We actually use `np.array` with the `copy` argument defaulting
+    to false, since this is more configurable than `np.asarray`.
+
+    Parameters
+    ----------
+    *args : strings
+        The names of the arguments to ensure are arrays.
+    **kwargs
+        Keyword arguments to pass to `np.array`. `copy` defaults to False,
+        so that it acts like `asarray`.
+
+    Example
+    -------
+    A function that adds a scaled version of an array to another array:
+
+    >>> @array_like('A', 'B')
+    >>> def A_plus_scaled_B(A, beta, B):
+    >>>     return A + beta * B
+
+    This function will work as expected even if `A` and `B` are lists.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.arguments = args  # the arguments to run `asarray` on
+        self.array_kwargs = dict(copy=False)
+        self.array_kwargs.update(kwargs)
+
+    def __call__(self, wrapped):
+        import inspect
+
+        # check that self.arguments are valid
+        argspec = inspect.getargspec(wrapped)
+        for arg in self.arguments:
+            if arg not in argspec.args:
+                raise ValueError("'%s' is not an argument of '%s'" %
+                                 (arg, wrapped.__name__))
+
+        # make the wrapper
+        def wrapper(_wrapped, _instance, args, kwargs):
+            callargs = inspect.getcallargs(_wrapped, *args, **kwargs)
+            for arg in self.arguments:
+                callargs[arg] = np.array(callargs[arg], **self.array_kwargs)
+            return wrapped(**callargs)
+
+        return FunctionWrapper(wrapped, wrapper)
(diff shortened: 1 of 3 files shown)


hunse:
Seems this breaks in Python 2.6. Maybe that's a problem, though we were planning on dropping 2.6 support anyway...

tbekolay:
Yeah; we made a Python 2.6 release, so we can drop support for it, I think. But that'd be a change that necessitates a bump to 2.1. But I think the node API change will do that anyway?

hunse:
Yeah, and we might have a learning rule API change in there too, based on the discussion in the other thread.

Do you think it makes sense to have a 2.1 branch? That way, we could merge this into 2.1, but not preclude a 2.0.1 if we've got a few small changes we want to roll out. Or I guess we could do it the other way around, and have a 2.0.1 based off the 2.0 tag. That would mean a little more cherry-picking/double merging, but then master is always the most cutting edge.

tbekolay:
Yeah... I dunno, I definitely don't want to do something that requires cherry-picking / double merging, etc. Maybe a 2.1 branch is a good compromise, though it may be a pain to rebase once we switch to 2.1.

My original plan was to go through the PR queue and fixup/merge anything that could be merged without needing a 2.1 bump, then do a 2.0.1 release. At that point, instead of switching master to 2.0.2-dev, we'd switch to 2.1-dev, at which point we can start merging 2.1 stuff. That would be the end of the line for 2.0 though. Which I'm kind of fine with... we don't really have the resources to justify maintaining multiple branches of development, and should instead make sure that each minor version upgrade is easy and well documented so we can continue moving forward and make sure people don't stick with old versions too long.

hunse:
I think this plan is okay as long as we're very prompt about getting 2.0.1 things merged and getting the release out. Since it's often easier to build on other people's work, if there are a lot of 2.1 PRs sitting around waiting to be merged, writing a new 2.1 feature would mean either not being able to build off those features, or having to manually put them all together and then have a PR that requires them all to be merged first.

I see 2.0.1 mostly as a patch to fix any critical bugs we discover before 2.1, so if nothing like that comes up, I'd be fine skipping it entirely. I'd hold all features until 2.1.

tbekolay:
Yeah; now that there's a workflow, releases are pretty trivial to do. I'm thinking that it might be worth dropping the whole idea of release candidates for the time being; might as well just keep bumping the version number up so that PyPI can serve it until we have enough users (or a standard set of models that support version 2.x) to be able to do robust testing of a release candidate.

Looking at https://github.com/nengo/nengo/blob/master/CHANGES.rst the only real bugfix since 2.0 is the fix for NumPy 1.6 compatibility, and what's in the queue now isn't critical for 2.0, so I could release 2.0.1 and switch to 2.1-dev today. Will discuss briefly in the dev meeting!

tbekolay:
Just to put online what we discussed in the dev meeting: I'll go ahead with a 2.0.1 release today so that we can start implementing / merging things for 2.1.

nengo/utils/numpy.py
```diff
@@ -201,3 +202,53 @@ def rmse(x, y, axis=None, keepdims=False):
...
+            if arg not in argspec.args:
+                raise ValueError("'%s' is not an argument of '%s'" %
+                                 (arg, wrapped.__name__))
+
+        # make the wrapper
+        def wrapper(_wrapped, _instance, args, kwargs):
+            callargs = inspect.getcallargs(_wrapped, *args, **kwargs)
+            for arg in self.arguments:
+                callargs[arg] = np.array(callargs[arg], **self.array_kwargs)
+            return wrapped(**callargs)
```
jgosmann: Does this work for functions with an `*args` argument (where not not all arguments might be named)?

jgosmann:
I think this needs a few tests of `array_like`, especially:
- [ ] Does it work for complex function signatures including `*args` and `**kwargs`.
- [ ] Can multiple `array_like` decorators applied to a single function (in case different arguments to the `np.array` function are required).

Probably also needs a changelog entry?

I can do those things when I have some time between other things, but if someone else wants to start on this, I'm fine with that.

Final question: Is the added complexity of the decorator worth the savings in typing? Instead of using the decorator, one could also write
```python
x = np.array(x, copy=False, ndim=1)
gain = np.array(gain, copy=False, ndim=1)
bias = np.array(bias, copy=False, ndim=1)
```
(and if `ndim` is not required `x = np.asarray(x)` would work)

hunse:
I'm taking a look right now.

hunse:
Ok, fixed it for complex function signatures, and added a test. Still haven't added a changelog, but let's make sure it's actually the solution we want first.

> Is the added complexity of the decorator worth the savings in typing?

My idea was that the decorator complexity is something that most people should never see, and it's more concise than doing all that checking at the beginning of the function. But it is another thing that a curious person would have to look into, and so there's an argument for just doing it manually. I like the decorator, but I'm not strongly committed to it and could be convinced to just do it manually for now.

jgosmann:
I also don't have a strong preference ...

tbekolay:
Dev meeting update: I kind of prefer doing it manually even though it's slightly longer... it feels more explicit, and yeah, I don't think it adds all that much length in the end. I'll do it the manual way in a separate branch to compare.

hunse:
The one other thing I like about the decorator is it "goes with" the function signature, so it seems more likely to me that people would copy it when implementing e.g. `gain_bias` in their own code.

Sometimes I really just miss typed arguments...

tbekolay:
The alternative implementation of this, #1197, was merged instead.

PR #634, opened 2015-01-27
Better error message in ElementwiseInc

hunse:
When we have a shape mismatch, display a better message
with the full shapes.


hunse:
Updated as per @tbekolay's feedback.

tbekolay:
Merging without a changelog update (change is pretty minor and will affect a very small number of people).

Issue #305, opened 2014-03-30
Desired semantics for connecting to a Node that has an output?


arvoelke:
What is supposed to happen if you connect to a `Node` that has specified an `output`? Currently this crashes in the builder:

``` python
import nengo

model = nengo.Network()
with model:
    a = nengo.Node(output=[0.5])
    b = nengo.Node(size_in=1, output=[-0.2])

    nengo.Connection(a, b)

    cb = nengo.Probe(b, filter=0.03)

sim = nengo.Simulator(model)
sim.run(1)
plot(sim.trange(), sim.data[cb])
```

Because `sig_in[b]` is never set in `build_node`. However, if we change the way we specify the output, then it's okay:

``` python
b = nengo.Node(size_in=1, output=lambda t, x: [-0.2])
```

Here, the probe gives `-0.2`.

I don't think there should be a difference between these two cases. Which case gives the desired behaviour (assuming we'll add validation for the first case in the future)? Are there some other considerations?

PR #640, opened 2015-02-04
Add RC setting for dtype

tbekolay:
Mostly, this just makes it possible to use `float16` and `float32` dtypes (using anything else wouldn't make much sense).

In order to see the effects of this change, I added a `--memory` option to `py.test` which spits out the total amount of memory used by the Python process running the test suite. This only works on Linux and Mac OS X for now.

To give a sense of what effect changing the dtype has on Nengo as a whole, here are the results from running the test suite on my machine.

`float64` (still the default):
- All tests pass
- Test suite takes ~120 seconds
- Test suite takes ~430 MiB of memory

`float32`:
- 2 tests fail (both because they check if two floats are exactly equal instead of just close; could fix in this PR)
- Test suite takes ~120 seconds
- Test suite takes ~290 MiB of memory

`float16`:
- 62 tests fail (Some `FloatingPointError`s in step functions, like due to NaNs or infs, other issues with float tolerances)
- Test suite takes ~140 seconds (I think all of this is overhead from `pytest` processing the failed tests to give verbose error messages)
- Test suite takes ~265 MiB of memory

So, it seems like running `float32`s is a pretty good option for machines with relatively little RAM! `float16` isn't feasible at the moment, though perhaps with some work it could be.

### DIFF
--- a/nengo-data/nengorc
+++ b/nengo-data/nengorc
@@ -26,6 +26,12 @@
 
 ### CONFIGURATION BEGINS HERE
 
+# --- Settings for controlling the precision of all calculations
+[precision]
+
+# Default bit precision for all NumPy arrays made by Nengo (e.g., bits: 64
+# would use float64/int64 dtypes). Must be one of '16', '32', or '64'. (string)
+#bits: 64
 
 # --- Settings for the decoder cache
 [decoder_cache]
--- a/nengo/builder/builder.py
+++ b/nengo/builder/builder.py
@@ -3,11 +3,11 @@ import warnings
 
 import numpy as np
 
-from nengo import rc
 from nengo.builder.signal import Signal, SignalDict
 from nengo.builder.operator import TimeUpdate
 from nengo.cache import NoDecoderCache
 from nengo.exceptions import BuildError
+from nengo.rc import rc
 
 
 class Model:
@@ -21,6 +21,9 @@ class Model:
         A name or description to differentiate models.
     decoder_cache : DecoderCache, optional
         Interface to a cache for expensive parts of the build process.
+    builder : `nengo.builder.Builder`, optional
+        A ``Builder`` instance to use for building. Defaults to a
+        new ``Builder()``.
 
     Attributes
     ----------
@@ -82,11 +85,13 @@ class Model:
         self.seeded = {}
 
         self.sig = collections.defaultdict(dict)
-        self.sig['common'][0] = Signal(0., readonly=True, name='ZERO')
-        self.sig['common'][1] = Signal(1., readonly=True, name='ONE')
+        self.sig['common'][0] = Signal(
+            np.array(0., dtype=rc.float_dtype), readonly=True, name='ZERO')
+        self.sig['common'][1] = Signal(
+            np.array(1., dtype=rc.float_dtype), readonly=True, name='ONE')
 
-        self.step = Signal(np.array(0, dtype=np.int64), name='step')
-        self.time = Signal(np.array(0, dtype=np.float64), name='time')
+        self.step = Signal(np.array(0, dtype=rc.int_dtype), name='step')
+        self.time = Signal(np.array(0, dtype=rc.float_dtype), name='time')
         self.add_op(TimeUpdate(self.step, self.time))
 
         self.builder = Builder() if builder is None else builder
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -12,6 +12,7 @@ from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError, ObsoleteError
 from nengo.neurons import Direct
 from nengo.node import Node
+from nengo.rc import rc
 from nengo.solvers import NoSolver, Solver
 from nengo.utils.numpy import is_iterable
 
@@ -62,19 +63,22 @@ def get_eval_points(model, conn, rng):
     if conn.eval_points is None:
         view = model.params[conn.pre_obj].eval_points.view()
         view.setflags(write=False)
+        assert view.dtype == rc.float_dtype
         return view
     else:
         return gen_eval_points(
-            conn.pre_obj, conn.eval_points, rng, conn.scale_eval_points)
+            conn.pre_obj, conn.eval_points, rng, conn.scale_eval_points,
+            dtype=rc.float_dtype)
 
 
-def get_targets(conn, eval_points):
+def get_targets(conn, eval_points, dtype=None):
+    dtype = rc.float_dtype if dtype is None else dtype
     if conn.function is None:
-        targets = eval_points[:, conn.pre_slice]
+        targets = eval_points[:, conn.pre_slice].astype(dtype)
     elif isinstance(conn.function, np.ndarray):
         targets = conn.function
     else:
-        targets = np.zeros((len(eval_points), conn.size_mid))
+        targets = np.zeros((len(eval_points), conn.size_mid), dtype=dtype)
         for i, ep in enumerate(eval_points[:, conn.pre_slice]):
             out = conn.function(ep)
             if out is None:
@@ -95,7 +99,7 @@ def build_linear_system(model, conn, rng):
             "This is because no evaluation points fall in the firing "
             "ranges of any neurons." % (conn, conn.pre_obj))
 
-    targets = get_targets(conn, eval_points)
+    targets = get_targets(conn, eval_points, dtype=rc.float_dtype)
     return eval_points, activities, targets
 
 
@@ -105,7 +109,7 @@ def build_decoders(model, conn, rng):
     bias = model.params[conn.pre_obj].bias
 
     eval_points = get_eval_points(model, conn, rng)
-    targets = get_targets(conn, eval_points)
+    targets = get_targets(conn, eval_points, dtype=rc.float_dtype)
 
     if conn.solver.weights and not conn.solver.compositional:
         # solver is solving for the whole weight matrix, so apply
@@ -158,8 +162,8 @@ def slice_signal(model, signal, sl):
     if isinstance(sl, slice) and (sl.step is None or sl.step == 1):
         return signal[sl]
     else:
-        size = np.arange(signal.size)[sl].size
-        sliced_signal = Signal(np.zeros(size), name="%s.sliced" % signal.name)
+        size = np.arange(signal.size, dtype=rc.float_dtype)[sl].size
+        sliced_signal = Signal(shape=size, name="%s.sliced" % signal.name)
         model.add_op(Copy(signal, sliced_signal, src_slice=sl))
         return sliced_signal
 
@@ -171,8 +175,8 @@ def build_solver(model, solver, conn, rng):
 
 @Builder.register(NoSolver)
 def build_no_solver(model, solver, conn, rng):
-    activities = np.zeros((1, conn.pre_obj.n_neurons))
-    targets = np.zeros((1, conn.size_mid))
+    activities = np.zeros((1, conn.pre_obj.n_neurons), dtype=rc.float_dtype)
+    targets = np.zeros((1, conn.size_mid), dtype=rc.float_dtype)
     # No need to invoke the cache for NoSolver
     decoders, solver_info = conn.solver(activities, targets, rng=rng)
     weights = decoders.T
@@ -249,7 +253,7 @@ def build_connection(model, conn):
         elif isinstance(conn.function, np.ndarray):
             raise BuildError("Cannot use function points in direct connection")
         else:
-            in_signal = Signal(np.zeros(conn.size_mid), name='%s.func' % conn)
+            in_signal = Signal(shape=conn.size_mid, name='%s.func' % conn)
             model.add_op(SimPyFunc(in_signal, conn.function, None, sliced_in))
     elif isinstance(conn.pre_obj, Ensemble):  # Normal decoded connection
         eval_points, decoders, solver_info = model.build(
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -394,7 +394,8 @@ class Copy(Operator):
         # If there are repeated indices in dst_slice, special handling needed.
         repeats = False
         if npext.is_array_like(dst_slice):
-            dst_slice = np.array(dst_slice)  # copy because we might modify it
+            # copy because we might modify it
+            dst_slice = np.array(dst_slice)
             if dst_slice.dtype.kind != "b":
                 # get canonical, positive indices first
                 dst_slice[dst_slice < 0] += len(dst)
@@ -696,7 +697,7 @@ class BsrDotInc(DotInc):
             mat_A = self.bsr_matrix((A, self.indices, self.indptr))
             inc = mat_A.dot(X)
             if self.reshape:
-                inc = np.asarray(inc).reshape(Y.shape)
+                inc = inc.reshape(Y.shape)
             Y[...] += inc
         return step_dotinc
 
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -4,7 +4,12 @@ import inspect
 import importlib
 import os
 import re
+try:
+    import resource
+except ImportError:  # pragma: no cover
+    resource = None  # `resource` not available on Windows
 import shlex
+import sys
 import warnings
 
 import matplotlib
@@ -55,6 +60,9 @@ class TestConfig:
 def pytest_configure(config):
     matplotlib.use('agg')
     warnings.simplefilter('always')
+
+    if config.getoption('memory') and resource is None:  # pragma: no cover
+        raise ValueError("'--memory' option not supported on this platform")
 
     if config.getoption('simulator'):
         TestConfig.Simulator = load_class(config.getoption('simulator')[0])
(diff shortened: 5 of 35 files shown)


hunse:
Lost my comments after rebase. See 3dcdb3a502122ec78aefc0ee205f3a9f3d9b475c and 3dcdb3a502122ec78aefc0ee205f3a9f3d9b475c.

hunse:
I've added a few commits that give Simulator a dtype, which can be set independently. It doesn't really reduce the `rc.get` calls, as I had hoped, but it does make it easy to change the Simulator dtype without touching the config file or changing the API dtype.

hunse:
This LGTM. 

The TravisCI error is a strange one, so I'd like to see if @tbekolay (or anyone else) has any intuition as to why it came up. It's basically a pep8 warning that we've never got before, and I can't reproduce on my own machine (my flake8 says it's up to date, too). We could just forcibly ignore the warning, but I'm curious why it's happening now.

jgosmann:
pep8 (the tool) changelog:
> 1.6.2 (2015-02-15)
> Changes:
>    * Added check for breaking around a binary operator. (Issue #197, Pull #305),

PEP8 (the proposal) says you should do

    var = (operand1 +
        operand2)

instead of

    var = (operand1
        + operand2)

hunse:
So apparently `pip install flake8 --upgrade` does not upgrade pep8... Thanks for finding this @jgosmann.

I'm going to throw this on the 'ignore' list for now, since 1) it's breaking all our PR checking, 2) we haven't worried about it in the past, and 3) in my opinion, we should continue to not worry about it.

hunse:
If everyone is happy with ignoring that PEP8 error, and @tbekolay is happy with my additions, then this is good to go.

tbekolay:
Hmm... I'm not sure about having the model create the signals. Will think about it a bit more though.

hunse:
It's the only way I could think of to have the dtype of signals created for a model default to that model's dtype. The signals are clearly associated with a model, so I see no problem with it.

hunse:
The other alternative is to just make sure that we set the dtype explicitly when making signals. So rather than `Signal(np.zeros(5), name='foo')` as we do now, we'd do `Signal(np.zeros(5, dtype=model.dtype), name='foo')`. (Note that the `dtype` parameter is on the signal value, not the signal constructor itself. I think this is better if we go this route.)

drasmuss:
That seems cleaner to me (and about equally error-prone), gets my vote.  The one advantage of passing `dtype` directly to the `Signal` constructor is that we could make it a mandatory argument instead of a kwarg, then that would force people to think about it whenever they create a Signal (and therefore less likely to forget to use `model.dtype`).

hunse:
FYI: rebased to master.

hunse:
Got this rebased. It passes most tests with float32 dtype. The ones it fails are mostly because we lose a tiny bit of precision on e.g. the synapse tests, which have tight tolerances.

I wanted to get it up so that people can take a look and provide feedback.

Still to do:
- [ ] Changelog entry
- [ ] The `--memory` option wasn't working for me, possibly because the pytest console output hook has changed (it didn't appear to be executing for me)
- [ ] One way that we could be slightly more efficient during the build process is making sure whenever we call e.g. `np.zeros` we pass in the dtype, so we don't end up making things in `float64` and casting to `float32`. However, I'm not sure if it's worth our time to implement that, and the extra code. I think the benefits would be fairly minimal. Opinions?
- [ ] Do we want to have all tests pass for `float32`? We could even add a float32 test to TravisCI. However, it might be easier to do this with the new `allclose` fixture with configurable tolerances. An alternative would be to have some tests unsupported for the float32 simulator (though that could be harder to do now that we've moved the unsupported tests to `setup.cfg`).

nengo/simulator.py
```diff
@@ -135,18 +131,15 @@ class Simulator:
 
     def __init__(
             self, network,
-            dt=0.001, seed=None, model=None, progress_bar=True, optimize=True,
-            dtype=global_dtype):
+            dt=0.001, seed=None, model=None, progress_bar=True, optimize=True):
```
hunse: The thing I don't like about this is that it makes the dtype harder to change within a file, for example if I want to make one simulator with float32 and one with float64 (for example, to look at how that effects performance). This is still possible by setting it directly on the `Model`, though, so I'm fine to have that as a backdoor way to do this, since it'll mostly be power users or developers wanting this feature.

The other reason I added the Simulator `dtype` originally is that I really don't like scripts that are not self-contained (i.e. rely on having certain RC file settings, etc.). I think it's much better to put such parameters in the script itself. I know it's possible to configure RC settings from within a script (see `test_dtype` here), but we don't actually have an example in [the docs](https://www.nengo.ai/nengo/nengorc.html). We should add an example or two and promote that as a way to use the system. Would there be any potential problems with changing the `dtype` right after Nengo import with this commit? It actually looks like it's pretty safe. This wasn't something I was thinking about when I first did this, and obviously the reason you switched to `rc.dtype`. We just have to make sure we don't use `rc.dtype` as a function argument default anywhere, I think. It might actually get tricky when we rebase this onto master, though, since there the `NdarrayParam` has a `dtype`.

tbekolay: As I mention in the commit message, I don't like having two different but interacting ways to specify a thing. I think in general we are wary of the RC system so we half use it, but we should just use it when we use it.

I can add some examples of changing RC settings in the docs, if that's important to you.

This is already rebased onto master.

hunse: Those examples don't have to happen now, just wanted to point that out.

I don't think that `NdarrayParam.dtype` will change in the case where someone changes the `rc.dtype` after importing nengo, since params get created on objects at import. The two solutions are make `NdarrayParam.dtype` a property and have it get `rc.dtype` if it's underlying value is `None`, or just not care. Thoughts?

tbekolay: I did the property thing.

tbekolay:
Pushed three squash commits. With those, this LGTM though I think we should probably get a third set of eyes to look at this and do the final merge (@drasmuss).

nengo/builder/operator.py
```diff
@@ -696,7 +698,7 @@ def step_dotinc():
             mat_A = self.bsr_matrix((A, self.indices, self.indptr))
             inc = mat_A.dot(X)
             if self.reshape:
-                inc = np.asarray(inc).reshape(Y.shape)
+                inc = np.asarray(inc, dtype=rc.sim_dtype).reshape(Y.shape)
```
drasmuss: `inc` should be automatically cast to `Y's` dtype in the `+=` below (if possible, and if it's not possible then casting it here won't help either).

nengo/conftest.py
```diff
@@ -419,10 +427,24 @@ def pytest_runtest_setup(item):
...
+    if resource and terminalreporter.config.option.memory:
+        # Calculate memory usage; details at
+        # http://fa.bianp.net/blog/2013/different-ways-to-get-memory-consumption-or-lessons-learned-from-memory_profiler/  # noqa, pylint: disable=line-too-long
+        rusage_denom = 1024.
+        if sys.platform == 'darwin':  # pragma: no cover
+            # ... it seems that in OSX the output is in different units ...
+            rusage_denom = rusage_denom * rusage_denom
+        mem = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / rusage_denom
+        terminalreporter.write_sep("=",
+                                   "total memory consumed: %.2f MiB" % mem)
```
drasmuss: I'm not sure what exactly is being reported here. Is it the maximum memory allocated during the test run, or the total memory allocated throughout the test run (which would probably be more than the maximum, due to garbage collection), or the memory allocated at the end of the rest run?

nengo-data/nengorc
```diff
@@ -72,6 +78,11 @@
 # --- Settings for the Nengo core simulator
 [nengo.Simulator]
 
+# NumPy dtype that will be used for all Simulator signals, if possible.
+# Note that Nengo is only expected to work for float dtypes.
+# If set to 'auto', will use the value in 'precision.dtype'. (string)
+#dtype: auto
```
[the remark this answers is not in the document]
tbekolay: Yeah I would also be cool with removing it; I kept it in because @hunse had it in his commit message, and I believe @arvoelke asked about it at some point, so I'd like their thoughts on whether it's necessary.

nengo/builder/builder.py
```diff
@@ -82,11 +88,13 @@ def __init__(self, dt=0.001, label=None, decoder_cache=None, builder=None):
         self.seeded = {}
 
         self.sig = collections.defaultdict(dict)
-        self.sig['common'][0] = Signal(0., readonly=True, name='ZERO')
-        self.sig['common'][1] = Signal(1., readonly=True, name='ONE')
+        self.sig['common'][0] = Signal(
+            np.array(0., dtype=rc.sim_dtype), readonly=True, name='ZERO')
```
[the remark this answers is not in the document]
tbekolay: This is what was in the commit before my squashes, to a certain extent. The reason I removed it is to be slightly more efficient -- what's the point of making a float64 array and then immediately making a new float32 copy? It would be different if it was something the user was passing us, but we are responsible for making these numpy array, so we might as well make setting the dtype a one-step process instead of a two-step process.

Of course, the way it was done before still required modifying all of these signals, so there was no benefit in terms of reducing the diff. Doing it like you suggest, of having `Signal.__init__` have a default, would cut down the diff, so I can see more of an argument for it, but it would still be slightly inefficient, so since I've gone through the effort of changing all of our dtypes, I'd prefer keeping it this way.

Unfortunately I can't think of a way to automatically detect when we're going to use the wrong dtype. There are legitimate reasons to use int dtypes (notable indices, which we do in a couples places) so we can't always say "hey, this should be a float". Then by the same logic, perhaps there are reasons why you might want a float64 for some operation regardless of the global dtype.

One alternative that comes to mind that would be easier to enforce would be to ask the user for a precision in bits rather than to ask for a dtype. That would make the implementation a little bit easier, and we wouldn't have to worry about super weird dtypes (which are already difficult to specify since we ask for a string that we pass to `np.dtype`). I'd be more comfortable doing the cast in `Signal.__init__` if we did it that way (though I would still keep all of these lines so that we don't make two arrays unnecessarily; casting in `Signal.__init__` would just be in case we missed it somewhere or if a user makes their own build function and forgets it there). But in that case we might also need to add something to `Signal` to differentiate float vs int signals?

nengo/builder/connection.py
```diff
@@ -158,8 +162,9 @@ def slice_signal(model, signal, sl):
     if isinstance(sl, slice) and (sl.step is None or sl.step == 1):
         return signal[sl]
     else:
-        size = np.arange(signal.size)[sl].size
-        sliced_signal = Signal(np.zeros(size), name="%s.sliced" % signal.name)
+        size = np.arange(signal.size, dtype=rc.sim_dtype)[sl].size
```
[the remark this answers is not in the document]
tbekolay: It doesn't actually matter since we only make this array to get its size.

nengo/builder/operator.py
tbekolay: I thought that with binary operations, NumPy will promote lower precision arrays to higher precision so that the result doesn't lose information. But maybe it doesn't matter when doing `+=`? I'm not sure but my inclination is to use the right dtype as quickly as possible...

nengo/conftest.py
tbekolay: I'm pretty sure that this is the amount of memory currently allocated to the Python process. Because of the way that Python does garbage collection, this should be a rough analogue to the maximum amount used during testing, though it's possible it never needed quite that much memory, only most of it. Python's GC doesn't give up any memory back to the OS, so it won't be less than the maximum used by any one test.

Sidenote, I would be OK removing this fixture from this PR; it was mostly implemented to make sure that the PR was having any effect at all. But now there's a test in here that all signals are the correct dtype, so that's enough to make sure it's having an effect, and if we really want to profile memory usage, we should probably use an actual memory profiler.

nengo-data/nengorc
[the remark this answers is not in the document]
hunse: For me, it's important to be able to make two simulators in the same script with different dtypes. However, I don't think that this line is necessary for that.

Aaron was asking about having a different dtype for build and simulation. This might be something we want to do in the future, but unfortunately it's never been supported by this PR.

So I think it's fine to get rid of this.

nengo/builder/operator.py
hunse: I don't think there's any reason not to cast here, though I think things would've been fine without it, too.

nengo/conftest.py
hunse: We check all signals, and I tried to check build outputs (e.g. eval points, encoders) as well, but I've never been certain that we've covered all arrays. It might be good to do some profiling at some point, though I'm fine to leave that as future work. When I was trying to estimate memory savings with `--memory`, using `float32` on the test suite didn't save as much as I might have hoped, and I'd like to get a better idea of true memory savings on a more practical example.

nengo/builder/builder.py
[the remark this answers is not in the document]
drasmuss: Moreso than reducing the diff, what strikes me is removing the repetitive `dtype=rc.sim_dtype` all throughout the code.  Whenever I see something like that it makes me think there has to be a way to avoid that 😄.  

The other, more significant, advantage is easing future maintainability.  In the current system, whenever we add a new Signal we need to remember to set `dtype=rc.sim_dtype`.  I am fairly certain we will not remember to do that every time, and then we'll end up with simulations with mixed dtypes.  And typically when we're adding a new Signal it's because we're adding a new feature/op, so it wouldn't be caught by the test that builds a model out of our existing ops and checks all their dtypes.  So again, maybe we will all remember to update that test whenever we add a new op, but I suspect not 🤷‍♂ .

> Unfortunately I can't think of a way to automatically detect when we're going to use the wrong dtype. There are legitimate reasons to use int dtypes (notable indices, which we do in a couples places) so we can't always say "hey, this should be a float". Then by the same logic, perhaps there are reasons why you might want a float64 for some operation regardless of the global dtype.

I wasn't thinking of automatic detection, I was thinking of someone accidentally passing an `initial_value` that doesn't match the specified `dtype` for the Signal.  So, e.g., you specify int and accidentally pass a float, or explicitly want `float64` but accidentally use `rc.sim_dtype`.  It's a minor benefit, since we usually create the initial value at the same time as the Signal so those kinds of errors are unlikely, but it wouldn't hurt.

Overall I suspect the cost of casting within the Signal init would be pretty minor (in terms of time and memory), since it's just a one-time cost during build time.  But if we did want to avoid that, we could just make the `initial_value` argument optional and add a `shape` parameter, so the Signal data would be specified abstractly with dtype/shape, and then the we'd call `np.zeros` (which we use in most cases) inside init.

nengo/builder/operator.py
drasmuss: > I thought that with binary operations, NumPy will promote lower precision arrays to higher precision so that the result doesn't lose information.

Since this is an in-place operation, it won't change the type of the base array (which is what we want).

> I don't think there's any reason not to cast here,

Allowing numpy to implicitly cast it as part of the += op will be faster than adding an explicit cast step.

nengo/builder/builder.py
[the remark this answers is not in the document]
hunse: The reason we pass a matrix rather than a shape is for signals whose initial values are not zero, e.g. weights, biases, etc. We could do it so that we pass a shape and an optional "init" value, but then I'd be worried for things like weights, we don't want to make an unnecessary extra copy if we can avoid it.

Personally, I think the cost with re-casting an array is very minimal, and it's not worth putting a lot of our dev time into avoiding it. Particularly, because I think with the large signals in this commit (i.e. weights), they should already be the right dtype due to having the right dtype on things passed in to the solver.

On that note, with these changes, we should be explicitly casting transforms to the right dtype in `builder/transforms.py`. I think in many cases the value returned by `transform.sample` will be right, but we need to make sure.

nengo/conftest.py
drasmuss: I'm fine leaving it or taking it out.  It does seem like we probably won't use it that much, but the code is already written and tested so 🤷‍♂ .

nengo/builder/builder.py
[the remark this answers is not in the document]
drasmuss: Yeah I wouldn't want to remove the ability to manually specify an `initial_value` for those reasons, this would just be an optional feature to avoid the re-casting in the common cases.  But agreed that I'm not sure how much of a difference it would actually make.

nengo/conftest.py
tbekolay: Feels like another thing we should separate into a pytest plugin.

drasmuss:
Added a commit to implement the things from discussion above.

After implementing it I realized that we don't actually need a `dtype` parameter, because we only rarely need a non-default dtype, and for those instances we're always passing an explicit `initial_value`.  So I just added a `shape` parameter, which gives a default all-zero initial value.

One option, rather than having a separate `shape` parameter, would be to merge that with the `initial_value` parameter (so if it's a tuple we interpret that as a shape, otherwise it's an initial_value).  I don't have a strong preference either way.

I didn't touch the `--memory` fixture.

tbekolay:
All looks good, I rebased/fixed up history then made one last fixup commit with things I noticed looking through the diff. Squash it if it looks good to you @drasmuss then merge away!

Issue #632, opened 2015-01-26
Error connection in learning can't use slicing on post


studywolf:
So when you have an error connection for learning, if you try to slice into the post population 

```  
File "/home/tdewolf/Dropbox/code/nengo/nengo/builder/operator.py", line 234, in make_step
    assert da in [1, dy] and dx in [1, dy] and max(da, dx) == dy
AssertionError
```

gets thrown. Pre-slicing OK, post-slicing not. 
You can get around it by creating a dummy population of the appropriate size to connect to, but that makes setting up learning connections even more of a pain. Here's the code that triggers it

```
import numpy as np
import nengo
from nengo.networks import EnsembleArray
nengo.log()

model = nengo.Network()

with model: 

    dim = 6
    state = nengo.Node(output=[0,]*dim)
    learning_signal = nengo.Node(output=[1,]*(dim/2))
    population1 = nengo.Ensemble(100, dim)
    population2 = nengo.Ensemble(100, dim)

    nengo.Connection(state, population1)
    err = nengo.Connection(learning_signal, population1[:dim/2], modulatory=True)
    nengo.Connection(population1, population2[:dim/2], 
                     function=lambda x: np.zeros(dim/2),
                     learning_rule_type = nengo.PES(err))

s = nengo.Simulator(model)
s.run(1)
```

Note: this also gets thrown when the post-population is the right size but you slice into it on the error connection: 

```
import numpy as np
import nengo
from nengo.networks import EnsembleArray
nengo.log()

model = nengo.Network()

with model: 

    dim = 6
    state = nengo.Node(output=[0,]*dim)
    learning_signal = nengo.Node(output=[1,]*(dim/2))
    population1 = nengo.Ensemble(100, dim)
    population2 = nengo.Ensemble(100, dim/2)

    nengo.Connection(state, population1)
    err = nengo.Connection(learning_signal, population1[:dim/2], modulatory=True)
    nengo.Connection(population1, population2, 
                     function=lambda x: np.zeros(dim/2),
                     learning_rule_type = nengo.PES(err))

s = nengo.Simulator(model)
s.run(1)
```

hunse:
This problem comes down to the fact that we do output slicing within the transform. So in the builder, the output of the `err` connection is 6-D, since the output of the transform is 6-D and includes the slice mapping. I think the best solution to this is to take the slice out of the transform, and add a `CopySlice` operator that copies the output of the transform in `err` to the correct dimensions of the target population.

[...]

Issue #366, opened 2014-06-01
Connect to a connection instead of modulatory=True


tbekolay:
@hunse suggested this in the learning PR:

> `error_conn = nengo.Connection(error, A_squared, modulatory=True)`
> 
> Could we instead write `error_conn = nengo.Connection(error, conn)`?

Since connections weren't an object in Nengo 1.4, it never occurred to me that this could be possible, but it is possible and perhaps is better syntax for what we do now with `modulatory=True`. I think so far we've never used a modulatory connection for anything but learning; are there other use cases for this?

hunse:
This is changed in #642. Error signals now connect into the learning rule, for example

    conn = nengo.Connection(pre, post, learning_rule_type=nengo.PES())
    nengo.Connection(error, conn.learning_rule)

PR #642, opened 2015-02-04
Learning unmodulated

hunse:
This makes it so that we connect directly to LearningRules instead of making modulatory connections, as discussed in #632. It fixes that problem.

The second commit switches the sign of the PES error, so that it is actually an error (i.e. the actual value minus the target value). This means that the PES learning rule moves in the opposite direction of the error, as one would expect (e.g., if the output is too large, make it smaller). This makes a lot more sense to me, and hopefully to everyone else as well. Now seemed like a good time to do it, because the changes from the first commit will require that everyone re-write their learning code anyway.

EDIT: this also addresses #366.


drasmuss:
I keep going back and forth on the error sign switch.  It makes for a pretty sharp discontinuity with previous models.  Also, the PES rule is essentially just a renaming of the delta rule, and the delta rule is usually written the way the PES rule is now (i.e. `target - actual`).  

But you're right that in that case, referring to the feedback signal as an "error" is kind of odd.  What we probably should have done is named it "feedback" instead of "error" or something.  I don't know whether it's better to change our nomenclature to match the implementation, or change the implementation to match the nomenclature.

celiasmith:
Every rule is essentially just a renaming of the delta rule ;)

Daniel Rasmussen wrote:

drasmuss:
In that it's all just gradient descent, yeah, although some are more similar than others :wink:.  And in general with gradient descent the convention is to use the error signal as in this PR, which is the main argument for the swap I think.  It's probably more likely that people used to things like backprop would be confused by the swapped error in the current implementation, as @hunse suggests.

hunse:
Do people have comments on this, or is it ready to go?

tbekolay:
I'll review this shortly. I'm considering the error signal sign flip uncontested!

nengo/utils/simulator.py
```diff
@@ -83,10 +83,6 @@ def validate_ops(sets, ups, incs):
     for node in sets:
         assert len(sets[node]) == 1, (node, sets[node])
 
-    # -- assert that only one op updates any particular view
-    for node in ups:
-        assert len(ups[node]) == 1, (node, ups[node])
```
tbekolay: Hm, is this OK? Why did we have this restriction in the first place?

tbekolay:
This was a bit old, so I gave it a rebase. Looking good!

Only two things I think before merge. First is the all-important changelog entry. I also had a thought about the `modulator` parameter... rather than completely remove all mentions of it here, how about for 2.1.x, if someone tries to set `modulator` to `True`, we raise a `ValueError` explaining the new API for learning rules, or at least pointing them to this PR (or a doc page on the learning API or something)? Definitely we will have some people upgrade Nengo without reading the changelog, so I think it'll be kinder to tell them about this change when we know it's happening (setting `modulatory=True`).

nengo/utils/simulator.py
hunse: I couldn't think of a good reason to have it. The original reason may have just been because we always used updates for filters, in which case the signal in the filter being updated should only ever be updated once. When you get into multiple updates per step, there is a possibility that the order of updates matters. I was going to say that that's not a problem if the updates are just increments, but that's not true. Any learning rule that depends on the magnitude of the connection weights will act differently if it's computed before or after another rule modifying the same weights. Maybe this is something we need to consider? We should at least be able to make this deterministic within a model, so I don't think it's too big of an issue.

hunse:
Adding that error sounds like a good idea.

arvoelke:
Updated the changelog and made it so that you get 

    Parameter is no longer supported; please refer to pull-request #642: modulatory connections removed

This introduced a new `AbandonedParam` parameter. My only concern is that this should not encourage people to abandon parameters liberally and destroy backwards compatibility.

arvoelke:
Should #643 be implemented here too, in order to contain these rather significant changes to PES to a single PR?

neworderofjamie:
Hello Nengoers, I'm trying to restore my SpiNNaker version of PES to it's former glory and the learning rate seems to have been scaled somehow. In the old learn_communications_channel example it was 1 and now it's 1e-6 - How has the calculation of the PES update changed?

THANKYOU!

drasmuss:
It was probably this change #527, where things were changed to scale with `dt`.  I know we had also talked about changing the learning rate to scale with the number of neurons in the pre-synaptic population (#643), but I'm not sure if that actually happened yet?

neworderofjamie:
The implementation I was working on at last years summer school multiplied the learning rate by dt (in seconds) - Is that the scaling in question or something more cunning involved?

drasmuss:
It's a little bit more complicated, in that what's actually happening is that the output activities are being scaled by `1/dt`.  So the `a(x)` in the PES equations are all larger than they were before, so the learning rate needs to be scaled down accordingly (@hunse made the change so he can double check that I'm not saying lies here).

arvoelke:
@neworderofjamie That sounds right, although the correct way to do it may depend on whether your spikes are scaled by `dt` (you can see how this was done in #527 as Dan said). Note that the above commit also divides the effective learning rate by `n_neurons` in the `pre` population, and this PR also flips the sign of the error signal. So the final rule is 



where `kappa` is the learning rate (invariant under `dt`), `n` is number of neurons in `pre`, `E` is the error signal, and `a` is the filtered pre-synaptic activity.

This PR just needs a final review now and then it should be good to go.

hunse:
Can we change `AbandonedParam` to `DeprecatedParam`?

Also, did we want to add scaling by number of pre neurons in this PR? Might as well change it now, and revise the learning rates *again*.

CHANGES.rst
```diff
@@ -34,6 +34,17 @@ Release History
   further improvements are pending, and the current implementation
   SHOULD NOT BE RELEASED!
   (`#652 <https://github.com/nengo/nengo/pull/652>`_)
+- ``LearningRuleType`` objects no longer take modulatory connections.
+  Instead, a connection should be made directly from the error population
+  to the learning rule. Also, the error's sign has been flipped for
+  the PES learning rule to be ``actual - target``.
```
hunse: "learning rule" -> "learning rule error"

CHANGES.rst
```diff
@@ -34,6 +34,17 @@ Release History
...
+  (`#642 <https://github.com/nengo/nengo/pull/642>`_)
+- The learning deltas to ``PES`` are now invariant under the number of neurons
```
hunse: I would say "The ``PES`` learning rule is now invariant under..."

arvoelke:
Sorry about throwing more onto the pile, but realized that the pre-activity should be filtered. It was just a one line change in the builder, and a new parameter for PES. Note that the parameter order for PES is different from BCM, otherwise all the examples (and possible models that people have already built), need to change their learning rate to a key-word argument.

neworderofjamie:
Ok, sorry for being slow here, but tracing all the changes that have occurred since last year through github is proving to be a slow process! can someone just confirm that the net effect of all of the changes is that:

1) The PES builder update has changed from:
```
pes.learning_rate * model.dt
```
To:
```
-pes.learning_rate * model.dt / n_neurons
```

So, sign has changed and learning is scaled by both number of pre-decoder neurons AND timestep.

2) Each spike adds 1/dt to the activity 

3) As of very recently, activity is filtered - I seem remember @tbekolay telling me this wasn't necessary for PES?

Don't 1 and 2 cancel each other out (mathematically if not in terms of your implementation)?

hunse:
@neworderofjamie That all sounds correct. 

You're right that the `dt` in 1 cancels with the `1 / dt` in 2. We wanted 2 so that spikes integrate to one, which means that you can filter spikes and get an approximation of the rate, which is really useful (it also made the code cleaner). Also, I think before there was an extra `dt` somewhere, because when I did #527 I also got the PES rule to be invariant to choice of `dt` (i.e. it acts like a proper differential equation).

As for the activity filtering, I would guess that it's not necessary (as you recall), since weight updates are integrated which acts as a form of low-pass filtering, averaging over any noise in the activities. So my guess is that filtering the activities just makes the weight updates a little less noisy. I'm not sure which is more biologically plausible. Perhaps @tbekolay can say more?

neworderofjamie:
Ok, that makes sense, but I think I must still be missing a factor somewhere. Ignoring the filtering for the moment, with the current default PES learning rate of 0.0001, cancelling out the model.dts, a 60 neuron ensemble multiplies its error signal by 0.0001 / 60 = 0.000001667, but with the previous default PES learning rate of 1.0, the same ensemble would multiple its error signal by 1.0 * 0.001 = 0.001 which differs by a factor of 600ish, but the examples work the same - Any thoughts?

hunse:
Aha, I got it! I was about to tell you that I couldn't figure it out either, but it just came to me.

Scaling the spikes by `1 / dt` not only affects the weight changes, it also means that the magnitude of the spikes going *through* the learned weights are scaled. So the weights themselves have to be a factor of `dt` smaller than the old weights did. That means that the old rule is actually a factor of 600 * 0.001 = 0.6 the speed of the new rule, which is approximately what I get if I look at `learn_communication_channel.ipynb` (i.e. the new one is slightly faster).

neworderofjamie:
@mundya, do you have any thoughts about implementing spike scaling on Spinnaker vs just dividing the learning rate by model.dt on the host (these small learning rates also have the bad habit of underflowing s16.15)

mundya:
> just dividing the learning rate by model.dt on the host

Would be my suggestion.

neworderofjamie:
Great, that would be my preference too.

nengo/utils/simulator.py
tbekolay: Yeah, I think this is something we need to consider... I'm sure the difference (in this case) is pretty small, but it could be large in some cases. I'm not sure how we'd order ops that both update the same signal... So yeah, I'd be keen to keep this assertion in and figure out another way to do these updates. I think the `PreserveValue` op from before was pretty good.

tbekolay:
@neworderofjamie @hunse: Yeah, the input activity filtering isn't necessary, but should improve learning in some cases. It should only hurt learning if the filter is especially long; you're basically spreading out when the weight update is applied. If you have no filter, then all that matters is the error signal at the exact moment of the spike; you'll make a somewhat large step in the direction of the error gradient. If you have a filter, then you take a more nuanced trajectory in error space; each decoder changes according to the integral of the filtered activity and local error, which is the dot product between the error and the encoding vector. In most cases, it's the difference between moving in a straight line versus moving in a mostly straight line, with possibly a little curve at the end. That curve may or may not be important, depending on your particular learning situation.

neworderofjamie:
Thanks Trevor, that makes total sense. I just added support for activity
filtering for Voja so it's a matter of matching whatever you guys do!
Obviously, only applying a PES update when a neuron spikes is better
CPU-wise, but it's not mega-important.

hunse:
Okay, brought PreserveValue back, as per @tbekolay's suggestion.

tbekolay:
OK! Apologies this took so long, but I've done a thorough review of this PR and I'm quite happy with it. Things I did:

- Rewrote changelog entries, and amended them into the commits in which the changes occurred.
- Reordered history a bit (should be more clear now) and cleaned up commit messages.
- Added 2519a1f1ee7d. It does a bit of refactoring to ensure correct behavior for any step function order (see the commit message for more details)
- Added a 0e9dcd2c047db30, which I'm proposing as a somewhat more generic `ObsoleteParam`.

@hunse and @arvoelke, take a look and tell me what you think.

studywolf:
So I was just playing around with this (looks great!) but I noticed that I can no longer do decoder learning connected to a Node as the post-object. The reason that I'm trying to do this is because the adaptation can generate some large values, and it's better for me to just relay those through a node straight to the plant being controlled rather than try to have some ensemble intermediary with radius=1e5 and 10 thousand neurons. In just normal nengo I can use direct mode, but for spinnaker that's not an option, so it would be great to be able to connect to nodes as the post object on learning!

Just looking at it it was a two line fix, so I made it and pushed it for discussion!

tbekolay:
This makes sense to me; we don't use any neural info from the post population in decoder PES, so why not. I wonder if we should revisit the `modifies` attribute in the learning rule types? Right now it's either `'Neurons'` `'Ensemble'` or both. Maybe it should be `'transform'` and `'decoders'`, since that's what the learning rule actually modifies.

arvoelke:
+1 to revisiting `modifies`. Voja modifies `encoders`, and so my solution in that PR was to have a `post_modifies`, but makes more sense to say what signals are being modified.

tbekolay:
OK, `'Neurons'` -> `'transform'` and `'Ensemble'` -> `'decoders'` in 44768d4. I think neither of the last two commits need changelog entries? Only people implementing learning rules would care...

hunse:
Added a fixup for neuron to ensemble learning connections. This should all be cleaned up in the `monolithic-transform` branch.

hunse:
Added a test for neuron to ensemble PES learning. I think this should be good to go, @tbekolay.

tbekolay:
Great; I'll look this over this afternoon!

tbekolay:
The new tests LGTM! @arvoelke mentioned that he wanted to give this a look over too. I'll probably merge tomorrow afternoon unless someone forcibly stops me.

tbekolay:
OK, this branch is now rebased to master, and the history's been squashed and whatnot. I think this is ready to merge; will do so in an hour or so unless anyone stops me!

mundya:
Awesome!

Issue #540, opened 2014-11-12
Adaptive neurons don't give a custom rate approximation


tcstewar:
Right now, when nengo computes decoders, it uses the rate approximation of the neuron model.  For AdaptiveLIF neurons, this means that it falls back on using the LIF rate approximation.  This means that the decoder will assume no adaptation occurs.  In Nengo 1.4, the AdaptiveLIF neuron had a rate approximation that indicated what the steady-state firing rate would be after adaptation, and the decoders were trained using that.  This meant that Ensembles would have a good steady-state response.

fragapanagos:
I have a [notebook](http://nbviewer.ipython.org/github/fragapanagos/notebooks/blob/master/neuron/adaptive_LIF_neuron.ipynb) in progress that examines the adaptive neuron. @tcstewar indeed there is no analytic solution as shown [here](http://nbviewer.ipython.org/github/fragapanagos/notebooks/blob/master/neuron/adaptive_LIF_neuron.ipynb#Steady-state-firing-rate).  Besides the magical "20" constant, there's also a problem with the magical "0.1" constant. It will lead to incorrect firing rates near threshold as shown [here](http://nbviewer.ipython.org/github/fragapanagos/notebooks/blob/master/neuron/adaptive_LIF_neuron.ipynb#Do-not-use-this-method-to-find-the-firing-rate-with-mean-feedback-approximation). Instead of iterative approximation using the mean feedback value approximation, you can use the bisection method  (binary search, also iterative) to find the exact firing rate as written [here](https://github.com/fragapanagos/notebooks/blob/master/neuron/utils/neuron.py#L118) (below it, [here](https://github.com/fragapanagos/notebooks/blob/master/neuron/utils/neuron.py#L189) is a version that uses scipy's bisection method).

fragapanagos:
The neuron model at https://github.com/Stanford-BIS/nengo_alif_steady_state addresses this issue. Should this issue be closed now?

fragapanagos:
@tcstewar for the neuron model at the above repository to work, it requires the membrane voltage clip parameter in the LIF math. Could that be introduced in the standard nengo LIF?

tcstewar:
> it requires the membrane voltage clip parameter in the LIF math. Could that be introduced in the standard nengo LIF?

Oops, I missed that...  I'll go poke at that (although it'd also be nice to have a test that would fail if the clipping wasn't happening). :)

fragapanagos:
[...]

PR #644, opened 2015-02-04
Ss alif

fragapanagos:
This pull request addresses https://github.com/nengo/nengo/issues/540 . The AdaptiveLIF neuron class in this branch implements custom gain_bias and rates functions based on the steady-state firing rate of the adaptive LIF neuron model. The current master branch AdaptiveLIF relies on the base LIF neuron's gain_bias and rates function.  This branch also modifies the step_math function to better match the theory (developed for continuous, not discrete time) and uses a version of the LIF neuron's step_math that doesn't clip the voltages below 0 to 0. The AdaptiveLIF unit tests were modified to check steady state firing rate predictions.

These changes cause the builder to use the  steady state firing rates instead of the initial transient firing rates of the AdaptiveLIF model when building decoders.

I can see a few of ways to go forward with this merge. The AdaptiveLIF model developed in this branch could be maintained as a separate model from the existing AdaptiveLIF model or could overwrite the existing AdaptiveLIF model. Maintaining as a separate model makes sense if there exist or users are planning to create models that leverage the current model's method of finding decoders. Otherwise, I think it makes more sense to overwrite. Alternatively, the models could be merged and have a parameter that let's the user decide which behavior to use.

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -276,7 +276,7 @@ class AdaptiveLIFRate(LIFRate):
         """Compute rates for input current (incl. bias)"""
         n = adaptation
         LIFRate.step_math(self, dt, J - n, output)
-        n += (dt / self.tau_n) * (self.inc_n * output - n)
+        n += -np.expm1(-dt / self.tau_n) * (self.inc_n * output - n)
 
 
 class AdaptiveLIF(AdaptiveLIFRate, LIF):
@@ -284,11 +284,11 @@ class AdaptiveLIF(AdaptiveLIFRate, LIF):
 
     probeable = ['spikes', 'adaptation', 'voltage', 'refractory_time']
 
-    def step_math(self, dt, J, output, voltage, ref, adaptation):
+    def step_math(self, dt, J, spiked, voltage, ref, adaptation):
         """Compute rates for input current (incl. bias)"""
         n = adaptation
-        LIF.step_math(self, dt, J - n, output, voltage, ref)
-        n += (dt / self.tau_n) * (self.inc_n * output - n)
+        LIF.step_math(self, dt, J - n, spiked, voltage, ref)
+        n += -np.expm1(-dt / self.tau_n) * (self.inc_n * spiked - n)
 
 
 class Izhikevich(NeuronType):
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -223,38 +223,18 @@ class LIFRate(NeuronType):
 class LIF(LIFRate):
     """Spiking version of the leaky integrate-and-fire (LIF) neuron model."""
 
+    clip = NumberParam(high=0.)
     probeable = ['spikes', 'voltage', 'refractory_time']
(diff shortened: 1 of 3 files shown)


fragapanagos:
@hunse good idea. Using clip is better than duplicating the step_math function. I've incorporated your suggestions in https://github.com/Stanford-BIS/nengo/commit/f0002202735abf962cde0389e6e1cc89d1be1e7a . In terms of "realism", I only added a high limit on clip rather than a low limit. Adding a low limit would mean the AdaptiveLIF steady state rate equations could be broken for certain inc_n / tau_n settings. I think if users wanted biological realism in their neuron model, they would be better served using a conductance based model.

tbekolay:
This looks great! If I recall, the math / motivation for this is in a notebook somewhere; @fragapanagos could you link to your notebook here so we can refer to it when reviewing? And from what I know (feel free to correct me, anyone) no one's made use of the AdaptiveLIF in a publication, so my vote would be to overwrite what we have with this implementation.

Then, two quick administrative things. First, when pulling changes from `master` into your branch, please use `rebase` instead of `merge` (`git pull --rebase` will do it if you're pulling). Second, we require contributors to assign copyright to ABR when contributing to Nengo. Read over [LICENSE.rst](https://github.com/nengo/nengo/blob/master/LICENSE.rst#contributor-assignment-agreement) and if you agree, then add your info to [CONTRIBUTORS.rst](https://github.com/nengo/nengo/blob/master/CONTRIBUTORS.rst). If you have questions about the legal stuff let us know; it's not easy text to parse, so if you want a human translation we'd be happy to oblige :rainbow:

fragapanagos:
@tbekolay Yes the math / motivation for this AdaptiveLIF implementation is developed in this [notebook](http://nbviewer.ipython.org/github/fragapanagos/notebooks/blob/master/neuron/adaptive_LIF_neuron.ipynb#Steady-state-firing-rate). I use a different notation in the notebook than in the code. The mapping between the code and the notebook is
 - inc_n in the code is \alpha_f in the notebook
 - tau_n in the code is \tau_f in the notebook
 - total input current J (after applying gain and bias) in the notebook is u in the notebook
 - the notebook uses x_t as the LIF threshold. The code uses x_t=1

Ah, duly noted with respect merging in / rebasing on master. Is there a way to undo the previous merge of master into this branch or is it stuck in the history now?

I read through the copyright agreement, but a human readable format would probably be helpful. I will pm to discuss this.

fragapanagos:
@tcstewar and @jjaguayo created a separate repository for this neuron model.
https://github.com/Stanford-BIS/nengo_alif_steady_state

From there, it's possible to install the neuron model and import it as separate from the default nengo AdaptiveLIF neuron. The reasoning for this separation being that it serves as an example for others who wish to define their own neuron models, and if a lot of people end up using the steady-state AdaptiveLIF model it can later be pulled into the main nengo repo. Therefore, I am closing this pull request.

PR #645, opened 2015-02-05
Explicitly specify Mathjax path

tbekolay:
I noticed that math wasn't rendering on https://pythonhosted.org/nengo/. This was because we were using `http` to access the MathJax CDN, but pythonhosted serves our docs through `https`. This is fixed in the upcoming v1.3 of Sphinx, but this PR makes it explicit so that it works now.

I've rebuilt and updated the pythonhosted docs, so it works now, FYI.


hunse:
Cool!

PR #647, opened 2015-02-05
Add analytics fixture.

jgosmann:
This allows to store additional data in test. A use case would be in
benchmarking and comparing different implementations/revisions. To
store analytics data add the following to your tests:

```python
def test_something(analytics):
    # code
    analytics.add_data('some_key', np.array([1, 2]), 'description')
```

You also have to set the NENGO_ANALYTICS environment variable to 1. The
data will be stored in the nengo.simulator.analytics subdirectory. Two
files will be generated for each test. The txt file contains the
description of the stored data and the npz file the actual data. It can
be loaded with numpy.load.

I am putting this PR out as basis for discussion. At the moment there is some code duplication between `Plotter` and `Analytics` which I would reduce before a merge into `master`.

The questions I would like to discuss:

1. File format: Numpy files are efficient and easily loadable in a Python stack. However, other file formats (e.g. CSV, JSON) might be easier to process in other programming language if someone wants to use those for analysis.
2. Environment variable for activation: To activate the storage of the data an environment variable has to be set. This follows what `Plotter` is doing. However, to me it seems (also in the case of `Plotter`) a bit obscure as `py.test --help` won't tell you about it, in contrast to options (like `--optional`).
3. Overwriting of files: Each execution of the test suite might overwrite the analytics files. It might be easier to compare different revisions if this doesn't happen (by including a timestamp or the Git hash somehow). But in that case it might produce too much unused files.

I will probably implement some helpers to compare the stored data once I have written some actual benchmarks.

### DIFF
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -9,7 +9,7 @@ import nengo.utils.numpy as npext
 from nengo.neurons import Direct, LIF, LIFRate, RectifiedLinear, Sigmoid
 from nengo.rc import rc
 from nengo.simulator import Simulator as ReferenceSimulator
-from nengo.utils.compat import ensure_bytes
+from nengo.utils.compat import ensure_bytes, is_string
 from nengo.utils.testing import Analytics, Plotter
 
 test_seed = 0  # changing this will change seeds for all tests
@@ -41,6 +41,27 @@ def RefSimulator(request):
     return ReferenceSimulator
 
 
+def recorder_dirname(request, name):
+    record = request.config.getvalue(name)
+    if is_string(record):
+        return record
+    elif not record:
+        return None
+
+    simulator, nl = ReferenceSimulator, None
+    if 'Simulator' in request.funcargnames:
+        simulator = request.getfuncargvalue('Simulator')
+    if 'nl' in request.funcargnames:
+        nl = request.getfuncargvalue('nl')
+    elif 'nl_nodirect' in request.funcargnames:
+        nl = request.getfuncargvalue('nl_nodirect')
+
+    dirname = "%s.%s" % (simulator.__module__, name)
+    if nl is not None:
+        dirname = os.path.join(dirname, nl.__name__)
+    return dirname
+
+
 @pytest.fixture
 def plt(request):
     """a pyplot-compatible plotting interface.
@@ -54,28 +75,29 @@ def plt(request):
     If you need to override the default filename, set `plt.saveas` to
     the desired filename.
     """
-    simulator, nl = ReferenceSimulator, None
-    if 'Simulator' in request.funcargnames:
-        simulator = request.getfuncargvalue('Simulator')
-    if 'nl' in request.funcargnames:
-        nl = request.getfuncargvalue('nl')
-    elif 'nl_nodirect' in request.funcargnames:
-        nl = request.getfuncargvalue('nl_nodirect')
-    plotter = Plotter(simulator, request.module, request.function, nl=nl)
-    request.addfinalizer(lambda p=plotter: p.__exit__(None, None, None))
+    dirname = recorder_dirname(request, 'plots')
+    plotter = Plotter(
+        dirname, request.module.__name__, request.function.__name__)
+    request.addfinalizer(lambda: plotter.__exit__(None, None, None))
     return plotter.__enter__()
 
 
 @pytest.fixture
 def analytics(request):
-    simulator, nl = ReferenceSimulator, None
-    if 'Simulator' in request.funcargnames:
-        simulator = request.getfuncargvalue('Simulator')
-    if 'nl' in request.funcargnames:
-        nl = request.getfuncargvalue('nl')
-    elif 'nl_nodirect' in request.funcargnames:
-        nl = request.getfuncargvalue('nl_nodirect')
-    analytics = Analytics(simulator, request.module, request.function, nl=nl)
+    """an object to store data for analytics.
+
+    Please use this if you're concerned that accuracy or speed may regress.
+
+    This will keep saved data organized in a simulator-specific folder,
+    with an automatically generated name. Raw data (for later processing)
+    can be saved with ``analytics.add_raw_data``; these will be saved in
+    separate compressed ``.npz`` files. Summary data can be saved with
+    ``analytics.add_summary_data``; these will be saved
+    in a single ``.csv`` file.
+    """
+    dirname = recorder_dirname(request, 'benchmarks')
+    analytics = Analytics(
+        dirname, request.module.__name__, request.function.__name__)
     request.addfinalizer(lambda: analytics.__exit__(None, None, None))
     return analytics.__enter__()
 
@@ -125,10 +147,13 @@ def pytest_generate_tests(metafunc):
 
 
 def pytest_addoption(parser):
-    parser.addoption('--benchmarks', action='store_true', default=False,
-                     help='Also run benchmarking tests')
-    parser.addoption('--plots', action='store_true', default=False,
-                     help='Also run plotting tests')
+    parser.addoption(
+        '--plots', nargs='?', default=False, const=True,
+        help='Save plots (optional with directory to save them in).')
+    parser.addoption(
+        '--benchmarks', nargs='?', default=False, const=True,
+        help='Also run benchmarking tests (optional with directory to save ' +
+        'the data in).')
     parser.addoption('--noexamples', action='store_false', default=True,
                      help='Do not run examples')
     parser.addoption(
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -62,6 +62,18 @@ def recorder_dirname(request, name):
     return dirname
 
 
+def parametrize_function_name(request, function_name):
+    suffixes = []
+    if 'parametrize' in request.keywords:
+        argnames = [
+            x.strip()
+            for x in request.keywords['parametrize'].args[0].split(',')]
+        for name in argnames:
+            suffixes.append('{0}={1}'.format(
+                name, request.getfuncargvalue(name)))
+    return '_'.join([function_name] + suffixes)
+
+
 @pytest.fixture
 def plt(request):
     """a pyplot-compatible plotting interface.
@@ -77,7 +89,8 @@ def plt(request):
     """
     dirname = recorder_dirname(request, 'plots')
     plotter = Plotter(
-        dirname, request.module.__name__, request.function.__name__)
+        dirname, request.module.__name__,
+        parametrize_function_name(request, request.function.__name__))
     request.addfinalizer(lambda: plotter.__exit__(None, None, None))
     return plotter.__enter__()
 
@@ -97,7 +110,8 @@ def analytics(request):
     """
     dirname = recorder_dirname(request, 'analytics')
     analytics = Analytics(
-        dirname, request.module.__name__, request.function.__name__)
+        dirname, request.module.__name__,
+        parametrize_function_name(request, request.function.__name__))
     request.addfinalizer(lambda: analytics.__exit__(None, None, None))
     return analytics.__enter__()
 
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -1,4 +1,5 @@
 import hashlib
+import inspect
 import os
 
 import numpy as np
@@ -69,8 +70,10 @@ def parametrize_function_name(request, function_name):
             x.strip()
             for x in request.keywords['parametrize'].args[0].split(',')]
         for name in argnames:
-            suffixes.append('{0}={1}'.format(
-                name, request.getfuncargvalue(name)))
+            value = request.getfuncargvalue(name)
+            if inspect.isclass(value):
+                value = value.__name__
+            suffixes.append('{0}={1}'.format(name, value))
     return '_'.join([function_name] + suffixes)
 
 
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -10,7 +10,7 @@ from nengo.neurons import Direct, LIF, LIFRate, RectifiedLinear, Sigmoid
 from nengo.rc import rc
 from nengo.simulator import Simulator as ReferenceSimulator
 from nengo.utils.compat import ensure_bytes
-from nengo.utils.testing import Plotter
+from nengo.utils.testing import Analytics, Plotter
 
 test_seed = 0  # changing this will change seeds for all tests
 
@@ -64,6 +64,20 @@ def plt(request):
     plotter = Plotter(simulator, request.module, request.function, nl=nl)
     request.addfinalizer(lambda p=plotter: p.__exit__(None, None, None))
     return plotter.__enter__()
+
+
+@pytest.fixture
+def analytics(request):
+    simulator, nl = ReferenceSimulator, None
+    if 'Simulator' in request.funcargnames:
+        simulator = request.getfuncargvalue('Simulator')
+    if 'nl' in request.funcargnames:
+        nl = request.getfuncargvalue('nl')
+    elif 'nl_nodirect' in request.funcargnames:
+        nl = request.getfuncargvalue('nl_nodirect')
+    analytics = Analytics(simulator, request.module, request.function, nl=nl)
+    request.addfinalizer(lambda: analytics.__exit__(None, None, None))
+    return analytics.__enter__()
 
 
 def function_seed(function, mod=0):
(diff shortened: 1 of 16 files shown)


tbekolay:
Notes from dev meeting:

- Look into [`pytest-regtest`](https://pypi.python.org/pypi/pytest-regtest) (a quick glance at the code makes me think this PR is probably better for our purposes, but worth a closer look)
- Possibly have a separate function for recording "raw" data (stuff in `sim.data`, for example), which is likely to be stored in `.npz`, and for recording "summary" data (statistical summaries of the raw data, for example), which is likely to be stored in `.csv` and appended to on each run
- This and `Plotter` should have its environment variables changed to `pytest` flags
- Possibly the raw data should be stored in a file that incorporates the test name and a timestamp, the csv in a file with the test name, and the timestamp at the start of each row

jgosmann:
pytest-regtest seems to only do a 1:1 match of the recorded data. But I want to get an output of the form improvement/regression/no change.

I am not sure about the usefulness of the timestamp because the mapping from timestamp to actual code is not (necessarily) straightforward. Maybe the following would work?

* Raw data in one directory which includes the timestamp per default or is named via the command line. That makes it easy to load the raw data in an automated way for comparison. It makes it also easy to delete old directories because files from different runs are not intermixed.
* CSV always in the same, but separate directory where the timestamp is just an additional column in the CSV

tbekolay:
I could also see the current git hash (`git rev-parse HEAD`) or the time of the current commit (`git show -s --format=%ci HEAD`) as more useful than the actual current time.

jgosmann:
Which might be misleading in case of uncommited changes ...

jgosmann:
Did the refactoring to reduce code duplication and remove the usage of environment variables. Also, added the possibility to store analytics summary data in a CSV file, but that is untested so far.

To save the plots use `py.test --plots --pyargs nengo`. To save them to some other location than the default use `py.test --plots=path --pyargs nengo`. To append the git commit hash to the path use ``py.test --plots=plots-`git rev-parse HEAD` --pyargs nengo`` (you can create an alias if it's too much to type). Edit that command to include whatever information you want to have in that path.

Same stuff works for the analytics/benchmarks with the `--benchmarks` flag.

jgosmann:
Rebased onto `skipif` (PR #629). That branch should be merged prior to this PR.

jgosmann:
Summary of this updated PR:

* Adding the `analytics` fixture.
* Removing the `plot` and `benchmark` mark and replacing them by a `noassertions` and/or `slow` mark. The `noassertions` test will only run depending on the used fixtures and the `--plots`/`--benchmarks` flags. This is essentially [what I proposed in this comment](https://github.com/nengo/nengo/pull/629#issuecomment-73922750).
* Register marks [as recommended in the py.test documentation](http://pytest.org/latest/example/markers.html). Allows to print them with `py.test --markers`.

At the moment we end up with one test which will never run (`test_subsolvers_L1` in `test_solvers.py`). This test should be changed to save it's output with the analytics fixture instead of just printing it.

jgosmann:
Changed the tests using `print` in `test_solvers.py` to use the `analytics` fixture. Also, updated the changelog. I think, this is ready for review now. :smile:

CHANGES.rst
```diff
@@ -27,6 +27,11 @@ Release History
 - SqrtBeta distribution gives distribution of semantic pointer elements.
   (`#414 <https://github.com/nengo/nengo/issues/414>`_,
   `#430 <https://github.com/nengo/nengo/pull/430>`_)
+- Various improvements of unit testing infrastructure.
+  (`#648 <https://github.com/nengo/nengo/pull/648>`_,
+  `#629 <https://github.com/nengo/nengo/pull/629>`_,
+  `#647 <https://github.com/nengo/nengo/pull/647>`_,
+  `#653 <https://github.com/nengo/nengo/pull/653>`_)
```
hunse: I was originally thinking we could leave these out of the changelog, since they're just testing related, but it's probably best to have them in for completeness.

jgosmann: I'm fine with doing it either way. Maybe we should add another section for stuff which is more relevant to developer than end users?

tbekolay: I think that devs would be reading all or most PRs / Issues / commit logs; the changelog points to the user-relevant subset of those. But I'm not too fussed either way if people want to include them.

nengo/tests/conftest.py
```diff
@@ -111,10 +137,13 @@ def pytest_generate_tests(metafunc):
...
 def pytest_addoption(parser):
-    parser.addoption('--benchmarks', action='store_true', default=False,
-                     help='Also run benchmarking tests')
-    parser.addoption('--plots', action='store_true', default=False,
-                     help='Also run plotting tests')
+    parser.addoption(
+        '--plots', nargs='?', default=False, const=True,
+        help='Save plots (optional with directory to save them in).')
+    parser.addoption(
+        '--benchmarks', nargs='?', default=False, const=True,
```
hunse: I would call this option `--analytics`, since it's not specific to benchmark tests (i.e., it controls whether the analytics fixture does anything in all tests). Also replace "benchmarks" with "analytics" pretty much everywhere else (e.g. "nengo.simulator.benchmarks" becomes "nengo.simulator.analytics").

hunse:
I discussed a few points with @jgosmann, so I'll summarize here. The main change is replacing pretty much every usage of "benchmarks" with "analytics", as I noted on the diff. I think when that is done, this is good to go.

There are a few other things that we might want to change at some point, but aren't mission-critical. For example, when using analytics, there is a .txt file produced to go along with every .npz file and provide some description. If no description is given, this is empty, so it's a bit confusing why it's there. Probably my first choice is just to put the description in the .npz file. Alternatively, I would at least have a little boilerplate in the .txt file, so it's at least a little clear why it's there (and it doesn't seem like some analytics failed to compute).

One other change would be to refine the purpose of the .csv summary file a bit more. My original idea was to have a human readable file that summarizes each of the test runs. This would include human readable time stamps (as opposed to raw `time.time()` values or whatever we have now), and better grouping for the lines (right now, one line is made for each attribute, so you have to look at the time stamp to see what lines go together).

The alternative is to have a machine readable .csv file that is easy to load up in Python or a spreadsheet for quick analysis. This is what @jgosmann was going for in the current iteration. One change I would make for this is would be to put quotes around the description, because otherwise an errant comma could throw off a .csv parser. Also, it seems to me it would be nice to have one line per test (rather than per attribute), but that raises issues about what happens if someone changes the test so it writes different attributes.

Lastly, it's not clear to me whether it should be possible to write dictionaries, arrays, lists, etc. as summary data attributes. These seem to go against the point of summary data, look bad in the .csv file, and aren't generally machine readable (for example by a spreadsheet program). For an example, take a look at `test_solvers.py:test_subsolvers_L2`. That said, should we actually limit what users can write, or simply try to set a good example by only writing scalars in our tests, or go somewhere in between with a warning. Again, the answer to this might depend on whether we're going for a machine-readable or human-readable file.

As I said, though, all these changes (except renaming "benchmarks" to "analytics") can wait for now, unless there's a clear consensus on some of the above and it's quick to do.

drasmuss:
I like putting the description in the npz file, seems like that would be tidiest and easiest to keep track of.

Normally I would favour the machine-readable approach (that's actually what I started out writing here).  But when I think of the use-case for that file, probably all we'll want to do 99% of the time is run the tests and then quickly open the csv to check the results.  It seems unlikely that anyone's going to be writing scripts to parse that file, so it does kind of make sense to focus on human readability.  But like you say, I wouldn't feel so strongly about that to hold things up here.

jgosmann:
* Removed all occurences of benchmark.
* Removed the changelog commit.

Will look into the possibility of storing text in an npz file.

As the npz file is machine readable, it might make sense to use something human readable instead of the CSV file. Does anyone want to propose a format?

The current implementation adds quotes and escaping in the CSV file automatically where necessary.

nengo/tests/conftest.py
```diff
@@ -111,10 +137,13 @@ def pytest_generate_tests(metafunc):
...
-    parser.addoption('--benchmarks', action='store_true', default=False,
-                     help='Also run benchmarking tests')
-    parser.addoption('--plots', action='store_true', default=False,
-                     help='Also run plotting tests')
+    parser.addoption(
+        '--plots', nargs='?', default=False, const=True,
+        help='Save plots (optional with directory to save them in).')
+    parser.addoption(
+        '--analytics', nargs='?', default=False, const=True,
+        help='Also run analytics tests (optional with directory to save ' +
```
hunse: I would say "Save analytics" instead of "Also run analytics tests", analogous to `--plots`.

nengo/tests/conftest.py
```diff
@@ -124,9 +153,21 @@ def pytest_addoption(parser):
...
-            ('plot', 'plots', "plots not requested"),
             ('slow', 'slow', "slow tests not requested")]:
         if getattr(item.obj, mark, None) and not item.config.getvalue(option):
             pytest.skip(message)
+
+    if getattr(item.obj, 'noassertions', None):
+        skip = True
+        skipreasons = []
+        for fixture_name, option, message in [
+                ('analytics', 'analytics', "benchmarks not requested"),
```
hunse: change "benchmarks" to "analytics"

tbekolay:
In terms of "human readable" formats (quotes used because they're not really that readable) there's JSON, YAML and XML. Pros for JSON: part of the standard library, so no new dependencies. Pros for YAML: a bit more readable, in my opinion. Pros for XML: none, IMO. If we're mostly storing a few scalars associated with string keys, I'd say JSON over CSV (since you can add more keys no problem).

hunse:
When I said human-readable, I meant *only* human readable, i.e. just a .txt file with the information printed. Then the question is what the text formatting should be (every attribute on one line, individual lines, etc.). 

I would suggest having everything on one (potentially long) line with the format "<timestamp> - <attribute>: <value>, <attribute>: <value>, ...". This makes it very easy to visually compare whether a value has changed or not. Perhaps `analytics.add_summary_data` could take a format string for the value. The downsides to this are the length of the line and the fact that it doesn't easily allow for description (we could do "<attribute> (<description>): <value>", but then the lines would get super long).

jgosmann:
PR #658 is another demonstration of how to use the analytics fixture.

jgosmann:
fuuuu :scream: I have rewritten history locally ...

tbekolay:
Hmm... how much?

jgosmann:
not sure, it seems you actually pushed over changes I already pushed ... apart from that it is probably only one commit

tbekolay:
Bah, sorry... I pulled this morning; hopefully nothing was lost?

jgosmann:
I still have my version. I think, I can fix it

tbekolay:
Lemme know if any issues arise. I'm pretty sure I didn't touch any commits other than my SQUASHME ones so hopefully it's easy to just cherry-pick those in the appropriate spots in your history!

jgosmann:
Should be fixed. The last commit addresses #669 because I was running in some limitations with the `saveas` approach (when loading analytics data in the `compare` branch). With crazy parametrizations the filename generation will fail (depending on whatever the string representation for the arguments is used), but for simple things it will work.

tbekolay:
Great! I'll take a look. I'm also looking into the test failures for IPython 3 right now, FYI.

I think the only remaining issue for this PR is the summary data format; probably something worth discussing at tomorrow's meeting? I wouldn't be opposed to merging it in with CSV for now though (if we add the quotation marks around the values as @hunse suggested earlier).

jgosmann:
Yeah, I don't really care about the summary format as I am not making use of it at the moment. So I'm fine with whatever people want.
:point_right: The Python CSV implementation already adds quotation marks where needed (and omits them where not needed-.

tbekolay:
> The Python CSV implementation already adds quotation marks where needed (and omits them where not needed

Oh, tight, didn't know that!

tbekolay:
I added a few commits with some additional changes. Summary (see commit messages for more details):

- Removed `plt.saveas` assignments, as `parametrize_function_name` handles it just as well
- Renamed "description" to "documentation" and "desc" to "doc" (this was a spur of the moment thing, happy to revert if people prefer description/desc)
- Use YAML instead of CSV for summary format (seems more readable to me)
- Only save summary/raw data if it exists (i.e., if you don't make any `add_raw_data` calls, it won't save an empty `.npz` file)

Here's the before and after for the summary format. LMK if you think it's an improvement.

CSV file before:

```
timestamp,name,value,doc
1425477873.404712,duration,1.5618319511413574,
1425478028.212792,duration,1.5737082958221436,
1425476656.6173716,duration,1.6549861431121826,how long to solve system
...
```

YAML file after:

```
- timestamp: 2015-03-04 09:55:05
  duration: 1.68086385727
- timestamp: 2015-03-04 09:55:09
  duration: 1.7604172229766846
- timestamp: 2015-03-04 10:11:31
  duration:
    doc: how long to solve system
    value: 1.8306658267974854
```

----

I think this is ready for merge, if people like these changes. #672 has to be merged first though. Oh, and I'm not sure about the reason for the change baafbe1c1a10403988a; @jgosmann can you enlighten me?

jgosmann:
Without baafbe1 `plt.saveas` won't work, because `get_filepath` uses `get_filename` which ignores `self.filename` and `plt.saveas`.

tbekolay:
Ah right you are! Thanks!

jgosmann:
The additional changes LGTM :cake:

tbekolay:
I've been doing a bit more thinking about the summary format, and I'm having second thoughts. Some of the solver tests use the summary data in a way I wasn't expecting; after hacking on it a bit, I thought "why not just have the user construct the string", something more like printing to the analytics object, which would show up in a .txt file. But then I realized we're just rediscovering logging, so perhaps we should just be using logging wherever we're adding the summary data in this branch, or just using what we have for raw data if it's very critical.

We can always revisit this later, but I think for the time being, I'm just going to rip out all of the summary data stuff in this branch and keep the raw data stuff as is. Does that sound good to those that care about this PR?

jgosmann:
Sounds good to me. I'm not using the summary stuff anyways right now. But @hunse might care ...

hunse:
You're right in that what we want for the summary stuff is essentially a logger. I think it would be nice to have a fixture that provides this logger, saving to an appropriately named file, etc. We could either do this as part of the `analytics` fixture, as we do now, or have a separate fixture.

I'm fine taking the summary out for now, but for me personally it is the more useful part of this PR, since it lets me run a test on a bunch of different branches and see how changes to the code affect performance. So I'd like to add something to do that at some point, and I think even an initial hack-ish solution is better than nothing. But if there's conflict on this, there's no point in holding back the rest of the analytics stuff.

tbekolay:
Yeah; I was looking at [pytest-capturelog](https://pypi.python.org/pypi/pytest-capturelog) and I wonder if we can just slightly adapt it to make it possible to spit out logged info to a file. Then we could remove the `nengo.log; pytest.main` block at the bottom of every test file and recommend people run `py.test` manually instead.

tbekolay:
OK, I rebased this to master. With the elimination of the summary stuff, the fixture isn't used anywhere in this code, so I added some additional tests (see the first `SQUASH` commit) to ensure that the fixture works for the time being. With that, I think this is good to merge.

I've looked at the source for pytest-capturelog and it's pretty simple / short; I think we can steal some of the ideas from it and make our own thing for spitting out logs to disk. I'll prep a PR for that as an alternative to the summary data.

hunse:
Squashed things up. Does this look good to everyone? I'm guessing we don't need a changelog entry, since it's just related to tests?

hunse:
> I've looked at the source for pytest-capturelog and it's pretty simple / short; I think we can steal some of the ideas from it and make our own thing for spitting out logs to disk. I'll prep a PR for that as an alternative to the summary data.

Sounds like a plan.

tbekolay:
> I'm guessing we don't need a changelog entry, since it's just related to tests?

Yep, @jgosmann had one originally and we had him remove it.

I forgot to amend the last commit message to remove references to summary data! If you could amend during merge, then I think it's good to go!

hunse:
Merged, but for some reason it's not letting me delete the branch.

tbekolay:
Weird... check out the [branches pages](https://github.com/nengo/nengo/branches); it says "You can't delete this branch because an open pull request depends on it"... weird!

jgosmann:
That would be #657.

hunse:
Ahh, I see. Well I guess we should rebase #657 to master, then, or rather make the PR from `compare` into `master` rather than from `compare` into `analytics`.

jgosmann:
Yeah, the branch has to be rebased and then the PR has to be changed to be against `master`.

PR #648, opened 2015-02-07
Better, faster .travis.yml

tbekolay:
Short version: with this PR, builds go from [8 - 14 minutes](https://travis-ci.org/nengo/nengo/builds/49638662) (with plenty of random failures) to [~3 minutes](https://travis-ci.org/nengo/nengo/builds/49803517) (haven't seen any random failures yet; `flake8`, which is usually what fails, runs in 26 seconds). They're also guaranteed to start faster, due to using recent TravisCI infrastructure upgrades.

----

Long version: This PR enables TravisCI to build and test Nengo faster. No functionality is lost; we still test on Python 2.6, and [coveralls info](https://coveralls.io/builds/1880935) is still maintained.

The two most significant ways in which we make TravisCI build faster are to

1. use `sudo: false` to enable container-based builds as described in [this blog post](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/) and [these docs](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) and
2. use Miniconda to download and install binaries in the home directory, so that we don't need to use `sudo`, and don't have to build anything from source.

Note that, since we're doing `sudo: false`, we could also enable caching; however, since we don't actually do any building from source (all the binaries are downloaded from `conda`; we only install a few things from `pip` and they're pure Python) we won't gain much in terms of speed because the cache is stored on Amazon S3, so it'd still have to be downloaded at the start. Plus, doing it this way without the cache ensures that we're testing with the most recent stable versions of these things.

A few additional things that I haven't yet done, but could do if people think they're worthwhile:

- Test with multiple versions of NumPy (we should maybe test with at least the earliest version we claim to support?)
- Test on Mac OS X (TravisCI has some Mac OS X VMs that we can use; you can't normally use them for Python, but since we install Python with `conda` we can use them)
  - Before anyone asks, they don't have any Windows VMs, but I'm going to make another PR to enable Nengo to be tested with [appveyor](http://www.appveyor.com/), which is basically TravisCI for Windows
- Add another target that builds the documentation to ensure there are no warnings (we can add in some more checkers to this later as well)
- Other ideas?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,3 @@
 [run]
-omit = nengo/tests/test*
+source = nengo
+omit = */tests/test*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,31 +1,58 @@
-language: python
-python:
-  - 2.6
-  - 2.7
-  - 3.3
-  - 3.4
-env:
-  global:
-    - PIP_WHEEL_COMMAND="pip install --find-links http://wheels.astropy.org/ --find-links http://wheels2.astropy.org/ --use-wheel --use-mirrors"  # Thanks astropy!
-matrix:
-  include:
-    - python: 2.7
-      env: TOXENV=pep8
-before_install:
-  - pip install --upgrade pip setuptools  # For wheel support
-install:
-  - "echo 'backend : Agg' > matplotlibrc"
-  - if [[ "$TOXENV" == "pep8" ]]; then pip install tox; fi
-  - if [[ "$TOXENV" != "pep8" ]]; then $PIP_WHEEL_COMMAND -r requirements.txt; fi
-  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.6" && "$TOXENV" != "pep8" ]]; then $PIP_WHEEL_COMMAND -r requirements-test-py26.txt; fi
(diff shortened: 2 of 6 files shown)


hunse:
+1 for testing other Numpy versions. We're currently failing on 1.7, but it might be just the way that some of the tests are written, more than actual Nengo (`np.linalg.norm` doesn't have the `axis` argument until around 1.8, I think, and we use that).

tbekolay:
OK, I've made the tests a tiny bit faster using `language=c`, and am now testing NumPy versions 1.6-1.9. I fixed the tests that failed on earlier NumPy versions.

I also added a commit to test that docs build. The doc building isn't perfect because I'm doing this through `python setup.py build_sphinx`, which doesn't support `-v` and `-W`, which would increase verbosity (I'm worried about the tests timing out due to no output) and fail on warnings, both of which would be nice. I'm making a quick PR for Sphinx to add those options to `build_sphinx`. I can either take that commit out of this PR until this works better, or leave it in as is, whatever people prefer!

hunse:
Wow, that is **way** faster! I would take out the documentation build test for now, and merge everything else.

tbekolay:
Yeah, I was pretty surprised myself! Props to Docker and Conda. I removed the doc building commit, and squashed the `language=c` one, so I think it's good to go!

hunse:
`flake8` used to take 7-12 minutes, now it takes 19 seconds. Just sayin'.

Issue #622, opened 2015-01-15
Possible proposal for an addition to the Node API


tcstewar:
As part of #616 and other previous verbal discussions, now and then it seems like people want to do something with the Node interface that isn't quite nicely supported by the current API.  Here is a proposal that I'm not sure I fully support, but I thought it might address some of the things people seem to want.

Currently, a Node's ```output``` value can either be a callable function or an array-like.  If it's an array-like, the Node always outputs that value.  If it's callable, then the backend should call the function regularly, passing in t (the current time) and optionally x (the sum of the inputs to the Node, if any).

As people have used Nodes for different things, there have been a few situations where people have wanted a little bit more power here.  Right now, there have been fairly straighforward workarounds, and so I'm not yet convinced that we need to add complexity to the Node API.  But, there have been three things people have expressed interest in:
 - responding in some way to reset() (which you can do right now by just detecting when t jumps back to 0)
 - getting access to the Simulator and Model object (which you can access now via manually passing it into the function)
 - getting access to the model's random number generator and/or seed (which you can access now via passing it into the function).

One way we could address all of this is by allowing people to set ```output``` to be a function factory instead of a function (or an arraylike).  That is, people could do this:

```python
def noise_maker(sim, model, rng):
    gain = rng.uniform(0, 1)
    freq = rng.uniform(0.5, 80.0)
    def my_func(t):
        return gain * np.sin(freq * t)
    return my_func
node = nengo.Node(noise_maker)
```

The backend would, when a simulation starts (or when a reset occurs), call this function, and then use the function that it returns from then on.  In this particular case, the result is a sine wave of a different frequency and amplitude, based on the seed of the Simulator.

This seems like a relatively simple change to the Node API.  It does complicate matters a fair bit for the GUI when it goes in to modify Nodes to allow for controllable sliders overriding the Node input values, but I think it's doable, if we decide there's significant advantage to having something like this over the existing Node system.  That said, I don't know what the use cases are.  I think this might address some of the aspects of #616, but I also think those aspects can be addressed reasonably well with the existing Node system.  This proposal here might be a bit cleaner, but if that's all we get out of it then I don't think we should add the complexity.

In any case, I thought I'd put this idea out there and see what people think.  In particular
 - Does this cover use cases that people want that are difficult with the current Node API?  What are those use cases?
 - How should the Node builder detect that it has been given a generator rather than a normal function?
 - Are there other parameters that the builder should pass in?
 - Are there completely different other proposals?

[...]

Issue #582, opened 2014-12-10
Reset system and rerun with different whitenoise signal


studywolf:
I was wanting to run a single simulation a bunch of times, with the same neurons but with random noise that was different between runs. ```sim.reset()``` also reset the noise, so I used this workaround: 

```
other_t = np.random.random()

with m:

    ...

    from nengo.utils.functions import whitenoise
    noise_func = whitenoise(0.1, 10, .001, dimensions=2)
    def func(t):
        print other_t
        return noise_func(t + other_t)
    noisenode = nengo.Node(output=func)

    ...

sim = nengo.Simulator(m)
for ii in range(50):
    sim.reset()
    other_t = np.random.random()
    sim.run(.75)
```

So this is mostly just a workaround post in case someone is looking to do this. It would be cool if there was a less hack way to do this as well!

PR #652, opened 2015-02-11
Updated Process interface

hunse:
As discussed in #622. See the first commit message for a summary of changes (there are a number).

The second commit fixes up the simulator reset so that it resets Processes.

For discussion, there are a number of name changes, so I want to make sure everyone is okay with those. Also, I wasn't sure whether we should try to "deprecate" the old `StochasticProcess.f` function so that 2.0 code can still run. I don't think it's quite as easy as just having that function return a reference to the process itself, so I would probably not bother with it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,12 @@ Release History
   ``min_voltage``. Voltages are clipped such that they do not drop below
   this value (previously, this was fixed at 0).
   (`#666 <https://github.com/nengo/nengo/pull/666>`_)
+- ``Process`` objects can now be passed directly as node outputs,
+  making them easier to use. The ``Process`` interface is also improved
+  and is currently the same as the ``Synapse`` interface. However,
+  further improvements are pending, and the current implementation
+  SHOULD NOT BE RELEASED!
+  (`#652 <https://github.com/nengo/nengo/pull/652>`_)
 
 **Behavioural changes**
 
@@ -50,6 +56,13 @@ Release History
 - Added ``utils.connection.eval_point_decoding`` function, which
   provides a connection's static decoding of a list of evaluation points.
   (`#700 <https://github.com/nengo/nengo/pull/700>`_)
+- Resetting the Simulator now resets all Processes, meaning the
+  injected random signals and noise are identical between runs,
+  unless the seed is changed (which can be done through
+  ``Simulator.reset``).
+  (`#582 <https://github.com/nengo/nengo/pull/582>`_,
+  `#616 <https://github.com/nengo/nengo/pull/616>`_,
+  `#652 <https://github.com/nengo/nengo/pull/652>`_)
 
 **Bug fixes**
 
--- a/examples/advanced/nef_summary.ipynb
+++ b/examples/advanced/nef_summary.ipynb
@@ -316,11 +316,11 @@
      "cell_type": "code",
      "collapsed": false,
      "input": [
-      "from nengo.processes import WhiteNoise\n",
+      "from nengo.processes import WhiteSignal\n",
       "\n",
       "model = nengo.Network(label=\"NEF summary\")\n",
       "with model:\n",
-      "    input = nengo.Node(WhiteNoise(1, high=5).f(rng=np.random.RandomState(60)))\n",
+      "    input = nengo.Node(WhiteSignal(1, high=5), size_out=1)\n",
       "    input_probe = nengo.Probe(input)\n",
       "    A = nengo.Ensemble(30, dimensions=1, max_rates=Uniform(80, 100))\n",
       "    nengo.Connection(input, A)\n",
@@ -376,7 +376,7 @@
      "input": [
       "model = nengo.Network(label=\"NEF summary\")\n",
       "with model:\n",
-      "    input = nengo.Node(WhiteNoise(1, high=5).f(rng=np.random.RandomState(60)))\n",
+      "    input = nengo.Node(WhiteSignal(1, high=5), size_out=1)\n",
       "    input_probe = nengo.Probe(input,)\n",
       "    A = nengo.Ensemble(30, dimensions=1, max_rates=Uniform(80, 100))\n",
       "    Asquare = nengo.Node(size_in=1)\n",
--- a/examples/learning/learn_communication_channel.ipynb
+++ b/examples/learning/learn_communication_channel.ipynb
@@ -40,7 +40,7 @@
       "%matplotlib inline\n",
       "\n",
       "import nengo\n",
-      "from nengo.processes import WhiteNoise"
+      "from nengo.processes import WhiteSignal"
      ],
      "language": "python",
      "outputs": []
@@ -51,7 +51,7 @@
      "input": [
       "model = nengo.Network()\n",
       "with model:\n",
-      "    inp = nengo.Node(WhiteNoise(60, high=5).f(d=2))\n",
+      "    inp = nengo.Node(WhiteSignal(60, high=5), size_out=2)\n",
       "    pre = nengo.Ensemble(60, dimensions=2)\n",
       "    nengo.Connection(inp, pre)\n",
       "    post = nengo.Ensemble(60, dimensions=2)\n",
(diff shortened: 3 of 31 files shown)


tcstewar:
Hmm, I'm not going to get a chance to look at this in detail until the weekend, but just to check, why did you end up going with the PythonClosure-state approach rather than the PythonClass-state approach?  I'd thought that's where the consensus was after the #622 discussion.  Also, why doesn't the ```step``` function take ```t``` as an input?  Isn't that necessary if it's run on a Simulator that doesn't guarantee a dt?

One thing to keep in mind is that we don't have to merge Process and NodeOutput.  If they do turn out to be close enough to be the same thing, then that can simplify things, but if there are particular reasons that the Process implementation you're proposing differs from the desired NodeOutput stuff, then we should keep them separate.  I still think they're pretty close, but I'm not as familiar with the inner workings of the Process stuff.

hunse:
> why did you end up going with the PythonClosure-state approach rather than the PythonClass-state approach? I'd thought that's where the consensus was after the #622 discussion.

I think you're right, we were leaning towards the PythonClass-state approach. I had actually forgotten that when I did it, and the closure approach is also what the synapse classes use, so there's a nice parallel there.

> Also, why doesn't the ```step``` function take ```t``` as an input?  Isn't that necessary if it's run on a Simulator that doesn't guarantee a dt?

Yes, I guess we should add that.

> One thing to keep in mind is that we don't have to merge Process and NodeOutput.

Very true. I did this PR based on what seemed to make sense for the existing Processes, without thinking much about how people might like to use them as node outputs. As it stands, what I have here should be quite powerful for doing various types of node outputs. But maybe we want a different interface; we could take the class approach for NodeOutputs, and the closure approach for Processes.

s72sue:
I was wondering when this will be reviewed and whether the Process interface in nengo2 is going to be updated? I need this interface to translate the HBB tutorials from nengo1.4 to nengo2.

tcstewar:
> I was wondering when this will be reviewed and whether the Process interface in nengo2 is going to be updated? I need this interface to translate the HBB tutorials from nengo1.4 to nengo2.

This will get reviewed eventually, but things are still up in the air about the approach we will take.  The PythonClosure approach is surprising to novice programmers, and the PythonClass approach is surprising to experienced programmers.  I'm still hoping for a clear winner one way or the other.

That said, it shouldn't be necessary for translating the HBB tutorials, as the WhiteNoise object still works as it is.  Once this PR gets sorted out, we should be able to clean up the tutorial a bit, but it should work fine without it (unless I'm not understanding the problem).

hunse:
The problem is, @s72sue needs *actual* white noise for the tutorials, to inject to neuron currents. This is something that I added in this PR.

In my opinion, we should go ahead and review and merge this, even if it's not the approach we ultimately want to take. I think it's considerably better than what we have now, and the user-facing side will look the same even if we later add a commit to change to the class approach (of course, if users have written their own processes, those would have to change, but we would have this resolved before 2.1 either way, so users would never have to see the intermediary step).

That said, I'm happy to split out parts of this commit if we really want to wait to decide. I'm just getting a little frustrated with some of my PRs sitting around for a long time, and getting stale, and then having to be rebased again, as well as not being able to use the improvements in further PRs and projects, etc.

tcstewar:
Yes, if this is causing frustration with how long it's taking, then we should definitely move ahead and sort something out with it quickly.  I'd taken the lack of discussion on this thread as an indication that it wasn't a high priority for anyone, and then lots of other things came up.

Can we meet tomorrow (thursday) in person and talk it through?  Basically, I want to get a clear understanding on what people making their own processes and what backend developers need to know about this system.  I definitely agree that it's be good to get something merged in soon, but I also think we need to be clear on what this means for backends too (both in terms is supporting Processes as noise input to neurons and in terms of supporting custom Processes as part of Nodes, if we want to keep that part).

tbekolay:
I'm available tomorrow to talk about it. Whenever!

hunse:
@tcstewar and I were thinking 10:00 am.

tbekolay:
Kay, sounds good.

hunse:
Okay, rebased and ready for review.

hunse:
Okay, made those changes! They took a little longer than I thought, but I figured out that WhiteSignal is actually naturally periodic (some weird Fourier stuff going on), so we don't need to do all that reflection magic, we can just loop through it periodically. This means that I'm less unhappy with the WhiteSignal now. I renamed `duration` to `period`, and if users pick a period that is shorter than their simulation it just means the signal will start to repeat, but nice and continuously so it should look okay.

celiasmith:
Is white allowing different kinds of white (e.g. Gaussian/Uniform?)

Eric Hunsberger wrote:

hunse:
WhiteNoise allows for different distributions (Gaussian, Uniform, etc., even UniformHypersphere). WhiteSignal, though, is the one that we generate in the Fourier domain, and I've never quite figured out the relationship between different white noise distributions and how they appear in the Fourier domain.

hunse:
Okay, added another test to make sure things stay deterministic in Travis CI. I think this is ready to go, fingers crossed.

tbekolay:
Cool, will quickly re-review and presumably merge this evening or early tomorrow morning.

tcstewar:
Looking good to me.

As a note to those following this discussion, the intent here is to get this merged in as it greatly improves the ability to control and generate noise in models.  Right now, it also allows you to attach these noise Processes to a Node and use it for generating signals.  The API for doing this is still a work in progress, and this PR is meant as a stepping stone in that direction.

One thing to keep in mind is that we should *NOT* do an official release until we get the Node part of this sorted out.  But this PR looks good to me as a step in the right direction.

tbekolay:
That reminds me actually, one thing that's missing is a changelog entry. It's a user-facing change so we should make one (or multiple in this PR's case) even if, before the release, we have the change or replace one of those entries.

I'll add them in the merge if they don't magically appear before I review / merge :rainbow:

hunse:
What's our general policy on the changelog? Is it the author's job or the merger's job to write it? I think we had said at one point it's the merger's job, but I think it makes more sense if the author does it, just so we have two pairs of eyes look at it just like everything else. My problem is I always forget.

hunse:
Okay, added changelog entries, and also referenced #582 in the commit message of 1d040f0.

jgosmann:
> WhiteSignal, though, is the one that we generate in the Fourier domain, and I've never quite figured out the relationship between different white noise distributions and how they appear in the Fourier domain.

From what I learned from Wikipedia it's a special property of Gaussian white noise vectors that they stay Gaussian white noise vectors under orthogonal transformations. That would mean the generation in the Fourier transform is probably only possible for Gaussian white noise.

hunse:
> That would mean the generation in the Fourier transform is probably only possible for Gaussian white noise.

But you can generate for example a uniform white noise signal and take the Fourier transform and get something, so it seems to me that there must be something in the frequency domain that unites all uniform white noise signals and makes them different from Gaussian white noise. That relationship could be quite complicated, though. Also, I'm not sure if this is at all useful to people practically, so I wouldn't worry too much about it. But it is an interesting question.

tbekolay:
This looks awesome! Made one FIXUP commit to not use mutable object as default args (see 6fd3682a902d7410b80ca17efca878bd0a1e30d9). Also made an issue to possibly make these immutable in the future (#719). @hunse LMK if the FIXUP commit looks good; if so I'll fixup and merge!

hunse:
Looks good. Having immutable objects would definitely be nice.

s72sue:
Hey Eric,

I was wondering how I can generate a uniform white noise i.e., chosen
uniformly from (-10, 10) ?  Based on what you are saying it seems that I
need to use WhiteNoise for this. But looking at the code, it looks like I
should be using WhiteSignal with a period of 20 and high=10.

Can you please clarify?

-Sue

hunse:
No, you definitely want WhiteNoise. Something like `nengo.processes.WhiteNoise(dist=nengo.dists.Uniform(-10, 10))`. You might have to play around with the amplitude, though, to get the results to be reasonable/simiar to Nengo 1.4. You want the range of your noise to be on the same order of magnitude as the signal you're adding it to. In your case, you're adding it to the neuron input current, so you probably want a range of [-1, 1] or even a bit smaller.

s72sue:
hmm..okay, thanks.

PR #653, opened 2015-02-17
Add pytest norecursedirs setting.

### LINKED ISSUE #650: py.test fails with existing build dir
My `py.test` was failing with the following error:
```
(nengodev)ctn02:nengo|better-product> py.test                                                                                            ? d31f153795
================================================================ test session starts =================================================================
platform linux2 -- Python 2.7.9 -- py-1.4.26 -- pytest-2.6.4
collected 0 items / 1 errors 

======================================================================= ERRORS =======================================================================
_________________________________________________________________ ERROR collecting . _________________________________________________________________
../../../.pyenv/versions/2.7.9/lib/python2.7/site-packages/py/_path/common.py:334: in visit
    for x in Visitor(fil, rec, ignore, bf, sort).gen(self):
../../../.pyenv/versions/2.7.9/lib/python2.7/site-packages/py/_path/common.py:380: in gen
    for p in self.gen(subdir):
../../../.pyenv/versions/2.7.9/lib/python2.7/site-packages/py/_path/common.py:370: in gen
    if p.check(dir=1) and (rec is None or rec(p))])
../../../.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/main.py:628: in _recurse
    ihook.pytest_collect_directory(path=path, parent=self)
../../../.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/main.py:166: in call_matching_hooks
    plugins = self.config._getmatchingplugins(self.fspath)
../../../.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/config.py:688: in _getmatchingplugins
    plugins += self._conftest.getconftestmodules(fspath)
../../../.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/config.py:521: in getconftestmodules
    mod = self.importconftest(conftestpath)
../../../.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/config.py:545: in importconftest
    raise ConftestImportFailure(conftestpath, sys.exc_info())
E   ConftestImportFailure: (local('/home/jgosmann/Documents/projects/nengo/nengo/tests/conftest.py'), (<class 'py._path.local.ImportMismatchError'>, ImportMismatchError('nengo.tests.conftest', '/home/jgosmann/Documents/projects/nengo/build/lib/nengo/tests/conftest.py', local('/home/jgosmann/Documents/projects/nengo/nengo/tests/conftest.py')), <traceback object at 0x7f5ed0553b48>))
```
After I did `rm -r build` it was working again. I think, the existence of that directory should make no difference (and if it has to, the error message is not helpful at all).

Not sure if it has something to do with me using `pyenv` and `virtualenv`.

jgosmann:
Fixes #650 by preventing py.test from looking into the build directory.
Also added some other directories where there should be no tests.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,3 +13,6 @@ ignore =
 
 [upload_sphinx]
 upload-dir = doc/_build/html
+
+[pytest]
+norecursedirs = .* *.egg build dist docs


setup.cfg
```diff
@@ -13,3 +13,6 @@ ignore =
 
 [upload_sphinx]
 upload-dir = doc/_build/html
+
+[pytest]
+norecursedirs = .git build dist docs
```
hunse: It would be nice to *only* look in `nengo` for tests, since right now it is the only directory with tests, but I can't find a way to do this.

hunse:
The default "norecursedirs" ignores all hidden folders and eggs (see [the docs](http://pytest.org/latest/customize.html)). Those seem like good things to ignore, so I've added them in. Otherwise looks good to me.

jgosmann:
I was looking at some other page of the documentation where the default was not stated ... but definitely a good idea to keep those values.

hunse:
Yeah, there are a few other default values (namely `'CVS', '_darcs', '{arch}'`) but we obviously don't need `CVS`, and I'm not sure what the other ones are for. Maybe we should keep them just in case?

jgosmann:
[`_darcs`](http://darcs.net/)
[`{arch}`](https://www.gnu.org/software/gnu-arch/)
They really put the most obscure RCS in there :octopus:

hunse:
Yeah, weird that they have those and not Git or SVN. Okay, if that's all those are, then this LGTM.

Issue #613, opened 2015-01-13
Specifying Probe outputs for high-speed export of data from a simulation


tcstewar:
While the default Probe system stores data in ```sim.data[probe]```, it would be useful to have a variety of options of where to send Probe data.  Backends would be free to implement this in a backend-specific manner, but the reference Nengo backend should provide useful functions that other backends might use in their implementations.  Importantly, the implementation should be focused of high-speed export of data from a running simulation (unlike the Node system, which is meant for generic input and output).

The following types of outputs are desired (all subclasses of ProbeOutput):
 - ProbeFile (data is appended to a .csv or .hdf5 file)
 - ProbeTexture (data is sent to an OpenGL texture)
 - ProbeUDP (data is sent over the network via UDP)

The suggested API is that Probes gain a new attribute ```output``` which can be set to a ProbeOutput.  The default value of "None" indicates the current ```sim.data[probe]``` approach should be used.  Multiple Probes might have the same ```output``` value, indicating this data should be merged as appropriate.  The ```label``` on a Probe can be used as an identifier for this merging (for example, the header of a ```.csv``` file).

Minimal example:

```python
model = nengo.Network(label='Minimal Example')
with model:
    a = nengo.Ensemble(100, 1, label='a')
    b = nengo.Ensemble(100, 1, label='b')

    file = nengo.ProbeFile()  # filename defaults to '%(model_name)s-%(timestamp)s.csv'
    nengo.Probe(a, output=file) # label defaults to a.label
    nengo.Probe(b, output=file) # label defaults to b.label
```

You should also be able to use the Config system:
```python
model = nengo.Network(label='Minimal Example')
model.config[Probe].output = nengo.ProbeFile()
with model:
    a = nengo.Ensemble(100, 1, label='a')
    b = nengo.Ensemble(100, 1, label='b')

    nengo.Probe(a)
    nengo.Probe(b)
```

[...]

PR #654, opened 2015-02-17
Probe outputs

hunse:
This sets up the framework for probe outputs as discussed in #613. The idea is that various subclasses of ProbeOutput can indicate whether a probe writes to a buffer, file, plot, etc. This PR sets that up by making Probes and ProbeOutputs things that are built, so it's easy to specify a builder function for different types of ProbeOutputs. Currently, there is just ProbeBuffer, which probes to a list like our old probes did. 

I've put this on top of my Process stuff because they both use the new `reset` function for Simulator, but otherwise it's independent of that PR.

### DIFF
--- a/examples/delay_node.ipynb
+++ b/examples/delay_node.ipynb
@@ -28,7 +28,7 @@
       "%matplotlib inline\n",
       "\n",
       "import nengo\n",
-      "from nengo.processes import WhiteNoise"
+      "from nengo.processes import WhiteSignal"
      ],
      "language": "python",
      "outputs": []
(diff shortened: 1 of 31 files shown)


jgosmann:
This PR is quite old. Not sure if it makes sense to review this before it gets rebased?

jgosmann:
I was trying to rebase this PR, but it is so old that it seems easier to just start over and reimplement this functionality.

tbekolay:
I'm going to close this as a revise and resubmit for the time being. Please **do not** delete the branch associated with this PR, as it will be a good reference / starting point for a future PR. The underlying issue here is currently under discussion at https://github.com/nengo/enhancement_proposals/pull/5 also, for those interested in this functionality.

PR #655, opened 2015-02-18
Ignore PEP8 error for binary op linebreaks

hunse:
Error W503 occurs when a line break occurs before a binary operator
instead of after it. We were ignoring this before `flake8` checked
it, so let's continue ignoring it.

See brief discussion at end of #640.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,5 +20,5 @@ commands = flake8 -v nengo
 
 [flake8]
 exclude = __init__.py
-ignore = E123,E133,E226,E241,E242,E731
+ignore = E123,E133,E226,E241,E242,E731,W503
 max-complexity = 10


tbekolay:
I'm in disagreement with PEP8 on this warning for sure (having the op at the start of the line is easier to read) so in favour of ignoring this. Merging.

jgosmann:
I never understood what the reason for that choice was ... (though, I usually follow it for the sake of consistency)

tbekolay:
I forget where I originally read it, but [this SE post](http://programmers.stackexchange.com/a/160647) does a good job explaining it... basically, in understanding something, the op is more salient than the LHS or RHS; in particular, the RHS needs the op for context, so separating the two is usually more confusing than helpful.

jgosmann:
Sorry, I meant the PEP8 choice with "that choice". To me breaking before the op seems more natural too. The SE post also does not really provide any good reasons for the PEP8 style. Basically most people seem to agree with us and there seem to be only two reasons for having the line break after the operator:

1. It follows English punctation.
2. It prevents errors with premature line endings.

I disagree with 1., because you would write a caculation like
```
  123
+ 456
-----
  579
```
and not like
```
123 +
456
-----
579
```

Point 2. could happen, but only if the indentation of next line matches the previous line and `pylint` would still warn you about it.

PR #657, opened 2015-02-23
Comparison tests for analytics data.

jgosmann:
This is a follow-up PR to #647.

It allows to mark test functions with `pytest.mark.compare` and indicates that the function will compare the data from two test runs producing analytics data. A test marked with `compare` will not run per default. Actually, it will be pruned from the collected tests and does not influence the success/skipped/failure statistics.

To run the comparison tests one has to pass `--compare [old-data] [new-data]` to `py.test`. This will prune all normal tests and only collect the comparison tests.

## How to write a comparison test

1. Mark it with `pytest.mark.compare`.
2. Name it `test_compare_<testname>` given that `test_<testname>` produced the relevant data. This naming scheme is necessary to automatically load the relevant analytics data. It would be possible to add the mark based on the test name, but maybe that's too much magic. It would also be possible to get rid of the `test_` prefix (but again it would be a bit more magic).
3. Add the `analytics_data` fixture as function argument. It will be a two element sequence with the same order as the arguments to `--compare`. By conventions the first one should be the older data.
4. Print some useful information (e.g., "X got Y% more accurate.").
5. Do an `assert` whether the metric improved. That way a failure will be reported when something got worse.

All output to stdout and stderr will also be shown for successful tests as there might be useful information (like by how much something got better).

I'll prepare another PR demonstrating such a comparison test.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,13 @@ Release History
 2.1.0 (unreleased)
 ==================
 
+**API changes**
+
+- Spiking ``LIF`` neuron models now accept an additional argument,
+  ``min_voltage``. Voltages are clipped such that they do not drop below
+  this value (previously, this was fixed at 0).
+  (`#666 <https://github.com/nengo/nengo/pull/666>`_)
+
 **Behavioural changes**
 
 - The ``probeable`` attribute of all Nengo objects is now implemented
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -223,14 +223,19 @@ class LIFRate(NeuronType):
 class LIF(LIFRate):
     """Spiking version of the leaky integrate-and-fire (LIF) neuron model."""
 
+    min_voltage = NumberParam(high=0)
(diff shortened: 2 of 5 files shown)


jgosmann:
Based on PR #647 and #657 I implemented a benchmark for the product network in PR #658.

Suprisingly it shows a much larger improvement than I expected:
```Multiplication improvement by 0.310131 (84%, p < 0.001)```

hunse:
Is this something that `py.test` suggested, or something you cooked up yourself? Either way, it's pretty cool, and something I never would have thought of doing with `py.test`.

I don't think I'd like to merge this without having #658 as well, though, since it's pretty essential in my opinion to have an example. So maybe we could just bundle this into #658?

jgosmann:
I came up withit myself. I am ok with bundling those PRs. We could even bundle it with #647. Or we can merge that first and then the other two. Or we can bundle the first commit of #658 with #647 and the other commit with this PR.

PR #660, opened 2015-02-24
Added Triangle synapse

hunse:
Though this FIR filter is not terribly efficient on normal hardware, it is apparently quite fast on dedicated digital hardware like DSPs.


tbekolay:
This LGTM! Added a test that makes a simulator, to ensure that works; also made some changes to synapse test naming, which made sense to me at the time. Also added a changelog entry. If these look good @hunse, I'll merge.

hunse:
I combined the triangle synapse tests into one, and renamed the tests. Maybe it was a bit too much work just to avoid those names, but I think it's a bit cleaner now. Otherwise looks great.

tbekolay:
Cool, works for me!

PR #666, opened 2015-02-26
Added min_voltage parameter to nengo.LIF

tcstewar:
This is needed for one version of the adaptive LIF neuron.  As mentioned https://github.com/nengo/nengo/pull/644#commitcomment-9593615 it was felt that having a parameter like this might also be useful for hyperpolarizing neurons too.

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -223,14 +223,19 @@ class LIFRate(NeuronType):
 class LIF(LIFRate):
     """Spiking version of the leaky integrate-and-fire (LIF) neuron model."""
 
+    min_voltage = NumberParam(high=0)
     probeable = ['spikes', 'voltage', 'refractory_time']
+
+    def __init__(self, tau_rc=0.02, tau_ref=0.002, min_voltage=0):
+        super(LIF, self).__init__(tau_rc=tau_rc, tau_ref=tau_ref)
+        self.min_voltage = min_voltage
 
     def step_math(self, dt, J, spiked, voltage, refractory_time):
 
         # update voltage using accurate exponential integration scheme
         dV = -np.expm1(-dt / self.tau_rc) * (J - voltage)
         voltage += dV
-        voltage[voltage < 0] = 0  # clip values below zero
+        voltage[voltage < self.min_voltage] = self.min_voltage
 
         # update refractory period assuming no spikes for now
         refractory_time -= dt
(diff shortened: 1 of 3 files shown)


mundya:
This adds to the "Nengo should have a spec document for backends" call.  Noted that I need to implement this.

tcstewar:
> This adds to the "Nengo should have a spec document for backends" call. Noted that I need to implement this.

Yup, we're definitely needing that sort of document.  I'm in the midst of doing that to a certain degree with this https://github.com/ctn-waterloo/nengo_reference backend by identifying a set of nengo tests that all backends should pass.  But there's also going to be optional features that a backend may or may not support (in which case there should also be ways of the backend indicating what it supports and doesn't support).  And it means we have to be much more careful about adding features that affect backends.

For this particular feature, I'd say it's completely optional for a backend to support it.  :)

nengo/neurons.py
```diff
@@ -225,12 +225,16 @@ class LIF(LIFRate):
 
     probeable = ['spikes', 'voltage', 'refractory_time']
 
+    def __init__(self, min_voltage=0, **kwargs):
```
hunse: This changes the order of parameters for LIF. Are we sure that no-one is doing `LIF(0.03, 0.003)` without the keywords? If they are, is it their own damn fault if this breaks their code?

tbekolay: We shouldn't change the order of parameters. And in general, I don't think we should use `**kwargs` if we can avoid it (and we can in this case)

hunse: Agreed. I'll make a fix.

hunse:
Okay, changed the parameter order. LGTM now.

tbekolay:
Cool, I'll take a quick look and merge. I will probably add some labels to the plots in `test_neurons`...

hunse:
I just added a maximum to `min_voltage`, making sure it doesn't go over 0. I think this makes sense since we reset the voltage to 0 after a spike, so it's kind of weird to have a `min_voltage` larger than zero. Does everyone agree?

tbekolay:
I think that makes sense assuming that the reset voltage is fixed. We could make that configurable too, though I'm not sure of any reason why anyone would particularly want to do that.

tbekolay:
Rebased to master, refactored the test and added a changelog entry. Will merge if @hunse and @tcstewar approve!

hunse:
I approve.

tcstewar:
Looks great to me.  Thanks for polishing this up!  :)

Issue #667, opened 2015-02-26
Error probing transform on connection


drasmuss:
To replicate:
```python
import nengo

with nengo.Network() as net:
    input = nengo.Node([0])

    a = nengo.Ensemble(10, 1)
    conn = nengo.Connection(input, a)
    
    nengo.Probe(conn, "transform")
    
sim = nengo.Simulator(net)
```
gives error:
```
Traceback (most recent call last):
  File "D:\Documents\officesvn\code\test.py", line 11, in <module>
    sim = nengo.Simulator(net)
  File "D:\Documents\nengogit\nengo\nengo\simulator.py", line 111, in __init__
    self.model.build(network)
  File "D:\Documents\nengogit\nengo\nengo\builder\builder.py", line 34, in build
    return Builder.build(self, *objs)
  File "D:\Documents\nengogit\nengo\nengo\builder\builder.py", line 74, in build
    cls.builders[obj_cls](model, obj, *args, **kwargs)
  File "D:\Documents\nengogit\nengo\nengo\builder\network.py", line 76, in build_network
    model.build(probe)
  File "D:\Documents\nengogit\nengo\nengo\builder\builder.py", line 34, in build
    return Builder.build(self, *objs)
  File "D:\Documents\nengogit\nengo\nengo\builder\builder.py", line 74, in build
    cls.builders[obj_cls](model, obj, *args, **kwargs)
  File "D:\Documents\nengogit\nengo\nengo\builder\probe.py", line 76, in build_probe
    synapse_probe(model, key, probe)
  File "D:\Documents\nengogit\nengo\nengo\builder\probe.py", line 37, in synapse_probe
    sig = sig[probe.slice]
  File "D:\Documents\nengogit\nengo\nengo\builder\signal.py", line 178, in __getitem__
    return self.__getitem__((item,))
  File "D:\Documents\nengogit\nengo\nengo\builder\signal.py", line 151, in __getitem__
    start, stop, stride = idx.indices(shape[ii])
IndexError: list index out of range
```

The problem is that the default transform is a zero dimensional array, so it breaks if you try to slice it (even if it's the default null slice).

hunse:
Zero-dimensional arrays are the bane of my existence :rage1: :rage2: :rage3: :rage4:

hunse:
We could probably just give transforms an `ndmin` of 1 or 2 in the parameter. This might actually simplify the code in some places.

tbekolay:
Fixed in #671.

Issue #665, opened 2015-02-25
Probing decoders on a non-decoded connection


drasmuss:
Here's a simple script to illustrate the problem:
```python
import nengo

with nengo.Network() as net:
    a = nengo.Node([0])
    b = nengo.Ensemble(100, 1)
    c = nengo.Ensemble(100, 1)

    conn1 = nengo.Connection(b, c)
    conn2 = nengo.Connection(a, b)
    conn3 = nengo.Connection(b.neurons, c.neurons)

    nengo.Probe(conn1, "decoders")
    nengo.Probe(conn1, "notprobeable")
    nengo.Probe(conn2, "decoders")
    nengo.Probe(conn3, "decoders")


sim = nengo.Simulator(net)
```

The first probe works fine, as it should.  The second probe is caught by our validation and gives a good error, as it should.  The third/fourth probe just fail at build time with `KeyError: 'decoders'`.  

The problem is that `Connection` says that you can probe `decoders`, by putting that in its `probeable` list, but that's actually only conditionally true.  So anything that relies on the `probeable` list doesn't work in that case (I figured this out because I was using `nengo.utils.probe.probe_all`, which looks though `probeable`).

Possible solutions
1) Just throw some patch fixes on the places where this is a problem (perhaps add some special case validation, fix `probe_all`).  I actually think this is probably the best idea.  It's clearly an error to probe the decoders on a non-decoded connection, so I don't think we should bend over backwards to accommodate the fact that users might do that.
2) Rework the `probeable` system to allow conditional probeables.  This only seems worth doing if we think there are other cases where this might be a problem, but since we haven't run into any yet, I'd be inclined to avoid this.

tbekolay:
Fixed in #671.

PR #671, opened 2015-03-02
Improved probing of Connection transform and decoders

drasmuss:
Fix for #665 and #667 

The only substantive change this makes is that `probeable` is no longer a `Parameter`.  And I think the only real difference that makes is that you can't use the config system to adjust it.  But I think it was always a bit weird to have `probeable` be a parameter like that.  It's really an internal variable, not a model parameter, and I don't think there's any good reason why the user would want to configure it.  I would actually be in favour of changing all the other `probeable` lists to be the same, but I didn't put that in here.

### DIFF
--- a/nengo/builder/probe.py
+++ b/nengo/builder/probe.py
@@ -33,10 +33,8 @@ def synapse_probe(model, key, probe):
         raise ValueError("Attribute '%s' is not probable on %s."
                          % (key, probe.obj))
 
-    if isinstance(probe.slice, slice):
+    if probe.slice is not None:
         sig = sig[probe.slice]
-    else:
-        raise NotImplementedError("Indexing slices not implemented")
 
     if probe.synapse is None:
         model.sig[probe]['in'] = sig
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -7,7 +7,7 @@ from nengo.ensemble import Ensemble
 from nengo.learning_rules import LearningRuleType, LearningRuleTypeParam
 from nengo.node import Node
 from nengo.params import (Default, BoolParam, DistributionParam, FunctionParam,
-                          IntParam, ListParam, NdarrayParam)
+                          IntParam, NdarrayParam)
 from nengo.solvers import LstsqL2, SolverParam
 from nengo.synapses import Lowpass, SynapseParam
 from nengo.utils.compat import is_iterable, iteritems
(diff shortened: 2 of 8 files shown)


nengo/builder/probe.py
```diff
@@ -33,10 +33,8 @@ def synapse_probe(model, key, probe):
         raise ValueError("Attribute '%s' is not probable on %s."
                          % (key, probe.obj))
 
-    if isinstance(probe.slice, slice):
+    if probe.slice is not None:
         sig = sig[probe.slice]
-    else:
```
drasmuss: There was no way this `else` was ever being reached, and as far as I can tell indexing slices works fine, so I think this was just left in by accident?

tbekolay: [`git blame`](https://github.com/nengo/nengo/blame/99312f30b5720e297e855a5d6b46c1808b5de6f8/nengo/builder.py#L1243) says this was @hunse's doing, so he can comment on it... seems fine to me!

tbekolay:
Agreed; `probeable` makes much more sense as a property. I'd be down for making the other probeables properties too (doesn't have to be this PR though). LGTM!

tbekolay:
Rebased to master and added commits to make other `probeable`s properties. Also added some short tests to ensure #665 and #667 are fixed.

I think this is ready to merge once #672 is in; @drasmuss, feel free to merge #672 if it looks good to you!

tbekolay:
Removed `ListParam` in 2638892. Merging if these changes are acceptable to @drasmuss!

drasmuss:
looks good!

nengo/builder/probe.py
hunse: Does this work with list indexing?

drasmuss: As far as I could tell it did, unless there is some weird indexing method I was missing.

hunse: So from what I can tell it doesn't work with list indexing:
```python
a = nengo.Ensemble(100, 1)
ap = nengo.Probe(a.neurons[[0, 1, 5]], 'voltage')
```

drasmuss: Seems to be something weird with neurons
```python
a = nengo.Ensemble(10, 3)
p = nengo.Probe(a[[0,2]])
```
works fine.

hunse: That's because probing an ensemble makes a "connection_probe", as will probing neurons. Since this uses the Connection builder code, which supports advanced slicing, it works fine. It's only when you make a "synapse_probe" (the function in which the line in question is located), for example probing the voltages on neurons, that you run into problems with advanced slicing.

drasmuss: Ah alright, we can put the warning back in then.  Or I guess we could implement it, but that seems like more work :sweat:

hunse: I would put the exception back in for now.

PR #673, opened 2015-03-03
Automatically prepend network labels to object labels

drasmuss:
I found I was adding this boilerplate to every network I was creating, so I thought perhaps others were too and it was worth automating.  All it does is prepend the network label (if it exists) to the object label (if it exists).  So object "a" in network "A" gets labeled "A_a".

I made it so it doesn't apply to the top-level network, because it isn't really useful to add the same prefix to every object in the model.  This way you just get different prefixes for the different subnetworks.

I considered adding a flag to turn this on and off, but decided against it.  You know I'm always on my crusade to reduce the number of parameters, and I think in this case we can avoid it.  a) This only affects labels, which don't really matter. b) I think people will want it enabled almost all the time. c) It can kind of be disabled, by setting the network or object label to `None`.  So the only user that isn't satisfied is someone who wants a label for their subnetwork, wants a label for their object in that subnetwork, but really doesn't want the object label to indicate the subnetwork.  

Also snuck in a few typo fixes :neckbeard:.

Examples:
```python
with nengo.Network(label="net1"):
    a = nengo.Ensemble(10, 1, label="a")
    print a.label # a

    with nengo.Network(label="net2"):
        b = nengo.Ensemble(10, 1, label="b")
        print b.label # net2_b

        with nengo.Network(label="sub2"):
            c_0 = nengo.Ensemble(10, 1, label="c0")
            print c_0.label # net2_sub2_c0

        with nengo.Network():
            c_1 = nengo.Ensemble(10, 1, label="c1")
            print c_1.label # c1

    with nengo.Network():
        d = nengo.Ensemble(10, 1, label="d")
        print d.label # d

    ea = nengo.networks.EnsembleArray(5, 1, label="array")
    print ea.output.label # array_output
    print ea.ensembles[0].label # array_0
```

### DIFF
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -24,7 +24,7 @@ class NetworkMember(type):
         inst.__init__(*args, **kwargs)
         if add_to_container:
             Network.add(inst)
-        inst._initialized = True  # value doesn't matter, just existance
+        inst._initialized = True  # value doesn't matter, just existence
         return inst
 
 
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -144,7 +144,7 @@ class Connection(NengoObject):
     But, we could create either of these two connections.
 
         nengo.Connection(node[0], ensemble)
-        nengo.Connection(ndoe[1], ensemble)
+        nengo.Connection(node[1], ensemble)
 
     Parameters
     ----------
--- a/nengo/network.py
+++ b/nengo/network.py
@@ -111,6 +111,12 @@ class Network(object):
         if not isinstance(network, Network):
             raise RuntimeError("Current context is not a network: %s" %
                                network)
+
+        if (len(Network.context) > 1 and
+                hasattr(obj, 'label') and obj.label is not None and
+                network.label is not None):
+            obj.label = network.label + "_" + obj.label
+
         for cls in obj.__class__.__mro__:
             if cls in network.objects:
                 network.objects[cls].append(obj)
--- a/nengo/networks/ensemblearray.py
+++ b/nengo/networks/ensemblearray.py
@@ -54,8 +54,6 @@ class EnsembleArray(nengo.Network):
 
         self.config[nengo.Ensemble].update(ens_kwargs)
 
-        label_prefix = "" if label is None else label + "_"
-
         self.n_neurons = n_neurons
         self.n_ensembles = n_ensembles
         self.dimensions_per_ensemble = ens_dimensions
@@ -73,8 +71,7 @@ class EnsembleArray(nengo.Network):
 
             for i in range(n_ensembles):
                 e = nengo.Ensemble(
-                    n_neurons, self.dimensions_per_ensemble,
-                    label=label_prefix + str(i))
+                    n_neurons, self.dimensions_per_ensemble, label=str(i))
 
                 nengo.Connection(self.input[i * ens_dimensions:
                                             (i + 1) * ens_dimensions],


drasmuss:
The one thing I wasn't sure of is the SPA networks.  I don't really know the code, but I know there can be some pretty complex network structures.  If SPA people think this wouldn't be good there, we can add the flag to disable it.

tcstewar:
Hmm, this isn't how I've been using labels at all, but that could be because I've been focusing on situations where labels are meant for hierarchical visual depictions of the model, so I definitely don't want this auto-prepending thing going on.

That said, I can believe that there are some times where something like this would be handy.  How about a helper function that can be called that will go ahead and do this to an entire Network?

```python
with model:    
   ...  # go make everything

nengo.utils.network.add_parent_labels(model)
```

That would make it explicit what's happening (although a better name might be needed), and avoids the weirdness that this fails:

```python
with model:
    a = nengo.Ensemble(100, 1)
    a.label = 'a'
    assert a.label == 'a'    # this passes
    with nengo.Network(label='net'):
        b = nengo.Ensemble(100, 1)
        b.label = 'b'
        assert b.label == 'b'    # this fails unexpectedly
```

drasmuss:
Well, actually your example would pass.  If you explicitly set the label, it will stay as whatever you set it, this is only applied when you pass a label to the object constructor.  But regardless, if there is a nonzero number of people who don't want this, then we should make it optional.

I hadn't thought of the GUI, but I can see that in the display there you wouldn't want this, since the subnetworks are already indicated visually.  

To give you an idea of my use-case, the main thing I use labels for is to get useful debugging messages.  So when I have a model with 10 different networks in it, and I get a message saying "can't connect Node 'output' to ...", that isn't very helpful without knowing what network 'output' is in.  Or if you're plotting probe data and you want to be reminded what the probe is targeting, having a more informative label for the probe is good.

So the question is how optional we want this to be.  My personal inclination would be to have it be something that is on by default, and make it easy for the GUI to disable it (rather than the helper function approach).  But I think it's just a matter of which would be preferred by the majority of people.

tcstewar:
> Well, actually your example would pass. If you explicitly set the label, it will stay as whatever you set it, this is only applied when you pass a label to the object constructor. But regardless, if there is a nonzero number of people who don't want this, then we should make it optional.

Ah, weird....  do we have any other cases where setting an argument in the constructor does something different from setting the parameter value after the constructor?  I hadn't even considered that it'd be treated differently here.

> To give you an idea of my use-case, the main thing I use labels for is to get useful debugging messages. So when I have a model with 10 different networks in it, and I get a message saying "can't connect Node 'output' to ...", that isn't very helpful without knowing what network 'output' is in. Or if you're plotting probe data and you want to be reminded what the probe is targeting, having a more informative label for the probe is good.

Hmm.  I think my problem with this is that it's attempting to store implicit hierarchy information inside the label itself.  This leads to all sorts of weird problems if I do something like change the label on a network after things have been built inside it.  I think the ideal solution would be to have a way of computing a full-path label (```print ens.full_label``` which would return something like ```net1->net2->a``` which is computed by traversing the hierarchy upward, so it's always up to date), but that would require nengo objects knowing what Network they are in, which we currently don't support.  We could probably also support it by just keeping an extra parameter around and automatically updating it when labels are changed and whatnot, but that's also got some tricky conditions.

Really, I think this all comes down to what ```label``` is for.  I'd been thinking of it as a label to use when depicting the item.  I hadn't been thinking of it as a pseudo-unique identifier (that's not guaranteed to be unique) for interpreting error messages.  For that case, I'd generally found that the stack trace does a pretty good job of telling me what object is causing the problem.  And for Probeing, I tend to use the ```label``` on the Probe itself.

drasmuss:
>Ah, weird.... do we have any other cases where setting an argument in the constructor does something different from setting the parameter value after the constructor?

Well, the whole config system works that way, e.g.
```python
with network:
    a = nengo.Ensemble(10, 1)
    print a.intercepts
    with config_1:
        a.intercepts = Default
        print a.intercepts
```
Not exactly the same obviously, but in general Nengo is full of things where statements will have different effects depending on their order and placement in the code.  That doesn't mean it's a good thing in this case, I'm just saying it's not inconsistent :wink:.

>  This leads to all sorts of weird problems...

This is kind of what I meant with my point a) above.  Labels are intended to be these things that sacrifice rigour for convenience and flexibility, that's why we separated them from the object identity.  And I see this as kind of in that spirit.  That is, while it's possible that this could introduce "bugs" in obscure cases (if it's possible for a label to have bugs), I think it makes up for that with convenience in the normal use-cases.

Another way to say it is that, exactly as you say, determining an object's location in the network hierarchy is not trivial.  And storing that information is something we've tried to avoid, precisely because it is error prone and complicated.  Nevertheless, clearly in many instances it is handy to have that information, in particular when you're trying to visualize an object's location in the hierarchy without the benefit of a GUI.  That's all this is trying to achieve.  It's not trying to add hierarchical location tracking to Nengo, or create pseudo-unique identifiers, it's just making labels a bit more informative.

All that being said, I've already agreed that it should be optional.  You didn't say explicitly, but do you think now that it shouldn't be an option at all?  Or that it should be built-in, but off by default?  Or that it should be separated from the main code into a helper function?  I'll assume that you don't like that option of "on by default" that I suggested :wink:.  I just want to be clear what the suggested change is before I start working on it.

tcstewar:
> That doesn't mean it's a good thing in this case, I'm just saying it's not inconsistent

:)  Fair enough.  

> All that being said, I've already agreed that it should be optional. You didn't say explicitly, but do you think now that it shouldn't be an option at all? Or that it should be built-in, but off by default? Or that it should be separated from the main code into a helper function? I'll assume that you don't like that option of "on by default" that I suggested . I just want to be clear what the suggested change is before I start working on it.

I'm very much not sure how I think about this.  I currently have an extremely strong resistance to having it "on by default", but my reasons for that are embedded in a very different mindset about what labels are for, and so if I'm wrong about that, then maybe that resistance will go away.  

That said, this still feels to me like something that should be explicitly turned on.  My first instinct of doing it as a helper function that iterates through the Network changing the labels won't always help with your error message case, since the helper function likely won't get called until after the Connections are made.    But I could see it as an option that could be set on Networks:

```python
model = nengo.Network(prepend_labels=True)
with model:
    ...
```

That way it's an easy enough option for people to turn on and off if they want it, and people using the GUI won't have to worry about it.  However, if someone does load up a model that has that parameter set into the GUI, then I don't see a clean way for the GUI to ignore it.

One not-all-that clean way to do it might be to pick a fairly unique join string (```->``` should be pretty good) and have the graphical display always set to only draw the label text after the last ```->```.  If we do something like that, then it could ignore the setting if it is on.  But it still feels like a very different use of the label to me.  

So, mostly, I'm quite uncertain.  :)  Way too many options, and none of them feel right to me yet.

hunse:
If this is something that we want, I would rather keep the label as the label, and have a property that can get the "full label" (with the parent labels prepended). We could also have an option that turns on full labels for debugging (since this seems to be @drasmuss main use case), or maybe we want this on by default.

tcstewar:
> If this is something that we want, I would rather keep the label as the label, and have a property that can get the "full label" (with the parent labels prepended). We could also have an option that turns on full labels for debugging (since this seems to be @drasmuss main use case), or maybe we want this on by default.

That's in line with my thinking as well, if we want a feature like this.  I think I'd lean towards error and warning messages always using the "full label".

This might be added to a list of reasons why we might want a "parent" reference in our graph....  I'd still really prefer to avoid that, though.  I think this example is the first strong reason I've seen for wanting that parent pointer.

tbekolay:
I'd prefer avoiding parent references if possible too... but if it can't be avoided, we should ensure that we keep a weak reference to the parent rather than a strong one.

tcstewar:
Ah, I just remembered the other reason I've run into in the past for wanting a parent pointer -- figuring out the parent network of an endpoint of a Connection so I can draw the Connection lines correctly in the GUI when a Network is set to "don't draw anything inside this Network".  Still, there's only mildly awkward ways to do that without a parent pointer.

drasmuss:
I think this use-case on its own is not enough to motivate adding the parent reference, I'd rather just be left with the current labels.  So perhaps we can just table this, and revisit it if we decide to add the parent reference later.

PR #675, opened 2015-03-04
Refactor test_examples.py

tbekolay:
Opting for standard test functions, rather than pytest hooks. This should make this file more readable, and a tiny bit easier to add new slow tests. This also uses the `slow` mark rather than manually calling `pytest.skip`, which means that the skip reason will be the same as other slow tests.

### DIFF
--- a/nengo/tests/test_examples.py
+++ b/nengo/tests/test_examples.py
@@ -1,4 +1,3 @@
-from glob import glob
 import os
 
 import pytest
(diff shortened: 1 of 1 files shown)


jgosmann:
Maybe rename `_test_noexceptions` to `assert_noexceptions`? That doesn't require the awkward `_` and won't be confused with an actual test function.

tbekolay:
Good call, will do!

tbekolay:
Done!

jgosmann:
LGTM :cake:

Issue #576, opened 2014-12-04
Potential cache performance regression


jgosmann:
I'm under the impression that the build performance suffered with PR #558. This might depend on the number of items in the cache. It might only be a problem on traditional HDD.

Maybe it's just my computer that has issues.

jgosmann:
The faster cache has been merged, should we close this issue?

Seanny123:
Did you end up addressing @drasmuss's concerns in your cache heuristic? If not, maybe close this issue and open a different one?

jgosmann:
We're not using a heuristic ... it's difficult to find a good one because it is quite system dependent. It's also difficult to come up with a universal benchmark.
I've run a few benchmarking trials with the improved cache and it seems the situations where cache hits are slower than no cache seem to be quite rare now. The only combination I've found is around 150 neurons and 5 dimensions in one ensemble. But feel free to run your own benchmarks (you can use `test_cache.py` as a template).

Cache misses are more expensive than no cache, but this will always be the case and that's the trade off you pay for faster repeated builds.

Seanny123:
If we've decided to not use a heuristic, then I would say you could close the issue. If someone creates a more critical benchmark, they can open up a new issue.

jgosmann:
I suppose, that's what I decided for now. Also this issue was about something else originally.

PR #677, opened 2015-03-06
Faster cache

jgosmann:
Instead of caching the decoder solver result based on the activity matrix and targets, this PR changes the caching to be based on neuron type, gain, bias, etc. and the targets. This avoids the time intensive calculation of the activity matrix in case of a cache hit.

This required a little bit restructering in the connection builder. Not sure if `build_decoders` and `solve_for_decoders` are the best names, but I haven't come up with any better naming so far. Also, `solve_for_decoders` takes eight arguments which is a lot, but so far I didn't have a good idea to reduce that count.

Further, this PR removes the cache benchmark notebook and adds "regular" benchmark tests based on the `analytics` fixture as well as a comparison of multiple runs.

On my office computer this PR makes cache hits in most cases faster than no cache. On my Macbook Pro the break even point is at about 250 neurons (for 1 dimension, 1000 evaluation points). Cache misses are always slower then having no cache at all (as before), but writing the data to the disk does not seem to be the main bottleneck.


hunse:
This needs a `rebase --onto` since the first half of commits are now in master. I think 5673541 is the first new commit, but I'll let @jgosmann do the rebase just in case.

tbekolay:
In case it helps, lately I've found that it's somewhat easier / less magical to just do an interactive rebase to master, and skip (comment out) all the commits that you know are in master, rather than doing the `--onto` stuff.

jgosmann:
#678 is not in master yet.

hunse:
Ahh, yes, I forgot that was part of another PR! I guess we should do that first :smile:

drasmuss:
I guess the downside of this is that it's less backend agnostic (i.e., all the parts that go into the backend activity calculation have to go into the hash key).  But that's probably inevitable, optimization always tends to have that effect (and this is all part of the backend anyway).  Would it be possible to have it fall back on just using the activities matrix?  I don't think it's a big priority if it would significantly complicate things, but if it'd be easy then that might be nice to have around as something that other backends like neurogrid could reuse.

jgosmann:
Rebased this to current master.
The `build_linear_system` function is still there, but not cached. It is not used by the reference backend, but maybe we want to keep it for other backends?
The `build_decoders` function is the new function which is cached.
I don't think it is worth to keep the old caching method based on the activity matrix as it only pais of for really large ensembles.

jgosmann:
This PR also addresses #576.

tcstewar:
> The build_linear_system function is still there, but not cached. It is not used by the reference backend, but maybe we want to keep it for other backends?

I'm using that function in one of the other backends, but I think I could adjust it to use build_decoders.  It'd be nice to keep both functions around for a little bit -- maybe just mark build_linear_system as deprecated?

jgosmann:
bump

nengo/builder/connection.py
```diff
@@ -53,6 +57,50 @@ def build_linear_system(model, conn, rng):
...
+    try:
+        decoders, solver_info = model.decoder_cache.wrap_solver(
+            solve_for_decoders)(
+                conn.solver, conn.pre_obj.neuron_type, gain, bias, x, targets,
+                rng=rng, E=E)
+    except ZeroActivityError:
+        raise ZeroActivityError(
+            "Building %s: 'activities' matrix is all zero for %s. "
+            "This is because no evaluation points fall in the firing "
+            "ranges of any neurons." % (conn, conn.pre_obj))
```
tcstewar: Should this raise an exception, or should it just return a decoder that's all zeros?  It seems weird to me that adjusting neuron properties could suddenly cause an exception to be raised.

jgosmann: [We're currently raising an exception](https://github.com/nengo/nengo/pull/677/files#diff-d88c987696c16673f33089794ccd0bbfR51) so this just keeps the existing behaviour and just names the exception so that it is possible to catch it without catching all `RuntimeError`.

I think, it should at least give a warning. But I can't imagine a case where a all zero activities matrix would not indicate a definitely broken model.

tcstewar: > We're currently raising an exception so this just keeps the existing behaviour

Good point.  Never mind, this should be kept with existing behaviour.

tbekolay:
I just started to review this but realized that combining the transform and decoders into a single thing (weights) changed a bunch of stuff here. Do you want me to attempt this rebase @jgosmann or do you want to handle it?

jgosmann:
I'll take a stab at it.

jgosmann:
rebased

tbekolay:
OK, running this branch is promising! For me, there's no significant differences when varying dimension, but if you vary neurons or evaluation points, cache hits are faster, but cache misses are slower after commit 34e6fb1e5c38b72817fcdca0f.

tbekolay:
Plotted results:

test_cache test_compare_cache_benchmark_varying_param d

test_cache test_compare_cache_benchmark_varying_param m

test_cache test_compare_cache_benchmark_varying_param n

Note that these aren't the total build times, but the difference in the build times before / after 34e6fb1. I'm going to review the code now, and in doing so I'll update the y label to reflect this.

jgosmann:
This looks like you compared cdc0d0261126f068c5054d76fe14527e37187fcf vs 34e6fb1e5c38b72817fcdca0f0319e707efd07c7. But the changes in cdc0d0261126f068c5054d76fe14527e37187fcf temporarily disabled the usage of the cache as it introduces some refactoring in the builder and the next commit adjusts the cache to those changes (so all comparisons are done against no cache, even the cache miss comparisons).

If I compare 0ac45c9698bbe0ef72937f84c571098ccda0e417 vs 34e6fb1e5c38b72817fcdca0f0319e707efd07c7, the times for cache hits decrease while everything else stays pretty constant.

test_cache test_compare_cache_benchmark_varying_param d

test_cache test_compare_cache_benchmark_varying_param m

test_cache test_compare_cache_benchmark_varying_param n

tbekolay:
Right you are! I did originally run the benchmarks on that commit, but I noticed that the next commit changed the tests so I reran them to make sure they keys matched. Anyhow, I reran them and here are my results now.

test_cache test_compare_cache_benchmark_varying_param d

test_cache test_compare_cache_benchmark_varying_param m

test_cache test_compare_cache_benchmark_varying_param n

So yeah, improvements on cache hits! No impact on misses. I'll do a quick look through of the code now.

tbekolay:
Code is looking good! I made some clarity / style suggestions in fixup commits... LMK what you think @jgosmann; they're all optional.

jgosmann:
LGTM :cake: 

Do you know about [`--fixup` and `--autosquash`](http://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html)?

tbekolay:
Whoa, I did not know about those! My mouth was literally agape the whole time I read that blog post! I must hand it to @hunse for coming up with a workflow that's compelling enough that git's devs implemented features to support it :rainbow: :octopus:

Merging!

jgosmann:
:yum: 
You can also set
```
[rebase]
    autosquash = true
```
in your git config.

tbekolay:
Oh my god. This is the best!

Seanny123:
@jgosmann that's... so... beautiful...

jgosmann:
:nail_care:

PR #678, opened 2015-03-09
Comparison tests for analytics data.

jgosmann:
Github does not allow to change the branch a PR is made against. Had to close #657 and reopen a new PR. Copied the description, but not the comments.

--------------------------------------

This is a follow-up PR to #647.

It allows to mark test functions with `pytest.mark.compare` and indicates that the function will compare the data from two test runs producing analytics data. A test marked with `compare` will not run per default. Actually, it will be pruned from the collected tests and does not influence the success/skipped/failure statistics.

To run the comparison tests one has to pass `--compare [old-data] [new-data]` to `py.test`. This will prune all normal tests and only collect the comparison tests.

## How to write a comparison test

1. Mark it with `pytest.mark.compare`.
2. Name it `test_compare_<testname>` given that `test_<testname>` produced the relevant data. This naming scheme is necessary to automatically load the relevant analytics data. It would be possible to add the mark based on the test name, but maybe that's too much magic. It would also be possible to get rid of the `test_` prefix (but again it would be a bit more magic).
3. Add the `analytics_data` fixture as function argument. It will be a two element sequence with the same order as the arguments to `--compare`. By conventions the first one should be the older data.
4. Print some useful information (e.g., "X got Y% more accurate.").
5. Do an `assert` whether the metric improved. That way a failure will be reported when something got worse.

All output to stdout and stderr will also be shown for successful tests as there might be useful information (like by how much something got better).

I'll prepare another PR demonstrating such a comparison test.

### DIFF
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -1,6 +1,7 @@
 import hashlib
 import inspect
 import os
+import re
 
 import numpy as np
 import pytest
@@ -120,6 +121,15 @@ def analytics(request):
 
 
 @pytest.fixture
+def analytics_data(request):
+    paths = request.config.getvalue('compare')
+    function_name = parametrize_function_name(request, re.sub(
+        '^test_[a-zA-Z0-9]*_', 'test_', request.function.__name__, count=1))
+    return [Analytics.load(
+        p, request.module.__name__, function_name) for p in paths]
+
+
+@pytest.fixture
 def logger(request):
     """a logging.Logger object.
 
@@ -188,6 +198,9 @@ def pytest_addoption(parser):
         '--analytics', nargs='?', default=False, const=True,
         help='Save analytics (can optionally specify a directory for data).')
     parser.addoption(
+        '--compare', nargs=2,
+        help='Compare analytics results (specify directories to compare).')
+    parser.addoption(
         '--logs', nargs='?', default=False, const=True,
         help='Save logs (can optionally specify a directory for logs).')
     parser.addoption('--noexamples', action='store_false', default=True,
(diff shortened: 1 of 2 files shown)


nengo/tests/conftest.py
```diff
@@ -119,6 +120,15 @@ def analytics(request):
     return analytics.__enter__()
 
 
+@pytest.fixture
+def analytics_data(request):
+    paths = request.config.getvalue('compare')
+    function_name = parametrize_function_name(request, re.sub(
+        '^test_[a-zA-Z0-9]*_', 'test_', request.function.__name__, count=1))
```
hunse: If we always name our tests `test_compare_*`, then this will always be replacing `test_compare_` with `test_`, right? Maybe it's better to keep it general like this, but it took me a while to figure out what this was doing.

hunse:
@jgosmann, could you give a command-line example of using this to compare two things? I'm just wondering a) do we have a way, when running the original tests, to route the analytics output to a particular directory, and b) do I then give the directory names to `--compare`, or the individual file names?

jgosmann:
```bash
py.test --compare old-data-dir new-data-dir nengo
```

hunse:
I meant a full example, including running the original tests.

jgosmann:
Updated the example.
a) Yes, output for analytics and plots can be routed to specific directories (or you can rename the directory afterwards)
b) You give the directory names to `--compare`. The correct files will be picked based on the test function name.

nengo/tests/conftest.py
jgosmann: Yes, the replacement works that way. Not sure whether we want to keep it general.

hunse:
Cool

nengo/tests/conftest.py
```diff
@@ -178,6 +189,10 @@ def pytest_addoption(parser):
 
 
 def pytest_runtest_setup(item):
+    if (item.config.getvalue('compare') and
+            not getattr(item.obj, 'compare', None)):
```
hunse: I would prefer `getattr(item.obj, 'compare', None) is None` here (mostly for readability), unless there's some reason it might be False or some other value that would evaluate as false.

Also, what is this `if` block testing for (i.e. why would we hit it)? Does it just skip the test if it's not marked "compare"? A comment explaining what this is doing would be nice.

jgosmann: I don't think that block is needed at all. It was intended for filtering comparison and non-comparison tests, but that is already happening in `pytest_collection_modifyitems`.

jgosmann:
Is there anything more to fix or can we get this reviewed and merged? (To be able to address #677 which is blocked by this PR right now.)

hunse:
I think everything looked fine to me when I looked at it a month ago. I was waiting for someone else to review/merge it, but it seems like we've kind of fallen out of the two-person review-merge process lately, so I can go ahead and merge it.

tbekolay:
Whoops, sorry about that. For the record, :+1:

PR #679, opened 2015-03-10
Error on attempted pickling of Nengo objects

### LINKED ISSUE #674: Loading a pickled network
Loading a pickled network results in an infinite recursion, bouncing between these two operations:
```
File "d:\documents\nengo\nengo\config.py", line 91, in get_param
    if key in self._extraparams:
File "d:\documents\nengo\nengo\config.py", line 44, in __getattribute__
    param = self.get_param(key)
```
Are we even trying to support pickling though?  If not, we should probably raise an exception when someone tries to pickle a network.

hunse:
Fixes #674. We now throw an error when trying to pickle a Network,
since the default unpickling doesn't work.

I also added errors for NengoObjects and ObjViews, since we haven't
tested pickling with them, so we might as well not allow it, rather
than allowing something that may break.

### DIFF
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -68,6 +68,12 @@ class NengoObject(with_metaclass(NetworkMember)):
             e.args = (arg0,) + e.args[1:]
             raise
 
+    def __getstate__(self):
+        raise NotImplementedError("Nengo objects do not support pickling")
+
+    def __setstate__(self, state):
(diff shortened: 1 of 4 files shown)


tbekolay:
LGTM! It would be good to support serialization at some point, so made a Big Job issue for that. Will merge in a bit (so speak now if anyone has issues with this!)

PR #681, opened 2015-03-10
Simplified signal logic

drasmuss:
I was looking into implementing advanced indexing for probing, which prompted me to dig through the Signal code a bit.  And it struck me that we were basically doing a lot of work there in order to manually recreate numpy's indexing/slicing.  So what this PR does is strip all that out, and put the load back on numpy.  It's functionally equivalent (unless I've screwed up somewhere), but hopefully the additions/deletions ratio is convincing that it's a lot simpler.

This also made it easy to implement advanced indexing for signals (I separated that into its own commit).  So now you can do neat things like
```python
a = nengo.Ensemble(10, 1, encoders=np.random.choice([1, -1], size=(10, 1)))
indices = (a.encoders == 1).squeeze()
p = nengo.Probe(a.neurons[indices], "voltage")
```

Since this touches the core of the simulator, we should be really sure that it doesn't break anything.  It passes all the unit tests, and I ran a few of my models on it, but if anyone else has some other complicated models they could test with, that'd be great.

Also, I'm not sure if NengoOCL was depending on something specific in SignalView that this will cause to break?  I think it really is basically equivalent, so hopefully not, but perhaps @hunse can comment.


drasmuss:
Also, after this change, `SignalDict` is doing very little.  `signals[sig]` is basically equivalent to `sig.value`.  So we could probably get even more simplification by going through and stripping that out, but I wanted to get people's thoughts on this initial change first.

hunse:
This definitely breaks NengoOCL, the question is how much work would it be to get it up and running again. I'm sure James was aware he was duplicating Numpy functionality when he wrote the Signal and SignalView code, which makes me feel like he had a good reason for doing it. Whether that reason is still a good reason, or we can get away without all that overhead, is not clear to me. But I'm very wary about this change, and I think we need to show that it could work with NengoOCL before we go ahead with it.

tbekolay:
My understanding is that Signals were designed to contain, essentially, the metadata associated with a Numpy array so that a backend could construct and organize buffers in the most efficient way possible based on what's best for that backend, and what operations will be done on those buffers. Since Signals are (/were/are supposed to be) just metadata, the indexing/slicing stuff had to be rewritten, and should be faster than doing it on actual buffers.

For better or worse, the desire to be able to reset signals back to their initial values (and to have initial values other than 0) meant that we ended up assigning a Numpy array to each Signal even if the backend construct buffers manually. I'm not sure there's a way around that. Since we have those arrays around anyway, I personally don't see the harm in leveraging their indexing/reshaping capabilities so that we don't have to reimplement out own; but, I'm sure there are good arguments for keeping them separate. One thing we'll want to be very careful with, if we do go with this approach, is that we should endeavour to reduce the number of copies whenever possible, as memory's going to be the big limitation if we have all these numpy arrays around in addition to whatever the backend creates.

But yeah, personally I'm in favor of this PR (probably because I haven't had to deal with OCL stuff yet). First, because it's conceptually simpler, and second because I believe this will also fix an issue that I was trying to deal with a little while ago that has to do with strides. A bit off-topic, but the way we were tracking strides breaks in development versions of NumPy (and in the up-coming 1.10 release of NumPy) because of relaxed stride checking (see [1](http://docs.scipy.org/doc/numpy/release.html#npy-relaxed-strides-checking), [2](https://github.com/numpy/numpy/issues/5085), [3](https://github.com/scipy/scipy/issues/4197), [4](https://github.com/numpy/numpy/issues/4681) for more info).

hunse:
A couple lines that are unfortunately missed by our tests are the `shares_memory_with` checks in `nengo.utils.simulator.validate_ops`. We might be able to do this kind of checking with Numpy arrays, I'm not sure.

This lack of coverage comes from the fact that we're not slicing signals. Basically what this PR is saying is "we're not slicing signals, so let's get rid of that functionality". Before we do that, we should make sure that it is not something that we'll want later on. As long as we don't want this, then I should probably be able to rewrite OCL to accommodate this change without too much hassle.

The reason that we're not using the slicing feature of SignalViews is that we usually want to support [advanced indexing](http://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#advanced-indexing) in addition to basic indexing. Numpy doesn't create views when advanced indexing, and neither can we. This means that we find other ways of doing things that support both basic and advanced indexing; for example, in connections we build the indexing into the transform (see `nengo.utils.builder.full_transform`). This isn't the most efficient way of doing things, and it also means that we just don't support advanced indexing in some places (e.g. when probing things that don't involve making a connection, such as neuron voltages, see the recent discussion in #671).

I've been wondering for a while whether we want to have advanced indexing on a Signal create a view. This would be very handy, but it means that the SignalView would have to carry around indices that map it back to the base signal. This starts to feel like we're recreating sparse arrays. So I haven't come up with a good way of doing this yet, or any other way of allowing for advanced indexing in our code without having to special-case it everywhere.

drasmuss:
This PR includes advanced indexing (with views rather than copies) and slicing for signals, if we're talking about the same thing.

hunse:
> This PR includes advanced indexing (with views rather than copies) and slicing for signals, if we're talking about the same thing.

I see the code that does that. This is going to cause problems when you get into the `make_step` function of an op, because currently when you reference into `signals` with a Signal you get a Numpy array back, and there's no way to return a Numpy array that's a view with advanced indexing. But that could possibly be fixed by having `signals` return Signals instead, which could remember their advanced indices.

The bigger problem comes in OCL, because now I have to start carrying around the indices, whereas before I could just create a view with the correct offset and strides. Also, since slices and advanced indices (i.e. iterables) are fundamentally different data types, it's kind of annoying to figure out how to store them efficiently in memory. That's not to say it's impossible, it's just a very big rewrite of OCL.

As I said above, though, it would be very handy to have advanced indexing built right into our Signals, if we can figure out a way to pull this off.

tbekolay:
I have a perhaps naive question about advanced indexing. Is it ever the case that something with advanced indexing could instead be done with multiple calls with simple indexing or slicing?

My impression is that the main use cases are to 1) send a single dimension's value to multiple dimensions in the post and to 2) probe irregularly spaced dimensions. But both of these can be done with multiple calls.

Use case 1:

```python
pre = nengo.Ensemble(10, 1)
post = nengo.Ensemble(20, 2)
# Advanced indexing
nengo.Connection(pre[[0, 0]], post)
# Simple indexing
nengo.Connection(pre, post[0])
nengo.Connection(pre, post[1])
```

Use case 2:

```python
# Advanced indexing
nengo.Probe(ten_d_ensemble[[3, 6, 7]])
# Simple indexing
nengo.Probe(ten_d_ensemble[3])
nengo.Probe(ten_d_ensemble[6:8])
```

Is this correct or am I missing use cases? If no, I might be inclined to just not do advanced indexing if it means we get speedups and/or it makes backend code simpler.

drasmuss:
>This is going to cause problems when you get into the make_step function of an op, because currently when you reference into signals with a Signal you get a Numpy array back, and there's no way to return a Numpy array that's a view with advanced indexing.

Yeah, the advanced indexing only works as long as you're referencing the ndarray through the signal, otherwise you end up with a copy.  As you say, I don't think there's any way around that without doing a lot more changes to the simulator logic.

>The bigger problem comes in OCL, because now I have to start carrying around the indices, whereas before I could just create a view with the correct offset and strides.

Another option would be to just not support advanced indexing in NengoOCL.  Since it's all backend code, we can be more flexible in where it's supported I think.

drasmuss:
>Is this correct or am I missing use cases? If no, I might be inclined to just not do advanced indexing if it means we get speedups and/or it makes backend code simpler.

Yeah, it's just a convenience thing.  Although keep in mind that what we're supporting here is advanced indexing of signals, not nengo objects.  So while it's true that nengo indexing tends to be pretty simple, it's probably easier to imagine cases in the backend code where you have complex indexing of some signal (e.g. a bunch of ensembles that you've grouped into a single signal) that you'd like to not take 10 different calls.

But yeah, I think we could definitely do without it if it's problematic.  You can take a look at just the first commit for what the change looks like without that.

hunse:
I think I'd prefer to keep advanced indexing out of Signals. We can add an `indexed_copy` op if we want, that can take a signal and a set of advanced indices, and copy them into a new signal (we could have advanced indexing into the target signal, as well). I think this would address most of our use cases, and not require big re-writes to the reference simulator or OCL.

drasmuss:
:+1: sounds good

drasmuss:
Alright, updates made

- rebased onto master
- removed advanced indexing
- fixed `nengo.utils.simulator.validate_ops`
- added indexed copy operation (just added optional indexing to the existing `Copy`)
- re: #688, made it so that read-only arrays don't get duplicated

I also added something new, which is to remove `SignalDict`.  As I mentioned, it really wasn't doing anything in the new implementation, so if we can avoid passing around an unnecessary data structure that seems for the best.  The only substantial change is that resetting is now done through the `model`, and I slightly changed how time gets passed to nodes.

hunse:
I'm guessing that the original idea behind SignalDict is that the signals that change are tied to the model. So you could build a model once, and then make a bunch of different simulators that simultaneously run copies of the model. Is that something that we might want?

drasmuss:
No use cases for that come to my mind anyway.  Any time that someone wants to run multiple instances of a model, they probably want something to be different between the copies (different random seeds, etc.), otherwise why bother having multiples.  Also I think all that SignalDict would be contributing to that use-case is providing a single collection of all the signals that need to be copied, which is still available (it's `model.all_signals` now).

hunse:
My idea is that you could use different *simulator* seeds, to have different instances of random noise, or use Oliver's `rebuild_node` function or something like it to rebuild a node without rebuilding the whole model.

The difference in my mind is that SignalDict is a mapping from signals to values, so you can have many such mappings. The signals themselves are just storing their initial values. In your system, the value is stored right in the signal, so each signal can only have one value (there can't be more than one mapping). 

Additionally, it causes problems on different backends, where I want to use the same builder, but my actual data will lie on a device. In the old system, there was a clear division between these sets of data: the initial values would be in the Signals on the host, and the runtime values would be in the SignalDict on the device. In your system, the runtime values are created on the host as part of the build process. So if I use the same builder in Nengo OCL, I would have to copies of the data on the host (initial value and runtime value), as well as something equivalent to a SignalDict mapping Signals to my device data.

drasmuss:
All you would have to do is write your own version of `Signal` for NengoOCL (which is easy to do now since `Signal` is really simple) that creates the runtime values on the device, instead of having to write your own version of `SignalDict` to do the same.  I really don't think there's much difference in functionality between the two approaches, it's just shuffling things around.

hunse:
How do you get that version of Signal used throughout the builder?

drasmuss:
I guess I don't know NengoOCL, but I would have thought that could be monkeypatched in without too much trouble, e.g.
```python
import nengo.builder

def init(self, value, name=None):
    print "I'm a new signal!"
    
    # replace this with something equivalent on NengoOCL
    self._value = np.array(value, copy=False, order='C', dtype=np.float64) 
    

nengo.builder.Signal.__init__ = init
```

hunse:
Sure, something like that might be possible. It scares me, though, because a) it seems like it could introduce some hard-to-find bugs, and b) whenever I see something like that, I immediately wonder how the underlying code can be written better to avoid it.

To me, there's a big conceptual difference between storing signal runtime values in Signals, versus in a SignalDict. The former connects them to the builder (Signals are builder objects), where the latter connects them to the simulator. Connecting them to the simulator makes more conceptual sense to me. And the fact that one way of doing it requires monkey patching suggests that something is conceptually wrong, and puts more burden on the user because they have to realize that Signal's init function is being monkey patched. If they miss that one line, they could spend a long time trying to crawl trhough Signal's actual init function trying to figure out how the stuff gets allocated on the device.

I have no problem with simplifying SignalDict, if that can be done, but getting rid of it entirely removes a distinction between the builder and the simulator that I think is useful to have.

drasmuss:
I have the opposite intuition, to me having all the information associated with a `Signal` associated with a single object makes more sense.  For example, here's the only example that changes in this PR (the `rectified_linear` example where we show the user how to make a step function)

Previously:
```python
    J = signals[self.J]
    output = signals[self.output]
    def step():
        output[...] = np.maximum(0, J)
    return step
```

After removing SignalDict:
```python
def make_step(self, dt, rng):
    J = self.J.value
    output = self.output.value
    def step():
        output[...] = np.maximum(0, J)
    return step
````

The latter just seems a lot more readable.  It's clear that a `Signal` is a thing that has a value, there's no `signals` parameter that makes no sense unless you go and look at the internal logic of the `Simulator`.  And I think that's pretty much true throughout that commit, the modified code is simpler and more readable.

I think the original reason we had `SignalDict`, and the conceptual distinction you refer to, was that it was going to be the common language shared by all backends.  So you could build once and then reuse it, as you describe.  But now that we have more experience in backend development, it doesn't seem like that's the case (NengoOCL is the only one that uses it).  So to me it seems worth it to get rid of some of that generality in order to simplify the code base, but I think it's definitely true that different people will have different opinions on that depending on their priorities.

Also I'm sure we could come up with a solution that doesn't involve monkey patching, that's just the simplest way that I could think of off the bat.

drasmuss:
Also I can put together a PR for NengoCL, I don't mean to just offload that work onto you.  I just don't want to start on that before we decide whether to go ahead with this or not.

hunse:
To me, a Signal is not a thing that has a value, just as an Operator does not do anything. They're both just describing the computation graph of the system.

If we're going to go your route, why do we need operators at all? Why not just make the step functions right away, and add them to a list?

I think there's lots of reasons to keep these separate. It's what makes the re-use of the builder in Nengo OCL possible. Just because Nengo OCL is currently the only backend outside the reference that uses the Builder is not a good reason to get rid of these features. I forsee the FPGA backend, and any supercomputer backend, also wanting to use the Builder.

> Also I'm sure we could come up with a solution that doesn't involve monkey patching

I'm not so sure. If you can think of something, I might reconsider this a bit. But to me, it seems like there's a conceptual problem, and usually when your code is not organized in the same way as your concepts, anything you do to get around it is going to be a "monkey patch". We want the builder and simulator separate (or at least I do), so I think the code has to reflect that.

celiasmith:
Eric Crawford has written a BlueGene backend... could ask for his input.

Eric Hunsberger wrote:

drasmuss:
>I'm not so sure. If you can think of something, I might reconsider this a bit.

As another example, we could use the approach you already implemented in #640 (having the model create Signals).

e2crawfo:
Just chiming in to say nengo_mpi uses the Builder as well.

jgosmann:
I was just reminded of the NEURON simulator bindings I wrote during the summer school which use the Builder as well. I will have to take a look how that code is affected by the proposed changes.

jgosmann:
It doesn't seem that any of the changes will break the NEURON bindings in a way that could not be fixed easily.

hunse:
I think this is superseded by #858, but I'll let @drasmuss decide and close if appropriate.

PR #682, opened 2015-03-10
Add `logger` fixture for saving test logs

tbekolay:
And a corresponding py.test `--logs` command line option. See the commit message for more details.

This is a WIP because I get some really weird test failures at unexpected times. E.g., if I run all of the unit tests, `test_rates` fails with
```
>       logging.StreamHandler.__init__(self, stream=TextIO())
E       TypeError: unbound method __init__() must be called with StreamHandler instance as first argument (got CaptureLogHandler instance instead)
```
but if I run this test individually, it runs fine. I'm investigating now, but figured I would make this PR to get opinions on the interface, and whether this is sufficient to replace what the analytics summary data was doing.

### DIFF
--- a/nengo/networks/tests/test_actionselection.py
+++ b/nengo/networks/tests/test_actionselection.py
@@ -1,5 +1,4 @@
 import numpy as np
-import pytest
 
 import nengo
 
@@ -48,8 +47,3 @@ def test_thalamus(Simulator, plt, seed):
 
     assert output[0] > 0.8
     assert np.all(output[1:] < 0.01)
-
-
-if __name__ == "__main__":
-    nengo.log(debug=True)
-    pytest.main([__file__, '-v'])
--- a/nengo/utils/tests/test_testing.py
+++ b/nengo/utils/tests/test_testing.py
@@ -1,13 +1,9 @@
 import errno
-import logging
 import os
 
 import pytest
 
-import nengo
-from nengo.utils.testing import Analytics, Timer
-
-logger = logging.getLogger(__name__)
+from nengo.utils.testing import Analytics, Logger, Timer
 
 
 def test_timer():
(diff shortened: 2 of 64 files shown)


nengo/networks/tests/test_actionselection.py
```diff
@@ -49,7 +49,6 @@ def test_thalamus(Simulator, plt, seed):
     assert output[0] > 0.8
     assert np.all(output[1:] < 0.01)
 
-
 if __name__ == "__main__":
-    nengo.log(debug=True)
-    pytest.main([__file__, '-v'])
+    import sys
+    pytest.main(sys.argv)
```
hunse: Do we really need these? Can't we just tell people to call `py.test test_whatever.py`?

nengo/utils/tests/test_testing.py
```diff
@@ -59,6 +55,31 @@ def test_analytics_norecord():
         analytics.get_filepath(ext='npz')
 
 
+def test_logger_record():
+    logger_obj = Logger('nengo.simulator.logs',
+                        'nengo.utils.tests.test_testing',
+                        'test_logger_record')
```
hunse: Could we have a `logger` fixture that returns a logger configured like this (and with parametrized function names, too)?

nengo/networks/tests/test_actionselection.py
tbekolay: I'm okay taking them out if that's the consensus. If that's what people want, I'll write up a page in the docs with more details on running unit tests...

nengo/utils/tests/test_testing.py
hunse: And also have it call `logger_obj.__enter__()` and make a callback for exit, à la `plt`?

tbekolay: There is one, this is unit testing the guts of that fixture.

nengo/networks/tests/test_actionselection.py
hunse: If other people use them, that's fine. I just don't, personally...

nengo/utils/tests/test_testing.py
hunse: Ahh, cool, missed that.

hunse:
I think this should do most of the stuff that I wanted. It would be cool to add a function to utils that gets the current git information (if you're running in a git repo), so that you can print things like the current branch and commit (hash), since it's easy to forget what timestamps mean.

nengo/networks/tests/test_actionselection.py
jgosmann: me neither

tbekolay:
Okay! I figured out the failed tests; `logging.shutdown()`, which is called in `nengo/utils/test_logging.py`, was causing the fixture to fail in weird ways. I changed those tests, as well as the `nengo.log` function, to hopefully play nicer with other things trying to use loggers.

I also removed all of the `if __name__ == "__main__"` blocks (and made #683 to improve documentation on running unit tests), and made #685 to track @hunse's suggestion.

With that, I think this is ready for srs review.

hunse:
I think that addresses all my comments. I scanned through it again and it looks good to me.

PR #689, opened 2015-03-18
Fix implementation of readonly signals

tbekolay:
I didn't think too hard about this fix, but it at least allows master to pass unit tests. Please make sure the unit tests pass before merging in the future... lots has changed from when that PR was created.

### DIFF
--- a/nengo/tests/test_builder.py
+++ b/nengo/tests/test_builder.py
@@ -231,8 +231,8 @@ def test_commonsig_readonly(RefSimulator):
     net = nengo.Network(label="test_commonsig")
     sim = RefSimulator(net)
     for sig in itervalues(sim.model.sig['common']):
-        sim.signals.init(sig, sig.value)
-        with pytest.raises(ValueError):
+        sim.signals.init(sig)
+        with pytest.raises((ValueError, RuntimeError)):
             sim.signals[sig] = np.array([-1])
-        with pytest.raises(ValueError):
+        with pytest.raises((ValueError, RuntimeError)):
             sim.signals[sig][...] = np.array([-1])
--- a/nengo/builder/signal.py
+++ b/nengo/builder/signal.py
@@ -34,6 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import numpy as np
 
+import nengo.utils.numpy as npext
 from nengo.utils.compat import StringIO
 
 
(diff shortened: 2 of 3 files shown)


nengo/tests/test_builder.py
```diff
@@ -231,7 +231,7 @@ def test_commonsig_readonly(RefSimulator):
     net = nengo.Network(label="test_commonsig")
     sim = RefSimulator(net)
     for sig in itervalues(sim.model.sig['common']):
-        sim.signals.init(sig, sig.value)
+        sim.signals.init(sig)
         with pytest.raises(ValueError):
```
tbekolay: Looks like in NumPy 1.6 this is a RuntimeError, so this should be:
```
        with pytest.raises((ValueError, RuntimeError)):
```

hunse:
Sorry, I thought I had rebased it not too long ago, so I made the bad assumption that it was fine.

tbekolay:
No worries, the fix wasn't too difficult. If running the test suite (I think around 3 minutes on my machine) is annoying as part of the merge workflow, perhaps we should look into ways to make it faster. Maybe 3 levels of tests rather than 2 (those marked and not marked as `slow`)?

hunse:
Yeah, it's really not too bad. I'm just lazy.

It would be nice to have an easier way to run the Travis tests, though, since as we saw here, there are some things that fail only on earlier Numpy.

tbekolay:
This is true. Could set it up through `tox`, but the downside of that is that `tox` only really allows for pip installing; for linux I have to install numpy from source for it to be fast and I don't know how to get that working through `tox`... could just set up different virtualenvs manually though, that's how I test on linux. Here's the bash function I use (annotated):

```bash
# It's nice to see the command echo'd
exe() { echo "\$ $@" ; "$@" ; }

# The "$@" passes command line args to py.test, so I can invoke like this:
#   test_nengo --logs --analytics
# to run py.test --pyargs nengo --logs --analytics on python 2 and 3
test_nengo() {
    exe workon nengo
    # Only generate plots once
    exe py.test --plots --pyargs nengo "$@"
    exe workon nengo3
    exe py.test --pyargs nengo "$@"
    exe workon nengo
    exe flake8 nengo
}
```

jgosmann:
(Using [`set -x`](http://www.faqs.org/docs/abs/HTML/options.html) instead of prefixing every line might be easier for echoing the commands.)

tbekolay:
`set -x` creates way too much noise because it also echoes all of the innards of these commands since `workon` is a shell function, not a command that creates a subshell. :hamburger:

PR #693, opened 2015-04-02
Change progress bar instantiation for IPython >= 3 compatibility.

jgosmann:
* Add RC option to set progress bar.
* Remove IPython notebook detection (cannot be done with IPython
  notebook >3 aka Jupyter)
* Add `activate_ipynb_features()` function.

Todo:
* [x] Document new RC options
* [x] Document `activate_ipynb_features()`
* [x] Add `activate_ipynb_features()` to examples?
* [ ] Document how to run `activate_ipynb_features()` automatically when a kernel in a notebook is started.

We are still deploying the progress bar front-end code in IPython notebook in the non-recommended way:

> If you distribute your custom widget, you may be using display and Javascript to publish the widget’s Javascript to the front-end. That is no longer the recommended way of distributing widget Javascript. Instead have the user install the Javascript to his/her nbextension directory or their profile’s static directory.

This is for backwards compatibility (the recommended way does not work prior to IPython 3). It would be possible to deploy the code in the recommended way for IPython 3 only and in the old way otherwise. But that might require some code duplication. Thus, I would vote to keep it how it is at the moment and switch over at a later point when we drop support for older IPython versions. The downsides of the old way are:
* Front-end code might not be loaded in time (and progress bar shows not up)
* Progress bar might break when the page is reloaded.

### DIFF
--- a/examples/dynamics/oscillator.ipynb
+++ b/examples/dynamics/oscillator.ipynb
@@ -1,7 +1,7 @@
 {
  "metadata": {
   "name": "",
-  "signature": "sha256:b0817b1b084b9473b13c94874aa954454cd615b1c4bc9cf52cd1f2666dd33284"
+  "signature": "sha256:a4fc96b278943b8e2b2f76a246ebe4cc62283b89dd05f650fff5fb2a429dbc42"
  },
  "nbformat": 3,
  "nbformat_minor": 0,
@@ -23,9 +23,11 @@
       "import matplotlib.pyplot as plt\n",
       "%matplotlib inline\n",
       "\n",
-      "import nengo"
+      "import nengo\n",
+      "%load_ext nengo.ipynb"
      ],
      "language": "python",
+     "metadata": {},
      "outputs": []
     },
     {
--- a/nengo/ipynb.py
+++ b/nengo/ipynb.py
@@ -11,12 +11,107 @@ This IPython extension cannot be unloaded.
 """
 
 from nengo.rc import rc
-from nengo.utils.progress import IPython2ProgressBar, IPythonProgressWidget
+from nengo.utils.ipython import get_ipython, has_ipynb_widgets
+from nengo.utils.progress import ProgressBar, timestamp2timedelta
+
+if has_ipynb_widgets():
+    from IPython.html.widgets import DOMWidget
+    from IPython.display import display
+    import IPython.utils.traitlets as traitlets
+else:
+    DOMWidget = object
 
 
 def load_ipython_extension(ipython):
     IPythonProgressWidget.load_frontend()
     if rc.get('progress', 'progress_bar') == 'auto':
         rc.set('progress', 'progress_bar', '.'.join((
-            'nengo.utils.progress',
-            IPython2ProgressBar.__name__)))
+            __name__, IPython2ProgressBar.__name__)))
+
+
+class IPythonProgressWidget(DOMWidget):
+    """IPython widget for displaying a progress bar."""
+
+    # pylint: disable=too-many-public-methods
+    _view_name = traitlets.Unicode('NengoProgressBar', sync=True)
+    progress = traitlets.Float(0., sync=True)
+    text = traitlets.Unicode(u'', sync=True)
+
+    FRONTEND = '''
+    require(["widgets/js/widget", "widgets/js/manager"],
+        function(widget, manager) {
+      if (typeof widget.DOMWidgetView == 'undefined') {
+        widget = IPython;
+      }
+      if (typeof manager.WidgetManager == 'undefined') {
+        manager = IPython;
+      }
+
+      var NengoProgressBar = widget.DOMWidgetView.extend({
+        render: function() {
+          // $el is the DOM of the widget
+          this.$el.css({width: '100%', marginBottom: '0.5em'});
+          this.$el.html([
+            '<div style="',
+                'width: 100%;',
+                'border: 1px solid #cfcfcf;',
+                'border-radius: 4px;',
+                'text-align: center;',
+                'position: relative;">',
+              '<div class="pb-text" style="',
+                  'position: absolute;',
+                  'width: 100%;">',
+                '0%',
+              '</div>',
+              '<div class="pb-bar" style="',
+                  'background-color: #bdd2e6;',
+                  'width: 0%;',
+                  'transition: width 0.1s linear;">',
+                '&nbsp;',
+              '</div>',
+            '</div>'].join(''));
+        },
+
+        update: function() {
+          this.$el.css({width: '100%', marginBottom: '0.5em'});
+          var progress = 100 * this.model.get('progress');
+          var text = this.model.get('text');
+          this.$el.find('div.pb-bar').width(progress.toString() + '%');
+          this.$el.find('div.pb-text').text(text);
+        },
+      });
+
+      manager.WidgetManager.register_widget_view(
+        'NengoProgressBar', NengoProgressBar);
+    });'''
+
+    @classmethod
+    def load_frontend(cls):
+        """Loads the JavaScript front-end code required by then widget."""
+        get_ipython().run_cell_magic('javascript', '', cls.FRONTEND)
+
+
+class IPython2ProgressBar(ProgressBar):
+    """IPython progress bar based on widgets."""
+
+    supports_fast_ipynb_updates = True
+
+    def __init__(self, task="Simulation"):
+        super(IPython2ProgressBar, self).__init__(task)
+        self._widget = IPythonProgressWidget()
+        self._initialized = False
+
+    def update(self, progress):
+        if not self._initialized:
+            display(self._widget)
+            self._initialized = True
+
+        self._widget.progress = progress.progress
+        if progress.finished:
+            self._widget.text = "{0} finished in {1}.".format(
+                self.task,
+                timestamp2timedelta(progress.elapsed_seconds()))
+        else:
+            self._widget.text = "{progress:.0f}%, ETA: {eta}".format(
+                progress=100 * progress.progress,
+                eta=timestamp2timedelta(progress.eta()))
--- a/nengo/ipynb.py
+++ b/nengo/ipynb.py
@@ -11,20 +11,25 @@ This IPython extension cannot be unloaded.
 """
 
 from nengo.rc import rc
-from nengo.utils.ipython import get_ipython, has_ipynb_widgets
+from nengo.utils.ipython import has_ipynb_widgets
 from nengo.utils.progress import ProgressBar, timestamp2timedelta
 
 if has_ipynb_widgets():
-    from IPython.html.widgets import DOMWidget
+    import IPython
+    if IPython.version_info[0] <= 3:
+        from IPython.html.widgets import DOMWidget
+        import IPython.utils.traitlets as traitlets
+    else:
+        from ipywidgets import DOMWidget
+        import traitlets
     from IPython.display import display
-    import IPython.utils.traitlets as traitlets
 else:
     DOMWidget = object
 
 
 def load_ipython_extension(ipython):
-    IPythonProgressWidget.load_frontend()
-    if rc.get('progress', 'progress_bar') == 'auto':
+    if has_ipynb_widgets() and rc.get('progress', 'progress_bar') == 'auto':
+        IPythonProgressWidget.load_frontend(ipython)
         rc.set('progress', 'progress_bar', '.'.join((
             __name__, IPython2ProgressBar.__name__)))
 
@@ -49,6 +54,8 @@ class IPythonProgressWidget(DOMWidget):
 
       var NengoProgressBar = widget.DOMWidgetView.extend({
         render: function() {
+          // Work-around for messed up CSS in IPython 4
+          $('.widget-subarea').css({flex: '2 1 0%'});
           // $el is the DOM of the widget
           this.$el.css({width: '100%', marginBottom: '0.5em'});
           this.$el.html([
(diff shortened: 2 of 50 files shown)


tbekolay:
For interest's sake, SymPy's equivalent function for doing this is `sympy.init_printing()`. It used to be an IPython extension, loaded with `%load_ext sympy.interactive.ipythonprinting`, or `%load_ext sympyprinting` ([from here](http://nbviewer.ipython.org/github/jrjohansson/scientific-python-lectures/blob/master/Lecture-5-Sympy.ipynb); more details [here](http://docs.sympy.org/dev/tutorial/printing.html)). Their use case is different from ours, but it might be useful to use similar names.

jgosmann:
In our case it is not really any printing functionality that gets initialized ...

jgosmann:
Are we still aiming to support Python 2.6 in Nengo 2.1? I'm using the `importlib` module which does not exist in 2.6.

tbekolay:
We haven't made any changes to drop support yet; ideally we would have at least one release of Nengo OCL that works on 2.6, but if Nengo OCL can target 2.0.1 then we can drop it.

jgosmann:
Can we add `importlib` to the 2.6 dependencies? [It should be pip installable.](https://pypi.python.org/pypi/importlib)

hunse:
Nengo OCL is working for most features on 2.0.1. You can take a look at [which tests I'm skipping](https://github.com/nengo/nengo_ocl/blob/master/nengo_ocl/tests/test_sim_ocl.py) to see which features aren't supported. I think for the most part, they're pretty minor, so I think we can go ahead with a release.

In #633 I drop support of 2.6, basically just by removing the 2.6 tests.

jgosmann:
@tbekolay Do you know why they moved from an extension to a function in SymPy? I was thinking about changing `activate_ipynb_features()` to an extension to be loaded with `%load_ext nengo.ipynb`.

The only way to automatically load the IPython progress bar (instead of the text version), seems to be to set an environment variable in `ipython_notebook_config.py` which is read out by Nengo. It would make things more convenient, but is somewhat magic and might cause non-obvious bugs? Otherwise `activate_ipynb_features()`/`%load_ext nengo.ipynb` has to be run in every notebook or the terminal progress bar is used (which looks not as nice and might blow up the notebook if a lot of those are created).

jgosmann:
I dug around and found [an answer to my question](https://code.google.com/p/sympy/issues/detail?id=3914). As in our case it only ever makes sense to run the function within IPython, an extension should be fine.

tbekolay:
Nice, I can finally close this tab ;) I didn't know off the top of my head so I'd have dug around as you did. That issue also linked to [this one](https://github.com/sympy/sympy/issues/7003), so if they manage to figure that out then we can use their solution perhaps?

jgosmann:
Nah, we can't. :(
They just print static stuff which is reasonably well supported. But there is no way to know whether widgets are supported or to provide a (non-static) fall back if widgets are not supported.

tbekolay:
Fist shake! :fist: :sake:

jgosmann:
This should be ready for review now.

jgosmann:
bump

jgosmann:
Note to self: Add changelog entries:
* IPython extension
* RC options
* breaks python 2.6 compatibility

jgosmann:
Or do we prefer to add `importlib` to the requirements for Python 2.6 instead of breaking the compatibility (it's pip installable)?

tbekolay:
I'd prefer adding `importlib` for now; we might as well keep Py 2.6 for 2.1 since it's still working. Break compatibility in 2.2.

tcstewar:
> I'd prefer adding importlib for now; we might as well keep Py 2.6 for 2.1 since it's still working. Break compatibility in 2.2.

That's my preference as well.  :)  It'd be nice to not require importlib as well for Py 2.6, but it's not too bad....

jgosmann:
> It'd be nice to not require importlib as well for Py 2.6, but it's not too bad....

The importlib code is just 38 lines of code. We could include it in Nengo?

tbekolay:
Eh... we already have some Python 2.6 specific dependencies (`ordereddict` for example), and I don't think we have many Py 2.6 users anyhow. Mostly it's Travis CI running it in Python 2.6, and that's pretty trivial. In this case I think we're mostly trying to reduce programmer effort, and right now the least-effort solution is to just add importlib to the Py 2.6 dependency list.

jgosmann:
Where is this list kept? There's `requirements-test-py26.txt`, but it is not just a test requirement.

tbekolay:
Just that file; we're not explicitly tracking it in `setup.py`. `tox` and TravisCI install the `test-26` requirements, so they work, but it's only recently that setuptools has support for version-specific `install_requires`. I don't think we have any actual Python 2.6 users so it hasn't come up, and since we'll drop support in 2.2 I don't think it's really worth the effort of adding it to `setup.py`.

jgosmann:
rebased, added changelog entries

jgosmann:
> Just that file

needed to go into `.travis.yml` too

tbekolay:
> needed to go into .travis.yml too

Ah right, I stand corrected.

nengo/ipynb.py
```diff
@@ -0,0 +1,126 @@
+"""IPython extension that activates special IPython notebook features of Nengo.
+
+At the moment this only activating the improved progress bar.
```
tbekolay: activating -> activates

tbekolay:
Just looked over this code, it's looking good! I mean, aside from the gross stuff we're forced to do with 3 version of IPython to support.

I just want to manually test it out with IPython 2, 3 and Jupyter. But once I'm done that, is there anything else you want to do before this gets merged in @jgosmann?

Also, is everything okay with `%load_ext nengo.ipynb%` to enable the fancy progress bar? We have to do this with a `%load_ext` call, so I'm not asking whether we should make this automatic (we can't), just asking if `nengo.ipynb` is memorable enough.

examples/dynamics/oscillator.ipynb
```diff
@@ -154,4 +156,4 @@
    "metadata": {}
   }
  ]
-}
```
hunse: In some of these files, the endline at the end of the file has been removed, and in others it's been added. We should probably decide which way we want it, and then change the cleanup devscript to do it that way. I think when I wrote the devscript, I put in the linebreak at the end because I don't like the little :no_entry_sign: that GitHub puts if you don't have it (command line git also gets mad), but I think IPython doesn't put one when you edit a file (and will even actively remove it), so perhaps we should go with no linebreak.

jgosmann:
> I just want to manually test it out with IPython 2, 3 and Jupyter.

I did that, but feel free to test it yourself.

> But once I'm done that, is there anything else you want to do before this gets merged in @jgosmann?

Nope.

examples/dynamics/oscillator.ipynb
jgosmann: > I put in the linebreak at the end because I don't like the little :no_entry_sign: that GitHub puts if you don't have it

It is less likely to look at the actual raw source of the notebooks than to look at Python files. So the little :no_entry_sign: might not be as much of a problem in these files.

jgosmann: I can't even figure out how to display the raw source of a notebook within Github (you either get it rendered or the pure data without the Github interface).

tbekolay: I agree, we should go with no linebreak for the time being. I think at a certain point, we should upgrade all out notebooks to the newer format (v4). I wouldn't be surprised if newer IPythons (or Jupyter) save with a newline at the end of the file, but if they don't, then we can vote / adopt a policy either way.

tbekolay:
OK! I tested this out on all three versions and it works great! I'll fix that typo in the merge, and since no one's suggested an alternative to `nengo.ipynb` I'm going to assume no one feels too strongly about it.

examples/dynamics/oscillator.ipynb
hunse: @jgosmann: in the diffs is always where I see it. I can't find a way to just view notebooks, either.

tbekolay:
OK, I've rebased to master and moved the history around a bit, added in some fixup / squash commits. @jgosmann, take a look, and if it looks good I'll merge!

hunse:
Could we squash a bit more before merge, too? Six commits still seems like a lot.

tbekolay:
I could see 9a93fb84c6d7bef514d37719a1d4503d7a6989a1, cfb5933dfb77091e668e8801 and af3ec0ee73de20c5d974093c9 being squashed into the first commit.

hunse:
I don't know the commits in detail, but at a glance that sounds good, assuming it's not too much work to move them around. I would squash in c4a6c34 too, because it's just keeping the examples up-to-date with the new requirements, right?

hunse:
Then it would just be the Jupyter stuff that's separate, which I think makes sense.

tbekolay:
It's true that c4a6c34 is just keeping examples up to date, but since it touches every example, it results in a big diff, which can be annoying to page through in the terminal. But I could see the argument for including it, so I'm pretty ambivalent.

hunse:
> since it touches every example, it results in a big diff, which can be annoying to page through in the terminal.

Fair enough. I'm fine with either way too.

tbekolay:
Will let @jgosmann decide ;)

jgosmann:
I prefer more and smaller commits.

tbekolay:
So should we leave it at the 6 it is now then?

jgosmann:
Would be my vote.

tbekolay:
Yeah, I'm good with that. @hunse, acceptable?

hunse:
The changelog commit should definitely be squashed, IMO.

Also, cfb5933 not only makes the change it purports to, but also does some refactoring. It's things like that which make me prefer fewer commits in some cases, because it seems less surprising to me that a commit changing the whole ProgressBar system would refactor things, versus a commit that says it changes when a warning is shown. Furthermore, 9a93fb8 unindents `IPythonProgressWidget`, and then cfb5933 moves it to another file. I feel like the diff would be cleaner if they're both in the same commit.

tbekolay:
For the sake of concreteness, here are the two versions:

- [less commits](https://github.com/nengo/nengo/compare/ipy3-less)
- [more commits](https://github.com/nengo/nengo/compare/ipy3-more)

Separated, the three commits have 262 insertions and 221 deletions. Squashed, the one commit has 187 insertions and 152 deletions.

hunse:
I think you forgot to squash the changelog commit in the fewer commits branch.

As a compromise, I'd be happy with more commits just with 6912d66 and 7abaa4a, as I did [here](https://github.com/nengo/nengo/compare/ipy3-mid).

tbekolay:
> I think you forgot to squash the changelog commit in the fewer commits branch.

Right you are! Fixed.

> I'd be happy with more commits just with 6912d66 and 7abaa4a squashed

That's fine with me too. The new commit message needs to be amended though, it's `nengo/ipynb.py` not `utils/ipynb.py`.

jgosmann:
> Also, cfb5933 not only makes the change it purports to, but also does some refactoring.

The refactoring is necessary to get rid of that warning mentioned in the commit message.

In general I find it easier to understand the changes in smaller commits than one big commit and smaller commits are also more helpful with tools like `git bisect` (as long as each commit in itself is consistent and leaves the repository in a working state).

I'm ok with squashing the changelog commit.

"Fixes to the IPythonProgressWidget" implies that `IPythonProgressWidget` was broken, but that was not the case I think. But the actual changes (detailed in the commit message) are more refactoring to keep the code simple and hide a warning.

jgosmann:
btw: "Remove IPython notebook detection (cannot be done with IPython notebook >3 aka Jupyter)" is not 100% correct. It cannot be done with IPython == 3 (not yet jupyter) either.

tbekolay:
My view is that the person making the PR has final say on the history, as they know better the intention behind each commit. That's why I like adding fixup / squash commits and getting a +1 before actually squashing / fixuping them (although admittedly I do a fair bit of squashing without an OK when a commit doesn't pass unit tests without a further commit). So, my vote would be to squash the changelog commit only.

> It cannot be done with IPython == 3 (not yet jupyter) either.

I'll amend that commit message, whichever version we move forward with!

jgosmann:
Also, I think it's better to have changes in code and movement of code to other places in separate commits as it makes for a more helpful diff for the changes. Though I admit, removing identation and moving code around might not be as critical.

The two commits I might squash in [the long history](https://github.com/nengo/nengo/compare/ipy3-more) are d81d2c7 and 6912d66 because the first of these directly caused the necessity of the second in a sense (to be able to set the IPython progress bar as a default in the config file).

hunse:
I probably use the word "fixes" too much; I use it whenever I change something. As I said, feel free to amend the message.

I understand the reasoning behind smaller commits, but there are reasons for larger ones, too. I like them because they make the history easier to read, and easier to tell what's going on. If commits depend on commits around them for context, I think that's bad (not that that's the case here). Obviously there's a balance; if commits are too small, then the history is too long and difficult to work with, and if they're too large, then any given commit is too hard to analyse. The question is where that balance is.

I think in general, we should always put changelog entries in the commits making the changes, not by themselves. I think that will help guide the size of commits, because if you're making a change that only requires one changelog entry, then it's more difficult to split it across multiple commits (though of course you could, if for example the first commit makes the change and then the following are just refactoring, etc.).

For this one, I'm fine with whatever.

jgosmann:
> If commits depend on commits around them for context, I think that's bad (not that that's the case here)

That's a case where non-fast-forward merges can make sense. (If the other option of squashing those commits together would end up with an unreasonable large commit.)

> I think in general, we should always put changelog entries in the commits making the changes, not by themselves.

I agree.

hunse:
> That's a case where non-fast-forward merges can make sense.

Yeah, I would be open to something like that. I think that can still be considered a linear history, if the branch is always rebased to master before merging.

tbekolay:
I did the squash @jgosmann suggested in [the long history](https://github.com/nengo/nengo/compare/ipy3-more). Is that version OK with everyone? Looks like the best compromise to me. Will bring that over here and merge if everyone's good with it.

hunse:
Fine by me.

jgosmann:
LGTM :cake:

PR #697, opened 2015-04-13
Fix networks

hunse:
Some minor fixes to EnsembleArray and Product. The first commit removes an unused node in Product, and allows any iterable of functions to be used in `EnsembleArray.add_output`.

The second commit might be a bit more controversial. It just removes the passthrough input nodes in Product and replaces them with ObjViews.

### DIFF
--- a/nengo/networks/ensemblearray.py
+++ b/nengo/networks/ensemblearray.py
@@ -106,7 +106,7 @@ class EnsembleArray(nengo.Network):
         sizes = np.zeros(self.n_ensembles, dtype=int)
 
         if is_iterable(function) and all(callable(f) for f in function):
-            if len(function) != self.n_ensembles:
+            if len(list(function)) != self.n_ensembles:
                 raise ValueError("Must have one function per ensemble")
 
             for i, func in enumerate(function):


tcstewar:
Nifty...  I'm cautiously in favour of replacing passthrough input nodes with ObjViews.  Can anyone think of any loss of functionality here?  I assume we can still do prod.A.size_in and prod.A.size_out.  And even the user interface stuff should be able to detect ObjViews and visually display them similarly to PassthroughNodes, if we decide we want that....

tbekolay:
It never occurred to me to replace passthrough nodes with ObjViews, but to me it seems like a great idea! In my mind it seems like passthrough nodes are more flexible, but I can't actually come up with a good use case of passthrough nodes that couldn't be done with ObjViews instead. If we choose to go this way, could we get rid of passthrough nodes in general? That would mean that it's no longer necessary to optimize them out on any backend.

tbekolay:
I think I forgot about the fact that nodes can project to multiple objects, while objviews can only project to one. Probably can't get rid of them completely, but I think it's good to favour objviews when possible.

hunse:
I think we still need passthrough nodes for e.g. the output of an EnsembleArray, because that requires multiple connections.

tcstewar:
> I think I forgot about the fact that nodes can project to multiple objects, while objviews can only project to one.

Yikes...  really?  So what happens if I do this?

```python
model = nengo.Network()
with model:
    a = nengo.Ensemble(100, 3)
    a0 = a[0]
    a1 = a[1]
    a2 = a[2]
    
    nengo.Connection(a0, a1)
    nengo.Connection(a0, a2)
    nengo.Connection(a1, a2)
    nengo.Connection(a1, a0)
    nengo.Connection(a2, a0)
    nengo.Connection(a2, a1)
```

tbekolay:
No, I mean you can't make an ObjView that will project to two separate objects. I.e.,

```python
a = nengo.Ensemble(100, 3)
b = nengo.Ensemble(100, 3)
c = a[0] + b[0]
nengo.Connection(a, c)
```
Though you can of course do this with nodes.

tcstewar:
> No, I mean you can't make an ObjView that will project to two separate objects.

Ah, got it...  very good point.

tbekolay:
Reminder to @tcstewar : check this out in the GUI

Seanny123:
Did the Product network segment of this pull request become obsolete because of Jan's new Product network that's now merged in to `master`?

tcstewar:
> Reminder to @tcstewar : check this out in the GUI

Right now in the GUI this is a little confusing, in that the GUI doesn't know about ObjView objects, so they don't show up.  I think it would be possible to add a new handler in the GUI for ObjView objects (in addition to Networks, Nodes, and Ensembles), if we decide we want to do this sort of thing.  One complication is it would also mean a new sort of Connection, since we'd want to draw a line between the ObjView and whatever object it's a view on.  But that's doable, if we decide we want this.

That said, there's one important functionality problem I just realized right now: you can't do a slice off an ObjView.  So right now, this change loses functionality, as you can no longer slice into ```prod.A``` or slice out of ```prod.output```.  That's a pretty important thing to be able to do.  But it would also be a pretty easy thing to add to an ObjView, I think.

> Did the Product network segment of this pull request become obsolete because of Jan's new Product network that's now merged in to master?

Yup, it's now obsolete.  However, even with the new approach, there's still the question of whether we should use passthrough nodes or ObjViews.  Right now, though, given the lack of ability to do slices on ObjViews, I'd definitely stick with passthrough Nodes.

Seanny123:
In that case, should I take the EnsembleArray modification, delete the rest and mark for merge?

tcstewar:
> In that case, should I take the EnsembleArray modification, delete the rest and mark for merge?

Yup, that sounds right to me.

Seanny123:
@tcstewar made the change, ready to merge

tcstewar:
Looks fine to me.  :)

tbekolay:
> there's still the question of whether we should use passthrough nodes or ObjViews

Are there other networks where this is a possibility? In `Product` we have a transform on the connection now, so it's not possible, but it'd be good to look at other networks to see if that's a possibility.

In any case, made #818 to add the ability to slice ObjViews. I'll merge this now.

PR #699, opened 2015-04-16
Spa subvocab

xchoo:
Redid implementation of vocabulary subsets. Subsets now reference the parent vocabulary, and handles transform_to correctly (doesn't add new terms as it did before). Also added a read_only flag for vocabularies to prevent unwanted addition of items to vocabularies.

### DIFF
--- a/nengo/spa/vocab.py
+++ b/nengo/spa/vocab.py
@@ -3,7 +3,7 @@ import warnings
 import numpy as np
 
 import nengo
-from nengo.exceptions import SpaParseError, ValidationError
+from nengo.exceptions import ReadonlyError, SpaParseError, ValidationError
 from nengo.spa import pointer
 from nengo.utils.compat import is_iterable, is_number, is_integer, range
 
@@ -81,6 +81,8 @@ class Vocabulary(object):
         self.include_pairs = include_pairs
         self._identity = None
         self.rng = rng
+        self.readonly = False
+        self.parent = None
 
     def create_pointer(self, attempts=100, unitary=False):
         """Create a new semantic pointer.
@@ -150,6 +152,11 @@ class Vocabulary(object):
 
         The pointer value can be a SemanticPointer or a vector.
         """
+        if self.readonly:
+            raise ReadonlyError(attr='Vocabulary',
+                                msg="Cannot add semantic pointer '%s' to "
+                                    "read-only vocabulary." % key)
+
         if not key[0].isupper():
             raise SpaParseError(
                 "Semantic pointers must begin with a capital letter.")
@@ -343,18 +350,32 @@ class Vocabulary(object):
             terms in this list that do not exist in the Vocabularies will
             be created.
         """
-        if keys is None:
-            keys = list(self.keys)
-            for k in other.keys:
-                if k not in keys:
-                    keys.append(k)
+        # If the parent vocabs of self and other are the same, then no
+        # transform is needed between the two vocabularies, so return an
+        # identity matrix.
+        my_parent = self if self.parent is None else self.parent
+        other_parent = other if other.parent is None else other.parent
 
-        t = np.zeros((other.dimensions, self.dimensions), dtype=float)
-        for k in keys:
-            a = self[k].v
-            b = other[k].v
-            t += np.outer(b, a)
-        return t
+        if my_parent is other_parent:
+            return np.eye(self.dimensions)
+        else:
+            if keys is None:
+                if self.readonly and other.readonly:
+                    keys = [k for k in self.keys if k in other.keys]
+                elif self.readonly:
+                    keys = list(self.keys)
+                elif other.readonly:
+                    keys = list(other.keys)
+                else:
+                    keys = list(self.keys)
+                    keys.extend([k for k in other.keys if k not in self.keys])
+
+            t = np.zeros((other.dimensions, self.dimensions), dtype=float)
+            for k in keys:
+                a = self[k].v
+                b = other[k].v
+                t += np.outer(b, a)
+            return t
 
     def prob_cleanup(self, similarity, vocab_size, steps=10000):
         """Estimate the chance of successful cleanup.
@@ -394,6 +415,37 @@ class Vocabulary(object):
         pcorrect = (1 - perror1) ** vocab_size
         return pcorrect
 
+    def extend(self, keys, unitary=False):
+        """Extends the vocabulary with additional keys.
+
+        Creates and adds the semantic pointers listed in keys to the
+        vocabulary.
+
+        Parameters
+        ----------
+        keys : list of strings
+            List of semantic pointer names to be added to the vocabulary.
+
+        unitary : bool or list of strings, optional
+                If True, all generated pointers to be unitary.  If a list of
+                names, any pointer whose name is on the list will be forced to
+                be unitary when created.
+        """
+        if is_iterable(unitary):
+            if is_iterable(self.unitary):
+                self.unitary.extend(unitary)
+            else:
+                self.unitary = list(unitary)
+        elif unitary:
+            if is_iterable(self.unitary):
+                self.unitary.extend(keys)
+            else:
+                self.unitary = list(keys)
+
+        for key in keys:
+            if key not in self.keys:
+                self[key]
+
     def create_subset(self, keys):
         """Returns the subset of this vocabulary.
 
@@ -407,21 +459,28 @@ class Vocabulary(object):
             new vocabulary.
 
         """
-
-        # Make new vocabulary object
-        result = Vocabulary(self.dimensions,
+        # Make new Vocabulary object
+        subset = Vocabulary(self.dimensions,
                             self.randomize,
                             self.unitary,
                             self.max_similarity,
                             self.include_pairs,
                             self.rng)
 
-        # Make a copy of the desired keys
+        # Copy over the new keys
         for key in keys:
-            result.add(key, self[key])
+            subset.add(key, self.pointers[key])
 
-        # Return the result
-        return result
+        # Assign the parent
+        if self.parent is not None:
+            subset.parent = self.parent
+        else:
+            subset.parent = self
+
+        # Make the subset read only
+        subset.readonly = True
+
+        return subset
 
 
 class VocabularyParam(nengo.params.Parameter):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -182,6 +182,10 @@ Release History
   evaluate the filter response to sine waves of given frequencies. This can
   be used to create Bode plots, for example.
   (`#945 <https://github.com/nengo/nengo/pull/945>`_)
+- ``nengo.spa.Vocabulary`` objects now have a ``readonly`` attribute that
+  can be used to disallow adding new semantic pointers. Vocabulary subsets
+  are read-only by default.
+  (`#699 <https://github.com/nengo/nengo/pull/699>`_)
 
 **Bug fixes**
 
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -80,7 +80,7 @@ class DecoderCache(object):
 
     Parameters
     ----------
-    read_only : bool
+    readonly : bool
         Indicates that already existing items in the cache will be used, but no
         new items will be written to the disk in case of a cache miss.
     cache_dir : str or None
@@ -93,8 +93,8 @@ class DecoderCache(object):
     _LEGACY = 'legacy.txt'
     _LEGACY_VERSION = 0
 
-    def __init__(self, read_only=False, cache_dir=None):
-        self.read_only = read_only
+    def __init__(self, readonly=False, cache_dir=None):
+        self.readonly = readonly
         if cache_dir is None:
             cache_dir = self.get_default_dir()
         self.cache_dir = cache_dir
@@ -252,7 +252,7 @@ class DecoderCache(object):
                 logger.debug("Cache miss [{}].".format(key))
                 decoders, solver_info = solver_fn(
                     solver, neuron_type, gain, bias, x, targets, rng=rng, E=E)
-                if not self.read_only:
+                if not self.readonly:
                     with open(path, 'wb') as f:
                         nco.write(f, solver_info, decoders)
             else:
--- a/nengo/spa/tests/test_vocabulary.py
+++ b/nengo/spa/tests/test_vocabulary.py
@@ -30,7 +30,7 @@ def test_include_pairs(rng):
     v['D']
     assert v.key_pairs == ['A*B', 'A*C', 'B*C', 'A*D', 'B*D', 'C*D']
 
-    v = Vocabulary(12, include_pairs=True)
+    v = Vocabulary(12, include_pairs=True, rng=rng)
     v['A']
     v['B']
     v['C']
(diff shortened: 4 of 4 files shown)


xchoo:
@tcstewar Please review! =D

nengo/spa/vocab.py
```diff
@@ -81,6 +81,10 @@ def __init__(self, dimensions, randomize=True, unitary=False,
         self.include_pairs = include_pairs
         self._identity = None
         self.rng = rng
+        self.read_only = False
+
+        # The parent of a non-subset vocabulary is itself
+        self.parent = self
```
tbekolay: Hmm... this seems like a red flag to me that this implementation is overly complex. Do you really need parent references? It makes managing memory and ensuring that these object get garbage collected very complicated...

xchoo: Oh. Hah. This is so that I can shortcut the comparison in the transform_to function. XD

jgosmann: Doesn't Python garbage collection handle cyclical references?

tbekolay: It's optional and [not guaranteed](https://docs.python.org/2/reference/datamodel.html#objects-values-and-types). It's not so much the garbage collection that I'm worried about, it just seems like an antipattern to me so my eyebrow instinctively raises. If there's absolutely no other way to do this, then so be it, but I suspect there's another way that ends up being simpler and easier.

nengo/spa/vocab.py
```diff
@@ -81,6 +81,8 @@ def __init__(self, dimensions, randomize=True, unitary=False,
         self.include_pairs = include_pairs
         self._identity = None
         self.rng = rng
+        self.read_only = False
+        self.parent = None
```
xchoo: @tbekolay There. I removed the cyclic reference. Check the ```transform_to``` function to see if that will freak out the garbage collector.

tcstewar:
I think I'm closer to understanding what this Subset stuff is being used for, but I there are still a few questions in my mind:

- What is the use case for adding terms to a VocabularySubset?  All the examples I've heard of the subset use (helpers for plotting, and creating AssociativeMemorys) don't require adding terms.  It seems to me like a VocabularySubset should be read_only.  That would eliminate the behaviour where adding terms to a subset automatically adds those terms to the parent (which seems a bit surprising to me, but that could just be because I'm not seeing the use case).
- Should VocabularySubset be a subclass of Vocabulary, or should we get rid of it entirely and just have ```create_subset``` return a suitably configured Vocabulary instance?
- What are the use cases for ```extend_subset```?  If this functionality is needed, shouldn't it be in the base class?

xchoo:
- What is the use case for adding terms to a VocabularySubset?
I was thinking of use cases outside of the networks we build in nengo. It is conceivable that people can use the subsets to do purely mathematical operations on vocabularies. And I can see that in some cases (sequential programming, or reusing the same subset for different purposes), subsets can be extended. I very nearly put in the ability to remove items from the vocabulary, but I didn't (yet.. haha)

- Should VocabularySubset be a subclass of Vocabulary, or should we get rid of it entirely and just have ```create_subset``` return a suitably configured Vocabulary instance?
Even with the read_only functionality of the Vocabularies, this will create problems when connecting two modules with different subsets. Suppose I had this:

```python
num = spa.Buffer(vocab=num_subset)  # num_subset = {'ONE', 'TWO', 'THR'} of vocab
pos = spa.Buffer(vocab=pos_subset)  # pos_subset = {'P1', 'P2', 'P3'} of vocab
mem = spa.Buffer(vocab=vocab)  # vocab = {'ONE', 'TWO', ..., 'P1', 'P2', ...} + PAIRS

actions = spa.Actions("mem = num * pos")
cortical = spa.Cortical(actions)
```

In this case, because everything is (essentially) using the one vocabulary, there should be no transforms between either of the connections. If the subset returned a read-only vocabulary, the conv_effect will add a transform from (e.g. the num buffer) ```{'ONE', 'TWO', 'THR'}``` of ```num_subset``` to ```{'ONE', 'TWO', 'THR'}``` of ```vocab``` (which is not an identity matrix), which then adds noise to the calculations. 

- What are the use cases for extend_subset? If this functionality is needed, shouldn't it be in the base class?
See point 1 for use cases. Should it be in the base class? Maybe? I just added it as a shortcut for ```vocab.parse('+'.join(keys))```.

tcstewar:
> I was thinking of use cases outside of the networks we build in nengo. It is conceivable that people can use the subsets to do purely mathematical operations on vocabularies. And I can see that in some cases (sequential programming, or reusing the same subset for different purposes), subsets can be extended. I very nearly put in the ability to remove items from the vocabulary, but I didn't (yet.. haha)

Hmm...  I have a fairly strong resistance to adding functionality based on possible conceivable future uses, rather than on actual use-cases that have come up.  I'd prefer to add functionality as we need it, rather than trying to guess what functionality we might want in the future.

tcstewar:
> In this case, because everything is (essentially) using the one vocabulary, there should be no transforms between either of the connections.

Ah, I think this example helps clarify things for me.  In the style of programming I was thinking of, I would definitely have each of those using the same vocab, since, as you point out, they are all using the same vocabulary, and so I think the code should actually indicate that.  However, the way you're using vocabs, you're wanting to indicate something about the eventual visual display: you only want ```{'ONE', 'TWO', 'THR'}``` to be shown for num and ```{'P1', 'P2', 'P3'}``` to be shown for pos.  And so you're defining subsets to do that.  Is that a fair summary of what you're doing?

If so, I've always been working from the assumption that if you want identity transforms then you should be using the same vocabulary on your components.  And if you want to control aspects of the visual display, you should do that outside of the core model code, since it's just about the visual display.

So I'm a bit worried about this change.  I think we should be working to make the ```spa``` module clearer and with less automatic magic (or at least a small set of particular magic that we are explicit about), rather than adding new magic.  This change feels like new magic to me, but it could just be because I'm not thinking about it in the right way.

tcstewar:
> Should VocabularySubset be a subclass of Vocabulary, or should we get rid of it entirely and just have create_subset return a suitably configured Vocabulary instance?
> Even with the read_only functionality of the Vocabularies, this will create problems when connecting two modules with different subsets.

On re-reading your comment, I meant to refer to a different issue with this question.  I was meaning that VocabularySubset shouldn't exist.  There'd still be the ```parent``` reference in Vocabulary, though, so ```transform_to``` would still be perfectly capable of handling this case (indeed, that's exactly what it does right now).

tcstewar:
> Should it be in the base class? Maybe? I just added it as a shortcut for vocab.parse('+'.join(keys)).

I think that's a useful functionality to have. I was just wondering why it was only being put in the VocabularySubset -- that looked to me like it was an indication that it was something you only wanted to do in a Subset.  But I'm also still unconvinced that Subset should exist at all, so I'm a bit weird.  ;)  But we certainly do that ```vocab.parse('+'.join(keys))``` thing a fair bit, and this would be much clearer, so I think it should be added to Vocabulary.

xchoo:
Looking at my code, I actually don't use subsets within the model itself. So I've merged the subset logic (parent + the read only code) back into the Vocabulary class. I'm also of the opinion that within the model, all of the vocabularies (for modules using the same vocabulary) should not be using subsets, but it is a concept that should be better documented with the spa modules. I can see new users using subsets for modules because "this module should only be used for XXX subset" (when really it doesn't make a difference).

nengo/spa/vocab.py
```diff
@@ -379,6 +400,20 @@ def prob_cleanup(self, similarity, vocab_size, steps=10000):
...
+        Creates and adds the semantic pointers listed in keys to the
+        vocabulary.
+
+        Parameters
+        ----------
+        keys : list of strings
+            List of semantic pointer names to be added to the vocabulary.
+
+        """
+        self.parse('+'.join(keys))
```
tcstewar: for clarity, it might be nicer to implement this as

```python
for k in keys:
    self[k]
```

or perhaps even

```python
for k in keys:
    if k not in self.keys:
        self.add(k, self.create_pointer())
```

or even

```python
def extend(self, keys, unitary=False):
    for k in keys:
        if k not in self.keys:
            self.add(k, self.create_pointer(unitary=unitary))
```

tcstewar:
> Looking at my code, I actually don't use subsets within the model itself. So I've merged the subset logic (parent + the read only code) back into the Vocabulary class. I'm also of the opinion that within the model, all of the vocabularies (for modules using the same vocabulary) should not be using subsets, but it is a concept that should be better documented with the spa modules. I can see new users using subsets for modules because "this module should only be used for XXX subset" (when really it doesn't make a difference).

I like the look of it merged back in like that.  :)  And I think you're completely right that this subset concept is something that needs serous documentation and clarity.  That said, I think this current PR is a clear improvement (the read_only parameter seems very useful).  I think in a future PR we might look into moving away from subsets and towards explicit specification of sets of keys (but keeping the same underlying Vocabulary, so we'd do things like specifying what terms to show in the display, or specifying what terms to include in the AssociativeMemory).  But that's for future work.  :)

xchoo:
The associative memory has already been re-coded to use a vocabulary and list of keys (see #702).

tcstewar:
> The associative memory has already been re-coded to use a vocabulary and list of keys (see #702).

Sweet!  :)

nengo/spa/vocab.py
```diff
@@ -379,6 +400,33 @@ def prob_cleanup(self, similarity, vocab_size, steps=10000):
...
+        Creates and adds the semantic pointers listed in keys to the
+        vocabulary.
+
+        Parameters
+        ----------
+        keys : list of strings
+            List of semantic pointer names to be added to the vocabulary.
+
+        """
+        if is_iterable(unitary):
```
tbekolay: What does this if/elif block do exactly? To me it reads like `unitary` can be a boolean or an iterable. Is that true? If so, why?

nengo/spa/vocab.py
```diff
@@ -328,18 +334,33 @@ def transform_to(self, other, keys=None):
...
-            b = other[k].v
-            t += np.outer(b, a)
-        return t
+        # If the parent vocabs of self and other are the same, then no
+        # transform is needed between the two vocabularies, so return an
+        # identity matrix.
+        my_parent = self if self.parent is None else self.parent
+        other_parent = other if other.parent is None else other.parent
+
+        if my_parent == other_parent:
```
tbekolay: Is short-circuiting this check the only reason to keep the parent links around? If so, I strongly suggest we remove the parent references. They can cause lots of problems down the road, and in the best case it's an example of premature optimization. Does this function really take that much time?

nengo/spa/vocab.py
xchoo: That is exactly how it reads. It reflects the behaviour of the ```unitary``` parameter in the constructor where you can either pass it a boolean (in which case, all vectors created will be / not be unitary), or a list of strings (in which case only semantic pointers that are within said list will be unitary).

nengo/spa/vocab.py
xchoo: No. The parent links are there so that if you try to connect two modules where the vocabulary of both modules have the same parent (or if a child is connecting to a parent), the transform returned is an identity transform. Without this check, this function will return a not-quite identity transform adding noise to the connection.

nengo/spa/vocab.py
jgosmann: I think, that should be documented because it is not obvious that `unitary` excepts a list of strings.

hunse:
@xchoo, @tcstewar: What's the status of this? @tcstewar, are you happy with it?

hunse:
Can this be rebased and @xchoo and @tcstewar sign off so we can get it done?

xchoo:
Busy at the moment. Earliest I can look at this is next monday.

tbekolay:
I've rebased this and squashed the commits together. Definitely the commit message should be looked at, but other than that, would be good for a re-review from @tcstewar and/or @xchoo!

nengo/spa/vocab.py
```diff
@@ -147,6 +149,17 @@ def add(self, key, p):
...
+        if self.read_only:
+            if self.parent is None:
+                # If this vocabulary is not a subset vocabulary, throw an error
+                raise ValueError("Cannot add semantic pointer '%s' to "
+                                 "read-only vocabulary." % key)
+            else:
+                # If this vocabulary is a subset vocabulary, apply operation on
+                # parent vocab.
+                self.parent.add(key, p)
+                return
```
tcstewar: Why is this fall-back here for the read_only case?  This allows this sort of thing to happen, which seems confusing to me:

```python
import nengo.spa

v = nengo.spa.Vocabulary(16)
v.extend(['A', 'B', 'C'])
v2 = v.create_subset(['A', 'B'])
v2.add('Z', v2.create_pointer())

print 'v.keys', v.keys
print 'v2.keys', v2.keys
```

```
v.keys ['A', 'B', 'C', 'Z']
v2.keys ['A', 'B']
```

It seems to me that it should raise a ValueError, rather than trying to create it in the parent.

nengo/spa/vocab.py
```diff
@@ -334,18 +347,32 @@ def transform_to(self, other, keys=None):
...
+            if keys is None:
+                if self.read_only and other.read_only:
+                    keys = [k for k in self.keys if k in other.keys]
+                elif self.read_only:
+                    keys = list(self.keys)
+                elif other.read_only:
+                    keys = list(other.keys)
+                else:
+                    keys = list(self.keys)
+                    keys.extend(other.keys)
```
tcstewar: If a key shows up in both lists, then it'll now be added *twice*!  This should probably be done as a ```set```.

nengo/spa/vocab.py
xchoo: This is to handle the cases where something is added to a vocabulary that is a subset of another vocabulary. Basically because of the magic that happens in the `transform_to` function (it add keys from one vocabulary to another) of the `Vocabulary` class. 

If we made it a `ValueError` this would fail:
```
vocab1 = Vocabulary(32)
vocab1.parse('A+B+C')
vocab2 = vocab1.create_subset(['A', 'B'])

with spa.SPA() as model:
    model.state_in = spa.State(vocab=vocab1)
    model.am = spa.AssociativeMemory(vocab2)
    model.state_out = spa.State(vocab=vocab1)

    cortical_actions = spa.Actions('am=state_in', 'state_out=am')
    
    model.cortical = spa.Cortial(cortical_actions)
```

The transform_to system will try to add `'C'` to `vocab2` and fail with the `ValueError`.

xchoo: I'll add a note that this might not actually happen with the conditional code below. But there may be cases where this will happen (i.e. the `transform_to` magic adds keys from one vocab to another)

jgosmann: I think we should get rid of the `transform_to` magic and be explicit about transforms. But that should be the scope of another PR.

tcstewar: As per the discussion in the meeting, I think the solution for now is to adjust this PR such that if it's read only then it always raises the ValueError, but then also modify transform_to so that it handles that gracefully (basically if a vocab is read only and it doesn't have a key, then skip that key rather than trying to generate it).

tcstewar: And yes, in the long run, we want to get rid of the magic transform_to and be explicit about it when making the SPA Actions, but that's a separate PR.

tcstewar:
This looks good to me now!  Seems to work well, and is a nice improvement over the current situation.  In the long run I still need to get rid of the auto-transform, but that's for the future.  :)

tbekolay:
Cool cool... is there anything in here worth mentioning in the changelog?

tbekolay:
OK, I added a few fixups here, to change the ValueError to ReadonlyError, rename read_only to readonly to be consistent with the rest of Nengo and add a changelog entry. @xchoo and @tcstewar I'll merge if these look good to both of you!

xchoo:
:+1:

xchoo:
You can remove 
"- Adding items to subset vocabularies (i.e. read-only vocabularies
  with a parent vocabulary) now applies this action to the parent
  vocabulary instead of throwing an exception."
from the commit message. That change was reverted.

tbekolay:
Thanks, will do :)

tcstewar:
LGTM!

PR #700, opened 2015-04-16
Added function to evaluate connection's decoding

hunse:
This required a minor change to the builder, ~~which may break derivative builders~~.

Also fixed a bug in `nengo.utils.ensemble.tuning_curves` to avoid input points being modified.

EDIT: I refactored this to keep the `build_linear_system` function, so that nothing should break in other builders.

### DIFF
--- a/nengo/utils/connection.py
+++ b/nengo/utils/connection.py
@@ -57,3 +57,49 @@ def target_function(eval_points, targets):
     return {'function': function,
             'eval_points': eval_points,
             'scale_eval_points': False}
+
+
+def eval_point_decoding(conn, sim, eval_points=None):
+    """Get the targets and actual decoded values for a set of eval points.
+
+    This function evaluates the static decoding (i.e. using the neuron type's
+    `rates` function) of a connection for a given set of evaluation points.
+
+    Parameters
+    ----------
+    conn : Connection
+        The Connection to evaluate the decoding of.
+    sim : Simulator
+        A Nengo simulator storing the built connection.
+    eval_points : array_like (N, E) (optional)
+        An N x E array of evaluation points to evaluate the decoding for, where
+        N is the number of points and E is the dimensionality of the input
+        ensemble (i.e. `conn.size_in`). If None (default), use the connection's
+        training evaluation points.
+
+    Returns
+    -------
+    eval_points : ndarray (N, E)
+        A shallow copy of the evaluation points used. E is the dimensionality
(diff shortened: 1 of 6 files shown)


nengo/utils/connection.py
```diff
@@ -57,3 +57,49 @@ def function(x):
...
+    eval_points : ndarray (N, E)
+        A shallow copy of the evaluation points used. E is the dimensionality
+        of the connection input ensemble (i.e. `conn.size_in`).
+    targets : ndarray (N, D)
+        The target function value at each evaluation point.
+    decoded : ndarray (N, D)
+        The decoded function value at each evaluation point.
+    """
+    from nengo.builder.ensemble import get_activities
+    from nengo.builder.connection import get_targets
```
hunse: Something was breaking when I put these imports at the top of the file, so I put them here.

tbekolay: Circular imports, I'd guess.

hunse: Maybe, but the exception didn't read like that.

tbekolay: ¯\_(ツ)_/¯

jgosmann:
I foresee merge conflicts with #677 ...

hunse:
Likely, but I don't think they should be too hard to resolve. I didn't change too much here.

I'm more worried about all the other PRs on the stack right now. I forsee a lot of merge conflicts in the near future.

nengo/utils/connection.py
tbekolay: Looked into it... I suppose it's not a circular import at the module level, but it is on the package level... this causes the `utils` package to depend on `builder`, but `builder` already depends on `utils` so neither can be imported first. So yeah, I think putting the import here is the right call.

tbekolay:
This LGTM. Needs a changelog entry, and a test if it's not too much trouble.

hunse:
Added a changelog entry and a test.

tbekolay:
Made a few changes; give them a look @hunse and I'll merge if they look good to you.

As an aside, the Sigmoid neuron type gives interesting results for this!

utils test_connection test_eval_point_decoding

hunse:
LGTM. You're impressed with the sigmoid because the errors are smaller and more regular than for e.g. LIF?

tbekolay:
Not impressed, just interested that it's such a regular pattern. Nothing impresses me anymore...

PR #706, opened 2015-04-19
Fix #704.

### LINKED ISSUE #704: Failure running unit tests.
While `py.test --pyargs nengo` seems to work, a pure `py.test` quits with this error:
```
(nengodev)ctn02:nengo|master> py.test                                                                                                   ✭? dc9fecbc5d
================================================================ test session starts =================================================================
platform linux2 -- Python 2.7.9 -- py-1.4.26 -- pytest-2.6.4
plugins: cov
collecting 618 itemsINTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/main.py", line 84, in wrap_session
INTERNALERROR>     doit(config, session)
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/main.py", line 121, in _main
INTERNALERROR>     config.hook.pytest_collection(session=session)
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/core.py", line 413, in __call__
INTERNALERROR>     return self._docall(methods, kwargs)
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/core.py", line 424, in _docall
INTERNALERROR>     res = mc.execute()
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/core.py", line 315, in execute
INTERNALERROR>     res = method(**kwargs)
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/main.py", line 125, in pytest_collection
INTERNALERROR>     return session.perform_collect()
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/main.py", line 548, in perform_collect
INTERNALERROR>     config=self.config, items=items)
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/core.py", line 413, in __call__
INTERNALERROR>     return self._docall(methods, kwargs)
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/core.py", line 424, in _docall
INTERNALERROR>     res = mc.execute()
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/core.py", line 315, in execute
INTERNALERROR>     res = method(**kwargs)
INTERNALERROR>   File "/home/jgosmann/Documents/projects/nengo/nengo/tests/conftest.py", line 237, in pytest_collection_modifyitems
INTERNALERROR>     compare = config.getvalue('compare') is None
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/config.py", line 840, in getvalue
INTERNALERROR>     return self.getoption(name)
INTERNALERROR>   File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/_pytest/config.py", line 836, in getoption
INTERNALERROR>     raise ValueError("no option named %r" % (name,))
INTERNALERROR> ValueError: no option named 'compare'

==================================================================  in 1.04 seconds ==================================================================
```

jgosmann:
The conftest.py file is not discovered in a subdirectory if that
directory is not passed as argument to py.test. It is recommended to
keep it in the project root (or top level test) directory:
http://pytest.org/latest/plugins.html#plugin-discovery-order-at-tool-startup

### DIFF
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -190,26 +190,6 @@ def pytest_generate_tests(metafunc):
             "nl_nodirect", [LIF, LIFRate, RectifiedLinear, Sigmoid])
 
 
-def pytest_addoption(parser):
-    parser.addoption(
-        '--plots', nargs='?', default=False, const=True,
-        help='Save plots (can optionally specify a directory for plots).')
-    parser.addoption(
-        '--analytics', nargs='?', default=False, const=True,
-        help='Save analytics (can optionally specify a directory for data).')
-    parser.addoption(
-        '--compare', nargs=2,
-        help='Compare analytics results (specify directories to compare).')
-    parser.addoption(
-        '--logs', nargs='?', default=False, const=True,
-        help='Save logs (can optionally specify a directory for logs).')
-    parser.addoption('--noexamples', action='store_false', default=True,
-                     help='Do not run examples')
-    parser.addoption(
-        '--slow', action='store_true', default=False,
-        help='Also run slow tests.')
-
-
 def pytest_runtest_setup(item):
     for mark, option, message in [
             ('example', 'noexamples', "examples not requested"),
--- a/nengo/tests/options.py
+++ b/nengo/tests/options.py
@@ -1,0 +1,18 @@
+def pytest_addoption(parser):
+    parser.addoption(
+        '--plots', nargs='?', default=False, const=True,
+        help='Save plots (can optionally specify a directory for plots).')
+    parser.addoption(
+        '--analytics', nargs='?', default=False, const=True,
+        help='Save analytics (can optionally specify a directory for data).')
+    parser.addoption(
+        '--compare', nargs=2,
+        help='Compare analytics results (specify directories to compare).')
+    parser.addoption(
+        '--logs', nargs='?', default=False, const=True,
+        help='Save logs (can optionally specify a directory for logs).')
+    parser.addoption('--noexamples', action='store_false', default=True,
+                     help='Do not run examples')
+    parser.addoption(
+        '--slow', action='store_true', default=False,
+        help='Also run slow tests.')
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,6 +17,7 @@ ignore = E123,E133,E226,E241,E242,E731,W503
 max-complexity = 10
 
 [pytest]
+addopts = -p nengo.tests.options
 norecursedirs = .* *.egg build dist docs
 markers =
     example: Mark a test as an example.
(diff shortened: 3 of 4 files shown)


jgosmann:
This just seemed like it would work, because of an old `.pyc` file that was not deleted.

tbekolay:
I think probably the only way to make this work would be to make a `conftest.py` in the top-level that just does `from nengo.tests.conftest import *`...

jgosmann:
That seems to lead to other problems.

jgosmann:
I came up with another fix which seems to work as intended and solves other problems too.

tbekolay:
Oh nice! Does this work because `tests/options.py` is a specially named file, or because of what's in `setup.cfg`? Would putting `addopts = -p nengo.tests.conftest` give the same result without having to move the options stuff? My main concern with moving the options is that it means that backends will have to import this stuff into their projects too, so the less files they have to import the better.

tbekolay:
Whoops, just read the commit message now. Questions answered!

jgosmann:
The fix seems to break tox ... *sigh*

jgosmann:
Fixed tox ... there is is still a problem with the examples. They are not installed and thus the directory is not found.

hunse:
Could we check for the examples directory and skip the examples tests if we can't find it?

jgosmann:
We could, but I think we should rather ensure that the examples get installed to the right location and we do the right thing to discover that location.

jgosmann:
Also to make sure that the examples are not silently skipped in the tests.

jgosmann:
I think, the examples problem is not really related to what is fixed here and is a bug which was obscured by other buggy stuff. The best is probably to go forward with reviewing and merging this PR while discussing the solution to the examples problem in a different issue. The problem also occurs only when running tox, py.test and Travis-CI work.

tbekolay:
This LGTM except for the first commit message. The subject line should be readable without having to go to Github to see what issue #704 is (though the body of the message should say that this fixes #704). Once that's amended I'll merge!

tbekolay:
Or, I'm happy to amend it if you've got a busy day ;)

jgosmann:
done :cake:

Issue #705, opened 2015-04-17
Query supported features of backend


jgosmann:
I'm working on making it easier to run the Nengo tests against different backends and it occurred to me that it might be useful if it were possible to query a backend for supported features in a standardized way. That would make it possible …
* … to write models in a backend agnostic way by checking for certain features and do some stuff to work around a missing feature.
* … to detect automatically which tests have to be passed by the backend.

tbekolay:
+1! I think we could probably figure out a decent syntax for this.

The hard part is going to be actually enumerating out all of the possible features that could exist, and how they interact... it's probably possible, but that would definitely be a Big Job. We also need to make sure that this additional documentation requirement is beneficial for backend implementers, or else it will be seen as a hurdle they have to jump through for no reason.

mundya:
As an external backend developer: +1

> The hard part is going to be actually enumerating out all of the possible features that could exist, and how they interact...

The problem I can see here is that the supported features can be quite specific/arcane.  For example, SpiNNaker currently doesn't support connections into neurons unless the connection is from some output/decoded output with a transform that delivers the same value to every neuron (this is for global inhibition support).  While I intend to support neuron->neuron connections soon it is very unlikely that we'll ever do x->neurons with anything but a global-inhibition-like transform.  I can't think of a pleasant way of expressing this.

tcstewar:
> The problem I can see here is that the supported features can be quite specific/arcane. For example, SpiNNaker currently doesn't support connections into neurons unless the connection is from some output/decoded output with a transform that delivers the same value to every neuron (this is for global inhibition support). While I intend to support neuron->neuron connections soon it is very unlikely that we'll ever do x->neurons with anything but a global-inhibition-like transform. I can't think of a pleasant way of expressing this.

I wonder if it might make sense to directly base these queries on the backend in terms of just which nengo/tests it passes.  So you could do something like ask if a particular backend passes ```nengo/tests/test_connection/test_neurons_to_ensemble```.  Something like ```if nengo_spinnaker.supports(nengo.tests.test_connection.test_neuron_to_ensemble)```.

It might be possible to set that up such that a backend developer wouldn't have to do anything extra to support that.  That way, instead of coming up with a new way of describing all possible things a backend might support, we just directly use the test suite itself as that ontology.   (and if there are any backends which, like SpiNNaker, require a finer division in terms of supported features, then that just indicates new tests to add).

[...]

PR #709, opened 2015-04-21
Pytest simulator and neuron argument

jgosmann:
(Based on PR #706.)

Adds `--simulator`, `--ref-simulator`, and `--neurons` arguments to py.test. They set the simulator, reference simulator, and neuron types under test.

Should the reference simulator be settable? Shouldn't it be always nengo.Simulator? [But nengo_ocl seems to set it to OclSimulator.](https://github.com/nengo/nengo_ocl/blob/master/nengo_ocl/tests/test_sim_ocl.py#L39)

This is a step towards less boiler plate code to run other backends against the Nengo tests. A PR for nengo_ocl will follow once #705 is resolved.

### DIFF
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -1,5 +1,6 @@
 import hashlib
 import inspect
+import importlib
 import os
 import re
 
@@ -11,15 +12,40 @@ import nengo
 import nengo.utils.numpy as npext
 from nengo.neurons import Direct, LIF, LIFRate, RectifiedLinear, Sigmoid
 from nengo.rc import rc
-from nengo.simulator import Simulator as ReferenceSimulator
 from nengo.utils.compat import ensure_bytes, is_string
 from nengo.utils.testing import Analytics, Logger, Plotter
 
-test_seed = 0  # changing this will change seeds for all tests
+
+class TestConfig(object):
+    """Parameters affecting all Nengo tests.
+
+    These are essentially global variables used by py.test to modify aspects
+    of the Nengo tests. We collect them in this class to provide a
+    mini namespace and to avoid using the ``global`` keyword.
+
+    The values below are defaults. The functions in the remainder of this
+    module modify these values accordingly.
+    """
+
+    test_seed = 0  # changing this will change seeds for all tests
+    Simulator = nengo.Simulator
+    RefSimulator = nengo.Simulator
 
 
 def pytest_configure(config):
     matplotlib.use('agg')
+
+    if config.getoption('simulator'):
+        TestConfig.Simulator = load_class(config.getoption('simulator')[0])
+    if config.getoption('ref_simulator'):
+        refsim = config.getoption('ref_simulator')[0]
+        TestConfig.RefSimulator = load_class(refsim)
+
+
+def load_class(fully_qualified_name):
+    mod_name, cls_name = fully_qualified_name.rsplit('.', 1)
+    mod = importlib.import_module(mod_name)
+    return getattr(mod, cls_name)
 
 
 @pytest.fixture(scope="session")
@@ -29,7 +55,7 @@ def Simulator(request):
     Please use this, and not ``nengo.Simulator`` directly. If the test is
     reference simulator specific, then use ``RefSimulator`` below.
     """
-    return ReferenceSimulator
+    return TestConfig.Simulator
 
 
 @pytest.fixture(scope="session")
@@ -40,7 +66,7 @@ def RefSimulator(request):
     Other simulators may choose to implement the same API as the
     reference simulator; this allows them to test easily.
     """
-    return ReferenceSimulator
+    return TestConfig.RefSimulator
 
 
 def recorder_dirname(request, name):
@@ -50,7 +76,7 @@ def recorder_dirname(request, name):
     elif not record:
         return None
 
-    simulator, nl = ReferenceSimulator, None
+    simulator, nl = TestConfig.RefSimulator, None
     if 'Simulator' in request.funcargnames:
         simulator = request.getfuncargvalue('Simulator')
     if 'nl' in request.funcargnames:
@@ -169,7 +195,7 @@ def rng(request):
     This should be used in lieu of np.random because we control its seed.
     """
     # add 1 to seed to be different from `seed` fixture
-    seed = function_seed(request.function, mod=test_seed + 1)
+    seed = function_seed(request.function, mod=TestConfig.test_seed + 1)
     return np.random.RandomState(seed)
 
 
@@ -180,7 +206,7 @@ def seed(request):
     This should be used in lieu of an integer seed so that we can ensure that
     tests are not dependent on specific seeds.
     """
-    return function_seed(request.function, mod=test_seed)
+    return function_seed(request.function, mod=TestConfig.test_seed)
 
 
 def pytest_generate_tests(metafunc):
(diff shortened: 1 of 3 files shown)


tbekolay:
This is very nice, definitely easier than the current method for backends to get access to Nengo's tests.

The main thing that I think this doesn't address is that when testing a backend, you don't want to waste time testing all the frontend stuff. The `load_tests` function which we use now will filter out any test that doesn't contain `Simulator` (or `RefSimulator` for some backends).

Definitely it's a fixable thing though. Here are a few options off the top of my head...

1. Add a `--frontend` option (or `--nofrontend` or `--backend` or something), which if not set, will skip any tests that don't accept the `Simulator` fixture (or `RefSimulator` and `Simulator` if `--ref-simulator` is set).
2. Skip tests that don't accept the `Simulator` or `RefSimulator` fixture if `--simulator` or `--ref-simulator` is explicitly set (or, similarly, skip them if the simulator used is not `nengo.Simulator`).

The other thing I'm a little concerned about is setting the neuron types. It seems like parsing that string is going to be pretty brittle; right now we say that there's one argument, which is a comma separated string. But if there's a space in that string it's going to break, right? I think that's going to trip people up. It's also not clear to me how this works if we were to, say, set the `--neurons` argument in Nengo's `setup.cfg`, and also set it in Nengo OCL's `setup.cfg`. Which takes precedence? Do the `setup.cfg`s always be evaluated, or do you have to be in the correct directory?

> Should the reference simulator be settable? Shouldn't it be always nengo.Simulator?

"Reference Simulator" is maybe not the best term here... essentially we've taken this to mean any simulator that uses the Signal / Ops stuff that the reference simulator uses. So, it should be testing things that the reference builder / simulator does that not all simulators have to do exactly the same, but if a backend uses the Nengo reference builder and therefore expects to implement the signals / ops the same way that the reference simulator does (as Nengo OCl and I believe the MPI backend do) then it should presumably do exactly the same thing as the reference simulator.

jgosmann:
1. Selecting a subset of tests for backends shouldn't be a problem. We just have to agree on how that is done.
2. We probably should trim the neuron type strings from whitespace after splitting.
3. Only one `setup.cfg`/`pytest.ini` will be read by pytest. [Which file exactly depends on the paths passed to `py.test`.](https://pytest.org/latest/customize.html) Note that tests loaded from a package with `--pyargs` will not add to the list of paths for this and the current working directory seems to be always included. Thus, `py.test --pyargs nengo`, `py.test --pyargs nengo .`, `py.test --pyargs nengo nengo_ocl/tests`, `py.test nengo_ocl`, `py.test`, ... in the `nengo_ocl` directory all load the `nengo_ocl/setup.cfg`. What might fail (in the sense that you have to pass the arguments in `setup.cfg` yourself) is if you try to run the backend tests from another directory which doesn't seem like a huge issue to me.

tbekolay:
1. Agreed. I think my +1 is for option 2: skip frontend tests `if Simulator is not nengo.Simulator`. A bit magical, but hopefully in a good way.
2. It's not the stripping that I'm worried about (though we should also do that) it's how the shell is going to parse the `py.test` invocation. Right now, we say that `neurons` is a single argument. But if someone says `py.test --neurons nengo.LIF, nengo.LIFRate` it will be parsed as two separate arguments unless they do `py.test --neurons "nengo.LIF, nengo.LIFRate"`. I suspect this will be confusing to many people... but I'm not too sure the way around it. I think it's just raising a red flag to me that this is not an easy thing to specify at the command line (unless there's a clever way to do it that I'm not aware of).
3. Good to know! I'll add this to the testing docs.

jgosmann:
People should know how their shell parses their arguments. We could add to the documentation that they should avoid spaces ...
Otherwise we could pass the name of a file specifying the neuron types. But that seems unnecessarily complex to me. Maybe there is a way to add additional parameters to `pytest.ini`/`setup.cfg` files that can be read out. But that doesn't seem to be documented. :/

hunse:
If I'm testing a backend, I don't want to have to specify neuron types every time I run the tests. It seems to me that backends should say what neuron types they support/want to test with, and that's what we use.

Maybe we can still have an argument to override the default neuron types for that backend, but in that case it would be a pretty advanced and rarely used feature, so I'm not worried if we have to document it in a bit more detail somewhere.

jgosmann:
You don't have to specify the neuron types each time you run the test if you put them into `pytest.ini`/`setup.cfg`.
Having the backend indicate supported neuron types would work, too. But that alone would not allow to test additional neuron types (for example, biospaun/nrnengo defines new neuron types, but uses the reference backend).

tbekolay:
Reminder to @tbekolay : try this out with `nengo_distilled`

hunse:
Got this rebased and fixed some small errors.

I've been trying it out on Nengo OCL (you have to fold in the `expose-synapses` branch to this one on the Nengo side, and check out the `processes` branch on the OCL side). It seems that there's some strange memory leak, since part way through the tests I start getting OCL errors when creating new contexts, saying "out of host memory" (though there is still a lot of memory available on my machine). The difference is when I run the tests through `test_sim_ocl.py` in the Nengo OCL project, it creates one context and uses this for all the Simulators, whereas using the `--simulator` arg creates many. I'm not sure how to address this.

hunse:
Oh, I see there is some stuff to use the same context in nengo/nengo_ocl#65. I'll check that out.

jgosmann:
> The difference is when I run the tests through test_sim_ocl.py in the Nengo OCL project, it creates one context and uses this for all the Simulators, whereas using the --simulator arg creates many.

Only creating one context if possible might be good to do for performance reasons. But it seems to me that creating multiple contexts shouldn't be a problem if it is implemented correctly (and old contexts are closed) and that the memory leak hints and a bug in this area.

hunse:
Yeah, it has something to do with the contexts not getting garbage collected. If I do this in a script, it runs fine:

    import pyopencl as cl
    for i in range(1000):
        ctx = cl.create_some_context()

but if I do this, it gives me the "out of host memory" error:

    import pyopencl as cl
    contexts = []
    for i in range(1000):
        contexts.append(cl.create_some_context())

I'm guessing that it might have to do with how `pytest` deals with fixtures. It might keep some reference to the simulator in the fixture, stopping it from being collected.

tbekolay:
Hm, interesting... figuring this out could be quite helpful for SpiNNaker too.

hunse:
I just tried this out again, and I'm still getting that `OUT_OF_HOST_MEMORY` error. In the interest of getting 2.1.0 out, I'm fine merging this, since all it does is add some options, and if you run without those options, everything works exactly as it does currently. But to get this actually working with Nengo OCL, some more work is going to be required, so maybe we'll have to make an issue (whether on the Nengo side or the Nengo OCL side, I don't know).

tbekolay:
Is closing / removing OCL context something that can be done in `Simulator.close`?

hunse:
> Is closing / removing OCL contexts something that can be done in `Simulator.close`?

I think I remove all handles, so the context should get GC'd automatically. I'll have to look in more detail to see why they're not.

tbekolay:
I refactored a tiny bit in 8ae721172457eb3e, but aside from that this looks good to me. If the refactoring looks OK to you @jgosmann and @hunse I'll merge!

hunse:
Yeah, I like that refactoring. LGTM.

PR #711, opened 2015-04-21
Unsupported backend features attribute

jgosmann:
(Based on PR #706 and #709.)

Adds the `unsupported` class attribute to `Simulator` as a standardized way to define unsupported features and skip those tests automatically.

See #705 for related discussion.

(This PR "conflicts" #710.)

### DIFF
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -3,6 +3,7 @@ import inspect
 import importlib
 import os
 import re
+from fnmatch import fnmatch
 
 import matplotlib
 import numpy as np
@@ -222,13 +223,14 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize("nl_nodirect", nodirect)
 
 
-def pytest_runtest_setup(item):
+def pytest_runtest_setup(item):  # noqa: C901
     rc.reload_rc([])
     rc.set('decoder_cache', 'enabled', 'False')
     rc.set('exceptions', 'simplified', 'False')
 
     if not hasattr(item, 'obj'):
         return
+
     for mark, option, message in [
             ('example', 'noexamples', "examples not requested"),
             ('slow', 'slow', "slow tests not requested")]:
@@ -236,7 +238,6 @@ def pytest_runtest_setup(item):
             pytest.skip(message)
 
     if getattr(item.obj, 'noassertions', None):
-        skip = True
         skipreasons = []
         for fixture_name, option, message in [
                 ('analytics', 'analytics', "analytics not requested"),
@@ -244,11 +245,18 @@ def pytest_runtest_setup(item):
                 ('logger', 'logs', "logs not requested")]:
             if fixture_name in item.fixturenames:
                 if item.config.getvalue(option):
-                    skip = False
+                    break
                 else:
                     skipreasons.append(message)
-        if skip:
+        else:
             pytest.skip(" and ".join(skipreasons))
+
+    if 'Simulator' in item.fixturenames:
+        for test, reason in TestConfig.Simulator.unsupported:
+            # We add a '*' before test to eliminate the surprise of needing
+            # a '*' before the name of a test function.
+            if fnmatch(item.nodeid, '*' + test):
+                pytest.xfail(reason)
 
 
 def pytest_collection_modifyitems(session, config, items):
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -55,6 +55,15 @@ class ProbeDict(Mapping):
 class Simulator(object):
     """Reference simulator for Nengo models."""
 
+    # 'unsupported' defines features unsupported by a simulator.
+    # The format is a list of tuples of the form `(test, reason)` with `test`
+    # being a string with wildcards (*, ?, [abc], [!abc]) matched against Nengo
+    # test paths and names, and `reason` is a string describing why the feature
+    # is not supported by the backend. For example:
+    #     unsupported = [('test_pes*', 'PES rule not implemented')]
+    # would skip all test whose names start with 'test_pes'.
+    unsupported = []
+
     def __init__(self, network, dt=0.001, seed=None, model=None):
         """Initialize the simulator with a network and (optionally) a model.
 


nengo/tests/conftest.py
```diff
@@ -232,6 +233,11 @@ def pytest_runtest_setup(item):
         if skip:
             pytest.skip(" and ".join(skipreasons))
 
+    if 'Simulator' in item.fixturenames:
+        for test, reason in _Simulator.unsupported:
+            if re.match(test, item.name):
```
jgosmann: `item.name` is not necessarily unique. It might be better to use `item.nodeid`, but that adds a dependence on the actual filename because the strings look like `nengo/tests/test_ensemble.py::test_constant_scalar[nl0]` and they get weird for tests in classes: `nengo/tests/test_something.py::TestSomething::()::test_foo`

jgosmann:
And nengo/nengo_ocl#65 shows it in action for a backend.

nengo/tests/options.py
```diff
@@ -0,0 +1,24 @@
+def pytest_addoption(parser):
```
hunse: Why did this move here?

nengo/tests/conftest.py
hunse: We could also write our own function to make a unique name for test functions. I do something like that above in `function_seed`, and it would be great to use the same code both here and there if possible. For this feature, it's especially important that the unique name is easily readable and writeable. I actually quite like `item.nodeid` for function names, though you're right it gets weird for classes (though I'm not sure how much we have to worry about classes, since I don't think we ever use them).

nengo/tests/options.py
jgosmann: To make it importable for other backends without importing `conftest.py`. [See also.](https://github.com/nengo/nengo_ocl/pull/65)

nengo/tests/conftest.py
hunse: Looking at nengo/nengo_ocl#65, I think we should definitely switch to `item.nodeid` before merging this. Otherwise, we might be skipping a lot of tests we don't mean to (there are a number of tests in different files that have the same name, so this is a real issue). We don't use tests in classes at all, AFAIK, so I don't think that we need to be worried about that. If we do decide to use tests in classes down the road, then we can always write a function that takes `item.nodeid` and formats it more nicely. But for now I think `item.nodeid` is fine.

nengo/simulator.py
```diff
@@ -58,6 +58,12 @@ def __len__(self):
 class Simulator(object):
     """Reference simulator for Nengo models."""
 
+    unsupported = [('test_constant_scalar', 'who cares')]
```
hunse: Is this supposed to have content in it? Or is that just an example that we should remove/put in the docstring below?

nengo/conftest.py
```diff
@@ -210,12 +231,17 @@ def pytest_runtest_setup(item):
...
-                    skip = False
+                    break
                 else:
                     skipreasons.append(message)
-        if skip:
+        else:
             pytest.skip(" and ".join(skipreasons))
 
+    if 'Simulator' in item.fixturenames:
+        for test, reason in _Simulator.unsupported:
```
hunse: Is there a better way to do this search? I'm guessing it's not a big performance issue, but we are searching through all the unsupported things for each test. I guess because we want to use regex we have to do it this way? Nothing better is coming to mind for me, but it's something to think about, and maybe fix down the road.

hunse:
Okay, I looked this over, and other than the question about that one unsupported test in the reference simulator, this looks good.

tbekolay:
Note that this is based on #709 so that has to be looked at first.

nengo/simulator.py
jgosmann: That's probably a left over from testing ...

nengo/conftest.py
jgosmann: The answer to this might also depend on how long do we expect the list of unsupported features to be?

But as long as we stick to regexes we probably have to match each regex against each test name.

One minor optimization could be to compile all of the regexes in a first pass. At the moment we compile each regex for each test.

jgosmann: Hm, compiling the regexes beforehand would complicate things a little bit and `re.match` apparently caches a certain number of recent regexes (how big is that number? I have no idea).

So I'm not sure if that's worth it.

hunse: No, probably not. I'm guessing we'd need at least an order of magnitude more tests before this would even be noticeable (I mean even if we did everything here perfectly optimally), so no point sinking time into it.

tbekolay: By default, `re` caches 100 patterns, so for most simulators we're good. This value is modifiable, but it's a private variable (`re._MAXCACHE`; doing `re._MAXCACHE = 1000` raises the limit to 1000 patterns). Could be worth setting it to approximately the number of tests we have? In a `try/except` since it's a private variable.

nengo/simulator.py
```diff
@@ -55,6 +55,12 @@ def __len__(self):
 class Simulator(object):
     """Reference simulator for Nengo models."""
 
+    unsupported = []
+    """Defines features unsupported by a simulator. The format is a list of
```
tbekolay: I recall that this is a convention for documenting variables in `pydoc`, but we don't use it anywhere else, so it feels a little out of place... Should we roll this into the `Simulator` docstring instead? It's missing a lot, so don't worry about making it complete, but just moving this to the class docstring I think would be better as I don't think `Sphinx` reads these.

Also, to render as code, you need two backticks before and after the string in ReST.

tbekolay:
With #709 merged, I rebased this branch to master. It all looks good to me, except the `unsupported` docstring which I commented on inline.

nengo/simulator.py
hunse: I'm not sure if we want this in the simulator docstring, because it's for developers not users, and I think we should keep the docstrings targeted at users. It would also be good to give some examples of how it's used in practice; we could promote Nengo OCL as one of these examples once that stuff is merged (which won't happen until Nengo 2.1.0 is released). So maybe just turn this into a comment for now, and we can worry about better documentation down the road? I could even see a "backend developer" document of some sort that takes developers through all the utilities and things they should know for developing a new backend.

tbekolay: Right, that's a good point. I made https://github.com/nengo/nengo.github.io/issues/7 to track the backend dev doc; for now turning into a comment makes sense.

tbekolay:
Made 856224823602ace30 to make into a comment and add an example. @jgosmann if this looks good to you I'll merge :rainbow:

nengo/simulator.py
```diff
@@ -55,11 +55,12 @@ def __len__(self):
 class Simulator(object):
     """Reference simulator for Nengo models."""
 
+    # 'unsupported' defines features unsupported by a simulator.
+    # The format is a list of tuples of the form `(test, reason)` with `test`
+    # being a regular expression matched against Nengo test names, and `reason`
+    # is a string describing why the feature is not supported by the backend.
+    # E.g., unsupported = [('test_pes*', 'PES rule not implemented')]
```
hunse: I don't think this is a good example. Remember this is regex. So isn't this going to look for "test_pe" with 0 or more esses at the end?

hunse: Yeah, I think wildcard matching would be enough, and easier to write.

jgosmann: +1

tbekolay: One more question about this... we match on the `nodeid`, which includes the path (an example nodeid is `nengo/tests/test_learning_rules.py::test_pes_weight_solver`). This is good, I think, as then we can xfail on files in addition to test names. However, things like `test_pes*` won't work with `fnmatch` because the path comes before `test_pes` in the nodeid.

I can think of a few options in response to this:

1. Keep it as is, give an example nodeid in the comment for `unsupported`, and make the example `*test_pes*` instead of `test_pes*`.
2. Automatically add `*` to the start and end of the test strings in `unsupported` (e.g., `test_pes*` is interpreted the same as `test_pes` which is the same as `*test_pes`).
3. Instead of wildcard matching, just do a substring test (i.e., `test in item.nodeid`). This is what `py.test -k` does (sort of; `-k` allows multiple substrings with boolean operators, so something like 'test_pes or test_unsupervised` would match both).

Any opinions? I'm not sure I have a strong preference between the three choices...

hunse: I like option 1 because it gives the most control, but I'm also used to having to put those stars myself. I don't like adding the star to the end in option 2; I feel like we have a lot of tests like `test_something` and `test_something_else`, and I might only want to match the first one (actually this is one thing I don't like about the way `-k` works on `py.test`). But I don't think there's really any downside to automatically adding a star to the beginning, and it might be a good idea to add it automatically. I'm against option 3 for the same reason as option 2.

So I'd be fine with option 1 or modified option 2 only adding the star to the beginning.

tbekolay:
I went with the modified option 2, adding a `*` at the start; see ba0fc583ffb8cd08df548. I also added a comment about why we add a `*` (and I made sure that having multiple `*`s at the start of the string makes no difference). The example in the comment (which I also updated) works now, so I think this one's all ready. Will merge if ba0fc583ffb8cd is ok with @jgosmann or @hunse.

PR #712, opened 2015-04-22
Prune fixture output directories in pytest

tbekolay:
I had a problem running #706, and it turned out to be because I had logs saved in a `nengo.simulator.logs` directory. Apparently `pytest` has the ability to run doctests from ReST files, and it think that .txts could contain ReST, which is fair. In any case, I added the output directories we make in test running to the `norecursedirs` list in `setup.cfg`, so now running `py.test` in the toplevel Nengo directory works when you have logs saved.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,7 +18,7 @@ max-complexity = 10
 
 [pytest]
 addopts = -p nengo.tests.options
-norecursedirs = .* *.egg build dist docs
+norecursedirs = .* *.egg build dist docs *.analytics *.logs *.plots
 markers =
     example: Mark a test as an example.
     noassertions: Mark a test without assertions. It will only be run if plots or analytics data are produced.


hunse:
Cool

jgosmann:
The logs directory can be changed, so there might be other directories with such files around. It might be better to change the glob pattern by adding `--doctest-glob=''` to `addopts` in `pytest.ini`.

tbekolay:
That'd make sense if we never want to do doctests... which perhaps we don't. Or, we could make it `*.rst`, which is what we store restructured text files in anyhow?

jgosmann:
Yeah, that works too. As long as we don't store doctests in txt files.
Or we could change the log file extensions to something else.

tbekolay:
Very true, `.log` is common for log files I hear.

PR #658, opened 2015-02-23
Product benchmark

jgosmann:
This implements a benchmark for the product network based on [my tech report](http://nbviewer.ipython.org/github/ctn-archive/technical-reports/blob/master/Precise-multiplications-with-the-NEF.ipynb).

It also demonstrates how to write a comparison test (PR #657).

PR #713, opened 2015-04-22
Product benchmark

jgosmann:
PR #658 against `master` this time.

### DIFF
--- a/nengo/networks/tests/test_product.py
+++ b/nengo/networks/tests/test_product.py
@@ -1,8 +1,12 @@
+from __future__ import print_function
+
 import numpy as np
+import pytest
 
 import nengo
 from nengo.utils.compat import range
-from nengo.utils.numpy import rmse
+from nengo.utils.functions import HilbertCurve
+from nengo.utils.numpy import rmse, maxint
 
 
 def test_sine_waves(Simulator, plt, seed):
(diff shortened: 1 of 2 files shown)


nengo/networks/tests/test_product.py
```diff
@@ -35,3 +35,100 @@ def test_sine_waves(Simulator, plt, seed):
     plt.xlim(right=t[-1])
 
     assert rmse(AB[:len(offset), :], sim.data[p][offset, :]) < 0.2
+
+
+class HilbertCurve(object):
```
hunse: Could we want this for other things? Is it worth putting it somewhere else?

jgosmann: I don't know ... do we need a space filling curve somewhere else?

hunse: I'm not sure if we need it right now, but I could see it being useful for other tests. Also, it seems kind of strange or at least surprising having it right in the middle of this test file.

jgosmann: Yeah, any suggestion where to put it? `nengo.utils.weird_things_that_do_not_belong_anywhere_else`? :wink: 
Maybe `nengo.utils.functions`?

hunse: Yeah, it's being used as an input function, so `nengo.utils.functions` seems appropriate.

jgosmann:
Moved and documented `HilbertCurve`, fixed an error while doing it.

hunse:
Why is this marked as "needs changes"?

Seanny123:
@hunse I think I meant to mark it as `needs rebase` and then derped. Sorry.

hunse:
Rebased and squashed this. Also added a commit to remove the convoluted seed generation that got in there somehow. If @jgosmann is good with this, I think it's good to go.

tbekolay:
LGTM!

jgosmann:
Doesn't look good to me. Now all the trials have the same seed, but they should have different seeds.

hunse:
Oh, I see, you wanted something like this.

hunse:
All I changed now is to just get an rng rather than get a seed to make an rng.

jgosmann:
LGTM :) :cake:

jgosmann:
Won't need a changelog entry, will it?

PR #714, opened 2015-04-22
Spa thalamus routing fix

tcstewar:
spa.Actions that caused routing to a module that has multiple inputs didn't work.  For example:

```python
spa.Actions('dot(buffer, A) --> cmp_A=A, cmp_B=B')
```

This was just due to ```spa.get_module``` not handling this case.  This PR fixes it and adds a test to make sure it works.

### DIFF
--- a/nengo/spa/spa.py
+++ b/nengo/spa/spa.py
@@ -88,7 +88,11 @@ class SPA(nengo.Network):
 
     def get_module(self, name):
         """Return the module for the given name."""
-        return self._modules[name]
+        if name in self._modules:
+            return self._modules[name]
+        elif '_' in name:
+            module, name = name.rsplit('_', 1)
+            return self._modules[module]
 
     def get_default_vocab(self, dimensions):
         """Return a Vocabulary with the desired dimensions.
--- a/nengo/spa/tests/test_thalamus.py
+++ b/nengo/spa/tests/test_thalamus.py
@@ -127,6 +127,56 @@ def test_routing(Simulator, seed, plt):
     assert valueC[2] < 0.2
 
 
+def test_nondefault_routing(Simulator, seed, plt):
+    D = 3
+    model = spa.SPA(seed=seed)
+    with model:
+        model.ctrl = spa.Buffer(16, label='ctrl')
+
+        def input_func(t):
+            if t < 0.2:
+                return 'A'
+            elif t < 0.4:
+                return 'B'
+            else:
+                return 'C'
+        model.input = spa.Input(ctrl=input_func)
+
+        model.buff1 = spa.Buffer(D, label='buff1')
+        model.buff2 = spa.Buffer(D, label='buff2')
+        model.cmp = spa.Compare(D)
+
+        node1 = nengo.Node([0, 1, 0])
+        node2 = nengo.Node([0, 0, 1])
+
+        nengo.Connection(node1, model.buff1.state.input)
+        nengo.Connection(node2, model.buff2.state.input)
+
+        actions = spa.Actions('dot(ctrl, A) --> cmp_A=buff1, cmp_B=buff1',
+                              'dot(ctrl, B) --> cmp_A=buff1, cmp_B=buff2',
+                              'dot(ctrl, C) --> cmp_A=buff2, cmp_B=buff2',
+                              )
+        model.bg = spa.BasalGanglia(actions)
+        model.thal = spa.Thalamus(model.bg)
+
+        compare_probe = nengo.Probe(model.cmp.output, synapse=0.03)
+
+    sim = Simulator(model)
+    sim.run(0.6)
+
+    vocab = model.get_output_vocab('cmp')
+    data = sim.data[compare_probe]
+    similarity = np.dot(data, vocab.parse('YES').v)
+
+    valueA = np.mean(similarity[150:200], axis=0)  # should be [1]
+    valueB = np.mean(similarity[350:400], axis=0)  # should be [0]
+    valueC = np.mean(similarity[550:600], axis=0)  # should be [1]
+
+    assert valueA > 0.6
+    assert valueB < 0.3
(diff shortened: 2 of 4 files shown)


xchoo:
Flake8's complaining (unused import). XD

tcstewar:
Oops...  damnit, pep8 doesn't check for that and that's what I've been checking against...  oops..  let's see how this version goes.

nengo/spa/spa.py
```diff
@@ -88,7 +88,11 @@ def __setattr__(self, key, value):
 
     def get_module(self, name):
         """Return the module for the given name."""
-        return self._modules[name]
+        if name in self._modules:
+            return self._modules[name]
+        elif '_' in name:
+            module, name = name.rsplit('_', 1)
```
xchoo: So this means that we can't have inputs and outputs names with an underscore in them, yes? This should be documented somewhere. (If we use an lsplit here, then we can't use module names with underscores.. haha)

nengo/spa/thalamus.py
```diff
@@ -188,6 +189,8 @@ def add_route_effect(self,
             subdim = self.subdim_channel
             if isinstance(target_module, nengo.spa.Buffer):
                 subdim = target_module.state.dimensions_per_ensemble
+            elif isinstance(source_module, nengo.spa.Buffer):
+                subdim = source_module.state.dimensions_per_ensemble
```
xchoo: On a related note, what if it connects to ```spa.Memory```. Should there be logic for that here too?

nengo/spa/tests/test_thalamus.py
```diff
@@ -127,6 +127,56 @@ def input_func(t):
...
+
+        node1 = nengo.Node([0, 1, 0])
+        node2 = nengo.Node([0, 0, 1])
+
+        nengo.Connection(node1, model.buff1.state.input)
+        nengo.Connection(node2, model.buff2.state.input)
+
+        actions = spa.Actions('dot(ctrl, A) --> cmp_A=buff1, cmp_B=buff1',
+                              'dot(ctrl, B) --> cmp_A=buff1, cmp_B=buff2',
+                              'dot(ctrl, C) --> cmp_A=buff2, cmp_B=buff2',
```
xchoo: I'd also like to see a test where the source is non-default option.

nengo/spa/spa.py
tcstewar: Yup, that should be documented somewhere.  In fact, I really don't like the underscore convention here, but nothing else was coming to mind.

nengo/spa/tests/test_thalamus.py
tcstewar: > I'd also like to see a test where the source is non-default option.

I don't think we have any built-in modules with multiple sources....

nengo/spa/thalamus.py
tcstewar: ```spa.Memory``` is a subclass of ```spa.Buffer```.  So no logic needed  :)

nengo/spa/tests/test_thalamus.py
xchoo: I think the '.' operator will work, but I'm not sure how well it will play with the way we are evaluating the action strings.

tcstewar: That does feel like a better option, and it might just be a simple option to add to that evaluation system.  But I haven't looked at that code since @arvoelke rewrote it all.  :)

nengo/spa/spa.py
jgosmann: Would `.` instead of `_` work? It should not occur as part of a module name, should it? And it seems like the natural syntax to me.

tcstewar: Yes, it would work, but it would require adding that functionality to the spa/action_objects system, and that's well outside the scope of this PR.

Note: this PR is about fixing a bug in the spa thalamus system that failed on ```module_whatever``` situations, which is what the rest of the spa system was built with.  It'd be a separate PR to go change everything over to ```module.whatever``` instead of ```module_whatever```.

nengo/spa/tests/test_thalamus.py
xchoo: Whoops. Added the line note to the wrong line! Haha
Yeah. We don't have any modules with non-default sources.... Perhaps we need to make one?

nengo/spa/thalamus.py
xchoo: Ah. Righto. But this does mean that if we are using custom modules, this code won't be able to check for them, yes?

nengo/spa/spa.py
tcstewar: And, just to clarify my ambiguous comment above:

> So this means that we can't have inputs and outputs names with an underscore in them, yes? 

Yes, but that's always been the case.  That's not something new to this PR.

xchoo: Whoopa. See comments in the next block of code (test_thalamus.py). (Accidentally replied in the wrong area)

nengo/spa/thalamus.py
tcstewar: What do you want it to do in those cases?  What it does right now is use the Buffer's geometry for storing information if it can find it, and otherwise it falls back on the parameter its been told to use.

I could see just getting rid of that check and always using the subdim_channel parameter.....  but that'd be a performance change from this PR, and my goal here was just to fix a bug that's breaking Spaun.  :)

tbekolay:
This fix is needed for Spaun, so I gave it a look over. Seems OK? I added a changelog entry (see the FIXUP commit) and fixed up the history a bit; @tcstewar, make sure the changelog entry looks OK. If so, then I'll merge this.

Also, let this comment be a reminder to any who care about this PR, if you want to give it a review, now's the time ;) There was some other discussion about SPA stuff, so if you want to keep that going, make a new issue for it.

PR #717, opened 2015-05-11
Always warn in unit tests in case they run twice.

### LINKED ISSUE #703: Fix unit testing of warnings.
We have some unit tests verifying that certain warnings are issued. However, these do not take into account that a warning might be configured to be only issued once. Under certain circumstances tests might run twice (see issue nengo/nengo_ocl#63) and thus failing in the second run.

[About unit testing warnings.](https://docs.python.org/2/library/warnings.html#testing-warnings)

Tests failing:
* test.nengo.tests.test_ensemble.test_eval_points_number_warning
* test.nengo.networks.tests.test_ensemblearray.test_neuronconnection[nl0]

jgosmann:
Fixes #703.

### DIFF
--- a/nengo/networks/ensemblearray.py
+++ b/nengo/networks/ensemblearray.py
@@ -124,7 +124,7 @@ class EnsembleArray(nengo.Network):
         output = nengo.Node(output=None, size_in=sizes.sum(), label=name)
         setattr(self, name, output)
 
-        indices = np.zeros(len(sizes) + 1)
+        indices = np.zeros(len(sizes) + 1, dtype=int)
         indices[1:] = np.cumsum(sizes)
         for i, e in enumerate(self.ea_ensembles):
             nengo.Connection(
--- a/nengo/params.py
+++ b/nengo/params.py
@@ -3,7 +3,7 @@ import weakref
 
 import numpy as np
 
-from nengo.utils.compat import is_integer, is_number, is_string
+from nengo.utils.compat import is_integer, is_ndarray, is_number, is_string
 from nengo.utils.numpy import compare
 from nengo.utils.stdlib import checked_call
 
@@ -132,7 +132,7 @@ class NumberParam(Parameter):
         super(NumberParam, self).__init__(default, optional, readonly)
 
     def __set__(self, instance, value):
-        if isinstance(value, np.ndarray) and value.shape == ():
+        if is_ndarray(value) and value.shape == ():
             value = value.item()  # convert scalar array to Python object
         super(NumberParam, self).__set__(instance, value)
 
--- a/nengo/utils/compat.py
+++ b/nengo/utils/compat.py
@@ -83,6 +83,11 @@ def is_iterable(obj):
     return isinstance(obj, collections.Iterable)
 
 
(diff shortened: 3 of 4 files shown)


jgosmann:
bump, this might also be something for 2.1.0

tbekolay:
I added a commit to fix some Python 3.4-specific warnings that were being raised on TravisCI (and locally for me). LMK what you think @jgosmann; I'll merge on your :+1:.

hunse:
The order in warning filters matters, so I would append our filter to the end of the filter list, rather than the front (`warnings.simplefilter` does the latter). That way, we're still ignoring deprecation warnings and any other warnings in the default filters.

```python
warnings.filters.append(('always', None, Warning, None, 0))
```

jgosmann:
`simplefilter` has an append argument. Not sure about which filter order we want, though.

hunse:
It probably doesn't really matter. The only thing I'm worried about is if we're counting warnings (which we should probably avoid in unit tests anyway, at least counting warnings of all types), if we let some superfluous warnings through, it could mess up the test. Since the point of this PR is to ensure that warnings can be thrown more than once, I think it's better to do just that and nothing more.

tbekolay:
Personally I think that the more warnings thrown the better... the fixes introduced here mean that we avoid bug reports down the road. I feel like it would be nice to actually log all warnings so that we can look at them, but maybe not have them fail tests as they're doing here. Does that mean we should refactor `WarningCatcher` to be more like `pytest.raises`?

tbekolay:
Added two commits to this PR, one to use `append=True` (which allows the tests to pass even without the Python 3.4 NumPy fixes) and one to make an `is_ndarray` helper function.

hunse:
> Personally I think that the more warnings thrown the better... the fixes introduced here mean that we avoid bug reports down the road.

It should still be possible to set a flag and run all the tests with all warnings, shouldn't it?

> I feel like it would be nice to actually log all warnings so that we can look at them, but maybe not have them fail tests as they're doing here. Does that mean we should refactor `WarningCatcher` to be more like `pytest.raises`?

I wrote `warns` to be like `raises` (and it's actually in `pytest` [now](https://github.com/pytest-dev/pytest/pull/897)), and that's really what we should be using, not counting warnings.

In general, though, I think our tests should look like user code. If we blanket enable all warnings, and then go and test a `DeprecationWarning`, a user might copy that block of code and wonder why they're not also seeing the warning. I'd rather have the filter set in the test itself, so that it's clear to the user that they need the filter to see the warning.

tbekolay:
A quick googling didn't reveal any way of setting a flag to see all warnings, but I'm sure it's possible. Something for later on, perhaps.

As for the `warns` vs. `WarningsCatcher`, I made #802 so we don't forget to use `pytest.warns` once it's released; for now, I think this PR does what it says it will do and we should merge it in if @jgosmann is happy with the proposed changes.

jgosmann:
Any other proposed changes than what is already committed I forgot about? If not, this should be ready to merge.

tbekolay:
I'm good with it; will merge in the morning!

Issue #720, opened 2015-05-20
Config system is slow?


hunse:
nengo_viz is using the config system for storing stuff. I recently noticed that generating the config files can be pretty slow, and it seems the problem is in how we access our config attributes. Specifically, we try to get the attribute and if it's not there, catch the error and do something else. I think it's this error handling that is slowing everything down (see [here](https://github.com/nengo/nengo/blob/master/nengo/config.py#L37)). Is there some way that we could do it without having to do error handling?

tbekolay:
Hm, is this the result of profiling? I'd be surprised if doing `getattr` is the cause of slowing things down, or the `try / except`... but yeah, you can probably do a `hasattr` instead of `try / except` instead. I'd be curious to see the profiling results regardless!

hunse:
I'm trying hasattr and getting weird stack overflows.

tbekolay:
Addressed in #721!

PR #721, opened 2015-05-20
Faster config system using getattr

hunse:
By overloading `__getattr__` instead of `__getattribute__` in the
config system, we avoid throwing and catching AttributeErrors all
the time. Since `__getattr__` is only called if the attribute is
not found normally, it is actually what makes sense in our situation
since we were overloading `__getattribute__` to do exactly this.

Addresses #720.

### DIFF
--- a/nengo/config.py
+++ b/nengo/config.py
@@ -27,9 +27,12 @@ class ClassParams(object):
     """
 
     def __init__(self, configures):
-        self._extraparams = {}
-        self._configures = configures
         assert inspect.isclass(configures)
+        self._configures = configures
+        self._extraparams = {}
+        self._default_params = tuple(
+            attr for attr in dir(self._configures)
+            if is_param(getattr(self._configures, attr)))
 
     def __contains__(self, key):
         return self in self.get_param(key)
@@ -62,7 +65,7 @@ class ClassParams(object):
     def __str__(self):
         name = self._configures.__name__
         lines = ["Parameters configured for %s:" % name]
-        for attr in list(self.default_params) + list(self.extra_params):
+        for attr in self.params:
             if self in self.get_param(attr):
                 lines.append("  %s: %s" % (attr, getattr(self, attr)))
         if len(lines) > 1:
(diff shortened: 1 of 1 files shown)


hunse:
I added a couple more commits. Caching the `default_params` is actually the key, and has made generating the config files way faster. The time to generate a config file in `nengo_viz` for Spaun has gone from ~2.8s to ~0.35s with these commits, with the `__getattr__` change responsible for ~1s, and the `default_params` change responsible for ~1.5s. I don't think the `get_param` change makes a big difference.

tbekolay:
These changes look good to me. Anything else you want to address in this PR @hunse or is it ready to go? Could maybe use a changelog entry, though it's not really a change most users will care about, so not essential.

nengo/config.py
```diff
@@ -139,16 +133,13 @@ def __init__(self, configures, clsparams):
...
-        except AttributeError:
-            if key in self._clsparams.default_params:
-                raise
-            param = self._clsparams.get_param(key)
-            if self in param:
-                return param.__get__(self, self.__class__)
-            return getattr(self._clsparams, key)
+    def __getattr__(self, key):
+        if key in self._clsparams.default_params:
+            raise AttributeError()
```
hunse: This needs a message.

hunse: I'm not totally sure why this is an error, so I'm not totally sure what to put. Before, it just got passed through from the `super(InstanceParams, self).__getattribute__(key)` call, so it would say something like "AttributeError: 'InstanceParams' object has no attribute 'radius'". Is this what we want?

hunse:
I think this is pretty much good to go, without a changelog entry because everything user-facing (and even developer-facing) should work exactly the same. Just that question about the AttributeError message.

nengo/config.py
tbekolay: This is an error because we don't want there to be two separate and non-interacting ways to set an object's attributes. Using radius as an example, if this wasn't an error, then these would both be possible:
```python
ens.radius = 2.0
net.config[ens].radius = 3.0
```
You might think that these lines both change the ensemble's radius, but they do not; the second line won't have any effect. We raise this error so that the second line is not possible.

hunse: But even if I comment out the first line in your example it still doesn't work, because of the AttributeError in `InstanceParams.__setattr__`. I'm not totally clear on how `InstanceParams` is supposed to work.

hunse: Never mind, I looked at the config notebook and I think it makes sense now. Any ideas on a good error message?

hunse:
Okay, took a stab at some better error messages for trying to read or write built-in attributes on InstanceParams.

tbekolay:
These look better than the ones that I was continually rewriting! I would like to remove the word "Just", maybe replace with "Please" ("Just" makes it seem like this is an easy thing, you should know this already, but it's actually a pretty subtle, complex thing).

hunse:
Done!

tbekolay:
Great, I'll merge!

PR #722, opened 2015-05-27
Pass x as copy to nodes.

### LINKED ISSUE #716: Non-deterministic undocumented Node behaviour
For nodes with a ```size_in``` > 0, storing a copy of the input to the function can lead to unexpected behaviour. 

This code illustrates the problem:
```python
import numpy as np

import nengo
from nengo.utils.functions import piecewise


class DirectBuffer(object):
    def __init__(self, dim=1):
        self.x = np.zeros(dim)

    def set_x(self, t, x):
        self.x = x              # Causes a problem
        # self.x = np.array(x)    # Causes no problems

    def get_x(self, t):
        return self.x


    stim = nengo.Node(piecewise({0: 1, 0.5: 0.5, 1: -1}))

    db = DirectBuffer(1)
    n1 = nengo.Node(output=db.set_x, size_in=1)
    n2 = nengo.Node(output=db.get_x)

    nengo.Connection(stim, n1, synapse=0.005)

    p1 = nengo.Probe(stim)
    p2 = nengo.Probe(n2)

for r in range(10):
    sim = nengo.Simulator(model)
    sim.run(1.5)

    plt.subplot(2, 1, 1)
    plt.plot(sim.trange(), sim.data[p1])
    plt.ylim([-1.05, 1.05])
    plt.title('stim')
    plt.subplot(2, 1, 2)
    plt.plot(sim.trange(), sim.data[p2])
    plt.ylim([-1.05, 1.05])
    plt.title('out')
    plt.show()
```

In the ideal case, the output of ```n2``` matches the signal generated by ```stim```. However, in a majority of the runs, the output of ```n2``` is 0.

jgosmann:
Closes #716.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,10 @@ Release History
 - The ``probeable`` attribute of all Nengo objects is now implemented
   as a property, rather than a configurable parameter.
   (`#671 <https://github.com/nengo/nengo/pull/671>`_)
+- Node functions receive ``x`` as a copied NumPy array (instead of a readonly
+  view).
+  (`#716 <https://github.com/nengo/nengo/issues/716>`_,
+  `#722 <https://github.com/nengo/nengo/pull/722>`_)
 
 **Improvements**
 
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -334,13 +334,8 @@ class SimPyFunc(Operator):
         t_in = self.t_in
         t_sig = signals['__time__']
 
-        args = []
-        if self.x is not None:
-            x_sig = signals[self.x].view()
-            x_sig.flags.writeable = False
-            args += [x_sig]
-
         def step():
+            args = () if self.x is None else (np.copy(signals[self.x]),)
             y = fn(t_sig.item(), *args) if t_in else fn(*args)
             if output is not None:
                 if y is None:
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -40,6 +40,9 @@ Release History
 - The ``probeable`` attribute of all Nengo objects is now implemented
   as a property, rather than a configurable parameter.
   (`#671 <https://github.com/nengo/nengo/pull/671>`_)
+- Node functions receive ``x`` as a copied NumPy array (instead of a readonly
+  view).
+  (`#626 <https://github.com/nengo/nengo/issues/716>`_)
```
tbekolay: Is it 626 or 716...? Presumably 722 should be in there too

nengo/builder/operator.py
```diff
@@ -334,13 +334,9 @@ def make_step(self, signals, dt, rng):
         t_in = self.t_in
         t_sig = signals['__time__']
```
tbekolay: This extra blank line is giving a flake8 error now

tbekolay:
Aside from the inline comments, this looks good to me!

jgosmann:
Fixed those. :cake:

tbekolay:
This LGTM! Does anyone have any objections to passing x as a copy? The discussion in #716 seemed to favour copying, so if no one objects by lunch time or thereabouts, I'll merge! :rainbow:

PR #724, opened 2015-05-28
lsmr solver test requires scipy>=0.11

arvoelke:
These tests were broken on my machine since my scipy is a couple years out of date. Fix proposed by @hunse and verified to skip on my machine.


jgosmann:
LGTM

PR #726, opened 2015-05-29
Connection seed was not being set

### LINKED ISSUE #725: Connection seed not being set in __init__
If I set a seed when initializing a connection (ie conn = nengo.Connection(a,b,seed=5)) the seed is still left as default None. I think the line "self.seed = seed" is missing in connection.py

hunse:
Fixes #725.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -70,6 +70,8 @@ Release History
 
 **Bug fixes**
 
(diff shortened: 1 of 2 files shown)


tbekolay:
Whoops! That was silly of us. Merging!

PR #727, opened 2015-06-02
Vector Oja unsupervised encoder learning rule

arvoelke:
This is a rewritten implementation based on code from the `voja` branch, commit 0e9ddabe4f0f8a084918d1e0e87f88981260f591. 

The above branch/rule appeared in [A spiking neural model of episodic memory encoding and replay in hippocampus](http://compneuro.uwaterloo.ca/publications/trujillo2014a.html), and exists here without any significant changes.

One difference is this now uses the unmodulated learning syntax. Another difference is the implementation is done with pure signals and ops. Lastly, it's invariant under `dt` (so previously used learning rates need to be adjusted).

The PR includes some unit tests and an example notebook to demonstrate use and correctness. Currently waiting on #642 (`learning-unmodulated` branch).

### DIFF
--- a/examples/learning/learn_associations.ipynb
+++ b/examples/learning/learn_associations.ipynb
@@ -1,0 +1,330 @@
+{
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 2",
+   "language": "python",
+   "name": "python2"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 2
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython2",
+   "version": "2.7.3"
+  },
+  "name": ""
+ },
+ "nbformat": 3,
+ "nbformat_minor": 0,
+ "worksheets": [
+  {
+   "cells": [
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "# Nengo Example: Learning new associations\n",
+      "\n",
+      "Being able to learn an input-output mapping (or a _heteroassociative memory_) is useful for storing and recalling associations. This is also a task required by more complicated models that require some notion of long-term memory.\n",
+      "\n",
+      "In a perfect world, the PES rule could be applied to learn this mapping from examples. However, when two distinct inputs cause the same neurons to fire, their decoded values will depend on one another. This leads to difficulty when trying to store multiple indpendent associations in the same memory.\n",
+      "\n",
+      "To solve this problem, a vector-space analog of Oja's rule, dubbed Vector-Oja's rule (or simply _Voja's rule_) was proposed. In essence, this unsupervised learning rule makes neurons fire selectively in response to their input. When used in conjunction with properly-chosen intercepts (corresponding to the largest dot-product between pairs of inputs), this approach makes it possible to scalably learn new associations in a spiking network.\n",
+      "\n",
+      "Voja's rule works by moving the encoders of the active neurons toward the current input. This can be stated succinctly as,\n",
+      "\n",
+      "$$\n",
+      "\\Delta e_i = \\kappa a_i (x - e_i)\n",
+      "$$\n",
+      "\n",
+      "where $e_i$ is the encoder of the $i^{th}$ neuron, $\\kappa$ is a modulatory learning rate (positive to move towards, and negative to move away), $a_i$ is the filtered activity of the $i^{th}$ neuron, and $x$ is the input vector encoded by each neuron. To see how this is related to Oja's rule, substituting $e_i$ with the row of weights $W_i$, $x$ for the pre-synaptic activity vector $b$, and letting $s = 1 / a_i$ be a dynamic normalizing factor, gives\n",
+      "\n",
+      "$$\n",
+      "\\Delta W_i = \\kappa a_i (b - s a_i W_i)\n",
+      "$$\n",
+      "\n",
+      "which is the update rule for a single row using Oja. For more details, see [Learning large-scale heteroassociative memories in spiking neurons](http://compneuro.uwaterloo.ca/publications/voelker2014a.html).\n",
+      "\n",
+      "This notebook will lead the reader through a basic example of building a network that can store and recall new associations."
+     ]
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "## Step 1: Configure some example data\n",
+      "\n",
+      "First, we will setup some keys (inputs) and values (outputs) for our network to store and recall."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "import numpy as np\n",
+      "import matplotlib.pyplot as plt\n",
+      "%matplotlib inline\n",
+      "\n",
+      "import nengo"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": null
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "num_items = 5\n",
+      "\n",
+      "d_key = 2\n",
+      "d_value = 4\n",
+      "\n",
+      "rng = np.random.RandomState(seed=7)\n",
+      "keys = nengo.dists.UniformHypersphere(surface=True).sample(num_items, d_key, rng=rng)\n",
+      "values = nengo.dists.UniformHypersphere(surface=False).sample(num_items, d_value, rng=rng)"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": null
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "An important quantity is the largest dot-product between all pairs of keys, since a neuron's intercept should not go below this value if it's positioned between these two keys. Otherwise, the neuron will move back and forth between encoding those two inputs."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "intercept = (np.dot(keys, keys.T) - np.eye(num_items)).flatten().max()\n",
+      "print(\"Intercept: %s\" % intercept)"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": null
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "## Step 2: Build the model\n",
+      "\n",
+      "We define a helper function that is useful for creating nodes that cycle through keys/values."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "def cycle_array(x, period, dt=0.001):\n",
+      "    \"\"\"Cycles through the elements\"\"\"\n",
+      "    i_every = int(round(period/dt))\n",
+      "    if i_every != period/dt:\n",
+      "        raise ValueError(\"dt (%s) does not divide period (%s)\" % (dt, period))\n",
+      "    def f(t):\n",
+      "        i = int(round((t - dt)/dt))  # t starts at dt\n",
+      "        return x[(i/i_every)%len(x)]\n",
+      "    return f"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": null
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "We create three inputs: the keys, the values, and a modulatory learning signal. The model is run continuously in two phases: the first half learns the set of associations, and the second tests recall.\n",
+      "\n",
+      "The learning signal will be set to 0 to allow learning during the first phase, and -1 to inhibit learning during the second phase.\n",
+      "\n",
+      "The memory is confined to a single ensemble. Roughly speaking, its encoders will hold the keys, and its decoders will hold the values. "
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "# Model constants\n",
+      "n_neurons = 200\n",
+      "dt = 0.001\n",
+      "period = 0.3\n",
+      "T = period*num_items*2\n",
+      "\n",
+      "# Model network\n",
+      "model = nengo.Network()\n",
+      "with model:\n",
+      "    \n",
+      "    # Create the inputs/outputs\n",
+      "    stim_keys = nengo.Node(output=cycle_array(keys, period, dt))\n",
+      "    stim_values = nengo.Node(output=cycle_array(values, period, dt))\n",
+      "    learning = nengo.Node(output=lambda t: -int(t>=T/2))\n",
+      "    recall = nengo.Node(size_in=d_value)\n",
+      "    \n",
+      "    # Create the memory\n",
+      "    memory = nengo.Ensemble(n_neurons, d_key, intercepts=[intercept]*n_neurons)\n",
+      "    \n",
+      "    # Learn the encoders/keys\n",
+      "    voja = nengo.Voja(post_tau=None, learning_rate=5e-2)\n",
+      "    conn_in = nengo.Connection(stim_keys, memory, synapse=None,\n",
+      "                               learning_rule_type=voja)\n",
+      "    nengo.Connection(learning, conn_in.learning_rule, synapse=None)\n",
+      "    \n",
+      "    # Learn the decoders/values, initialized to a null function\n",
+      "    conn_out = nengo.Connection(memory, recall, learning_rule_type=nengo.PES(1e-3),\n",
+      "                                function=lambda x: np.zeros(d_value))\n",
+      "    \n",
+      "    # Create the error population\n",
+      "    error = nengo.Ensemble(n_neurons, d_value)\n",
+      "    nengo.Connection(learning, error.neurons, transform=[[10.0]]*n_neurons,\n",
+      "                     synapse=None)\n",
+      "    \n",
+      "    # Calculate the error and use it to drive the PES rule\n",
+      "    nengo.Connection(stim_values, error, transform=-1, synapse=None)\n",
+      "    nengo.Connection(recall, error, synapse=None)\n",
+      "    nengo.Connection(error, conn_out.learning_rule)\n",
+      "    \n",
+      "    # Setup probes\n",
+      "    p_keys = nengo.Probe(stim_keys, synapse=None)\n",
+      "    p_values = nengo.Probe(stim_values, synapse=None)\n",
+      "    p_learning = nengo.Probe(learning, synapse=None)\n",
+      "    p_error = nengo.Probe(error, synapse=0.005)\n",
+      "    p_recall = nengo.Probe(recall, synapse=None)\n",
+      "    p_encoders = nengo.Probe(conn_in.learning_rule, 'scaled_encoders')"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": null
+    },
+    {
+     "cell_type": "heading",
+     "level": 2,
+     "metadata": {},
+     "source": [
+      "Step 3: Running the model"
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "sim = nengo.Simulator(model, dt=dt)\n",
+      "sim.run(T)\n",
+      "t = sim.trange()"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": null
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "## Step 4: Plotting simulation output\n",
+      "\n",
+      "We first start by checking the keys, values, and learning signals."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "plt.figure()\n",
+      "plt.title(\"Keys\")\n",
+      "plt.plot(t, sim.data[p_keys])\n",
+      "plt.ylim(-1, 1)\n",
+      "\n",
+      "plt.figure()\n",
+      "plt.title(\"Values\")\n",
+      "plt.plot(t, sim.data[p_values])\n",
+      "plt.ylim(-1, 1)\n",
+      "\n",
+      "plt.figure()\n",
+      "plt.title(\"Learning\")\n",
+      "plt.plot(t, sim.data[p_learning])\n",
+      "plt.ylim(-1.2, 0.2);"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": null
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "Next, we look at the error during training and testing. In the top figure, the error being minimized by PES goes to zero for each association during the training phase. In the bottom figure, the recall error is close to zero, with momentary spikes each time a new key is presented."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "train = t<=T/2\n",
+      "test = ~train\n",
+      "\n",
+      "plt.figure()\n",
+      "plt.title(\"Value Error During Training\")\n",
+      "plt.plot(t[train], sim.data[p_error][train])\n",
+      "\n",
+      "plt.figure()\n",
+      "plt.title(\"Value Error During Recall\")\n",
+      "plt.plot(t[test], sim.data[p_recall][test] - sim.data[p_values][test]);"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": null
+    },
+    {
+     "cell_type": "markdown",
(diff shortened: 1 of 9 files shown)


arvoelke:
@neworderofjamie For your reference. The notebook `examples/learning/learn_associations.ipynb` should be the best starting point for seeing the Voja/PES combination in action. The `SimVoja` class from 0e9ddab gives a compact implementation of the rule. This PR implemented the rule using signals/ops, and so it may be more difficult to use as reference, but both use the same math.

neworderofjamie:
Thank you for this. Sorry for the slow response, I'm out of the office until Friday when I'll look at this in more detail and maybe even start implementing!

neworderofjamie:
I'm finally getting into implementing this and I've got one question, is it y or y^2!? The two references you point at say:

In the builder version:

learning * (outer(y, x) - y*e)

and , in the SimVoja version:

outer(y, x) - s y^2 e

arvoelke:
In that SimVoja comment, `s` is `1/y`, so it cancels a `y` to give you the same thing. This is just to show the relationship to Oja. The first thing is what you want to follow, or the notepad description which is in terms of individual neurons and perhaps easier to conceptualise.

neworderofjamie:
Yes, both easier to conceptualise and easier to write in C :) Thank you

neworderofjamie:
Aaron, making slow onward progress - Just one question:

In your example, is this meant to be a inhibitory connection?

```
nengo.Connection(learning, error.neurons, transform=[[10.0]]*n_neurons,
                     synapse=None)
```

It seems to be currently causing weirdness for Nengo SpiNNaker - @mundya, learning is a Node.

arvoelke:
Yep. The idea is that a learning signal of `0` is the default learning, and `-1` is used to inhibit both Voja and PES. This basically injects large amounts of negative current into all of the `error` neurons. The value of 10 is pretty arbitrary.

neworderofjamie:
Ok, cool -  It seems like the node->inhibition pathway doesn't get properly mapped to SpiNNaker

mundya:
> Ok, cool - It seems like the node->inhibition pathway doesn't get properly mapped to SpiNNaker

That should work, but I'll have a check as to what's happening.

arvoelke:
Try removing the intercepts argument from the error population? It's not actually necessary (my comment there is wrong).

neworderofjamie:
One more question - Should the ability to specify no post activity filtering with post_tau=None also exist for PES (which has, after all been proven to work without filtering)

tbekolay:
We could / should do that; we should also accept synapse types rather than numbers. But for now I think you can achieve this by setting `post_tau=0.0`.

neworderofjamie:
Yeah, that makes sense - Not filtering offers a (possibly significant)
performance benefit on SpiNNaker so I'll treat tau=0 and tau=None as
special cases for both rules.

tbekolay:
Not sure if you have an explicit `dt`, but if you do then you can also check for `tau < dt`, which is a bit more flexible :) We do this in the reference builder.

neworderofjamie:
Excellent idea!

arvoelke:
Yep to all the above. We check for `tau < .03 * dt` in [nengo/synapses.py](https://github.com/nengo/nengo/blob/5ece98d1e19ca04996cc8f4ec091f6eea9274fc1/nengo/synapses.py#L110). The `Connection` class automatically casts numerical `tau` to this `Lowpass` object, which will then build a delta filter if that check passes at build time. Some time down the road, it would probably be good to support the parent `LinearFilter` class (I'm guessing supporting the custom `Synapse` would not be feasible).

neworderofjamie:
@arvoelke - am I right in thinking that, in your example, the learning signal is 0 when Voja should be learning in your example and -1 when it shouldn't? From the build_voja code, I'm struggling to see how  that get combined with the constant learning rate parameter to modulate learning..

arvoelke:
That is right. The relevant lines are basically here:

    learning = Signal(np.zeros(rule.size_in), name="Voja:learning")
    assert rule.size_in == 1
    model.add_op(Reset(learning, value=1.0))
    model.sig[rule]['in'] = learning  # optional connection will attach here

The last line is specific to the way we are doing modulatory connections now. This tells the builder where the connection to the `learning_rule` should go. This connected value (`learning`) is added to a default value of one. A few lines down this gets scaled into `scaled_learning`, which then scales the delta.

arvoelke:
Blegh. I'm finding it impossible to cherry-pick 34040c3bbb955a28507b95a68d5dee7398f02bb2 into master now. 
 - `validate_rule` assumes the `pre_obj` always matters. But I need `modifies = ['encoders']` to validate iff `isinstance(conn.post_obj, Ensemble)`. The simplest way around this is to add some custom logic to the top of `validate_rule` and short-circuit the function, but this makes the semantics of `modifies` quite muddy.
 - `build_learning_rule` assumes you can infer what the delta is going to modify just by looking at `conn.solver`, `pre_obj`, and `post_obj`. But I need the delta to modify the encoders, and knowing this at this point in the build depends only on the fact that `modifies = ['encoders']` or `isinstance(rule_type, Voja)`. Also, `get_pre_ens` assumes the `pre_obj` is an `Ensemble`. Again, the simplest way around this would be to use separate logic at the very top and short-circuit the function.
 - `build_learning_rule` also assumes that `delta` does not depend on the signal being modified. But in this case the `delta` applied to the encoders depends on the current encoders (i.e. there's a cycle). The previous solution was to use `as_update=True` in `ElementwiseInc`, but that no longer exists.

The first few problems can be uncomfortably hacked away, but the last point is especially debilitating. Not sure where to go with this. It seems a bunch of things in these two functions need to be reworked if we want to take encoder learning seriously. We may also need to bring back `as_update`?

tbekolay:
> I need modifies = ['encoders'] to validate iff isinstance(conn.post_obj, Ensemble)

Why not just have the learning rule modify the pre encoders instead of the post encoders? Or does that not match what the learning rule requires? Either way, I would think that the `modifies` should be explicit; either `pre.encoders` or `post.encoders`. I think having that would also address your second point.

> build_learning_rule also assumes that delta does not depend on the signal being modified

That's not true; look at the `Oja` rule, which uses `transform` in the `step` function. The reason why this works and doesn't cause a cycle is because the `delta` is calculated based on the *current* value of transform, but doesn't actually get used to update the transform until the start of the next timestep. So, you update the transform based on the previous timestep's deltas, and then calculate new deltas.

What you will have to do is use the `PreserveValue` op for the `encoders`, just like we do for the `transform`. This allows us to carry on the value from the previous timestep, and then increment in the correct order.

arvoelke:
> Why not just have the learning rule modify the pre encoders instead of the post encoders? Or does that not match what the learning rule requires? 

In theory, yes, but wouldn't this be really strange? You'd be attaching a learning rule to a connection, but that learning rule wouldn't modify anything to do with that connection. It would be modifying (all of) the connections that go into that connection's pre. So if I had,

    x = nengo.Ensemble(100, 1)
    conn_a = nengo.Connection(a, x)
    conn_b = nengo.Connection(b, x)
    nengo.Connection(x, output, learning_rule_type=nengo.Voja())

Then you'd be modifying the encoders on `conn_a` and `conn_b`. It seems much more intuitive to say,

    x = nengo.Ensemble(100, 1)
    conn_a = nengo.Connection(a, x, learning_rule_type=nengo.Voja())
    nengo.Connection(x, output)

To modify the encoders of `conn_a` into `x`.

> either `pre.encoders` or `post.encoders`. I think having that would also address your second point.

I don't think it should ever be `pre.encoders` for the above reason. And if `encoders` always refers to the post, this doesn't help either of the first two points. I think the proper way to go is to endow clearer semantics on `modifies`. I can give a clearer proposal.

> What you will have to do is use the `PreserveValue` op for the `encoders`

Ah, okay. This thing is new to me. I'll take a look to see how that works.

arvoelke:
Bump. The first two points still need discussion. I am happy to give a proposed rework of `modifies`.

tbekolay:
Sure, what's your proposal for `modifies`?

arvoelke:
Suggestions are actually blocked on #729, which looks like it would simplify things considerably.

neworderofjamie:
Bump! What's the current status of this stuff? #729 looks to have been merged, but the loss of pre and post-modifies has left the SpiNNaker implementations of PES and Voja somewhat broken. For our purposes, it would be a simple change to do the necessary re-routing of error signals for encoder and decoder learning rules based around `modifies = ['encoders']`  and `modifies = ['decoders']` rather than  `pre_modifies` and `post_modifies`, but is this stuff is going to change again in the near future, I'll probably hold off.

hunse:
@arvoelke: Can you rebase this? Is there an obvious answer to the `modifies` question now?

arvoelke:
Sorry for the delay. I will look into reworking `modifies` this week.

arvoelke:
I managed to get over a few of the hurdles, but point (3) from above is still causing me issues.

The delta is applied to the encoders, while the delta also depends on the encoders, which causes a cycle. The encoders are marked with `PreserveValue`, but I still can't figure out how to make this work (previously I just used `as_update=True`). If anyone wants to take a look, the broken code is here: 84ad9b4a82bef6f105f45c26bb6b72abb8452061

arvoelke:
This should now be ready for review again. The original commit for this PR is now gone, but can be found here: https://github.com/nengo/nengo/commit/34040c3bbb955a28507b95a68d5dee7398f02bb2

@hunse took a brief look, and the working theory was that the cycle only works for PES because of the filtering done within the cycle. Direct cycles with basic ops and without any filtering still seem to be an issue, even with `PreserveValue`. This indeed seems to be the case, because adding a filter to `delta` resolved the issue and made all the tests pass.

Rather than try to sort out this problem, it was pointed out that there is no real advantage to implementing this with the basic set of ops. So, I just went back to using the original `SimVoja` code from the old branch. This simplifies implementation and makes the cycle a non-issue (for the same reason it's okay for the other learning rules).

hunse:
I added a commit that makes VOja more like the other learning rules, updating via a delta. I think the delta is important if we want to have e.g. multiple encoder learning rules acting on the same set of encoders.

nengo/builder/connection.py
```diff
@@ -190,9 +190,6 @@ def get_prepost_signal(is_pre):
         gain = model.params[conn.post_obj.ensemble].gain[post_slice]
         weights = multiply(gain, weights)
 
-    if conn.learning_rule is not None and weights.ndim < 2:
-        raise ValueError("Learning connection must have full transform matrix")
```
hunse: We should still have this check, I think, right?

arvoelke: I was meaning to ask you about that. The voja unit tests fail with this check still in, and I didn't look close enough to figure out why.

hunse: Well you only want this check if the learning is modifying the weights, which used to be assumed but is not necessarily the case now.

nengo/builder/learning_rules.py
```diff
@@ -83,6 +83,80 @@ def step_simoja():
         return step_simoja
 
 
+class SimVoja(Operator):
+    """Simulates a simplified version of Oja's rule in the vector space.
```
hunse: We typically don't have detailed docstrings like this for operators, since users aren't likely to see them. I'm wondering if there's a better place for some of this stuff, especially the little derivation. The notebook, perhaps?

nengo/builder/connection.py
arvoelke: Is it needed for the other learning rules considering the checks done to validate the LearningRuleType param?

hunse: Maybe not

nengo/learning_rules.py
```diff
@@ -197,6 +201,42 @@ def _argreprs(self):
...
+    Modifies an ensemble's encoders to be selective to its inputs.
+
+    A connection to the learning rule will provide a scalar weight for the
+    learning rate (minus 1). For instance, 0 is normal learning, -1 is no
+    learning, and less than -1 causes anti-learning or "forgetting".
+
+    Parameters
+    ----------
+    learning_rate : float, optional
+        A scalar indicating the rate at which decoders will be adjusted.
```
hunse: "decoders" -> "encoders"

nengo/learning_rules.py
```diff
@@ -197,6 +201,42 @@ def _argreprs(self):
...
+
+    Modifies an ensemble's encoders to be selective to its inputs.
+
+    A connection to the learning rule will provide a scalar weight for the
+    learning rate (minus 1). For instance, 0 is normal learning, -1 is no
+    learning, and less than -1 causes anti-learning or "forgetting".
+
+    Parameters
+    ----------
+    learning_rate : float, optional
```
hunse: I prefer the order to be the same as in the constructor.

hunse:
Other than my few comments relating to some docstrings, I'm happy with this now. I think it makes it easy to do encoder learning in the future, too, which will be handy.

nengo/learning_rules.py
arvoelke: Looking elsewhere in the repo, I can't find a consistent rule for the ordering. Sometimes it's neither alphabetical nor order of appearance. Perhaps we should pick a convention and then do a giant sweep over in another PR.

hunse: Yeah, I just glanced at Connection for reference, and some of its parameters listed in the docstring don't even exist anymore. Should definitely clean this up before release.

hunse: Added #851.

arvoelke:
I like all the changes you made. Looked over it and I don't have any comments. Thanks for the help! Resolved the docstring comments.

nengo/learning_rules.py
arvoelke: Done

nengo/builder/learning_rules.py
arvoelke: Moved to notebook

nengo/builder/connection.py
arvoelke: Resolution appears to be that this was moved into a condition for when the learning rule modifies the weights.

arvoelke:
@neworderofjamie - I should also check to see these changes aren't too inconvenient for you? To summarize, `modifies` is now one of "weights", "decoders", or "encoders" (instead of a list). I don't think you need to worry about weights (it is for BCM/Oja). The decoders option is only used by PES, which means it modifies either the decoders or the full weight matrix (the latter if the connection is between two sets of neurons or the solver is in weight mode). The encoders option is only used by Voja, which means it modifies the encoders of the post-synaptic ensemble. Basically, I think this just means you need to route either the encoders or the decoders for the respective value of modifies.

neworderofjamie:
That sounds ideal - If anything easier to handle than the list format.

studywolf:
bump

neworderofjamie:
Pleeeeease :)

hunse:
@arvoelke: Is this good to go? If so, I can get it rebased and merged.

arvoelke:
Yes, please. I did review your changes, but, we should probably have a third person look over the whole thing (since we both wrote code for the PR)? @tbekolay @jgosmann?

tbekolay:
I can look it over!

examples/learning/learn_associations.ipynb
```diff
@@ -0,0 +1,330 @@
+{
+ "metadata": {
+  "kernelspec": {
```
tbekolay: What version of IPython did you use to create this notebook? Our [existing notebooks](https://raw.githubusercontent.com/nengo/nengo/master/examples/basic/2d_representation.ipynb) don't have this metadata in them. Ideally, all of the notebooks would be created with the same version of IPython, which I think for the others is `IPython<3` (2.4.1 now I think).

examples/learning/learn_associations.ipynb
```diff
@@ -0,0 +1,330 @@
...
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "num_items = 5\n",
+      "\n",
+      "d_key = 2\n",
+      "d_value = 4\n",
+      "\n",
+      "np.random.seed(7)\n",
```
tbekolay: Making a [`RandomState`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.random.RandomState.html) instance is a bit safer than mucking with numpy's global random state, I think... at least, that's what we do to generate `rng` in the test suite.

examples/learning/learn_associations.ipynb
```diff
@@ -0,0 +1,330 @@
...
+     "source": [
+      "An important quantity is the largest dot-product between all pairs of keys, since a neuron's intercept should not go below this value if it's positioned between these two keys. Otherwise, the neuron will move back and forth between encoding those two inputs."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "intercept = (np.dot(keys, keys.T) - np.eye(num_items)).flatten().max()\n",
+      "print (\"Intercept: %s\" % intercept)"
```
tbekolay: Minor nitpick: no space between `print` and `(` (should look like a function for Py3).

examples/learning/learn_associations.ipynb
```diff
@@ -0,0 +1,330 @@
...
+      "    voja = nengo.Voja(post_tau=None, learning_rate=5e-2)\n",
+      "    conn_in = nengo.Connection(stim_keys, memory, synapse=None,\n",
+      "                               learning_rule_type=voja)\n",
+      "    nengo.Connection(learning, conn_in.learning_rule, synapse=None)\n",
+      "    \n",
+      "    # Learn the decoders/values, initialized to a null function\n",
+      "    conn_out = nengo.Connection(memory, recall, learning_rule_type=nengo.PES(1e-3),\n",
+      "                                function=lambda x: np.zeros(d_value))\n",
+      "    \n",
+      "    # Give the error connection positive intercepts so that it can be inhibited\n",
```
tbekolay: A minor pedantic point perhaps, but I don't believe that you need to get positive intercepts if you're using  inhibition through direct connections to neurons. Negative intercepts only affect your gain and bias, which is used to map input to currents. When you connect directly to neurons, you're bypassing the encoding process and just providing current to your neurons. All neuron types respond more for increasing current (up to a certain point for neurons that can futz out like `Izhikevich`), so I don't think setting intercepts makes a difference when directly inhibiting neurons.

nengo/builder/connection.py
```diff
@@ -213,14 +210,19 @@ def get_prepost_signal(is_pre):
...
+        targets = []
+        for r in itervalues(rule) if isinstance(rule, dict) else rule:
+            model.build(r)
+            targets.append(r.modifies)
+
+        if 'encoders' in targets:
+            model.add_op(PreserveValue(model.sig[conn.post_obj]['encoders']))
+        if 'decoders' in targets or 'weights' in targets:
+            if weights.ndim < 2:
+                raise ValueError("'transform' must be a matrix for learning")
```
tbekolay: Another minor nitpick: should we say "2-dimensional array" instead of matrix to avoid people accidentally giving us an `np.matrix` instance?

examples/learning/learn_associations.ipynb
hunse: Gain and bias still get applied in neuron-neuron connections. Since you're bypassing the encoders, you're right that the tuning curve will be generally monotonically increasing, but if you have a large negative intercept (positive bias), then it might take a very large negative current to actually shut that neuron off.

nengo/builder/connection.py
hunse: That's a good idea. What would happen if people give matrices? Maybe something to look into down the road. I wish that Numpy didn't have matrices.

examples/learning/learn_associations.ipynb
tbekolay: Right, I forgot about that! Though in this case we're using a weight of `10`, so it will definitely be below any intercepts, so I still think it's a bit confusing / unnecessary to set the intercepts.

nengo/builder/learning_rules.py
```diff
@@ -96,29 +152,36 @@ def get_post_ens(conn):
...
-        tag = "omega += delta"
-    elif isinstance(conn.pre_obj, Neurons):
-        delta = Signal(np.zeros((rule.size_in, pre.n_neurons)), name='Delta')
-        tag = "omega += delta"
+
+    # --- Set up delta signal
+    if rule.modifies == 'encoders':
+        if not conn.is_factored:
+            ValueError(
+                "Cannot perform encoder learning on an unfactored connection")
```
tbekolay: By unfactored do you mean a neuron-to-neuron connection? I think most users will not know what you mean by unfactored.

nengo/connection.py
```diff
@@ -36,24 +36,35 @@ def __set__(self, conn, rule):
...
+                    % (conn.post_obj.__class__.__name__, rule))
 
-        if not isinstance(conn.post_obj, (Ensemble, Neurons, Node)):
-            raise ValueError("'post' must be of type 'Ensemble', 'Neurons' "
-                             "or 'Node' for learning rule '%s' (got type '%s')"
-                             % (rule, conn.post_obj.__class__.__name__))
+        # If the rule modifies 'weights', then it must have full weights
+        if rule.modifies == 'weights' and conn.is_factored:
+            raise ValueError(
+                "Learning rule '%s' can not be applied to factored "
```
tbekolay: Again I think 'factored' is a confusing term. I think in the past we've used "decoded" connection versus "direct" or "neuron to neuron" connections, but I agree that those aren't super clear either...

nengo/tests/test_learning_rules.py
```diff
@@ -311,3 +311,91 @@ def check_rule(rule, conn, rule_type):
...
+    """Tests that voja changes active encoders to the input."""
+    n = 200
+    learned_vector = np.asarray([0.3, -0.4, 0.6])
+    learned_vector /= np.linalg.norm(learned_vector)
+    n_change = n // 2  # modify first half of the encoders
+
+    # Set the first half to always fire with random encoders, and the
+    # remainder to never fire due to their encoder's dot product with the input
+    intercepts = np.asarray([-1]*n_change + [0.99]*(n - n_change))
+    np.random.seed(seed)  # for UniformHyersphere
```
tbekolay: Here also, use the `rng` fixture rather than messing with global state.

tbekolay:
This is looking great! I made a few inline nitpicky comments.

One thing that I'd be curious for other people's input on is using the terms "factored" and "unfactored" for the connections between ensembles. To me, I think it's not at the right abstraction level for a typical user, as it requires you to have a pretty good grasp of Nengo internals / the NEF.

Possible alternatives: decoded vs. neuron-to-neuron / synaptic

I guess technically we only need one term because you can check `conn.is_decoded` or `not conn.is_decoded`.

hunse:
`is_decoded` works. We can't really say anything about what is connected to what, for example you can use a weight solver between ensembles and have a non-decoded ensemble-ensemble connection.

tbekolay:
Guess it depends on your mental model... I would still consider a non-decoded ensemble-ensemble connection a neuron-to-neuron connection. Semantics!

hunse:
I would too, but again the user needs some idea of how things are working
internally to know that.

neworderofjamie:
Hello, what's the ETA for getting this merged guys?

tbekolay:
I was hoping for more comments on the `factored` vs. `decoded` or some other nomenclature for differentiating the two types of connections. The rest of the changes are pretty minor nitpicks, so I can do them in the merge if @arvoelke is too busy with other stuff.

arvoelke:
> Hello, what's the ETA for getting this merged guys?

Sorry Jamie, this was my bad for not checking back on the review. Been a little swamped lately, but I will give it a shot Monday. Feel free to email me if you have questions/concerns.

examples/learning/learn_associations.ipynb
arvoelke: I used `ipython nbconvert --to notebook --nbformat 3 examples/learning/learn_associations.ipynb` precisely because I am using IPython 3 (nbformat 4), while our other notebooks are in IPython 2 (nbformat 3). If you look at line 22, it should be the correct format and version. See https://github.com/ipython/ipython/issues/8075. Thinking it's fine (just extra metadata)... In the future we should probably think about running this command automatically as a hook.

examples/learning/learn_associations.ipynb
arvoelke: Done

examples/learning/learn_associations.ipynb
arvoelke: Right, this was a mistake from earlier on in the PR that I never correct, see: "Try removing the intercepts argument from the error population? It's not actually necessary (my comment there is wrong)." Fixed

nengo/builder/connection.py
arvoelke: Done

nengo/tests/test_learning_rules.py
arvoelke: Done

examples/learning/learn_associations.ipynb
arvoelke: Good to know, thanks! Done

arvoelke:
Addressed @tbekolay comments. I decided to try `decoded` instead of `factored`, because that is really the intended semantics (IMO should not necessarily be tied to what it's connected to), and users should be more familiar with the former term.

arvoelke:
Bump

Seanny123:
Are you sure you rebased to master? It still shows conflicts need to be resolved on this end.

arvoelke:
Thanks for the heads-up. I rebased to a slightly out-of-date master. Should be good now.

hunse:
Yep, ready for @tbekolay to merge.

tbekolay:
Alrighty, doing so.

neworderofjamie:
:fireworks:

PR #729, opened 2015-06-09
Combined decoders and transform

hunse:
I combined `decoders` and `transform` in the connection builder. This makes things a lot cleaner, and it should be easier for users too, since they don't have to decide between probing 'transform' or 'decoders' when doing learning (right now it depends on the type of connection).

I also improved slicing, so we support slicing on neuron-to-neuron connections.

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -6,19 +6,19 @@ import nengo.utils.numpy as npext
 from nengo.builder.builder import Builder
 from nengo.builder.ensemble import gen_eval_points, get_activities
 from nengo.builder.node import SimPyFunc
-from nengo.builder.operator import DotInc, ElementwiseInc, PreserveValue, Reset
+from nengo.builder.operator import (
+    DotInc, ElementwiseInc, PreserveValue, Reset, SlicedCopy)
 from nengo.builder.signal import Signal
 from nengo.builder.synapses import filtered_signal
 from nengo.connection import Connection
 from nengo.ensemble import Ensemble, Neurons
 from nengo.neurons import Direct
 from nengo.node import Node
-from nengo.utils.builder import full_transform
 from nengo.utils.compat import is_iterable, itervalues
 
 
 BuiltConnection = collections.namedtuple(
-    'BuiltConnection', ['decoders', 'eval_points', 'transform', 'solver_info'])
+    'BuiltConnection', ['eval_points', 'solver_info', 'weights'])
 
 
 def get_eval_points(model, conn, rng):
@@ -54,6 +54,29 @@ def build_linear_system(model, conn, rng):
     return eval_points, activities, targets
 
 
+def multiply(x, y):
+    if x.ndim <= 2 and y.ndim < 2:
+        return x * y
+    elif x.ndim < 2 and y.ndim == 2:
+        return x.reshape(-1, 1) * y
+    elif x.ndim == 2 and y.ndim == 2:
+        return np.dot(x, y)
+    else:
+        raise ValueError("Tensors not supported (x.ndim = %d, y.ndim = %d)"
+                         % (x.ndim, y.ndim))
+
+
+def slice_signal(model, signal, sl):
+    assert signal.ndim == 1
+    if isinstance(sl, slice) and (sl.step is None or sl.step == 1):
+        return signal[sl]
+    else:
+        size = np.arange(signal.size)[sl].size
+        sliced_signal = Signal(np.zeros(size), name="%s.sliced" % signal.name)
+        model.add_op(SlicedCopy(signal, sliced_signal, a_slice=sl))
+        return sliced_signal
+
+
 @Builder.register(Connection)  # noqa: C901
 def build_connection(model, conn):
     # Create random number generator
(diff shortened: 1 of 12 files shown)


hunse:
Okay, this is ready for review!

tbekolay:
> I also improved slicing, so we support slicing on neuron-to-neuron connections.

Oh awesome, I ran into that bug recently :)

arvoelke:
Do we have any situations where we learn decoders on PES and then apply a transform? Would that still work? Say if I'm learning a scalar and then projecting that into a higher dimensional space. Or do we now need to apply the same transform to the error signal in this case?

hunse:
> Do we have any situations where we learn decoders on PES and then apply a transform? Would that still work? Say if I'm learning a scalar and then projecting that into a higher dimensional space. Or do we now need to apply the same transform to the error signal in this case?

I think you would need to apply the same transform on the error signal, or at least account for it in some way.

hunse:
Okay, this is rebased and ready for review.

arvoelke:
So, as I was suggesting above, I had been using PES assuming that the error is w.r.t. the decoded vector of dimension `size_mid`, not the transformed vector of dimension `size_out`. That's how it worked before, but we had no unit tests to assert this behaviour (so I guess I had been taking advantage of undefined behaviour). The old behaviour seemed more useful to me (e.g. projecting a learned scalar to each dimension in a product network), but I'm not really tied to it.

I've added a unit test to assert the new behaviour here (and fixed a line to make it pass): 91d59a3721e0ac52c2b470a994d962a2651d32f4 

Note: if you change line `173` to have no transform, then that's what would have passed on master. Feel free to cherry-pick this in if we're in agreement.

I'll take a look at the rest soon.

nengo/builder/connection.py
```diff
@@ -6,19 +6,30 @@
...
 from nengo.utils.compat import is_iterable, itervalues
 
 
 BuiltConnection = collections.namedtuple(
-    'BuiltConnection', ['decoders', 'eval_points', 'transform', 'solver_info'])
+    'BuiltConnection', ['eval_points', 'solver_info', 'weights'])
+
+
+def get_sliced_signal(signal, s, model):
+    assert signal.ndim == 1
```
arvoelke: - Rename to `sliced_signal` to be consistent with `filtered_signal`, and since `get` usually implies no side-effects
 - Docstring
 - Change assertion to exception

nengo/builder/operator.py
```diff
@@ -201,6 +201,42 @@ def step():
         return step
 
 
+class SlicedCopy(Operator):
+    """Copy from `a` to `b` with slicing: `b[b_slice] = a[a_slice]`"""
+    def __init__(self, a, b, a_slice=Ellipsis, b_slice=Ellipsis,
+                 kind='set', tag=None):
+        assert kind in ['set', 'inc', 'update']
```
arvoelke: It doesn't look like `kind='update'` is ever used in this code? So I'd suggest either making a unit test for `update` (and changing the assert to an exception), or replacing this with a boolean for `set` versus `inc`.

nengo/builder/connection.py
```diff
@@ -54,6 +65,18 @@ def build_linear_system(model, conn, rng):
     return eval_points, activities, targets
 
 
+def multiply(x, y):
```
arvoelke: Nit: docstring

nengo/builder/connection.py
```diff
@@ -107,83 +132,51 @@ def get_prepost_signal(is_pre):
...
-            decoders, solver_info = solver(
-                activities, targets, rng=rng,
-                E=model.params[conn.post_obj].scaled_encoders.T)
+            targets = multiply(targets, conn.transform.T)
+            E = model.params[conn.post_obj].scaled_encoders.T[post_slice]
+            decoders, solver_info = solver(activities, targets, rng=rng, E=E)
             model.sig[conn]['out'] = model.sig[conn.post_obj.neurons]['in']
-            signal_size = model.sig[conn]['out'].size
+            signal_size = conn.post_obj.neurons.size_in
+            post_slice = Ellipsis  # don't apply slice later
```
arvoelke: TIL...

arvoelke:
Gone over it pretty carefully and it looks good and well thought out. I'm fine with the renaming to `weights`, but regardless this all seems important enough to warrant the _all-important changelog entry_ (especially since this altered a bunch of our examples, and therefore our documentation).

hunse:
One thing to note is that with this PR, we no longer use `full_transform`. Should we remove it? This would affect #754 and #755.

nengo/builder/operator.py
hunse: Yeah, I put that in just because we could use such an op in the future. But I can remove it if people don't want it.

Also, re asserts: for things in the Builder, where the user should never get an error if we test things properly, asserts are fine IMO. This is just a sanity check for developers.

nengo/builder/connection.py
hunse: Do you want a docstring because every function should have a docstring, or because there's something about this function that you think is unclear? Other helper functions in this file do not have docstrings, but I'm happy to add one if things need clarifying.

nengo/builder/operator.py
arvoelke: If there's no code path executing it directly or indirectly, then a unit test should be added to assert its correctness. On the small off-chance that there's a subtle issue/bug with usage, then this may cause pain to the unsuspecting developer down the road who tries to use it assuming that it works. Likewise, future maintenance may take on an unnecessary burden to support something that is never used. So unless it's useful enough to motivate a unit test, it would be less work to omit for now.

nengo/builder/connection.py
arvoelke: Nit means an optional nitpick, so you don't have to change it. But, yes, I personally think other helper functions in this file should have docstrings. It takes 5 seconds to write the one-liner, and can significantly aid in the reuse of code and in quickly understanding the scope of helpers.

nengo/builder/operator.py
hunse: Or you can just read the code and know it works ;) 

But fair point. I'll just remove it.

hunse:
Addressed @arvoelke's suggestions. I kept the asserts in, because I think they should only ever be seen if *we* mess something up (i.e., users should never see them).

I also added a commit to remove the `as_update` argument from the `Copy` and `DotInc` operators, since it was unused in both cases.

arvoelke:
Cherry-pick 91d59a3721e0ac52c2b470a994d962a2651d32f4 and update the change log? Then it looks good to me.

arvoelke:
LGTM. Would be good for someone else to look this over now and get it in soon as it's a pretty big change.

tbekolay:
I'll give this a review tonight or early tomorrow!

tbekolay:
OK, reviewed! This looks great to me. Only a few issues, some of which I can do during the merge.

1. As commented inline, are you sure we need to remove `decoders` from `BuiltConnection`? I know Chris used this for a learning notebook recently...
2. Is it still possible to probe `'decoders'` and `'transform'`? If so, then I think we should allow this, but raise a deprecation warning and remove that ability in 2.2. If not, then perhaps we should allow this for now with the deprecation warning? In either case, we need a stronger changelog message, as it means end users will need to change their models (I can write this in the merge though).

I'll aim to merge this in today.

hunse:
My rationale for removing "decoders" from BuiltConnection was that it would be pretty rare for someone to want the decoders without the transform in it. I've started thinking about the "transform" and "function" on connections as going hand in hand (in fact, there's an argument that we shouldn't really even have "transform" on an ensemble->ensemble connection, because you can always build it in to "function", but let's not go there now since it makes problems for other connection types). Anyway, if you want the decoders for the "function" that your connection is computing, I think it makes the most sense if that includes the transform as well, since the transform is also part of the "function" you're computing.

As for probing "decoders" and "transform", it's not possible to probe them in the way you could before, which is why I chose "weights" as the name for the new matrix. That way, whether we're talking about BuiltConnection or probing, it's clear that "weights" is different from the old "transform" or "decoders". It would be good to have some error notifying users that "decoders" and "transform" are obsolete, if users try to probe them, but I can't think of any sensible way to map them to what's there now (mapping both "transform" and "decoders" to "weights" would cause problems, I think, because probes would appear to continue working but not return what they used to).

tbekolay:
Makes sense @hunse! I was working on adding a more helpful error message for when you try to probe decoders / transform, but it kind of became its own beast (see #781) so I'm happy to merge this as is. I'll give this a final lookover and history fixup tomorrow morning and then merge.

tbekolay:
Fixed up history. Squashed the 'removing decoders from BuiltConnection' commit into the main one, as without it most models won't build. Merging in a few minutes unless there are last minute objections!

PR #734, opened 2015-06-12
Do not apply pre-slice if connection has function

### LINKED ISSUE #733: what to use for weird slicing [::2]
```python
import numpy as np
import nengo
import nengo_spinnaker

model = nengo.Network(seed=1)
with model: 

    node_in = nengo.Node(output=[2,3,1,5])
    ens_out = nengo.Ensemble(n_neurons=200, dimensions=6)

    conn = nengo.Connection(node_in[[1,3]], ens_out[2:], function=lambda x: [x[0], x[1], x[0], x[1]])

print nengo.utils.builder.full_transform(conn)
```

raises the error

```bash
  print nengo.utils.builder.full_transform(conn)
  File "/home/tdewolf/Dropbox/code/nengo/nengo/utils/builder.py", line 49, in full_transform
    np.arange(size_in)[pre_slice]] = transform
ValueError: shape mismatch: objects cannot be broadcast to a single shape
```

which is a problem for nengo_spinnaker because it uses the ```full_transform``` method, i remember that we dealt with this at some point, thoughts? @tbekolay @arvoelke @hunse @mundya @tcstewar

hunse:
Since the pre-slice is already applied, either as part of the
Python function or when computing the decoders.

Fixes #733

### DIFF
--- a/nengo/utils/builder.py
+++ b/nengo/utils/builder.py
@@ -27,7 +27,8 @@ def full_transform(conn, slice_pre=True, slice_post=True, allow_scalars=True):
         If false, these scalars will be turned into scaled identity matrices.
     """
     transform = conn.transform
-    pre_slice = conn.pre_slice if slice_pre else slice(None)
+    pre_slice = (conn.pre_slice if slice_pre and conn.function is None else
+                 slice(None))
     post_slice = conn.post_slice if slice_post else slice(None)
 
     if pre_slice == slice(None) and post_slice == slice(None):


studywolf:
rebased, and then tested, works!

hunse:
Also note that #729 does slicing in a totally different way, and should avoid this problem completely in the reference simulator.

tbekolay:
Cool, this LGTM. And pretty small, so what harm can it do? Merging!

PR #736, opened 2015-06-13
LIF2C Model

epaxon:
Built a new neurons model that has 2 compartments and is capable of linear summation as well as multiplicative shunting through a shunting inhibitory channel. Modified Connection to take a new parameter known as target, which allows pre-synaptic inputs to influence any post-synaptic state variable. Target defaults to 'in', but can be changed to the name of another state variable. LIF2C multiplication is directly onto the neural firing rates, and does not support multiplication with negative numbers. This means that there are some issues when trying to multiply the underlying vector. Example in LIF2C basics.


tcstewar:
Nifty!  I'm looking forward to fiddling with this model a bit....

As for submitting it to the repository, we've tended to leave alternate neuron models as separate packages.  We're trying to stick with a core set of neuron models in the main nengo repository, as otherwise things get complicated with having multiple backends.  There's an example of a separate repository here https://github.com/jleugeri/nengo-adaptiveLN-models which was made by a student at a summerschool I went to last month.

However, with this neuron model it really looks like there's an indication of features we should consider adding to the core of nengo to make it easier to make neuron models.  The two that jump out at me are:

 - making Connections into an alternate Signal in the neuron model
 - getting access to the input current without the bias added.

Do we have a suggested approach for Connections into different inputs for the neuron?  This seems like something that would be needed for a variety of models.  I'm not sure what the right syntax should be... the two options that come to mind are ```Connection(a, b, target='dopamine')``` and ```Connection(a, b.dopamine)```.  And we should also support the direct neuron connection too (```Connection(a, b.neurons, target='dopamine')``` and ```Connection(a, b.neurons.dopamine)```

Thoughts?  Is there an easier way to do this in current nengo?  I couldn't see one....

epaxon:
Yep, there's definitely some niceties that could be added to make the NeuronModels more generalizeable. I'm not sure how to do the gain and bias -- perhaps in the builder there should be a way to store the gain and bias into the neuronmodel, instead of setting it manually in multiple places. My worry with saying b.dopamine, or b.neurons.dopamine, is that they are not visible until runtime, and there may be issues with the state variables being nested (i.e. b.dopamine would really be b.neurons.dopamine * encoders). With the target variable, this could all be handled in the builder of connection.

Seanny123:
@tbekolay @tcstewar should we move this to `nengo_extras` or do something else with it? It's been lying around in Core Nengo for a while.

tbekolay:
Mostly it's up to @epaxon. @epaxon, if you haven't made a repository for this neuron type already, we have a repository now, https://github.com/nengo/nengo_extras, that we've been using for things like extra neuron types. What do you think about putting this neuron type there?

If you want to make a PR on that repo, then let us know and we'll close this PR. If you're OK with us moving it there (with attribution of course), then leave a comment and we'll close this when we make the PR on nengo_extras.

epaxon:
Do whatever you guys think is best :-)

tbekolay:
Cool, I'll put moving this to `nengo_extras` on my todo list, but close in the meantime.

PR #743, opened 2015-06-16
Cyclic vector

hunse:
Added a function to generate cyclic vectors. These vectors leave a target vector unchanged after a specified number of convolutions.

### DIFF
--- a/nengo/utils/numpy.py
+++ b/nengo/utils/numpy.py
@@ -57,7 +57,7 @@ def array(x, dims=None, min_dims=0, readonly=False, **kwargs):
     return y
 
 
-def array_hash(a, n=100):
+def array_hash(a, n=10000):
     """Simple fast array hash function.
 
     For arrays with size larger than ``n``, pick ``n`` elements at random
@@ -67,7 +67,7 @@ def array_hash(a, n=100):
     if not isinstance(a, np.ndarray):
         return hash(a)
 
-    if a.size < n:
+    if a.size <= n:
         # hash all elements
         v = a.view()
         v.setflags(write=False)
--- a/nengo/spa/tests/test_utils.py
+++ b/nengo/spa/tests/test_utils.py
@@ -1,0 +1,69 @@
+import numpy as np
+import pytest
+
+from nengo.utils.compat import range
+from nengo.utils.numpy import array_hash
+from nengo.spa.utils import cyclic_vector
+
+
+def circconvk(a, b, k=1, invert_a=False, invert_b=False, axis=-1):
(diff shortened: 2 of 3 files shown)


jgosmann:
[There is also this slightly different method](https://github.com/ctn-waterloo/summerschool2015/blob/workingmemory/workingmemory/Generate%20circular%20vectors.ipynb) which guarantees the similarity structure of subsequent vectors, but uses one specific vector instead of generating a random one.

Seanny123:
Tried taking a shot at reviewing this, but I was way over my head mathematically. What resource did you derive this stuff from? Maybe I could read that to gain some sort of understanding? Alternatively, @jgosmann, you seemed to understand what was going on?

jgosmann:
I derived pretty much all of it myself. Probably used Wikipedia to look up some common knowledge like DFT matrices or Euler's formula. I could probably walk you through it at some point ... or maybe it's worth doing a presentation at a lab meeting about this and some related stuff?

Seanny123:
You'll have to ask Chris. I never have a good feel for what should be
presented during a lab meeting.

Seanny123:
Since this was asked about a lot during the last Summer School, should we make sure this gets merged before the next summer school? @tbekolay do you want to take a shot at this?

nengo/utils/numpy.py
```diff
@@ -57,7 +57,7 @@ def array(x, dims=None, min_dims=0, readonly=False, **kwargs):
     return y
 
 
-def array_hash(a, n=100):
+def array_hash(a, n=10000):
```
hunse: Note to reviewers: I changed this limit because hashing an array of this size is still very fast on my machine (~0.1 ms). The previous limit was just too small, resulting in more possibility for hash collisions.

jgosmann:
This might better fit into nengo_extras?

Also, this is just one of two methods and to me it is not clear which one should be preferred (if any). [Some details in my last lab presentation.](https://github.com/ctn-waterloo/internal_tech_reports/blob/master/Semantic%20Pointers%20and%20convolutions.ipynb)

hunse:
> Also, this is just one of two methods and to me it is not clear which one should be preferred

There's no reason we can't offer both. Just write yours up :)

jgosmann:
Documenting it will be somewhat tiresome ... but I should probably do that soon-ish (might not be before the summer school). But it would be good to have some other SPA changes I'm working on merged before.

We still have to agree on where to put it, though.

hunse:
Merged as nengo/nengo_extras#33

PR #751, opened 2015-06-19
Typos

studywolf:
phew! good catch @studywolf


studywolf:
wait damnit, did this not from master

PR #753, opened 2015-06-20
Fixed tests for seed bug in numpy 1.8.0

arvoelke:
Got some weirdness on my Windows 64-bit Python install:

    >>> seed = function_seed(function_seed)
    >>> seed
    581567418L
    >>> seed > npext.maxint
    False
    >>> np.random.RandomState(seed)
      File "mtrand.pyx", line 564, in mtrand.RandomState.__init__(numpy\random\mtrand\mtrand.c:5217)
      File "mtrand.pyx", line 600, in mtrand.RandomState.seed (numpy\random\mtrand\mtrand.c:5468)
    ValueError: object of too small depth for desired array
    >>> np.random.RandomState(int(seed))
    <mtrand.RandomState object at 0x0000000005D3B030>
    >>> int(seed) == seed
    True
    >>> np.__version__
    '1.8.0'

Basically, you can't pass longs to `RandomState`, even if they are `< npext.maxint`. This breaks most of the tests on my box atm. The fix is to cast it to `int`.

### DIFF
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -157,7 +157,10 @@ def function_seed(function, mod=0):
     hash_list = os.path.normpath(path).split(os.path.sep) + [c.co_name]
     hash_string = ensure_bytes('/'.join(hash_list))
     i = int(hashlib.md5(hash_string).hexdigest()[:15], 16)
-    return (i + mod) % npext.maxint
+    s = (i + mod) % npext.maxint
+    int_s = int(s)  # numpy 1.8.0 bug when RandomState on long type inputs
+    assert type(int_s) == int  # should not still be a long because < maxint
+    return int_s
 
 
 @pytest.fixture


nengo/tests/conftest.py
```diff
@@ -157,7 +157,10 @@ def function_seed(function, mod=0):
     hash_list = os.path.normpath(path).split(os.path.sep) + [c.co_name]
     hash_string = ensure_bytes('/'.join(hash_list))
     i = int(hashlib.md5(hash_string).hexdigest()[:15], 16)
-    return (i + mod) % npext.maxint
+    s = (i + mod) % npext.maxint
+    int_s = int(s)  # numpy 1.8.0 bug when RandomState on long type inputs
+    assert int_s == s  # no overflow because of mod by npext.maxint
```
jgosmann: This does not work as you would think. [The documentation of `int()` says](https://docs.python.org/2/library/functions.html#int)
>  If the argument is outside the integer range, the function returns a long object instead.
Thus, the assert will never be false.

jgosmann:
What surprises me is that you get a long int in the first place. Line 159 seems like it should be within normal int range and thus it should not produce a long int. But neither should `npext.maxint` be a long int. Can you check what type those expressions have?

nengo/tests/conftest.py
arvoelke: Ah okay. I could change the assert to `type(int_s) == int`, or just leave it out?

arvoelke:
Line 159 does produce a long (`type(i) == long`), because, as you were saying, `int` can return `long`. Although `type(npext.maxint) == int`, the modulo does not implicitly cast.

jgosmann:
I wonder, if it would be better to adjust the number of bytes we take from the hexdigest to fit into a normal int on all systems (or figure out the appropriate number for the system)?
Another thing: Would it be better to use `digest` instead of `hexdigest` and convert to an int with the `struct` package (this wouldn't require to convert to the hex string and parse it again).

nengo/tests/conftest.py
jgosmann: `type(int_s) == int`would probably be better.

arvoelke:
For your first question, the `+ mod` on line 160 could still make the int overflow into long territory. For your second question, this code is only used for testing, so I'm all for whatever assures us correctness with the least amount of changes.

jgosmann:
Sure, should be fine if you fix the assert.

arvoelke:
Fixed

arvoelke:
Bump. Currently have to do all my dev work with this manually patched in.

tbekolay:
[Travis-CI builds for NumPy 1.8](https://travis-ci.org/nengo/nengo/jobs/67902922). Why doesn't it fail in the same way that your environment fails? Is this a Windows issue?

arvoelke:
Probably has to do with 64-bit Windows and however numpy was compiled.

arvoelke:
To clarify, the problem of being passed longs should be present on all machines (see [for instance](https://groups.google.com/forum/#!topic/stan-users/Gyzesp5N4U4)), but the problem that `type(seed) == long` should be exclusive to my machine, which is what's being fixed here.

arvoelke:
Bump:grey_question:

tbekolay:
Your bump has been ... accepted! Merging now.

arvoelke:
Thanks! :smile:

PR #754, opened 2015-06-20
Support slicing with numpy arrays

arvoelke:
Previously if you tried to slice with a numpy array, you got some obscure error about needing `any()` or `all()` to compare the slice to `slice(None)`. This should make it work now.

Branched off #753.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -118,6 +118,8 @@ Release History
 - Added ``[progress]`` section to ``nengorc`` which allows setting
   ``progress_bar`` and ``updater``.
   (`#693 <https://github.com/nengo/nengo/pull/693>`_)
+- Numpy arrays can now be used as indices for slicing objects.
+  (`#754 <https://github.com/nengo/nengo/pull/754>`_)
 
 **Bug fixes**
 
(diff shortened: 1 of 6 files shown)


hunse:
Okay, I think this is good to go. The changes to `full_transform` aren't really necessary, since we don't use this in the builder anymore, but I think some backends might, so we might as well keep it up to date. `eq_none_slice` didn't seem like a generally useful function to me, so I just made it inline where it's used in `full_transform`.
Okay, I think this is good to go. The changes to `full_transform` aren't really necessary, since we don't use this in the builder anymore, but I think some backends might, so we might as well keep it up to date. `eq_none_slice` didn't seem like a generally useful function to me, so I just made it inline where it's used in `full_transform`.

hunse:
Okay, made the discussed changes.

tbekolay:
OK, LGTM. Should I squash the first two commits or keep them separate?

hunse:
I couldn't decide whether to squash or not, either. Probably best to squash, since with other changes to the repo, it's actually only the first commit that's needed for slicing with numpy arrays, so it might as well all be together.

PR #755, opened 2015-06-20
Fix full_transform for 1D vector

### LINKED ISSUE #750: Uncaught transform error
```python
with nengo.Network():
    a = nengo.Ensemble(300, 1)
    b = nengo.Ensemble(300, 1)
    nengo.Connection(a, b, transform=[1])
```
Should raise an error because the transform is a vector rather than a matrix or a scalar.

This is a problem in Nengo from Git on both my machine and Chris'.

mundya:
`full_transform` should never return a vector, but in one specific case this was possible.

```python
>>> conn = nengo.Connection(ens1, ens1, transform=[1])
>>> full_transform(conn)
np.array([ 1.])
```

This commit fixes this by identifying this special case and returning a scalar instead.

Fixes #750.

### DIFF
--- a/nengo/utils/builder.py
+++ b/nengo/utils/builder.py
@@ -36,7 +36,10 @@ def full_transform(conn, slice_pre=True, slice_post=True, allow_scalars=True):
             # transform is already full, so return a copy
             return np.array(transform)
         elif transform.size == 1 and allow_scalars:
-            return np.array(transform)
+            if transform.ndim == 1:
+                return np.array(transform[0])
+            else:
+                return np.array(transform)
 
     # Create the new transform matching the pre/post dimensions
     func_size = conn.function_info.size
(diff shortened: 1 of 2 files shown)


nengo/utils/builder.py
```diff
@@ -36,7 +36,10 @@ def full_transform(conn, slice_pre=True, slice_post=True, allow_scalars=True):
             # transform is already full, so return a copy
             return np.array(transform)
         elif transform.size == 1 and allow_scalars:
-            return np.array(transform)
+            if transform.ndim == 1:
+                return np.array(transform[0])
```
arvoelke: NTS: This is okay since `transform.size == 1` ensures there is only one element.

arvoelke:
LGTM

hunse:
Merged.

hunse:
Thanks @mundya!

PR #756, opened 2015-06-22
Build synapses

hunse:
Remove `filtered_signal` from the builder, and just build synapses directly.

### DIFF
--- a/nengo/builder/builder.py
+++ b/nengo/builder/builder.py
@@ -41,7 +41,11 @@ class Model(object):
         op.make_step(signals, self.dt, np.random)
 
     def has_built(self, obj):
-        """Returns true iff obj has been processed by build."""
+        """Returns true if obj has built parameters.
+
+        Some objects (e.g. synapses) do not add params and can be built more
+        than once (i.e. this function returns false), allowing re-use.
+        """
         return obj in self.params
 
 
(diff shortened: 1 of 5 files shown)


tbekolay:
Pushed a small fixup commit to avoid using the builtin variable name `input` (used `input_sig`).

I was initially a bit concerned that the synapse builder doesn't store the signal anywhere, but I think this approach is better because it allows the builder that's calling this builder to store the signal somewhere instead (synapses don't exist outside of other objects, after all).

So yeah, I'll merge if @hunse is ok with the variable name change!

hunse:
Personally, using `input` as a variable name doesn't bother me. No one should be using the `input` function anyway, and there's precedent in other packages, like this [code I copied from Scipy](https://github.com/nengo/nengo/blob/master/nengo/utils/filter_design.py#L288).

But if you want to change it, that's fine. I'll also change it above. Also, `input_sig` is not a keyword arg, so we probably should just use it positionally, which means the name is much less of an issue. I'll add that change.

hunse:
Okay, added a fixup.

tcstewar:
> Personally, using input as a variable name doesn't bother me.

As a tangential note, I've stopped using ```input``` as a variable name in all tutorials and nengo/nengo_gui example code, just because the syntax highlighters in most editors pick up on it and so it looks weird.  Much less of an issue in core code like this, but I thought I'd mention why I'm doing it elsewhere...  :)

tbekolay:
> syntax highlighters in most editors pick up on it

That's why I changed it too :) I think it's less of an issue here since it's an arg to a function, but yeah, doesn't hurt to change it I think. And using it positionally is good too. Merging!

PR #758, opened 2015-06-23
Dists, Processes, Synapses are now frozen

### LINKED ISSUE #746: Param defaults can be modified in-place
One example is the connection solver:
```python
conn = nengo.Connection(a, b)
conn.solver.weights = True
```
This actually changes the default solver, so **all** connections using the default solver now solve for weights. We could have similar problems with other parameters, too.

Two possible solutions are:
  1. Have parameters make a copy when they're being set to their default value. This requires that they know how to copy the default value.
  2. Make objects such as neuron types, solvers, etc. frozen after creation, so that trying to change an instance will result in an error.

hunse:
Distributions, Processes, and Synapses are now unmodifyable (frozen).
This means that they can be used as defaults without risk of a user
modifying them (fixes #746). They can also be compared and hashed.

### DIFF
--- a/nengo/params.py
+++ b/nengo/params.py
@@ -1,11 +1,12 @@
 import collections
-import weakref
+import inspect
 
 import numpy as np
 
-from nengo.utils.compat import is_array, is_integer, is_number, is_string
-from nengo.utils.numpy import compare
-from nengo.utils.stdlib import checked_call
+from nengo.utils.compat import (
+    is_array, is_integer, is_number, is_string, itervalues)
+from nengo.utils.numpy import array_hash, compare
+from nengo.utils.stdlib import WeakKeyIDDictionary, checked_call
 
 
 class DefaultType:
@@ -38,17 +39,25 @@ class Parameter(object):
         If true, the parameter can only be set once.
         By default, parameters can be set multiple times.
     """
-    def __init__(self, default=Unconfigurable, optional=False, readonly=False):
+    equatable = False
+
+    def __init__(self, default=Unconfigurable, optional=False, readonly=None):
         self.default = default
         self.optional = optional
+
+        if readonly is None:
+            # freeze Unconfigurables by default
+            readonly = default is Unconfigurable
         self.readonly = readonly
 
-        # use WeakKey dictionaries so items can still be garbage collected
-        self.defaults = weakref.WeakKeyDictionary()
-        self.data = weakref.WeakKeyDictionary()
+        # default values set by config system
+        self._defaults = WeakKeyIDDictionary()
+
+        # param values set on objects
+        self.data = WeakKeyIDDictionary()
 
     def __contains__(self, key):
-        return key in self.data or key in self.defaults
+        return key in self.data or key in self._defaults
 
     def __delete__(self, instance):
         del self.data[instance]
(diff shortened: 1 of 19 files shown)


tbekolay:
Dang, you're on a roll lately! Looking forward to checking this (and other) PRs out soon :D

hunse:
Is this good to go? The WeakKeyIDDictionary additions look good. I just rebased it to master.

tbekolay:
No, I haven't finished my review yet. The dict is all that I've looked at in detail so far.

hunse:
After the rebase, some of the cache tests are failing, I think because we don't have Parameters set up for pickling. I might need @jgosmann and @tbekolay to help figure out the best way to do this.

tbekolay:
Alrighty! Fun times. Will look into this after the progress bar.

hunse:
Made FrozenObjects pickleable for the cache, and added default strings for FrozenObjects as well as specific strings for Processes and some Distributions that hadn't been covered yet, as requested in #832.

Seanny123:
Other than the validation part of the Ndparam which I didn't understand, this looks good to me. That being said, I guess there's no point to rebasing until Trevor is done his thesis?

nengo/params.py
```diff
@@ -187,18 +233,37 @@ def __set__(self, instance, ndarray):
...
-        try:
-            ndarray = np.asarray(ndarray, dtype=np.float64)
-        except TypeError:
-            raise ValueError("Must be a float NumPy array (got type '%s')"
-                             % ndarray.__class__.__name__)
-
-        if ndarray.ndim != ndim:
+    def validate(self, instance, ndarray):  # noqa: C901
+        if isinstance(ndarray, np.ndarray):
+            ndarray = ndarray.view()
```
tbekolay: This makes it possible to have non-float arrays. Is this intentional? It seems useful some places, but dangerous in others.

nengo/params.py
```diff
@@ -187,18 +233,37 @@ def __set__(self, instance, ndarray):
...
+        else:
+            try:
+                ndarray = np.array(ndarray, dtype=np.float64)
+            except TypeError:
+                raise ValueError("Must be a float NumPy array (got type '%s')"
+                                 % ndarray.__class__.__name__)
+        if self.readonly:
+            ndarray.setflags(write=False)
+
+        if '...' in self.shape:
```
tbekolay: Maybe a comment here would be helpful... something like `# Convert '...' to the appropriate number of '*' dimensions` (if I'm reading this correctly)

tbekolay:
Rebased and re-reviewed. @hunse, could you give the okay on commit e6364b6883973f15? Looks ready to merge to me if that commit looks good to you (it's all pretty minor style things).

nengo/params.py
hunse: It seems strange to me to make the requirement for float, and specifically float64, at the parameter level. I would prefer to just keep whatever users give us. This can be especially useful for large data structures that we want to store but not necessarily cast to float64. You're right, though, that this means that backends have to be more careful about how they use the data in these parameters. Maybe this should be a separate change? I'm trying to remember why I changed it in this PR specifically.

hunse:
d3c30eb all looks good to me.

nengo/params.py
tbekolay: I think it's fine for now; I was thinking about it a bit more and we deal with this more explicitly during the build process, which makes more sense to me at the moment. I made #930 to look at this in more detail in the future.

tbekolay:
Just realized there's no changelog entry; I really need to make some checklists for that. Think anything is critical to note? I feel like it's mostly internal changes that won't affect users, but you never know... if something's critical, let me know and I'll add it to #918, which already touches the changelog.

PR #768, opened 2015-07-03
Added distribution for dot products / intercepts

arvoelke:
Subclassed `SubvectorLength` to get the distribution that I use for my adaptive cleanup (heteroassociative memory). This is important for setting the intercepts of neurons so that they fire with some chosen probability (see docstring for details).

Also added the inverse cdf to `SqrtBeta`, and some optional unit tests for the `SqrtBeta` distribution.

### DIFF
--- a/nengo/dists.py
+++ b/nengo/dists.py
@@ -11,8 +11,11 @@ class Distribution(object):
     The only thing that a probabilities distribution need to define is a
     ``sample`` function. This base class ensures that all distributions
     accept the same arguments for the sample function.
-
     """
+
+    def _sample_shape(self, n, d=None):
+        """Returns output shape for sample method."""
+        return (n,) if d is None else (n, d)
 
     def sample(self, n, d=None, rng=np.random):
         """Samples the distribution.
@@ -58,7 +61,7 @@ class PDF(Distribution):
         return "PDF(x=%r, p=%r)" % (self.x, self.pdf)
 
     def sample(self, n, d=None, rng=np.random):
-        shape = (n,) if d is None else (n, d)
+        shape = self._sample_shape(n, d)
         return np.interp(rng.uniform(size=shape), self.cdf, self.x)
 
 
@@ -98,7 +101,7 @@ class Uniform(Distribution):
             self.low, self.high, ", integer=True" if self.integer else "")
 
     def sample(self, n, d=None, rng=np.random):
-        shape = (n,) if d is None else (n, d)
+        shape = self._sample_shape(n, d)
         if self.integer:
             return rng.randint(low=self.low, high=self.high, size=shape)
         else:
@@ -139,7 +142,7 @@ class Gaussian(Distribution):
         return "Gaussian(mean=%r, std=%r)" % (self.mean, self.std)
 
     def sample(self, n, d=None, rng=np.random):
-        shape = (n,) if d is None else (n, d)
+        shape = self._sample_shape(n, d)
         return rng.normal(loc=self.mean, scale=self.std, size=shape)
 
 
@@ -178,7 +181,7 @@ class Exponential(Distribution):
         self.high = high
 
     def sample(self, n, d=None, rng=np.random):
-        shape = (n,) if d is None else (n, d)
+        shape = self._sample_shape(n, d)
         x = rng.exponential(self.scale, shape) + self.shift
         high = np.nextafter(self.high, np.asarray(-np.inf, dtype=x.dtype))
         return np.clip(x, self.shift, high)
@@ -286,13 +289,14 @@ class SqrtBeta(Distribution):
     --------
     SubvectorLength
     """
+
     def __init__(self, n, m=1):
         super(SqrtBeta, self).__init__()
         self.n = n
         self.m = m
 
     def sample(self, num, d=None, rng=np.random):
-        shape = (num,) if d is None else (num, d)
+        shape = self._sample_shape(num, d)
         return np.sqrt(rng.beta(self.m / 2.0, self.n / 2.0, size=shape))
 
     def pdf(self, x):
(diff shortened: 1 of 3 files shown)


arvoelke:
Also, I think the `invcdf` I added for `SqrtBeta` would be useful for setting the radius in an `EnsembleArray` in order to cover some proportion of inputs (as suggested in your paper)? Maybe this should be documented more clearly?

jgosmann:
Could definitely by useful (but I don't need it the paper/spaopt).

Seanny123:
@arvoelke why is this marked as "work in progress"? What else is left to be done to this pull request before it can be properly reviewed?

arvoelke:
Think I did that in case @jgosmann had suggestions based on his research, and because of the two todo's in the original post (better naming? changelog?). These can be subsumed by the review (so yes it's ready). :smile:

nengo/dists.py
```diff
@@ -294,6 +294,25 @@ def cdf(self, x):
...
+        ----------
+        y : ndarray
+            Cumulative probabilities in [0, 1].
+
+        Returns
+        -------
+        ndarray
+            Evaluation points `x` in [0, 1] such that `P(X <= x) = y`.
+        """
+        from scipy.special import betaincinv
```
Seanny123: Usually if we have a SciPy dependency, we throw a special error if it's not found. Should this be done here as well?

nengo/dists.py
```diff
@@ -309,11 +328,52 @@ class SubvectorLength(SqrtBeta):
...
 
 
+class CosineSimilarity(SubvectorLength):
+    """Distribution of dot products between random unit vectors.
+
+    This is equivalent to the distribution of a single coefficient in a unit
+    vector (i.e. a single dimension of `UniformHypersphere(surface=True)`),
+    which can be obtained from `SubvectorLength` by symmetry.
+
+    This can be used to calculate an intercept `c = invcdf(1 - p)` such that
```
Seanny123: A bit nit-picky, but I have a really hard time parsing `invcdf`. What does it mean? Can it be renamed?

arvoelke: It means "Inverse Cumulative Distribution Function". The `cdf` part is standard terminology. This method name has already been used elsewhere a couple places in the file.

nengo/dists.py
arvoelke: This is how it was done in a couple other places in this file.

nengo/dists.py
Seanny123: *facepalm*

jgosmann: Would it make sense to be [consistend with SciPy and use `ppf`](http://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.ppf.html#scipy.stats.rv_continuous.ppf)?

arvoelke: Done

arvoelke:
Renamed invcdf to ppf. Think this should be ready if no complaints.

nengo/tests/test_dists.py
```diff
@@ -134,6 +134,109 @@ def test_sqrt_beta(n, m):
...
+    x = np.arange(dt, 1+dt, dt)
+
+    dist = dists.SqrtBeta(n, m)
+
+    pdf = dist.pdf(x)
+    cdf = dist.cdf(x)
+    ppf = dist.ppf(cdf)
+
+    # The pdf should reflect the samples
+    np.random.seed(seed)
```
hunse: You shouldn't have to do this. Make a `numpy.random.RandomState` and pass it where you need it.

hunse: Actually, you can just get one with the `rng` fixture.

nengo/dists.py
```diff
@@ -335,6 +335,25 @@ def cdf(self, x):
             sq_x < 1., betainc(self.m / 2.0, self.n / 2.0, sq_x),
             np.ones_like(x))
 
+    def ppf(self, y):
+        """Inverse cumulative distribution function.
```
hunse: Say what PPF stands for here, to help people understand why it's called "ppf" (I'd never heard the term before scipy, and I honestly can't remember what it is right now, but it's in the scipy docs).

hunse:
Made a few comments. Go ahead and add the changelog entry, too (that's the submitter's job).

nengo/tests/test_dists.py
arvoelke: Done

nengo/dists.py
arvoelke: Done (percent point function)

arvoelke:
Added changelog and addressed above comments (thanks)!

arvoelke:
Bumping since this is needed by @mundya and @neworderofjamie this week.

nengo/dists.py
```diff
@@ -350,11 +369,52 @@ class SubvectorLength(SqrtBeta):
...
+
+    See also
+    --------
+    SqrtBeta
+    """
+
+    def __init__(self, dimensions):
+        super(CosineSimilarity, self).__init__(dimensions)
+
+    def sample(self, num, rng=np.random):
```
hunse: `sample` doesn't have the same signature as we typically use for `Distribution.sample`. We might as well take the extra `d` parameter like `SubvectorLength` does, if it's not too much trouble (and I don't think it should be).

arvoelke: Good point. Fixed this now, added a test, and removed some repeat logic in the process.

nengo/dists.py
```diff
@@ -350,11 +373,53 @@ class SubvectorLength(SqrtBeta):
...
     SqrtBeta
     """
+
     def __init__(self, dimensions, subdimensions=1):
         super(SubvectorLength, self).__init__(
             dimensions - subdimensions, subdimensions)
 
 
+class CosineSimilarity(SubvectorLength):
+    """Distribution of dot products between random unit vectors.
```
hunse: "Distribution of cosine similarity between two random vectors."
or maybe
"Distribution of the cosine of the angle between two random vectors."

then below: "The cosine similarity is given by the cosine of the angle between two vectors, which is equal to the norm of the dot product of the vectors, divided by the norms of the individual vectors".

arvoelke: Done. But "norm of the dot product" -> just "dot product" (note the result is signed, unlike `SubVectorLength`).

arvoelke:
Improved documentation to clarify the connection between this distribution and the cosine angle.

PR #777, opened 2015-07-08
fixed product network layout

Seanny123:
While trying to fix the `spa.Compare` network I noticed that the `Product` network displayed really weirdly in the `nengo_gui` so I fixed it. Basically all this does is remove this lonely node hanging around and make it more graphically logical to connect to the `Product` network. I thought it would affect more stuff, but apparently not.

Test with the following code:
```
import nengo
import numpy as np

model = nengo.Network()
D = 8
n_neurons = 20

with model:
    model.out = nengo.Node(size_in=D)
    model.ens_out = nengo.Node(size_in=16)

    model.prod = nengo.networks.Product(n_neurons, D)

    nengo.Connection(model.prod.output, model.out)
    nengo.Connection(model.prod.product.output, model.ens_out)
```

Before:
before_prod_fix

After:
after_prod_fix

### DIFF
--- a/nengo/networks/product.py
+++ b/nengo/networks/product.py
@@ -26,7 +26,8 @@ def Product(n_neurons, dimensions, input_magnitude=1, config=None, net=None):
                                     radius=input_magnitude * np.sqrt(2))
         nengo.Connection(net.A, net.product.input[::2], synapse=None)
         nengo.Connection(net.B, net.product.input[1::2], synapse=None)
-        net.output = net.product.add_output('product', lambda x: x[0] * x[1])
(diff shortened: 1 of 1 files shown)


jgosmann:
I think, it might be a better fix to [remove this line](https://github.com/nengo/nengo/blob/product-net-fix/nengo/networks/product.py#L22) and leave everything else as is. That should look like the upper network with just the dangling node removed (whereas in your version a passthrough node is inserted which seems unecessary).

jgosmann:
This PR will be superseded by #651 which completely changes the product network and which will look like this (for 3-dimensional vectors):
snapshot3

PR #779, opened 2015-07-09
Add clipped exponential distribution.

jgosmann:
This distribution is useful for thresholding in ensembles. It performs better than a uniform distribution of intercepts or setting all intercepts exactly to the threshold.

download
blue: fixed intercepts
green: uniform
red: exponential distribution
The ideal function would be a step function going from 0 to 1 at 0.3 (plus synaptic filtering).

I am planning to update the `AssociativeMemory` to use this distribution and maybe implement a thresholding network.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -139,6 +139,9 @@ Release History
   ensemble's encoders to fire selectively for its inputs. (see
   ``examples/learning/learn_associations.ipynb``).
   (`#727 <https://github.com/nengo/nengo/issues/727>`_)
+- Added a clipped exponential distribution useful for thresholding, in
+  particular in the AssociativeMemory.
+  (`#582 <https://github.com/nengo/nengo/pull/779>`_)
 
 **Bug fixes**
 
--- a/nengo/dists.py
+++ b/nengo/dists.py
@@ -143,6 +143,31 @@ class Gaussian(Distribution):
         return rng.normal(loc=self.mean, scale=self.std, size=shape)
 
 
+class Exponential(Distribution):
+    """An exponential distribution where high values are clipped.
+
+    Parameters
+    ----------
+    scale : float
+        The scale parameter (inverse of the rate parameter lambda).
+    shift : float, optional
+        Amount to shift the distribution by. There will be no values smaller
+        than this shift when sampling from the distribution.
+    high : float, optional
+        All values larger than this value will be clipped to this value.
+    """
+    def __init__(self, scale, shift=0., high=np.inf):
+        self.scale = scale
+        self.shift = shift
+        self.high = high
+
+    def sample(self, n, d=None, rng=np.random):
+        shape = (n,) if d is None else (n, d)
+        return np.clip(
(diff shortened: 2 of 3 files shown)


xchoo:
Oh. I'm working on the ```AssociativeMemory``` thing right now. :smile:

jgosmann:
I know of that problem, but I am not sure whether to add that tolerance in the distribution or wherever it is used (the required tolerance might depend on the usage?)

xchoo:
Can we add the tolerance as a parameter in the constructor?

jgosmann:
We could ...

hunse:
Or we could just use the minimum tolerance for the type:
```python
high = np.nextafter(self.high, np.asarray(-np.inf, dtype=exp_val.dtype))
```

jgosmann:
This might be the best solution. I didn't know about `nextafter` (it doesn't seem to be in the Numpy documentation).

xchoo:
Indeed. This is the first time i've seen ```nextafter```. :smile:

hunse:
This seems fine to me. At some point, we might need to move some of the more esoteric distributions in `dists.py` somewhere else; for example, this one could maybe go with the SPA stuff if it's mostly being used there. That would keep `dists.py` more for distributions that users are going to use. But I'm fine putting this in `dists.py` for now.

jgosmann:
> this one could maybe go with the SPA stuff if it's mostly being used there

I don't think that this distribution is specific to the SPA. You could as well build a scalar thresholding ensemble with it. Also, I don't think, that an exponential distribution is very esoteric. But I could see moving the `SqrtBeta` stuff around.

hunse:
This needs a changelog entry. Maybe mention briefly what it's useful for?

It would also be great to have a notebook example showing it off, but we can add this down the road. (Maybe have one notebook documenting all distributions? Though I feel like Gaussian and Uniform are more self-explanatory.) Also at some point we might want to consider building the Laplace distribution into this class (or making it a subclass or something), since it's very similar and takes the same parameters.

hunse:
Oh, also write the PDF in the docstring. You can probably copy it from the Numpy exponential distribution docstring, though frankly I'm not totally keen on all the LaTeX in their formatting, because it makes it hard to read if you just open the docstring in Python.

Seanny123:
@hunse added the docs you requested. Is this now ready to merge?

hunse:
Okay, I fixed up the docstring more. Can somebody look that over?

jgosmann:
LGTM

hunse:
@jgosmann and I talked, and decided to change the name of this distribution to `Exponential` (rather than `ClippedExpDist`), since if `high` is not set it can be used as a normal `Exponential` distribution.

tbekolay:
:+1: for exponential -- I was also going to mention that the `Dist` at the end is likely redundant based on context, so happy this got changed :)

As a sidenote, I've been using this distribution a lot lately for inhibitory populations, and it's extremely useful. Thanks @jgosmann! Now I can switch back to master!

Issue #759, opened 2015-06-23
spa.Actions unsupported operand


Seanny123:
When I try to parse a spa.Actions like:

```
cortical_actions = spa.Actions(
    'out = (color_in * shape_in) ~cue'
)
```

I get the error:

```
TypeError: Invalid operator in expression ' (color_in * shape_in) * ~cue': unsupported operand type(s) for *: 'Convolution' and 'Source'
```

Is there a reason this type of operation is unsupported? It seems possible to chain those networks together through order of operations, but maybe I'm missing something? Is it that the expression is somewhat ambiguous as to what the user wants to do?

tcstewar:
> Is there a reason this type of operation is unsupported?

The main reason is implementation -- it's annoying to implement all those possibilities.  Notice that ```dot(color, shape)``` also isn't supported.

That said, I think there's a design decision here.  After all, it's perfectly possible to implement what you described there by making a few buffers and doing CorticalActions.  Part of me thinks that might be the right way to do it, as it makes it more explicit what populations are being created.  So it's a matter of figuring out how explicit we want to be...

Seanny123:
So the reason I was interested in implementing it as just a bunch of
chained operations was that it avoided me creating another buffer, which
meant I used less neurons. However, this may have been misguided. Are
pass-through buffers ignored in simulation just like pass-through nodes?

Is there a reason this type of operation is unsupported?

The main reason is implementation -- it's annoying to implement all those
possibilities. Notice that dot(color, shape) also isn't supported.

That said, I think there's a design decision here. After all, it's
perfectly possible to implement what you described there by making a few
buffers and doing CorticalActions. Part of me thinks that might be the
right way to do it, as it makes it more explicit what populations are being
created. So it's a matter of figuring out how explicit we want to be...

—
Reply to this email directly or view it on GitHub
<https://github.com/nengo/nengo/issues/759#issuecomment-114683512>.

tcstewar:
> So the reason I was interested in implementing it as just a bunch of
> chained operations was that it avoided me creating another buffer, which
> meant I used less neurons

If you want it to be like a pass-through Node, then just set the spa.Buffer to direct=True.  That said, if people want these sorts of things, they could be implemented.  It is, however, tricky to handle all the cases, and I prefer a system that is clear about what it does and does not support, rather than one that supports more things but is unpredictable about what it does and does not support.  Mind you, I'm not sure we're clear yet about what we do and do not support, but that should be the goal....

[...]

PR #781, opened 2015-07-10
Make Nengo-specific Exceptions

tbekolay:
I was adding a helpful exception when trying to probe decoders or transform, which will be removed in #729, but in order to get that exception working nicely, it kind of ballooned into reworking the exceptions currently raised for validation errors. This should be less magical / more clear, but perhaps more verbose (that might be fixable though).

But yeah, this is also a work in progress, I'm just making this PR now because I'm going to refer to this in #729.

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -10,6 +10,7 @@ from nengo.builder.operator import (
 from nengo.builder.signal import Signal
 from nengo.connection import Connection
 from nengo.ensemble import Ensemble, Neurons
+from nengo.exceptions import BuildError
 from nengo.neurons import Direct
 from nengo.node import Node
 from nengo.utils.compat import is_iterable, itervalues
@@ -17,10 +18,6 @@ from nengo.utils.compat import is_iterable, itervalues
 
 BuiltConnection = collections.namedtuple(
     'BuiltConnection', ['eval_points', 'solver_info', 'weights'])
-
-
-class ZeroActivityError(RuntimeError):
-    pass
 
 
 def get_eval_points(model, conn, rng):
@@ -48,7 +45,7 @@ def build_linear_system(model, conn, rng):
     eval_points = get_eval_points(model, conn, rng)
     activities = get_activities(model, conn.pre_obj, eval_points)
     if np.count_nonzero(activities) == 0:
-        raise RuntimeError(
+        raise BuildError(
             "Building %s: 'activites' matrix is all zero for %s. "
             "This is because no evaluation points fall in the firing "
             "ranges of any neurons." % (conn, conn.pre_obj))
@@ -77,8 +74,8 @@ def build_decoders(model, conn, rng):
         decoders, solver_info = wrapped_solver(
             conn.solver, conn.pre_obj.neuron_type, gain, bias, x, targets,
             rng=rng, E=E)
-    except ZeroActivityError:
-        raise ZeroActivityError(
+    except BuildError:
+        raise BuildError(
             "Building %s: 'activities' matrix is all zero for %s. "
             "This is because no evaluation points fall in the firing "
             "ranges of any neurons." % (conn, conn.pre_obj))
@@ -90,7 +87,7 @@ def solve_for_decoders(
         solver, neuron_type, gain, bias, x, targets, rng, E=None):
     activities = neuron_type.rates(x, gain, bias)
     if np.count_nonzero(activities) == 0:
-        raise ZeroActivityError()
+        raise BuildError()
 
     if solver.weights:
         decoders, solver_info = solver(activities, targets, rng=rng, E=E)
@@ -108,7 +105,7 @@ def multiply(x, y):
     elif x.ndim == 2 and y.ndim == 2:
         return np.dot(x, y)
     else:
-        raise ValueError("Tensors not supported (x.ndim = %d, y.ndim = %d)"
+        raise BuildError("Tensors not supported (x.ndim = %d, y.ndim = %d)"
                          % (x.ndim, y.ndim))
 
 
@@ -134,13 +131,13 @@ def build_connection(model, conn):
         key = 'out' if is_pre else 'in'
 
         if target not in model.sig:
-            raise ValueError("Building %s: the '%s' object %s "
-                             "is not in the model, or has a size of zero."
+            raise BuildError("Building %s: the %r object %s is not in the "
+                             "model, or has a size of zero."
                              % (conn, 'pre' if is_pre else 'post', target))
         if key not in model.sig[target]:
-            raise ValueError("Error building %s: the '%s' object %s "
-                             "has a '%s' size of zero." %
-                             (conn, 'pre' if is_pre else 'post', target, key))
+            raise BuildError(
+                "Building %s: the %r object %s has a %r size of zero."
+                % (conn, 'pre' if is_pre else 'post', target, key))
 
         return model.sig[target][key]
 
@@ -223,7 +220,7 @@ def build_connection(model, conn):
             model.add_op(PreserveValue(model.sig[conn.post_obj]['encoders']))
         if 'decoders' in targets or 'weights' in targets:
             if weights.ndim < 2:
-                raise ValueError(
+                raise BuildError(
                     "'transform' must be a 2-dimensional array for learning")
             model.sig[conn]['weights'].readonly = False
             model.add_op(PreserveValue(model.sig[conn]['weights']))
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -10,14 +10,29 @@ from nengo.builder.operator import (
 from nengo.builder.signal import Signal
 from nengo.connection import Connection
 from nengo.ensemble import Ensemble, Neurons
-from nengo.exceptions import BuildError
+from nengo.exceptions import BuildError, ObsoleteError
 from nengo.neurons import Direct
 from nengo.node import Node
 from nengo.utils.compat import is_iterable, itervalues
 
 
-BuiltConnection = collections.namedtuple(
+BuiltConnection_ = collections.namedtuple(
     'BuiltConnection', ['eval_points', 'solver_info', 'weights'])
+
+
+class BuiltConnection(BuiltConnection_):
+    """Subclassing the namedtuple to provide better error messages."""
+    @property
+    def decoders(self):
+        raise ObsoleteError("decoders are now part of 'weights'. "
+                            "Access BuiltConnection.weights instead.",
+                            since="v2.1.0")
+
+    @property
+    def transform(self):
+        raise ObsoleteError("transform is now part of 'weights'. "
+                            "Access BuiltConnection.weights instead.",
+                            since="v2.1.0")
 
 
 def get_eval_points(model, conn, rng):
--- a/nengo/dists.py
+++ b/nengo/dists.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 import numpy as np
 
+from nengo.exceptions import ValidationError
 from nengo.params import (BoolParam, IntParam, NdarrayParam, NumberParam,
                           Parameter, Unconfigurable, FrozenObject)
 import nengo.utils.numpy as npext
@@ -51,20 +52,22 @@ class PDF(Distribution):
     p : vector_like (n,)
         Probabilities of the `x` points.
     """
-    x = NdarrayParam(shape='*')
-    p = NdarrayParam(shape='*')
+    x = NdarrayParam('x', shape='*')
+    p = NdarrayParam('p', shape='*')
 
     def __init__(self, x, p):
         super(PDF, self).__init__()
 
         psum = np.sum(p)
         if np.abs(psum - 1) > 1e-8:
-            raise ValueError("PDF must sum to one (sums to %f)" % psum)
+            raise ValidationError(
+                "PDF must sum to one (sums to %f)" % psum, attr='p', obj=self)
 
         self.x = x
         self.p = p
         if len(self.x) != len(self.p):
-            raise ValueError("`x` and `p` must be the same length")
+            raise ValidationError(
+                "`x` and `p` must be the same length", attr='p', obj=self)
 
         # make cumsum = [0] + cumsum, cdf = 0.5 * (cumsum[:-1] + cumsum[1:])
         cumsum = np.cumsum(p)
@@ -99,9 +102,9 @@ class Uniform(Distribution):
         If true, sample from a uniform distribution of integers. In this case,
         low and high should be integers.
     """
-    low = NumberParam()
-    high = NumberParam()
-    integer = BoolParam()
+    low = NumberParam('low')
+    high = NumberParam('high')
+    integer = BoolParam('integer')
 
     def __init__(self, low, high, integer=False):
         super(Uniform, self).__init__()
@@ -136,11 +139,11 @@ class Gaussian(Distribution):
 
     Raises
     ------
-    ValueError if ``std <= 0``
+    ValidationError if ``std <= 0``
 
     """
-    mean = NumberParam()
-    std = NumberParam(low=0, low_open=True)
+    mean = NumberParam('mean')
+    std = NumberParam('std', low=0, low_open=True)
 
     def __init__(self, mean, std):
         super(Gaussian, self).__init__()
@@ -211,7 +214,7 @@ class UniformHypersphere(Distribution):
         Default: False
 
     """
-    surface = BoolParam()
+    surface = BoolParam('surface')
 
     def __init__(self, surface=False):
         super(UniformHypersphere, self).__init__()
@@ -223,7 +226,7 @@ class UniformHypersphere(Distribution):
 
     def sample(self, n, d, rng=np.random):
         if d is None or d < 1:  # check this, since other dists allow d = None
-            raise ValueError("Dimensions must be a positive integer")
+            raise ValidationError("Dimensions must be a positive integer", 'd')
 
         samples = rng.randn(n, d)
         samples /= npext.norm(samples, axis=1, keepdims=True)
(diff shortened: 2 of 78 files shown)


tbekolay:
Note to self: Also add an `ObsoleteError` for getting the `'decoders'` in the `BuiltConnection`.

jgosmann:
> Note to self: Also add an ObsoleteError for getting the 'decoders' in the BuiltConnection.

Might `DeprecatedError` be a ebetter name? At least to me it seems that "deprecated" is more widely used than "obsolete".

hunse:
We discussed this somewhere, and @arvoelke pointed out that deprecated things are still usable, they'll just be removed in a future version (or otherwise made unusable in the future). He suggested "obsolete" to refer to these things that are unusable now.

hunse:
The [thesaurus suggests](http://www.thesaurus.com/browse/obsolete) AntediluvianError, HorseAndBuggyError, MothEatenError, or PasséError.

hunse:
The only other thing I can think of is RemovedError and RemovedParam (though it sounds a bit like we got rid of an error).

tbekolay:
Yeah, `Removed` is the only other real synonym for the reason @hunse pointed out, but Removed has too many other possible meanings, I think. Also, we already have the `ObsoleteParam` so if we want to rename the error, we have to rename the param too.

jgosmann:
Ah, I forgot about the distinction @arvoelke pointed out. With that `ObosoleteError` makes sense. :)

Seanny123:
I'd like an error and a warning for things that aren't biologically plausible. My motivation is that in SPA, there's a lot of operations that we don't allow that we cover with `NotImplemented` when really it should be `NeverWillBeImplemented` because it really isn't biologically plausible. 

For example, in `spa.Actions` doesn't parse every possible string you can throw at it because a lot of options would mean creating new networks, which isn't a direction we want to push the user in, as discussed in #759 .

tcstewar:
> I'd like an error and a warning for things that aren't biologically plausible. My motivation is that in SPA, there's a lot of operations that we don't allow that we cover with NotImplemented when really it should be NeverWillBeImplemented because it really isn't biologically plausible.
> 
> For example, in spa.Actions doesn't parse every possible string you can throw at it because a lot of options would mean creating new networks, which isn't a direction we want to push the user in, as discussed in #759 .

I think that'd be a SPA-specific exception rather than a Nengo-specific exception, since it's much more about SPA design philosophy (as much as that can be said to exist....)

tbekolay:
Note to self: add `SimulationClosedError` as per #859.

tbekolay:
This PR's ready for review. The main thing I don't like is the `ValidationError` in a4cdcc0190cce02e6ed326e7ac62 ... it's a better user experience, as it removes a bit of magic and gives informative messages in more situations. But for devs, it's kind of verbose / annoying. So, closer looks at that commit, and alternative suggestions would be appreciated!

hunse:
What don't you like about ValidationError? Just that it takes a lot of arguments, that the dev has to figure out? I think it's an improvement. It's really only core Nengo devs that will have to deal with it I think. People writing backends can avoid it, as can people writing Nengo extensions (e.g. new neruon types) since they can just use our parameters and throw ValueErrors if they need extra exceptions.

tbekolay:
The two things that irk me a bit about `ValidationError` are:

- Needing to access `__class__` a lot, which seems a bit clunky to me. But this could maybe be helped by accepting an instance and automatically doing `__class__` if it's not a class?
- The redundancy in specifying the same name twice for each param; e.g., `tau = NumberParam('tau', ...)`

But you're right, it's only Nengo devs that have to care about it, so I would say my second concern is no biggie. But I'll try out accepting an instance and introspecting a bit to do the `__class__` part automatically.

Also, what do you think about where `ValidationError` is raised? Often I don't really care to see the `validate` method, I mostly care about a few levels up the stack where I passed the wrong argument to the Nengo object. Should we reraise `ValidationError`s a few levels up the stack?

hunse:
> I'll try out accepting an instance and introspecting a bit to do the `__class__` part automatically.

Sounds good, I think that would be an improvement. I was trying to think of a case where that magic would be troublesome, but nothing is coming to me right now.

> Often I don't really care to see the `validate` method, I mostly care about a few levels up the stack where I passed the wrong argument to the Nengo object. Should we reraise `ValidationError`s a few levels up the stack?

I agree, you usually want to see it where the parameter is set. As long as the error message is more than clear enough for the user to fix it; if the message is unclear, then it can be nice to see the actual logic that's leading to the error.

How hard would it to be to reraise these things? If it's not too hard, I think it would be worth trying.

Overall, this looks good to me. I was thinking I should test things out and look at some of the error messages, and then I started wondering whether it's worth having a notebook that raises many/most of the errors we have in Nengo so it's easy to look at all the messages. We could organize it into sections based on file.

tbekolay:
> a notebook that raises many/most of the errors we have in Nengo so it's easy to look at all the messages

Yeah, sounds good. I think something like that's necessary to test the reraise logic anyhow so I'll give it a shot! Thinking it will be kind of like the strings notebook.

tbekolay:
OK, so I've been playing around with this a bit. It is possible to reraise specifically for validation errors for NengoObjects only, but it's not perfect. I would appreciate votes on which of the two possibilities are better from the user perspective!

#### Option 1: No reraising

In a bare Python interpreter

```pytb
>>> with nengo.Network():
...     ens = nengo.Ensemble(n_neurons=0, dimensions=1)
... 
  File "<stdin>", line 2, in <module>
  File "nengo/base.py", line 30, in __call__
    inst.__init__(*args, **kwargs)
  File "nengo/ensemble.py", line 89, in __init__
    self.n_neurons = n_neurons
  File "nengo/base.py", line 69, in __setattr__
    super(NengoObject, self).__setattr__(name, val)
  File "nengo/params.py", line 186, in __set__
    super(NumberParam, self).__set__(instance, value)
  File "nengo/params.py", line 80, in __set__
    self.validate(instance, value)
  File "nengo/params.py", line 217, in validate
    super(IntParam, self).validate(instance, num)
  File "nengo/params.py", line 200, in validate
    num))
nengo.exceptions.ValidationError: Ensemble.n_neurons: Value must be greater than or equal to 1 (got 0)
```

In IPython

```pytb
In [2]: with nengo.Network():
   ...:     ens = nengo.Ensemble(n_neurons=0, dimensions=1)
   ...:     
ValidationError                           Traceback (most recent call last)
<ipython-input-2-46a60ac68203> in <module>()
      1 with nengo.Network():
----> 2     ens = nengo.Ensemble(n_neurons=0, dimensions=1)
      3 

/home/tbekolay/Code/nengo/nengo/base.pyc in __call__(cls, *args, **kwargs)
     28         add_to_container = kwargs.pop('add_to_container', True)
     29         # Do the __init__ before adding in case __init__ errors out
---> 30         inst.__init__(*args, **kwargs)
     31         if add_to_container:
     32             Network.add(inst)

/home/tbekolay/Code/nengo/nengo/ensemble.pyc in __init__(self, n_neurons, dimensions, radius, encoders, intercepts, max_rates, eval_points, n_eval_points, neuron_type, gain, bias, noise, seed, label)
     87                  bias=Default, noise=Default, seed=Default, label=Default):
     88 
---> 89         self.n_neurons = n_neurons
     90         self.dimensions = dimensions
     91         self.radius = radius

/home/tbekolay/Code/nengo/nengo/base.pyc in __setattr__(self, name, val)
     67             val = Config.default(type(self), name)
     68         # try:
---> 69         super(NengoObject, self).__setattr__(name, val)
     70         # except ValidationError:
     71         #     exc_info = sys.exc_info()

/home/tbekolay/Code/nengo/nengo/params.pyc in __set__(self, instance, value)
    184         if is_array(value) and value.shape == ():
    185             value = value.item()  # convert scalar array to Python object
--> 186         super(NumberParam, self).__set__(instance, value)
    187 
    188     def validate(self, instance, num):

/home/tbekolay/Code/nengo/nengo/params.pyc in __set__(self, instance, value)
     78 
     79     def __set__(self, instance, value):
---> 80         self.validate(instance, value)
     81         self.data[instance] = value
     82 

/home/tbekolay/Code/nengo/nengo/params.pyc in validate(self, instance, num)
    215             raise ValidationError(instance.__class__, self.name,
    216                                   "Must be an integer; got '%s'" % num)
--> 217         super(IntParam, self).validate(instance, num)
    218 
    219 

/home/tbekolay/Code/nengo/nengo/params.pyc in validate(self, instance, num)
    198                         "" if self.low_open else "or equal to ",
    199                         self.low,
--> 200                         num))
    201             high_comp = 0 if self.high_open else 1
    202             if self.high is not None and compare(num, self.high) >= high_comp:

ValidationError: Ensemble.n_neurons: Value must be greater than or equal to 1 (got 0)
```

#### Option 2: Reraising

In a bare Python interpreter

```pytb
>>> with nengo.Network():
...     ens = nengo.Ensemble(n_neurons=0, dimensions=1)
... 
  File "<stdin>", line 2, in <module>
  File "nengo/base.py", line 30, in __call__
    inst.__init__(*args, **kwargs)
  File "nengo/ensemble.py", line 89, in __init__
    self.n_neurons = n_neurons
  File "nengo/base.py", line 73, in __setattr__
    reraise(exc_info[0], exc_info[1], None)
  File "<string>", line 2, in reraise
nengo.exceptions.ValidationError: Ensemble.n_neurons: Value must be greater than or equal to 1 (got 0)
```

In IPython

```pytb
In [2]: with nengo.Network():
   ...:     ens = nengo.Ensemble(n_neurons=0, dimensions=1)
   ...:     
ValidationError                           Traceback (most recent call last)
<ipython-input-2-46a60ac68203> in <module>()
      1 with nengo.Network():
----> 2     ens = nengo.Ensemble(n_neurons=0, dimensions=1)
      3 

/home/tbekolay/Code/nengo/nengo/base.pyc in __call__(cls, *args, **kwargs)
     28         add_to_container = kwargs.pop('add_to_container', True)
     29         # Do the __init__ before adding in case __init__ errors out
---> 30         inst.__init__(*args, **kwargs)
     31         if add_to_container:
     32             Network.add(inst)

/home/tbekolay/Code/nengo/nengo/ensemble.pyc in __init__(self, n_neurons, dimensions, radius, encoders, intercepts, max_rates, eval_points, n_eval_points, neuron_type, gain, bias, noise, seed, label)
     87                  bias=Default, noise=Default, seed=Default, label=Default):
     88 
---> 89         self.n_neurons = n_neurons
     90         self.dimensions = dimensions
     91         self.radius = radius

/home/tbekolay/Code/nengo/nengo/base.pyc in __setattr__(self, name, val)
     71             exc_info = sys.exc_info()
     72             logger.debug(traceback.format_tb(exc_info[2]))
---> 73             reraise(exc_info[0], exc_info[1], None)
     74 
     75     def __getstate__(self):

/home/tbekolay/Code/nengo/nengo/utils/compat.pyc in reraise(tp, value, tb)

ValidationError: Ensemble.n_neurons: Value must be greater than or equal to 1 (got 0)
```

----

As far as I can tell, there's no way to get rid of reporting that the exception is reraised (ideally the `self.n_neurons = n_neurons` line would be where the traceback starts) but since it's much shorter, I'm in favor of option 2. Note that I log the full traceback with `logger.debug` so that if someone's editing the `Parameter` code it's easier to debug.

hunse:
You're right, neither one seems perfect. But I agree that option 2 seems better because of the shorter traces. 

Would it be possible to have this as a config option? This seems like something that people might want to customize, defaulting to option 2.

tbekolay:
Yeah, it could be configurable. The `__setattr__` code will be kind of gross since it'll be a try/except in an if statement, but it's probably the best from the user perspective.

jgosmann:
slight preference for option 2

tbekolay:
OK, this is finally ready for re-review. Quick summary of changes:

1. I went with @hunse's suggestion and added an RC option; see 82a19456df2100b23f for that (I can split it into a separate commit if desired).
2. When doing the RC option, I realized that we weren't actually disabling the cache in unit tests. 3e8739b85870ce7 fixes that.
3. I simplified `ValidationError` a fair bit (see 679947ef6e4d65582d) by changing the order of arguments and making it possible to pass an instance or a class. The error raising code looks a lot better with this, IMO (mostly because the error message goes first).
4. I added a few more exception types so that almost every exception that we raise manually is a `NengoException` subclass. The only non-NengoExceptions raised are `NotImplementedError`s, and a few random errors in utility modules that we copied verbatim from elsewhere (these exceptions are pretty unlikely to be raised, and would be annoying to change if we update our version of these modules). If anyone sees any other examples of raising non-NengoException exceptions, that's a bug.
5. Finally, I added the notebook that @hunse suggested showing examples of all of the exceptions. A rendered version can be seen [here](http://nbviewer.jupyter.org/urls/dl.dropboxusercontent.com/u/6182929/exceptions.ipynb).

I think I'm pretty happy with everything, except for some of the changes I snuck into c23bb4ee00f75d6. If anyone has a better idea for where to put those, I'm all ears! In any case, tagging @hunse and @drasmuss for reviews.

hunse:
Okay, added a few comments. For the most part, it looks great!

tbekolay:
OK, I think I addressed all the comments except https://github.com/nengo/nengo/commit/0e1fbabf46f3c0b6d4fa739b595c39584c523975#commitcomment-16629707 (pending discussion) and https://github.com/nengo/nengo/commit/0344dba6ff31bce294c47b9abc4868d731b27e20#commitcomment-16629522.

nengo/dists.py
```diff
@@ -99,9 +102,9 @@ class Uniform(Distribution):
         If true, sample from a uniform distribution of integers. In this case,
         low and high should be integers.
     """
-    low = NumberParam()
-    high = NumberParam()
-    integer = BoolParam()
+    low = NumberParam('low')
```
drasmuss: This `name = Class('name')` theme makes me kind of :grimacing:.  It's exactly the kind of thing we were trying to avoid when we made `label` optional for other Nengo objects.  I can see why we want to make names mandatory here though, since `config.set_param(Param(...))` needs to have a name.  I don't have a great solution, just complaining.

The only thing that comes to mind is to go back to the old syntax of `set_param(name, parameter)`, (so we're sure to have a name), and then have `set_param` do `parameter.label = name`.  It's a bit awkward though, because it's not obvious what to do if they already set a label for the parameter.

drasmuss: Or I guess another option would be to just throw an error if they pass a parameter to `config.set_param` that doesn't have a label set.

tbekolay: Yeah, I don't like it either; there's a bit of discussion about it in the PR comments as I would rather avoid it, though @hunse is OK with it.

To me the main issue is not with `set_param`; for setting these class attributes, I believe that in order to avoid duplication we'd have to do some metaclass stuff since we would essentially be adding some custom stuff to `type.__setattr__` or some other magic method similar to that. Maybe it wouldn't be that bad since it would be part of the base `Distribution` or whatever, but if people try to use `Param`s in their own custom classes it'll be pretty surprising...

nengo/builder/connection.py
```diff
@@ -48,7 +60,7 @@ def build_linear_system(model, conn, rng):
     eval_points = get_eval_points(model, conn, rng)
     activities = get_activities(model, conn.pre_obj, eval_points)
     if np.count_nonzero(activities) == 0:
-        raise RuntimeError(
+        raise ZeroActivityError(
```
drasmuss: Could we just make this a `BuildError` (and then get rid of `ZeroActivityError`)?  I think the only reason we had `ZeroActivtyError` before was so that we could provide a more specific error message for that particular case (without catching all `RunTimeErrors`, but now that's already happening here thanks to the refactoring.

tbekolay: The thought crossed my mind! Since it's crossed 2 minds now I'll do it.

nengo/spa/pointer.py
```diff
@@ -76,14 +80,15 @@ def __isub__(self, other):
     def __mul__(self, other):
         """Multiplication of two SemanticPointers is circular convolution.
 
-        If mutliplied by a scaler, we do normal multiplication.
+        If mutliplied by a scalar, we do normal multiplication.
```
drasmuss: I like that you fixed `scaler`, but left `mutliplied` :japanese_ogre:

tbekolay: Haha shit I totally missed that one ;) will change *I guess*

drasmuss:
That's all the comments from me, looks good.

nengo/dists.py
jgosmann: I agree, that this is not the nicest syntax, but I think it is ok in this place:

* These parameter definitions are not exactly what most users will do on a regular basis.
* Both `name`s are in the same line. So it is hopefully easy to spot inconsistent names.
* There are other places in Python that follow the same style, namely `namedtuple` or dynamically defining a class with `type`.

hunse: +1 to @jgosmann. It's not ideal, but it's not near as bad as the `label` for NengoObjects situation, since that concerned every model ever, and this only concerns advanced users and devs. If someone can come up with a better solution in the future, by all means, but not worth holding up this PR for.

nengo/spa/pointer.py
hunse: What do you call it when two dogs of different breeds have babies? Mutliplication!

nengo/dists.py
drasmuss: Agreed, something for a future PR.

tbekolay:
OK just pushed some new stuff to address the typo and to remove `ZeroActivityError`. I think for the time being I'm going to leave `NetworkContextError` and `SpaModuleError`, but we can always come back to it in another issue / PR.

drasmuss:
Looks good to me.  I'll wait for a +1 from @hunse, then I can merge.  Do you have plans in mind for the history, or want me to just squash all the fixups?

tbekolay:
I think I would favor just squashing the fixups, as each commit has a pretty clear commit message that should be easy to ignore when looking through the commit log for non-exception stuff. But if other people think it's too much we could squash the exceptions from BuildError to SpaModuleError into one big commit (so 5 commits total, something like 5373328 bc633eb ba93b8d b3fcaab 51c149f). Reviewer's choice ;)

hunse:
I'm fine keeping things separate. My only caveat is I'd like the changelog entry part in e155633 to appear at the end, because it's describing the changes of all the commits.

tbekolay:
I'll split that out now and move it to the end.

tbekolay:
Done; marked it as a FIXUP on d470d344b0ab3bb1f6 so should probably add a note about that to its commit message.

PR #782, opened 2015-07-10
Compare network now returns a scalar

### LINKED ISSUE #775: spa.Compare output transform
The `spa.Compare` code has a [really weird output transform](https://github.com/nengo/nengo/blob/2ebda418e91b17f5c30f3ed66954b8f34d9670c1/nengo/spa/compare.py#L72).

Why does it do `vocab.parse("YES").v`? What's the motivation behind this transform anyways?

Seanny123:
To fix #775 :
- Make the `spa.Compare` network return a scalar. 
- Allow multiple scalars to be added together for BasalGanglia inputs

### DIFF
--- a/nengo/spa/basalganglia.py
+++ b/nengo/spa/basalganglia.py
@@ -71,6 +71,8 @@ class BasalGanglia(Module):
                         # enforced in DotProduct constructor
                         assert isinstance(c.item2, Source)
                         self.add_dot_input(i, c.item2, c.item1, c.scale)
+                elif isinstance(c, Source):
+                    self.add_scalar_input(i, c)
                 elif is_number(c):
                     self.add_bias_input(i, c)
                 else:
@@ -141,3 +143,34 @@ class BasalGanglia(Module):
         with self.spa:
             nengo.Connection(output, self.input[index:index+1],
                              transform=transform, synapse=self.input_synapse)
+
+    def add_scalar_input(self, index, source):
+        """ Add a scalar input that will vary over time.
+
+        This would be used for such a thing as the ouput of the Compare module.
+
+        Parameters
+        ----------
+        index : int
+            the index of the action
+        source : Source
+            the module output to read from
+        scale : float
+            a scaling factor to be applied to the result
+        """
+        output, _ = self.spa.get_module_output(source.name)
+        if output.size_out != 1:
+            raise NotImplementedError("Only sources with a dimension"
+                                      "of 1 can be scalar inputs")
+
+        try:
+            scale = float(source.transform.symbol)
+        except ValueError:
+            raise ValueError("Transform can only be a scalar value"
+                             " the value %s is invalid"
+                             % source.transform.symbol)
+
+        with self.spa:
+            nengo.Connection(output, self.input[index:index+1],
+                             transform=scale,
+                             synapse=self.input_synapse)
--- a/nengo/spa/compare.py
+++ b/nengo/spa/compare.py
@@ -3,7 +3,6 @@ import numpy as np
 import nengo
 from nengo.dists import Choice
 from nengo.spa.module import Module
-from nengo.utils.compat import range
 
 
 class Compare(Module):
@@ -34,6 +33,7 @@ class Compare(Module):
             vocab = dimensions
 
         self.output_scaling = output_scaling
+        self.dimensions = dimensions
 
         with self:
             self.compare = nengo.networks.EnsembleArray(
@@ -45,32 +45,23 @@ class Compare(Module):
 
             self.inputA = nengo.Node(size_in=dimensions, label='inputA')
             self.inputB = nengo.Node(size_in=dimensions, label='inputB')
-            self.output = nengo.Node(size_in=dimensions, label='output')
+            self.output = nengo.Node(size_in=1, label='output')
 
         self.inputs = dict(A=(self.inputA, vocab), B=(self.inputB, vocab))
-        self.outputs = dict(default=(self.output, vocab))
-
-        t1 = np.zeros((dimensions * 2, dimensions), dtype='float')
-        t2 = np.zeros((dimensions * 2, dimensions), dtype='float')
-        for i in range(dimensions):
-            t1[i * 2, i] = 1.0
-            t2[i * 2 + 1, i] = 1.0
+        self.outputs = dict(default=(self.output, None))
 
         with self:
-            nengo.Connection(self.inputA, self.compare.input, transform=t1)
-            nengo.Connection(self.inputB, self.compare.input, transform=t2)
-
-        def multiply(x):
-            return [x[0] * x[1]]
-        self.compare.add_output('product', function=multiply)
+            nengo.Connection(self.inputA,
+                             self.compare.input[::2], synapse=None)
+            nengo.Connection(self.inputB,
+                             self.compare.input[1::2], synapse=None)
+            self.compare.add_output('product', lambda x: x[0] * x[1])
 
     def on_add(self, spa):
         Module.on_add(self, spa)
 
-        vocab = self.outputs['default'][1]
-
-        transform = np.array([vocab.parse('YES').v] * vocab.dimensions)
-
         with self:
-            nengo.Connection(self.compare.product, self.output,
-                             transform=transform.T * self.output_scaling)
+            nengo.Connection(self.compare.product,
+                             self.output,
+                             transform=self.output_scaling *
+                             np.ones((1, self.dimensions)))
--- a/nengo/spa/tests/test_basalganglia.py
+++ b/nengo/spa/tests/test_basalganglia.py
@@ -11,13 +11,17 @@ def test_basal_ganglia(Simulator, seed, plt):
     with model:
         model.vision = spa.Buffer(dimensions=16)
         model.motor = spa.Buffer(dimensions=16)
+        model.compare = spa.Compare(dimensions=16)
 
+        # test all acceptable condition formats
         actions = spa.Actions(
             '0.5 --> motor=A',
             'dot(vision,CAT) --> motor=B',
             'dot(vision*CAT,DOG) --> motor=C',
             '2*dot(vision,CAT*0.5) --> motor=D',
             'dot(vision,CAT)+0.5-dot(vision,CAT) --> motor=E',
+            'dot(vision,PARROT) + compare --> motor=F',
+            '0.5*dot(vision,MOUSE) + 0.5*compare --> motor=G',
         )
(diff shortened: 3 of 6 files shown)


hunse:
Could a regular user of SPA---@tcstewar, @xchoo, anyone else (@jgosmann maybe?)---please review this?

nengo/spa/tests/test_basalganglia.py
```diff
@@ -11,13 +11,17 @@ def test_basal_ganglia(Simulator, seed, plt):
...
 
+        # test all acceptable condition formats
         actions = spa.Actions(
             '0.5 --> motor=A',
             'dot(vision,CAT) --> motor=B',
             'dot(vision*CAT,DOG) --> motor=C',
             '2*dot(vision,CAT*0.5) --> motor=D',
             'dot(vision,CAT)+0.5-dot(vision,CAT) --> motor=E',
+            'dot(vision,PARROT) + compare --> motor=F',
+            'dot(vision,MOUSE) + 0.5*compare --> motor=G',
```
xchoo: I think you want ```0.5 * dot(vision, MOUSE) + 0.5 * compare``` here.

Seanny123: Fixed.

xchoo:
I did a quick pass. Made a minor comment (made a separate issue for my more major comment #798).   @tcstewar should also take a pass at it.

nengo/spa/basalganglia.py
```diff
@@ -141,3 +143,36 @@ def add_dot_input(self, index, source, symbol, scale):
...
+            the index of the action
+        source : Source
+            the module output to read from
+        symbol : Source
+            the semantic pointer to compute the dot product with
+        scale : float
+            a scaling factor to be applied to the result
+        """
+        output, _ = self.spa.get_module_output(source.name)
+        if(output.size_out != 1):
```
tcstewar: This should be ```if output.size_out != 1:```

Seanny123: What's the rule of thumb for adding parentheses to if-statements? It's not mentioned in the PEP8 and my personal preference is to just add them everywhere.

nengo/spa/basalganglia.py
```diff
@@ -141,3 +143,36 @@ def add_dot_input(self, index, source, symbol, scale):
...
+            raise NotImplementedError("Only sources with a dimension"
+                                      "of 1 can be scalar inputs")
+
+        try:
+            scale = float(eval(source.transform.symbol))
+        except ValueError:
```
tcstewar: How can this ValueError ever occur?  If I try parsing a rule like ```compare*A --> whatever=something``` then I get a ```NameError``` rather than a ```ValueError```.  Should this be NameError instead?  Or is this a Python 2 versus Python 3 thing (I'm on Python 2)?

nengo/spa/basalganglia.py
xchoo: I am okay with having parentheses. If I do add them though, I leave a space between the ```if`` and the rest of the conditional.

nengo/spa/basalganglia.py
```diff
@@ -141,3 +143,36 @@ def add_dot_input(self, index, source, symbol, scale):
...
+            raise NotImplementedError("Transform can only be a scalar value"
+                                      "the value %s is invalid"
```
tcstewar: I'm not sure this should be a NotImplementedError.  Seems more like a ValueError to me.

Also, there should be a space at the end of one string or the beginning of the other (otherwise the error message contains "valuethe"....)

nengo/spa/basalganglia.py
```diff
@@ -141,3 +143,36 @@ def add_dot_input(self, index, source, symbol, scale):
...
+                                      % source.transform.symbol)
+
+        with self.spa:
+            nengo.Connection(output, self.input[index:index+1],
```
tcstewar: spacing around the ```+```

nengo/spa/compare.py
```diff
@@ -45,32 +45,23 @@ def __init__(self, dimensions, vocab=None, neurons_per_multiply=200,
...
-            nengo.Connection(self.inputA, self.compare.input, transform=t1)
-            nengo.Connection(self.inputB, self.compare.input, transform=t2)
-
-        def multiply(x):
-            return [x[0] * x[1]]
-        self.compare.add_output('product', function=multiply)
+            nengo.Connection(self.inputA,
+                             self.compare.input[::2], synapse=None)
+            nengo.Connection(self.inputB,
+                             self.compare.input[1::2], synapse=None)
```
tcstewar: Nice.  Glad to get rid of those weird transforms..... :)

nengo/spa/basalganglia.py
xchoo: Perhaps another test would be better? Like ```isnumeric()```, or a regex?

xchoo: Also, I think for the ```ValueError``` to work, you'll want ```float(source.transform.symbol)```?

nengo/spa/basalganglia.py
tcstewar: I'm not okay with having parentheses, unless they're needed (for things like having an expression go across multiple lines).  As far as I know we only use them in the nengo codebase when we need to go across multiple lines.   Anyone know if I am wrong on that?

nengo/spa/basalganglia.py
Seanny123: Currently investigating. Gimme a sec.

tcstewar:
Nice.  Other than the little things I commented on above, this looks nice to me, and could really simplify a lot of code.  Thank you for doing this!

nengo/spa/basalganglia.py
tcstewar: > Perhaps another test would be better? Like a regex?

Please not a regex!  ;)  I much prefer this sort of better-to-beg-forgiveness-than-ask-permission approach....  :)

nengo/spa/basalganglia.py
xchoo: Note: That will not be consistent with the rest of the code. Should all of them be changed?

nengo/spa/basalganglia.py
xchoo: Haha. I'm fine either way. :smile:

nengo/spa/basalganglia.py
Seanny123: A quick Google gave me nothing and `flake8` didn't set off any alarms, so I guess it's personal? I'm fine with noting it down on our personal style guide or even putting it into our flake8 checks.

xchoo: Yeah. Parentheses and ```if``` statements don't work so well together. I do use them if the logic gets complex. Also, you can't use them to split lines because:
```
if (cond1 or
    cond2):
    do_things = True
```
Will throw a flake8 error (all the tabs are aligned).

nengo/spa/basalganglia.py
tcstewar: Gah, never mind...  Hmm, weird, I thought pep8 enforced that spacing, but it doesn't look like it does...  Basically, I didn't notice that the other code was also bad this way.  It looks like it isn't technically required for pep8, but in the rest of the codebase we are putting spaces around operators, so we should probably be consistent here.  My instinct is that it doesn't matter and leave it how it is, but people have yelled at me for my coding style instincts in the past....  ;)

xchoo: I'll add it to issue #798.

nengo/spa/basalganglia.py
tcstewar: > A quick Google gave me nothing and flake8 didn't set off any alarms, so I guess it's personal? I'm fine with noting it down on our personal style guide or even putting it into our flake8 checks.

Hmm, all of the single-line python if statements I can find (in nengo or in other places) do not use parentheses.  Do you have examples where parentheses are used?  I'm curious what they are...

hunse: @tcstewar is right: no parentheses unless you need them. (The only exception I can think of is `flag = (a == b)`; in that case I often use parentheses to make it easier to read.)

nengo/spa/basalganglia.py
hunse: I kinda thought that was a thing too, but pep8 [actually suggests](https://www.python.org/dev/peps/pep-0008/#other-recommendations) using spaces to group operators by order of execution.

tcstewar: > I kinda thought that was a thing too, but pep8 actually suggests using spaces to group operators by order of execution. In a case like this, I'm fine either way. Whichever seems more readable, and I think what's there is quite readable.

Sweet!  I'm very glad to know that...  there's been a number of times where I've wanted to have some operators with spaces and some without (```a*b + c*d```) and I'd been forcing myself to not do that...   Good to know at least some of my style instincts are okay....  ;)

nengo/spa/compare.py
```diff
@@ -45,32 +45,23 @@ def __init__(self, dimensions, vocab=None, neurons_per_multiply=200,
 
             self.inputA = nengo.Node(size_in=dimensions, label='inputA')
             self.inputB = nengo.Node(size_in=dimensions, label='inputB')
-            self.output = nengo.Node(size_in=dimensions, label='output')
+            self.output = nengo.Node(size_in=1, label='output')
 
         self.inputs = dict(A=(self.inputA, vocab), B=(self.inputB, vocab))
         self.outputs = dict(default=(self.output, vocab))
```
tcstewar: Hmm, this output shouldn't have a vocab associated with it.  I think the right thing to do would be ```self.outputs = dict(default=(self.output, None))``` but I'm not sure if that breaks anything else....

tbekolay:
Ditto here, worth putting in the changelog? At first glance it seems a bit esoteric to bother mentioning...

jgosmann:
Not sure, I was annoyed a lot by getting back some weird vector from compare.

Seanny123:
What's the heuristic as to what to put into the changelog and what to leave out? This does technically make an API change.

jgosmann:
Made a PR #804 with changelog entries added.

tcstewar:
> Ditto here, worth putting in the changelog? At first glance it seems a bit esoteric to bother mentioning...

Part of me thinks this the sort of thing that should go in the changelog, since it is an API change (and one that's not backwards-compatible).  However, I'm pretty sure no one has used ```spa.Compare``` other than the people involved in this PR, so I'm not that worried about it....  (that's why I didn't push for doing a backwards-compatible or deprication approach to this change).

Issue #778, opened 2015-07-09
Combining small ensembles within the same layer


arvoelke:
In certain cases, we might repeat lots of small ensembles in a single "layer" (e.g. the vision network of @hunse, or my texture classifier), and this turns out to be very inefficient because lots of small numpy vector operations are way more expensive than their grouped equivalent.

See, for example:

```
import numpy as np

import nengo

def model_slow(n_ensembles, neurons_per_ensemble):
    with nengo.Network() as model:
        stim = nengo.Node(output=[1])
        output = nengo.Node(size_in=1)
        for i in range(n_ensembles):
            ens = nengo.Ensemble(neurons_per_ensemble, 1)
            nengo.Connection(stim, ens)
            nengo.Connection(ens, output, transform=1./n_ensembles)
        return model

def model_fast(n_ensembles, neurons_per_ensemble):
    with nengo.Network() as model:
        stim = nengo.Node(output=[1])
        output = nengo.Node(size_in=1)
        ens = nengo.Ensemble(n_ensembles*neurons_per_ensemble, 1)
        nengo.Connection(stim, ens)
        nengo.Connection(ens, output)
        return model


from timeit import Timer

n_ensembles = 1000
neurons_per_ensemble = 5
duration = 0.1
trials = 20

def bootstrap(method, n_ensembles, neurons_per_ensemble, duration):
    sim = nengo.Simulator(method(n_ensembles, neurons_per_ensemble))
    return lambda: sim.run(duration)    

for method in model_slow, model_fast:
    t = Timer(bootstrap(method, n_ensembles, neurons_per_ensemble, duration))
    print t.timeit(trials)
```

In this (rather dramatic) example, there is a 200x performance gain by grouping together all of the ensembles:

```
130.49983397
0.644406616342
```

It's also more accurate whenever the sub-ensembles are encoding correlated (or, in this case, the exact same) information, because the decoders should be solved for jointly (as a brief aside: this is the case if you use a `Product` network and set either `A` or `B` to be a single scalar [i.e. scaling a vector]).

In general this sort of optimization isn't always so easy to do (either manually or automatically).
- Need to know when not to use the `EnsembleArray`
- Can be tricky to set different distributions on the different sub-ensembles
- May want different synapses on different sub-ensembles (this one is the killer for me, while it should be theoretically possible to group different synapse models together into a more efficient LTI)
- Or just may not be obvious that an optimization could be done when the code is scattered throughout a large-scale model
- ...?

Also related to #738.

[...]

Issue #738, opened 2015-06-13
Optimizing away repeated decoders and/or sparse transforms


arvoelke:
I have ensembles `A` and `B`, which are both high dimension. 

Between them I create a few thousand connections, all from a single dimension in `A` to a single dimension in `B`. Many of these connections are coming from the same dimension in `A`.

`A` cannot be an ensemble array. And the connections all have different synapses, so the connections cannot be combined.

This is very slow (it wouldn't even build after an hour), because each connection is recomputing the same decoders, and using a large transform matrix with a single 1.

It was possible to speed up the build process ~100x, simply by creating an array of scalar nodes between `A` and `B`, and channelling each dimension of `A` to each "proxy" node. This reduces the number of decoder optimizations to the dimensionality of `A`, and the transforms to `B` become column vectors instead of matrices.

The same process could be repeated for `B`, to convert all the column vectors into scalars, again with the cost of an extra matrix. This does not help for my case though, since no connections are to the same dimensions in `B`, and so there are no repeated column vectors to consolidate at a proxy.

Automating these sorts of optimization procedures will of course be challenging. But the benefits are potentially enormous in some cases.

tcstewar:
Weird...  are we really recomputing ```pinv(dot(A.T, A))``` for every connection?  I thought that was kept in the cache....  or are we only caching decoders now?  In Nengo 1.4 we only computed that inverse once per ensemble, cached it, and then every decoder was just computed with just a matrix multiply from that (I don't think we even bothered caching the decoders).

Maybe the right approach is to set up some hooks into the cache system so that the solvers can make use of it as appropriate?

arvoelke:
Part of it could be that each connection had a different scalar on the transform. Not sure if decoder caching takes that into account, or if it's pre-multiplied.

Not sure if these sorts of optimizations can be solved at the decoder caching level, though. Whether this sort of consolidation is advantageous requires global knowledge of whether there will be connections that use similar decoders that can essentially share the same signal (both for building [memory and number of optimizations], and for simulating [shape and number of multiplies]).

tcstewar:
> Not sure if these sorts of optimizations can be solved at the decoder caching level, though. Whether this sort of consolidation is advantageous requires global knowledge of whether there will be connections that use similar decoders that can essentially share the same signal (both for building [memory and number of optimizations], and for simulating [shape and number of multiplies]).

[...]

PR #788, opened 2015-07-13
Heterogeneous synapses

arvoelke:
This is my temporary solution for #778. This also makes #738 a non-issue for me.

As far as I can see, the only thing actually stopping people from manually optimizing small ensembles together themselves, is the fact that one might need different synapses on different neurons. The problem is that, currently, in order to use different synapses, you need a different connection for each one. This results in lots of small numpy operations, which is relatively inefficient.

To solve this, I wrote this callable class, `HeteroSynapse`, which applies a list of filters to its input dimensions. To implement this efficiently, all of the filters are combined together into a single state-space LTI system. The resulting matrices/operations scale with `n*k` where `n` is the number of synapses, and `k` is the order of each synapse (but you can mix various orders if that suits your fancy). 

Even in the case where `n = 1` (i.e. a standard Nengo model), it is slightly more efficient than the reference implementation, since it doesn't have to iterate through the transfer function in a Python `for` loop, and avoids use of `deques` which have to constantly allocate/deallocate memory from non-contiguous blocks. For this (contrived) example with a single high-order synapse, it is more than twice as fast as the reference implementation, despite using more connections:

    import numpy as np
    import nengo
    from nengo.synapses import LinearFilter

    n_neurons = 100
    dims = 100
    order = 100
    p = np.zeros(order)
    p[0] = 1
    synapse = LinearFilter(p, p)

    def reference(n_neurons, dims, synapse):
            stim = nengo.Node(output=np.zeros(dims))
            x = nengo.Ensemble(n_neurons, dims)
            nengo.Connection(stim, x, synapse=synapse)
            return model

    def custom(n_neurons, dims, synapse):
            stim = nengo.Node(output=np.zeros(dims))
            syn = nengo.Node(size_in=dims, output=HeteroSynapse(synapse, dt=0.001))
            x = nengo.Ensemble(n_neurons, dims)
            nengo.Connection(stim, syn, synapse=None)
            nengo.Connection(syn, x, synapse=None)
            return model

    from timeit import Timer

    r = []
    for b in (reference, custom):
        sim = nengo.Simulator(b(n_neurons, dims, synapse))
        timer = Timer(lambda: sim.run(0.01, progress_bar=False))
        r.append(timer.timeit(50))

    print r[0] / r[1]

For cases where `n > 1`, it is an order of magnitude faster than making a separate connection for each synapse (even if you use the proxy "trick" from #738).

The included notebook verifies that it does the right thing. I'm currently using this to do:
 - More robust dynamics / temporal representation
 - Feature extraction in the frequency domain for classification of dynamic stimuli

Looking to gather opinions on the future of this. We could:
 - Integrate this into the builder?
 - Release it as an advanced example, as I've done here?
 - Create some utils for using this effortlessly within models?


arvoelke:
Moved to an [external library](https://github.com/arvoelke/nengolib).

PR #791, opened 2015-07-16
Exception when module isn't added to spa

### LINKED ISSUE #730: SPA network fails and it is non-obvious why
It took me a while to figure out why the following network wasn't working:
```python
model = spa.SPA(seed=84)
with model:
    input_buf = spa.Buffer(d, vocab=vocab)
    spa.Input(input_buf='B')
    
    p_input_buf = nengo.Probe(input_buf.state.output, synapse=0.01)
```

Turns out that `spa.Input` has to be assigned to something and that almost everything has to be prefixed by `model.`. Working version:

```python
model = spa.SPA(seed=84)
with model:
    model.input_buf = spa.Buffer(d, vocab=vocab)
    model.input_node = spa.Input(input_buf='B')#, current_buf=lambda t: 'A' if t < 0.3 else '0')
    
    p_input_buf = nengo.Probe(model.input_buf.state.output, synapse=0.01)
```

In non-SPA examples we don't do the prefixing with `model.` which makes it non-obvious why it is suddenly needed. That `spa.Input` has to be assigned to someting (even though it is never referred to afterwards) might also be not obvious.

Seanny123:
Fixes #730 by raising an exception when a module is added to a network, but not as an attribute

### DIFF
--- a/nengo/spa/input.py
+++ b/nengo/spa/input.py
@@ -29,6 +29,7 @@ class Input(Module):
         Input(vision=input1, task='X')
 
     """
+
     def __init__(self, label=None, seed=None, add_to_container=None, **kwargs):
         super(Input, self).__init__(label, seed, add_to_container)
         self.kwargs = kwargs
--- a/nengo/spa/spa.py
+++ b/nengo/spa/spa.py
@@ -48,9 +48,7 @@ class SPA(nengo.Network):
 
     These names can be used by special Modules that are aware of these
     names.  As an example, the Cortical module allows you to form connections
-    between these modules in ways that are aware of semantic pointers.
-
-    First, the example with a ``with`` statement::
+    between these modules in ways that are aware of semantic pointers::
 
         with example:
             example.a = spa.Buffer(dimensions=8)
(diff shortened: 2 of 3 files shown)


nengo/spa/input.py
```diff
@@ -31,13 +31,13 @@ def input1(t):
     """
     def __init__(self, label=None, seed=None, add_to_container=None, **kwargs):
         super(Input, self).__init__(label, seed, add_to_container)
+        print("HELLO")
```
jgosmann: Remove debug code ...

nengo/spa/tests/test_spa.py
```diff
@@ -0,0 +1,41 @@
...
+    # build a normal model that shouldn't raise a warning
+    with model:
+        model.buf = spa.Buffer(d)
+        model.input_node = spa.Input(buf='B')
+        # make sure errors aren't fired for non-spa modules
+        prod = nengo.networks.Product(10, 2)
+
+    with pytest.raises(ValueError):
+        model = spa.SPA(seed=seed)
+        # build a model that should raise a warning because no variable
```
jgosmann: These comments should say 'error' which different from a 'warning'.

jgosmann: Or do we want to change the error to a warning?

nengo/spa/spa.py
```diff
@@ -86,6 +133,20 @@ def __setattr__(self, key, value):
 
             value.on_add(self)
 
+    def __exit__(self, dummy_exc_type, dummy_exc_value, dummy_tb):
+        super(SPA, self).__exit__(dummy_exc_type, dummy_exc_value, dummy_tb)
+        module_list = self._modules.values()
```
jgosmann: By changing this to
```python
module_list = set(self._module.values())
```
the code below should become amortized linear. (Maybe it is even better to use `frozenset` instead of `set` because the set won't be changed?)

nengo/spa/tests/test_spa.py
tcstewar: I'd keep it as an error rather than a warning....

nengo/spa/spa.py
```diff
@@ -86,6 +133,20 @@ def __setattr__(self, key, value):
...
+        for net in self.networks:
+            print("net: %s" % (net))
+            # Since there are no attributes to distinguish what's been added
+            # and what hasn't, we have to ask the network
+            # Note: This is going to run O(n^2) which is a bit scary
+            # maybe there's some way to optomize more easily,
+            # to make sure Python is doing binary search?
+            if isinstance(net, Module) and not(net in module_list):
```
jgosmann:
```python
not(net in module_list)
```
can be expressed more readable as
```python
net not in module_list
```

nengo/spa/spa.py
```diff
@@ -86,6 +133,20 @@ def __setattr__(self, key, value):
 
             value.on_add(self)
 
+    def __exit__(self, dummy_exc_type, dummy_exc_value, dummy_tb):
+        super(SPA, self).__exit__(dummy_exc_type, dummy_exc_value, dummy_tb)
+        module_list = self._modules.values()
+        for net in self.networks:
+            print("net: %s" % (net))
```
jgosmann: more debug code to remove

Seanny123:
Thanks for the review Jan. I made the changes you suggested. Let me know if it's okay now and I'll squash it and mark it for merge.

BTW, did @tbekolay ever tell you that you're allowed to merge things into this repository now?

jgosmann:
> BTW, did @tbekolay ever tell you that you're allowed to merge things into this repository now?

I don't remember it?

One more thing: There is a bunch of code in the docstring. [That should probably formatted as code (so that it looks nice in the HTML documentation).](http://sphinx-doc.org/rest.html#source-code)

Seanny123:
Fixed the docstrings.

tbekolay:
Is there anything in here worth putting in the changelog?

jgosmann:
Was just about to ask the same question :P

jgosmann:
Made a PR #804 with changelog entries added.

Issue #742, opened 2015-06-16
InputGatedMemory should warn about auto-setting ptsc


Seanny123:
As discussed with Trevor and Xuan, a default post synaptic time constant is set when mem config is passed in that is different than when no mem config is passed in. The user should be warned when passing in a mem config without the post synaptic time constant set, since this may unexpectedly defy the expectations of the user, since the memory will stop acting as a memory. We also shouldn't over-ride the config of the user if no post synaptic time constant is set.

This decision is similar to the warning that occurs in the basal ganglia.

Seanny123:
@tbekolay @xchoo I actually can't find where that warning happens in the basal ganglia? Was that actually an old version?

xchoo:
Line 56-63 of ```actionselection.py``` in ```nengo\networks```

Seanny123:
I don't see how that triggers a warning. I was looking for another example of a warning. If there is no example of a warning, should this be opened to wider discussion before I put in new warnings?

tbekolay:
You're right; I think originally this was a warning, but then I changed it such that, if the user did not set a synapse, I temporarily set a synapse and then after the function is done, I unset the synapse with `del`. If you search in that file for all instances of the string `override` you can see how this logic plays out.

So, I guess you can use that kind of pattern rather than warning.

Seanny123:
I noticed that this over-riding isn't tested. Should it be? Should I make sure to write a test for mine?

tbekolay:
Yes, definitely test anything that isn't yet tested :)

tbekolay:
We no longer pass configs to the `InputGatedMemory` so this issue is moot!

Issue #790, opened 2015-07-15
Standardizing when, when not, and where to use the config system


xchoo:
I'm trying to reorganize the associative memory network code to take full use of the config system (for an example see: [link](https://github.com/nengo/nengo/issues/780#issuecomment-120173547)). The idea being that if you don't specify a configuration, the network will build using whatever defaults the network designer (i.e. me) has put in, and if you do specify a configuration, it will use the configuration given.

In the examples below, the 'defaults' for the custom network are: max_rates=Uniform(10, 50), and encoders=Choice([[1]]). These can be overridden by the user.

<hr>

Method 1 (the 'old' way of doing it) - Put in all the possible changes to the network in the function definition:

```
def my_net(max_rates=Uniform(10, 50), encoders=Choice([[1]])):
    ... # Make the network
    with network:
        ens = nengo.Ensemble(..., max_rates=max_rates, encoders=encoders)
    ... # make the rest of the network
```

<hr>

Method 2 (using kwargs, and modifying the dictionary) - Used in the current AssociativeMemory code:

```
def my_net(**ens_kwargs):
    ... # Make the network
    with network:
        e_kwargs = dict(ens_kwargs)
        e_kwargs['max_rates'] = ens_kwargs.get('max_rates', Uniform(10, 50))
        e_kwargs['encoders'] = ens_kwargs.get('encoders', Choice([[1]]))
        ens = nengo.Ensemble(**e_kwargs)
    ... # make the rest of the network
```

<hr>

Method 3 (using kwargs, and updating the config) - Used in EnsembleArray:

```
def my_net(**ens_kwargs):
    ... # Make the network
    with network:
        network.config[nengo.Ensemble].update(ens_kwargs)
        ens = nengo.Ensemble(...)     
    ... # make the rest of the network
```

<hr>

Method 4 (pass in the config to use) - Used in Product:

```
def my_net(config=None):
    ... # Make the network

    if config is None:
        config = nengo.Config(nengo.Ensemble)
        config[nengo.Ensemble].max_rates = Uniform(10, 50)
        config[nengo.Ensemble].encoders = Choice([[1]])

    with nested(config, network):
        ens = nengo.Ensemble(...)     
    ... # make the rest of the network
```

Alternatively, I think this will work?

```
def my_net(config=None):
    ... # Make the network

    if config is None:
        config = nengo.Config()

    my_config = nengo.Config(nengo.Ensemble)
[...]

PR #795, opened 2015-07-21
Synapse override (on hold until config stuff is resolved)

Seanny123:
As discussed in #742 , I moved the synapse-config overriding option into utils and added it to the InputGatedMemory.


xchoo:
In light of the recent discussion with the config system usage, should we postpone the merging of this PR?

Seanny123:
Do you mean #790?

xchoo:
Yup! :smile:

Issue #690, opened 2015-03-18
Change spa.Buffer name


celiasmith:
Terry and I have been discussing this in an email... Here's a summary so far.

Chris:

>> it's occurred
>> to me that maybe we should rename spa.Buffer to something else?  Buffer is
>> an odd word that we don't generally use in the SPA, especially since
>> WorkingMemory is it's own spa element.  Do you think just spa.Ensemble would
>> make sense? (Since that's usually what it is...)  or perhaps you have
>> another idea...

Terry:

> Hmm...  yup, I'd totally be onboard with changing that name.  The
> intent with the name Buffer is to indicate that it's transient storage
> -- if something isn't actively sending stuff into it, it won't store
> the value (that's the difference between a Buffer and a Memory). But
> the word "Buffer" really doesn't communicate that well.  I'm not sure
> I like spa.Ensemble, as I think that'd be prone to confusion.   But
> I'm not sure what's better.
>
> Another possibility would be to combine Buffer and Memory into one
> thing called Memory, and do something like:
>
> model.a = spa.Memory()                          # this is a Buffer
> model.b = spa.Memory(maintain=True)     # this is a Memory

Chris: 

As for the name thing, I also worried about Ensemble being confusing -- then I thought a) maybe the 'spa.' would help alleviate confusion enough and b) aren't these really extremely functionally similar to regular ensembles (and more so ensemble arrays).  So it seemed like the best option... given:

- I also thought about the 'memory' thing, but worried that they're really not like memories in any typical sense  (i.e. only last a couple ms is 'remembered')... any more than regular ensembles are I suppose. 

- I've been trying to come up with other designations, but none really strike me as good: representation (too generic); pointer (don't have to be compressed); network (don't really have internal structure); group/population (just synonyms for ensemble); neurons/vector (nope); CorticalEnsemble or other things longer than one word (too long ;) )

Anyway, that's where my thinking is at the moment... no resolution yet, but still feeling 'buffer' isn't quite right.

tbekolay:
The main problem with `spa.Ensemble` is that people could/would do `from nengo.spa import Ensemble` and then the resulting model code would be super confusing. `SPAEnsemble` would be better (though `spa.SPAEnsemble` is redundant, so probably no better then `Buffer`).

[...]

PR #796, opened 2015-07-21
Make spa.State to replace spa.Memory and spa.Buffer

### LINKED ISSUE #761: spa.Buffer and spa.Memory are annoying to connect to
Writing `buffer.state.output` is annoying and counter-intuitive. So as I discussed with Terry, I'm going to move the ensemble array into the SPA module and since this will cause an API change that will break a bunch of code, I'm going to create a new module, which brings about the delightful issue of renaming everything and deprecating the old stuff as discussed in #690.

Seanny123:
Fixes #761 and #690. Documentation fixes are still required.

### DIFF
--- a/nengo/spa/state.py
+++ b/nengo/spa/state.py
@@ -1,13 +1,12 @@
-import warnings
-
 import numpy as np
 
 import nengo
 from nengo.spa.module import Module
+from nengo.networks.ensemblearray import EnsembleArray
 
 
-class Buffer(Module):
-    """A module capable of representing a single vector, with no memory.
+class State(Module):
+    """A module capable of representing a single vector, with optional memory.
 
     This is a minimal SPA module, useful for passing data along (for example,
     visual input).
@@ -17,21 +16,23 @@ class Buffer(Module):
     dimensions : int
         Number of dimensions for the vector
     subdimensions : int
-        Size of the individual ensembles making up the vector.  Must divide
+        Size of the individual ensembles making up the vector. Must divide
         evenly into dimensions
     neurons_per_dimensions : int
         Number of neurons in an ensemble will be this*subdimensions
+    feedback : float
+        Gain of feedback connection. Set to 1.0 for perfect memory. Other
+        non-zero values will create a decaying memory
+    feedback_synapse : float, nengo.Synapse
+        The synapse on the feedback connection
     vocab : Vocabulary, optional
         The vocabulary to use to interpret this vector
-    direct : boolean
-        Whether or not to use direct mode for the neurons
     """
+
     def __init__(self, dimensions, subdimensions=16, neurons_per_dimension=50,
-                 vocab=None, direct=False, label=None, seed=None,
-                 add_to_container=None):
-        warnings.warn("Buffer is deprecated in favour of spa.State",
-                      DeprecationWarning)
-        super(Buffer, self).__init__(label, seed, add_to_container)
+                 feedback=0.0, feedback_synapse=0.1, vocab=None, label=None,
+                 seed=None, add_to_container=None):
+        super(State, self).__init__(label, seed, add_to_container)
 
         if vocab is None:
             # use the default one for this dimensionality
@@ -45,17 +46,24 @@ class Buffer(Module):
         subdimensions = min(dimensions, subdimensions)
 
         if dimensions % subdimensions != 0:
-            raise ValueError('Number of dimensions(%d) must be divisible by '
-                             'subdimensions(%d)' % (dimensions, subdimensions))
+            raise ValueError("Dimensions (%d) must be divisible by sub"
+                             "dimensions (%d)" % (dimensions, subdimensions))
 
         with self:
-            self.state = nengo.networks.EnsembleArray(
+            self.state_ensembles = EnsembleArray(
                 neurons_per_dimension * subdimensions,
                 dimensions // subdimensions,
                 ens_dimensions=subdimensions,
-                neuron_type=nengo.Direct() if direct else nengo.LIF(),
                 radius=np.sqrt(float(subdimensions) / dimensions),
                 label='state')
+            self.input = self.state_ensembles.input
+            self.output = self.state_ensembles.output
 
-        self.inputs = dict(default=(self.state.input, vocab))
-        self.outputs = dict(default=(self.state.output, vocab))
+        self.inputs = dict(default=(self.input, vocab))
+        self.outputs = dict(default=(self.output, vocab))
+
+        with self:
+            if feedback is not None and feedback != 0.0:
+                nengo.Connection(self.output, self.input,
+                                 transform=feedback,
+                                 synapse=feedback_synapse)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -50,6 +50,9 @@ Release History
   ``transform`` may have changed depending on the type of connection;
   it is now no longer possible to probe ``decoders`` or ``transform``.
   (`#729 <https://github.com/nengo/nengo/pull/729>`_)
+- The ``spa.State`` object replaces the old ``spa.Memory`` and ``spa.Buffer``.
+  These old modules are deprecated and will be removed in 2.2.
+  (`#796 <https://github.com/nengo/nengo/pull/796>`_)
 
 **Behavioural changes**
 
--- a/nengo/spa/__init__.py
+++ b/nengo/spa/__init__.py
@@ -8,6 +8,7 @@ from .input import Input
 from .memory import Memory
 from .pointer import SemanticPointer
 from .spa import SPA
+from .state import State
 from .thalamus import Thalamus
 from .utils import enable_spa_params, similarity
 from .vocab import Vocabulary
--- a/nengo/spa/memory.py
+++ b/nengo/spa/memory.py
@@ -1,3 +1,5 @@
+import warnings
+
 import nengo
 from nengo.spa.buffer import Buffer
 
@@ -12,13 +14,15 @@ class Memory(Buffer):
     synapse : float
         synaptic filter to use on recurrent connection
     tau : float or None
-        Effective time constant of the integrator.  If None, it should
+        Effective time constant of the integrator. If None, it should
         have an infinite time constant.
     """
 
     def __init__(self, dimensions, subdimensions=16, neurons_per_dimension=50,
                  synapse=0.01, vocab=None, tau=None, direct=False,
                  label=None, seed=None, add_to_container=None):
+        warnings.warn("Memory is deprecated in favour of spa.State",
+                      DeprecationWarning)
         super(Memory, self).__init__(
             dimensions=dimensions,
             subdimensions=subdimensions,
--- a/nengo/spa/spa.py
+++ b/nengo/spa/spa.py
@@ -111,7 +111,7 @@ class SPA(nengo.Network):
             # Since there are no attributes to distinguish what's been added
             # and what hasn't, we have to ask the network
             if isinstance(net, Module) and (net not in module_list):
-                raise ValueError("%s was not added as an attribute of"
+                raise ValueError("%s was not added as an attribute of "
                                  "the SPA network and won't be detected"
                                  % (net))
 
--- a/nengo/spa/tests/test_state.py
+++ b/nengo/spa/tests/test_state.py
@@ -1,0 +1,134 @@
+import numpy as np
+import pytest
+
+import nengo
+from nengo import spa
+
+
+def test_basic():
+    with spa.SPA() as model:
+        model.state = spa.State(dimensions=16)
+
+    input = model.get_module_input('state')
+    output = model.get_module_output('state')
+    assert input[0] is model.state.input
+    assert output[0] is model.state.output
+    assert input[1] is output[1]
+    assert input[1].dimensions == 16
+
+
+def test_neurons():
+    with spa.SPA() as model:
+        model.state = spa.State(dimensions=16, neurons_per_dimension=2)
+
+    assert len(model.state.state_ensembles.ensembles) == 1
+    assert model.state.state_ensembles.ensembles[0].n_neurons == 16 * 2
+
+    with spa.SPA() as model:
+        model.state = spa.State(dimensions=16, subdimensions=1,
+                                neurons_per_dimension=2)
+
+    assert len(model.state.state_ensembles.ensembles) == 16
+    assert model.state.state_ensembles.ensembles[0].n_neurons == 2
+
+
+def test_dimension_exception():
+    with pytest.raises(Exception):
+        with spa.SPA() as model:
+            vocab = spa.Vocabulary(16)
+            model.state = spa.State(dimensions=12, vocab=vocab)
+
+    with spa.SPA() as model:
+        model.state = spa.State(dimensions=12, subdimensions=3)
+
+
+def test_no_feedback_run(Simulator, seed):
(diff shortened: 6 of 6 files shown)


nengo/spa/state.py
```diff
@@ -0,0 +1,76 @@
...
+    neurons_per_dimensions : int
+        Number of neurons in an ensemble will be this*subdimensions
+    vocab : Vocabulary, optional
+        The vocabulary to use to interpret this vector
+    direct : boolean
+        Whether or not to use direct mode for the neurons
+    """
+
+    def __init__(self, dimensions, subdimensions=16, neurons_per_dimension=50,
+                 feedback=1.0, tau=0.1,
```
Seanny123: @tcstewar please ignore the messiness of this commit and direct your attention to this file. What do you think of the `feedback` parameter? Is it being used appropriately? Previously, in `spa.Memory` the "feedback" parameter was scaled based on tau and the synapse. Why was that and should it be done here?

tcstewar: Hmm...  I don't think a parameter called ```feedback``` should also change the input transform, as you have it here.  I'd expect a parameter called ```feedback``` to only affect the feedback.

In the previous code, the two parameters were ```synapse``` and ```tau```, where ```synapse``` was the "synaptic filter to use on recurrent connection" and ```tau``` was the "effective time constant of the integrator".  In other words, if ```tau=3.0``` then the memory would last for around 3 seconds (and if ```tau=None``` then it would try to be a perfect integrator with an effective time constant of infinity).  That's what all the scaling was about was attempting to make that happen.

Part of me really likes that way of organizing it.  But part of me also thinks that it was a bit overcomplicated, difficult to explain, and never really worked that well anyway (as in the actual time constant you get isn't all that close to what you asked for, and it's hard to know over what range of values it will be good).  And in practice, we always ended up doing something like "just set the recurrent transform to 0.95".

So all of that said, I think the simplest and clearest approach is to do a ```feedback``` parameter.  I think it should default to ```0.0``` (i.e. no feedback) and if people want a memory they should set it to ```1.0```.  And if hey want a decaying memory they can set it to ```0.95``` or whatever.  And the only thing that parameter should affect is that feedback (i.e. it should not affect the input scaling).

The other option which is a bit more complex is to just have the ```tau``` parameter, or perhaps ```memory_time``` and do the same ```transform = 1.0 - synapse/memory_time``` calculation (although that in itself only works for ```synapse``` values that are floats....)

So yeah, I'd lean more to the simple option.  The tau thing was a neat idea, but not worth in the long run, I think.

tcstewar: On a related note, I do think the synapse for the feedback connection should be a parameter, and I think the parameter name should be ```feedback_synapse```.  It should just be passed in to the ```synapse``` argument (i.e. it shouldn't assume it's a float and pass it in to ```synapse.LowPass()```, as you have here).

Seanny123: Just to make sure I understand what you're saying, I'm going to try and write the conclusion down in my own words.

Change the `feedback` parameter to `feedback_synapse` and add it directly to the `synapse` parameter of the feedback connection. By setting the `feedback_synapse` to 1 an almost perfect memory should be created, but the default should be 0.

Is that right?

tcstewar: Nope, not quite.

There are two things that have to be configured: what type of synapse it is, and how strong it is.  So there are two parameters: ```feedback``` and ```feedback_synapse```.  In the implementation, these parameters are only used in one place: on the recurrent Connection which will look like:

```python
nengo.Connection(self.state_ensembles.output,
                                self.state_ensembles.input,
                                transform=feedback, synapse=feedback_synapse)
```

Seanny123:
This is now ready for review. Somehow while rebasing to master I picked up an extra commit. Hopefully this will go away when I squash and someone merges, but let me know if there's anything special I need to do.

nengo/spa/state.py
```diff
@@ -0,0 +1,74 @@
...
+        Number of dimensions for the vector
+    subdimensions : int
+        Size of the individual ensembles making up the vector.  Must divide
+        evenly into dimensions
+    neurons_per_dimensions : int
+        Number of neurons in an ensemble will be this*subdimensions
+    vocab : Vocabulary, optional
+        The vocabulary to use to interpret this vector
+    direct : boolean
+        Whether or not to use direct mode for the neurons
```
tcstewar: We don't need ```direct``` as a parameter, as it should be done by the config system (as with the other Networks)

nengo/spa/state.py
```diff
@@ -0,0 +1,72 @@
...
+        Size of the individual ensembles making up the vector.  Must divide
+        evenly into dimensions
+    neurons_per_dimensions : int
+        Number of neurons in an ensemble will be this*subdimensions
+    vocab : Vocabulary, optional
+        The vocabulary to use to interpret this vector
+    """
+
+    def __init__(self, dimensions, subdimensions=16, neurons_per_dimension=50,
+                 feedback=0.0, feedback_synapse=0.01,
```
tcstewar: need documentation for ```feedback``` and ```feedback_synapse```

nengo/spa/state.py
```diff
@@ -0,0 +1,71 @@
...
+                label='state')
+
+        self.inputs = dict(default=(self.input, vocab))
+        self.outputs = dict(default=(self.output, vocab))
+
+        with self:
+            nengo.Connection(self.input, self.state_ensembles.input,
+                             synapse=None)
+
+            if(feedback != 0.0):
```
tcstewar: No parentheses on ``if``` unless needed for multiple lines...  :)

nengo/spa/state.py
```diff
@@ -0,0 +1,75 @@
...
+        Number of dimensions for the vector
+    subdimensions : int
+        Size of the individual ensembles making up the vector.  Must divide
+        evenly into dimensions
+    neurons_per_dimensions : int
+        Number of neurons in an ensemble will be this*subdimensions
+    feedback : float
+        The gain of the feedback connection. Set to 1.0 for perfect memory.
+        Other values will result in a memory that decays over time.
+    feedback_synapse : float, Nengo.Synapse
```
tcstewar: capitalization on ```nengo```....

nengo/spa/state.py
```diff
@@ -0,0 +1,75 @@
...
+                label='state')
+
+        self.inputs = dict(default=(self.input, vocab))
+        self.outputs = dict(default=(self.output, vocab))
+
+        with self:
+            nengo.Connection(self.input, self.state_ensembles.input,
+                             synapse=None)
+
+            if feedback != 0.0:
```
tcstewar: It would be nice to handle the case where someone sets ```feedback=None``` as well as ```feedback=0.0```.

tcstewar:
Looking really good, btw.  Thank you for putting everything together and doing this!

tcstewar:
(and yup, commit  dab4bcb really needs to be removed from this....   yay git....  :)  )

nengo/spa/state.py
```diff
@@ -0,0 +1,77 @@
...
+    subdimensions : int
+        Size of the individual ensembles making up the vector.  Must divide
+        evenly into dimensions
+    neurons_per_dimensions : int
+        Number of neurons in an ensemble will be this*subdimensions
+    vocab : Vocabulary, optional
+        The vocabulary to use to interpret this vector
+    direct : boolean
+        Whether or not to use direct mode for the neurons
+    """
```
xchoo: Documentation need updating.

nengo/spa/state.py
```diff
@@ -0,0 +1,77 @@
...
+
+        self.inputs = dict(default=(self.input, vocab))
+        self.outputs = dict(default=(self.output, vocab))
+
+        with self:
+            if(feedback == 0.0):
+                nengo.Connection(self.input, self.state_ensembles.input,
+                                 synapse=None)
+            else:
+                nengo.Connection(self.input, self.state_ensembles.input,
```
xchoo: What's this connection for?

Seanny123: Stop reviewing for a bit. When I was squashing I lost all my work.

nengo/spa/state.py
```diff
@@ -0,0 +1,77 @@
...
+    neurons_per_dimensions : int
+        Number of neurons in an ensemble will be this*subdimensions
+    vocab : Vocabulary, optional
+        The vocabulary to use to interpret this vector
+    direct : boolean
+        Whether or not to use direct mode for the neurons
+    """
+
+    def __init__(self, dimensions, subdimensions=16, neurons_per_dimension=50,
+                 feedback=1.0, feedback_synapse=0, tau=0.1,
```
xchoo: Do we want to use different names for ```feedback``` and ```tau```?

nengo/spa/state.py
xchoo: Oh. Haha. K. :smile:

nengo/spa/state.py
```diff
@@ -0,0 +1,72 @@
...
+    Parameters
+    ----------
+    dimensions : int
+        Number of dimensions for the vector
+    subdimensions : int
+        Size of the individual ensembles making up the vector. Must divide
+        evenly into dimensions
+    neurons_per_dimensions : int
+        Number of neurons in an ensemble will be this*subdimensions
+    float : float
```
xchoo: ```feedback``` (instead of ```float```)?

nengo/spa/state.py
```diff
@@ -0,0 +1,72 @@
...
+        Gain of feedback connection. Set to 1.0 for perfect memory. Other
+        non-zero values will create a decaying memory
+    feedback_synapse : float, nengo.Synapse
+        The synapse on the feedback connection
+    vocab : Vocabulary, optional
+        The vocabulary to use to interpret this vector
+    """
+
+    def __init__(self, dimensions, subdimensions=16, neurons_per_dimension=50,
+                 feedback=0.0, feedback_synapse=0.01, vocab=None, label=None,
```
xchoo: ```feedback_synapse=0.1``

Seanny123: Why? In `spa.Memory` the synapse defaults to `0.01`

nengo/spa/state.py
```diff
@@ -0,0 +1,72 @@
...
+        self.outputs = dict(default=(self.output, vocab))
+
+        with self:
+            nengo.Connection(self.input, self.state_ensembles.input,
+                             synapse=None)
+
+            if feedback != 0.0 and feedback is not None:
+                nengo.Connection(self.state_ensembles.output,
+                                 self.state_ensembles.input,
+                                 transform=feedback, synapse=feedback_synapse)
```
xchoo: There's no connection from ```self.state_ensembles.output``` to ```self.output```.
I'd use ```self.input = self.state_ensembles.input``` and ```self.output = self.state_ensembles.output```.

nengo/spa/state.py
xchoo: ```spa.Memory``` has a weird calculation (which includes a ```tau``` parameter). @tcstewar comments?

tcstewar: The previous discussion (that I'd still agree with) was that the ```tau``` parameter was a bad idea.

As for changing the default, yes, i'd agree with @xchoo that a better default would be 0.1.  I find myself changing it to that almost all the time....

Seanny123:
Alright, once again, I have the audacity to declare that I think I've fixed everything mentioned. Let me know if I should (carefully) squash and mark for merging.

Seanny123:
Squashed and ready to go.

Seanny123:
@tbekolay how do I enter this change into the changelog?

tbekolay:
Edit the `CHANGES.rst` file. Hopefully how to do that is pretty clear from the structure of that file, but let me know if anything is unclear!

Seanny123:
@tbekolay Done. Let me know if I obeyed the rules properly.

tbekolay:
RULES OBEYED. I think probably we just one entry; keep the API changes entry, and if you want, you can mention that `spa.Buffer` and `spa.Memory` are deprecated (will be removed in 2.2).

Seanny123:
@tbekolay I'm officially between a rock and a hard place in terms of trying to squash my last commit. Do you know how to deal with it? Jan say it's because somewhere along my catastrophic rebasing I ended with two parents to my second last commit. If you don't know how to fix it can we just pretend it's not a problem?

tbekolay:
I'll fix it in the merge, no worries!

Seanny123:
Thanks to Jan, we fixed the commit history and this is ready to merge!

tbekolay:
Do we have any examples that use buffers / memories? If so, we should switch them over to using `State` at some point... I was too lazy to craft the perfect `grep` statement, but if someone wants to look for that and finds anything, feel free to make an issue for it. But, that's not a blocking issue, so merging now!

tcstewar:
> Do we have any examples that use buffers / memories? If so, we should switch them over to using State at some point..

There are some in ```nengo_gui``` (some of the tutorials), but I want to wait on updating those until there's a ```nengo``` release.....

tbekolay:
Makes sense!

ikajic:
I'm just going through the above mentioned tutorials in `nengo_gui` and there are a few spots where `spa.Memory` and `spa.Buffer` are still used. Is it ok if I change them now?

Seanny123:
That would be awesome!

ikajic:
When using `spa.State` as Memory (by setting `feedback=1`), the default value of the optional argument `feedback_synapse` is `0.1`. But, in `spa.Memory` the same parameter (I assume) is `synapse=0.01`. Should this be changed in `spa.State` or should I keep on setting `feedback_synapse` to `0.01`?

tcstewar:
> Should this be changed in spa.State or should I keep on setting feedback_synapse to 0.01?

0.01 was a horrible choice for the default.  All the examples should work better with 0.1 (the new default)

ikajic:
I told them to work but they don't listen to me :)! The behavior with 0.1 seems to be less stable (e.g. in `22-spa-sequence-controlled.py` and when I change it to `0.01` it does what it should be doing.

tcstewar:
> I told them to work but they don't listen to me :)! The behavior with 0.1 seems to be less stable (e.g. in 22-spa-sequence-controlled.py) and when I change it to 0.01 it does what it should be doing.

Argh...  damnit, you're right.  With the increased synapse, the memory is *much* more stable, and so random noise never kicks it into the desired sequence....  grr...  we could add in a quick model.input=spa.Input(memory=lambda t: 'A' if t<0.1 else '0') to give it that boost, but that's ugly.  Hmm....

tcstewar:
I can't find a better solution for this example rather than doing what you suggested

```python
model.memory = spa.State(D, feedback=1, feedback_synapse=0.01)
```

That said, I'm still pretty sure that the best default for feedback_synapse is 0.1.  So I guess we should just do this and use it as an excuse to mention that synapse and its effects....

ikajic:
Ok, here is another one: adding a new rule `0.5 --> memory=A` also does it with `feedback_synapse=0.1`. Arguably, adding a new rule is less ugly (syntactically), but semantically it is still ugly. And we already use that sort of trick in `20-spa-actions.py` to turn down the random winner.

tcstewar:
> Ok, here is another one: adding a new rule 0.5 --> memory=A also does it with feedback_synapse=0.1.

I tried things like that, and they're really not stable for me on my computer.  Does it work all the time for you?

ikajic:
>I tried things like that, and they're really not stable for me on my computer. Does it work all the time for  you?

Yup. So far it worked every time I've tried.

PR #799, opened 2015-07-26
Added lower bound option to UniformHypersphere dist

xchoo:
Added low and high options to UniformHypersphere distribution.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,8 @@ Release History
 - Added a new example showing how adjusting ensemble tuning curves can
   improve function approximation.
   (`#1129 <https://github.com/nengo/nengo/pull/1129>`_)
+- Added a minimum magnitude option to ``UniformHypersphere``.
+  (`#799 <https://github.com/nengo/nengo/pull/799>`_)
 
 **Changed**
 
--- a/nengo/dists.py
+++ b/nengo/dists.py
@@ -1,4 +1,7 @@
 from __future__ import absolute_import
+
+import warnings
+
 import numpy as np
 
 from nengo.exceptions import ValidationError
@@ -222,17 +225,29 @@ class UniformHypersphere(Distribution):
         Whether sample points should be distributed uniformly
         over the surface of the hyperphere (True),
         or within the hypersphere (False).
+    min_magnitude : Number, optional (Default: 0)
+        Lower bound on the returned vector magnitudes (such that they are in
+        the range ``[min_magnitude, 1]``). Must be in the range [0, 1).
+        Ignored if ``surface`` is ``True``.
     """
 
     surface = BoolParam('surface')
+    min_magnitude = NumberParam('min_magnitude', low=0, high=1, high_open=True)
 
-    def __init__(self, surface=False):
+    def __init__(self, surface=False, min_magnitude=0):
         super(UniformHypersphere, self).__init__()
+        if surface and min_magnitude > 0:
+            warnings.warn("min_magnitude ignored because surface is True")
         self.surface = surface
+        self.min_magnitude = min_magnitude
 
     def __repr__(self):
-        return "UniformHypersphere(%s)" % (
-            "surface=True" if self.surface else "")
+        args = []
+        if self.surface:
+            args.append("surface=%s" % self.surface)
+        if self.min_magnitude > 0:
+            args.append("min_magnitude=%r" % self.min_magnitude)
+        return "%s(%s)" % (type(self).__name__, ', '.join(args))
 
     def sample(self, n, d, rng=np.random):
         if d is None or d < 1:  # check this, since other dists allow d = None
@@ -247,7 +262,8 @@ class UniformHypersphere(Distribution):
         # Generate magnitudes for vectors from uniform distribution.
         # The (1 / d) exponent ensures that samples are uniformly distributed
         # in n-space and not all bunched up at the centre of the sphere.
-        samples *= rng.rand(n, 1) ** (1.0 / d)
+        samples *= rng.uniform(
+            low=self.min_magnitude**d, high=1, size=(n, 1)) ** (1. / d)
 
         return samples
 
--- a/nengo/tests/test_dists.py
+++ b/nengo/tests/test_dists.py
@@ -3,6 +3,7 @@ import pytest
 
 import nengo.dists as dists
 import nengo.utils.numpy as npext
+from nengo.utils.testing import warns
 from nengo.exceptions import ValidationError
 
 
@@ -69,19 +70,15 @@ def test_exponential(scale, shift, high, rng):
         assert abs(np.mean(samples - shift) - scale) < ci
 
 
-@pytest.mark.parametrize("dimensions", [0, 1, 2, 5])
-def test_hypersphere(dimensions, rng):
-    n = 150 * dimensions
-    if dimensions < 1:
-        with pytest.raises(ValueError):
-            dist = dists.UniformHypersphere().sample(1, dimensions)
-    else:
-        dist = dists.UniformHypersphere()
-        samples = dist.sample(n, dimensions, rng=rng)
-        assert samples.shape == (n, dimensions)
-        assert np.allclose(np.mean(samples, axis=0), 0, atol=0.1)
-        hist, _ = np.histogramdd(samples, bins=5)
-        assert np.allclose(hist - np.mean(hist), 0, atol=0.1 * n)
+@pytest.mark.parametrize(
+    "min_magnitude,d", [(0, 1), (0, 2), (0, 5), (0.6, 1), (0.3, 2), (0.4, 5)])
+def test_hypersphere_volume(min_magnitude, d, rng):
+    n = 150 * d
+    dist = dists.UniformHypersphere(min_magnitude=min_magnitude)
+    samples = dist.sample(n, d, rng=rng)
+    assert samples.shape == (n, d)
+    assert np.allclose(np.mean(samples, axis=0), 0, atol=0.1)
+    assert np.all(npext.norm(samples, axis=1) >= min_magnitude)
 
 
 @pytest.mark.parametrize("dimensions", [1, 2, 5])
@@ -92,6 +89,16 @@ def test_hypersphere_surface(dimensions, rng):
     assert samples.shape == (n, dimensions)
     assert np.allclose(npext.norm(samples, axis=1), 1)
     assert np.allclose(np.mean(samples, axis=0), 0, atol=0.25 / dimensions)
+
+
+def test_hypersphere_dimension_fail(rng):
+    with pytest.raises(ValueError):
+        dists.UniformHypersphere(0).sample(1, 0)
+
+
+def test_hypersphere_warns(rng):
+    with warns(UserWarning):
+        dists.UniformHypersphere(surface=True, min_magnitude=0.1)
 
 
 @pytest.mark.parametrize("weights", [None, [5, 1, 2, 9], [3, 2, 1, 0]])


nengo/dists.py
```diff
@@ -156,8 +156,10 @@ class UniformHypersphere(Distribution):
 
     """
 
-    def __init__(self, surface=False):
+    def __init__(self, low=0, high=1, surface=False):
```
arvoelke: Docstring needs to be updated with these parameters, along with an explanation of how they effect the distribution.

arvoelke:
Missing unit test.

xchoo:
Haha. Yeah. I realized after making the pull request that I didn't do the tests for them. :laughing:

jgosmann:
Doesn't this make the `surface` parameter redundant because you could set `low = high = 1.`?

arvoelke:
The `low`/`high` is being applied before any scaling, so `low = high = 1` would just give the constant vector `1 / d`.

xchoo:
Yeah. There's some interaction between the ```low```, ```high``` and ```surface``` parameters that I didn't consider. I will fix that too.

tcstewar:
Hmm, if I'm interpretting this right, setting ```low=0.5``` and ```high=0.75``` would lead to sample points that have a radius between ```0.5^(1/D)``` and ```0.75^(1/D)``` (but still uniformly distributed).  Is that right?  If so, it might be clearer to actually specify the radius, so that the parameters have a clearer meaning.  Something like:

```python
samples *= rng.uniform(low=self.low ** d, high=self.high ** d, size=(n, 1)) ** (1.0 / d)
```

That said, I'm not sure what the situation is you want these parameters for, so maybe they're more useful the way you have them -- I just thought I'd point out the possibility.  :)

xchoo:
Oh. I think you are right. haha

xchoo:
Alrighty. I think I've addressed the issues listed here.

nengo/dists.py
```diff
@@ -156,12 +163,17 @@ class UniformHypersphere(Distribution):
...
     """
 
-    def __init__(self, surface=False):
-        self.surface = surface
+    def __init__(self, low=0, high=1, surface=False):
+        self.low = max(low, 0)
+        self.high = max(high, self.low)
+
+        if surface:
+            self.low = self.high
```
tcstewar: This logic (of the interaction between high, low, and surface) should be documented in the parameters

nengo/tests/test_dists.py
```diff
@@ -79,6 +79,28 @@ def test_hypersphere_surface(dimensions, rng):
     assert np.allclose(np.mean(samples, axis=0), 0, atol=0.25 / dimensions)
 
 
+@pytest.mark.parametrize("low,high", [(-2, -1), (-1, 1), (1, 2), (1, -1)])
```
tcstewar: What does it mean for low or high to be less than zero?  Doesn't that lead to imaginary numbers in the ```low ** (1/D)``` calculation?  Shouldn't there be a check that they are always great than zero?

nengo/dists.py
xchoo: It is? (well, maybe not the surface one, but the rest are?)

nengo/tests/test_dists.py
xchoo: There's no formal check, it just makes them 0 if you give numbers < 0. Do you think it will be a better idea to throw a value error instead?

jgosmann:
I don't like, the interaction of the arguments and the possible to pass nonsensical combinations. With `surface=False` the vector length is < `high`, but with `surface=True` the vector length is <= `high`. Also, with `surface=True` the value of `low` has no effect.

Not sure if there is a good way to work around this. Maybe deprecate `surface` to remove it in a future version? Or remove `high` and allow to pass in the value to `surface`?

tcstewar:
> I don't like, the interaction of the arguments and the possible to pass nonsensical combinations. With surface=False the vector length is < high, but with surface=True the vector length is <= high. Also, with surface=True the value of low has no effect.

That's my concern as well.  And right now, the only documentation about this interaction is that "Values provided out of this range will be corrected", which i don't think is clear enough about what this interaction will do.

I'm still not sure what the use case is for this change, but one possibility would be to get rid of ```high``` as a parameter.  In most of the use cases I can think of, ```high``` is redundant with ```radius```, and is just a multiply anyway.  So that might be sufficient.

@xchoo , what is the use case you're thinking of for this change?  Would it work without ```high```?

xchoo:
Yup! It would work without high. Although, if that's the case, I would move to rename UniformHypersphere to UniformUnitHypersphere.

jgosmann:
But it's only unit if the radius is `1` ...

xchoo:
If you leave out ```high```, the max magnitude of the distribution will be 1. In that case, if you set ```surface=True```, the distribution will ignore the ```low``` value and put all the points on the surface of the hypersphere.

xchoo:
(Also, the description of the class is ```Distributions over an n-dimensional unit hypersphere.```)

tcstewar:
> Although, if that's the case, I would move to rename UniformHypersphere to UniformUnitHypersphere.

I would vote against that renaming, just on backwards compatibility reasons.  :)

xchoo:
> I would vote against that renaming, just on backwards compatibility reasons. :)
I guessssssssssss.........

jgosmann:
You could allow a number as argument for `surface` (in addition to a boolean) which would then be treated as `high`.

xchoo:
But how would you indicate if you want to have the points on the surface or not? E.g., you want a hypersphere of radius 2, and all points on the surface.

jgosmann:
`low=2, surface=2'? (but this is somewhat inconsistent with `randint` where `low == high` is not allowed)

Maybe we should implement two distributions: One samples from the whole hypersphere (with arguments `low` and `high`) and the other one just from the surface (by internally using the first distribution from the whole hypersphere and normalizing) with one argument `radius`. For backward compatibility we would have to keep the current API too, but we could mark it as deprecated and remove it one (or a few more releases) later.

tcstewar:
Your documentation still doesn't indicate the interaction between surface and low.  In particular, if surface is True, then low is ignored.  Furthermore, your documentation still refers to the ```high``` parameter, which has been removed.

I will also note that, how it currently stands, setting surface to True gives exactly the same result as setting low to 1.0, and setting surface to False gives exactly the same result as setting low to 0.0.  This suggests that you could also get rid of the ```low``` parameter and just have ```surface``` accept a float or a boolean.  This would at least get rid of the confusion surrounding the interaction between the two variables.

xchoo:
> This suggests that you could also get rid of the low parameter and just have surface accept a float or a boolean. This would at least get rid of the confusion surrounding the interaction between the two variables.

But naming that parameter ```surface``` doesn't make any sense. The ```surface``` parameter name has been left there for compatibility. Ideally, I'd have a ```high``` and a ```low``` and no ```surface``` parameter.

tcstewar:
> One samples from the whole hypersphere (with arguments low and high) and the other one just from the surface (by internally using the first distribution from the whole hypersphere and normalizing) with one argument radius. For backward compatibility we would have to keep the current API too, but we could mark it as deprecated and remove it one (or a few more releases) later.

Hmm...  I'd like to avoid bringing ```radius``` into this discussion, as that's an Ensemble-level parameter and I could see also sorts of potential for confusion here.  

Also, before looking at an API change like that, I'd still like to know what the use case is for something like this.  I've never wanted to sample from a uniform hyper-doughnut before -- why do we want that capability now?

xchoo:
I do. Same logic as to why I use a Uniform distribution with a non-zero low when messing around with non standard intercepts.

xchoo:
(i.e. I want to provide only valid sample points within the intercept ranges)

tcstewar:
> But naming that parameter surface doesn't make any sense. The surface parameter name has been left there for compatibility. Ideally, I'd have a high and a low and no surface parameter.

It makes as much sense to me as ```low`` and ```high``` do.  I had to read your code in order to understand that it wasn't a bounding box, which was my initial assumption.

xchoo:
> It makes as much sense to me as ```low`` and ```high``` do. I had to read your code in order to understand that it wasn't a bounding box, which was my initial assumption.

I could use the name ```min_magnitude``` and ```max_magnitude```. I used ```low``` and ```high``` to keep it kinda consistent with the uniform distribution.

tcstewar:
> I could use the name min_magnitude and max_magnitude. I used low and high to keep it kinda consistent with the uniform distribution.

I would be way happier with ```min_magnitude``` rather than ```low```.  There's no need to keep it consistent with the parameters for the uniform distribution when you have to look at the implementation to realize that it uses a uniform distribution internally.  

And I still think ```max_magnitude``` isn't needed -- it's not worth the overhead of the complexity of describing the interactions between the parameters if there isn't a use case that requires it.

tcstewar:
> Same logic as to why I use a Uniform distribution with a non-zero low when messing around with non standard intercepts.

Ah, thank you.  That makes a lot more sense now that I know why this is desired...  :)

nengo/dists.py
```diff
@@ -148,20 +148,29 @@ class UniformHypersphere(Distribution):
...
+        such that the sampled vector magnitudes >= min_magnitude.
+        The default value and minimum value for min_magnitude is 0.
+        The maximum value for min_magnitude is 1. Values outside this range
+        will be clipped to this range.
     surface : bool
         Whether sample points should be distributed uniformly
         over the surface of the hyperphere (True),
         or within the hypersphere (False).
+        Note: Setting surface to True generates sample points from the
+        hypersphere surface irregardless of what the value of min_magnitude is.
```
tcstewar: Much clearer.  Although I will be ridiculously picky and object to the use of "irregardless"....  ;)

nengo/dists.py
```diff
@@ -148,20 +148,29 @@ class UniformHypersphere(Distribution):
...
 
-    def __init__(self, surface=False):
+    def __init__(self, min_magnitude=0, surface=False):
+        self.min_magnitude = min(max(min_magnitude, 0), 1)
         self.surface = surface
 
     def __repr__(self):
-        return "UniformHypersphere(%s)" % (
-            "surface=True" if self.surface else "")
+        return "UniformHypersphere(min_magnitude=%f%s)" % (
```
tcstewar: Might want to use ```%g``` here instead of ```%f```.  Not sure if we've been consistent on that elsewhere, though.

nengo/tests/test_dists.py
```diff
@@ -54,19 +54,41 @@ def test_gaussian(mean, std, rng):
...
+        # Check the distribution of sample points (only applicable when
+        # low < 1)
+        if min_magnitude < 1:
+            hist, _ = np.histogramdd(samples, bins=8)
+            # Use 8 bins to allow for the min_magnitude=0.5 case
+            assert np.allclose(hist[hist > 0] - np.mean(hist[hist > 0]), 0,
+                               atol=0.1 * n)
+
+        # Test if the distributions low and high values are set correctly
+        # wrt the provided low and high values
```
tcstewar: documentation should refer to min_magnitude, not low and high

nengo/dists.py
```diff
@@ -148,20 +148,29 @@ class UniformHypersphere(Distribution):
...
         or within the hypersphere (False).
+        Note: Setting surface to True generates sample points from the
+        hypersphere surface regardless of what the value of min_magnitude is.
         Default: False
 
     """
 
-    def __init__(self, surface=False):
+    def __init__(self, min_magnitude=0, surface=False):
+        self.min_magnitude = min(max(min_magnitude, 0), 1)
```
jgosmann: `np.clip` might be a bit easier to understand than the `min`/`max` combination?

jgosmann:
Personally, I would split up the unit test in multiple smaller functions to get rid of all the branching (I like to keep the individual tests as short and simple as possible; and I don't like to have a lot of logic in there that could have a bug itself). But if other people have a different opinion, I'm fine with leaving it as is.

jgosmann:
Oh, it would be also good to squash some of the commits before merging.

Seanny123:
Squashed and address Jan's comments. @jgosmann does this look good to you now?

jgosmann:
seems ok

hunse:
The commit message title is too long. It must be under 50 characters.

nengo/dists.py
```diff
@@ -189,20 +189,29 @@ class UniformHypersphere(Distribution):
...
         over the surface of the hyperphere (True),
         or within the hypersphere (False).
+        Note: Setting surface to True generates sample points from the
+        hypersphere surface regardless of what the value of min_magnitude is.
         Default: False
 
     """
 
-    def __init__(self, surface=False):
+    def __init__(self, min_magnitude=0, surface=False):
```
hunse: The new parameter should go after the existing `surface`, just in case people were using it positionally.

nengo/dists.py
```diff
@@ -189,20 +189,29 @@ class UniformHypersphere(Distribution):
...
+        such that the sampled vector magnitudes >= min_magnitude.
+        The default value and minimum value for min_magnitude is 0.
+        The maximum value for min_magnitude is 1. Values outside this range
+        will be clipped to this range.
     surface : bool
         Whether sample points should be distributed uniformly
         over the surface of the hyperphere (True),
         or within the hypersphere (False).
+        Note: Setting surface to True generates sample points from the
+        hypersphere surface regardless of what the value of min_magnitude is.
```
hunse: I'd remove this note, and in the description of `min_magnitude` say that it's ignored if `surface == True`.

nengo/dists.py
```diff
@@ -189,20 +189,29 @@ class UniformHypersphere(Distribution):
 
     Parameters
     ----------
+    min_magnitude : Number
+        The closed lower bound of the uniform hypersphere distribution;
+        such that the sampled vector magnitudes >= min_magnitude.
+        The default value and minimum value for min_magnitude is 0.
+        The maximum value for min_magnitude is 1. Values outside this range
+        will be clipped to this range.
```
hunse: Can this be more concise? e.g. "Lower bound on the returned vector magnitudes (i.e. magnitude >= min_magnitude). Must be in the range [0, 1]; defaults to 0. Ignored if `surface == True`."

nengo/dists.py
```diff
@@ -217,7 +226,9 @@ def sample(self, n, d, rng=np.random):
         # Generate magnitudes for vectors from uniform distribution.
         # The (1 / d) exponent ensures that samples are uniformly distributed
         # in n-space and not all bunched up at the centre of the sphere.
-        samples *= rng.rand(n, 1) ** (1.0 / d)
+        # samples *= rng.rand(n, 1) ** (1.0 / d)
```
hunse: Remove this commented out line.

nengo/tests/test_dists.py
```diff
@@ -68,19 +68,36 @@ def test_exponential(scale, shift, high, rng):
         assert abs(np.mean(samples - shift) - scale) < ci
 
 
-@pytest.mark.parametrize("dimensions", [0, 1, 2, 5])
-def test_hypersphere(dimensions, rng):
+@pytest.mark.parametrize("min_magnitude,dimensions",
+                         [(0, 1), (0, 2), (0, 5),
+                          (0.5, 1), (0.5, 2), (0.5, 5), (2, 1), (-2, 1)])
+def test_hypersphere(min_magnitude, dimensions, rng):
```
hunse: This is a weird test since it's checking a bunch of different things. Also, we lost the test case where `dimensions == 0`. I'd almost rather keep the original test as-is, and add a second test for the `min_magnitude` parameter.

Seanny123:
@hunse I fixed the rest of your comments and it's now passing all the tests. Thanks for being so throrough!

nengo/dists.py
```diff
@@ -192,17 +192,22 @@ class UniformHypersphere(Distribution):
...
-    def __init__(self, surface=False):
+    def __init__(self, surface=False, min_magnitude=0):
+        self.min_magnitude = np.clip(min_magnitude, 0, 1)
         self.surface = surface
 
     def __repr__(self):
-        return "UniformHypersphere(%s)" % (
-            "surface=True" if self.surface else "")
+        return "UniformHypersphere(min_magnitude=%s%r)" % (
+            ", surface=True" if self.surface else "", self.min_magnitude,)
```
hunse: This is still not right. Also, min_magnitude should only appear if it's different from the default.

nengo/tests/test_dists.py
```diff
@@ -68,19 +68,44 @@ def test_exponential(scale, shift, high, rng):
...
-            dist = dists.UniformHypersphere().sample(1, dimensions)
+    dist = dists.UniformHypersphere(min_magnitude=min_magnitude)
+    samples = dist.sample(n, dimensions, rng=rng)
+    assert samples.shape == (n, dimensions)
+    assert np.allclose(np.mean(samples, axis=0), 0, atol=0.1)
+
+
+@pytest.mark.parametrize("min_magnitude,dimensions",
+                         [(0.5, 1), (2, 1), (-2, 1)])
+def test_hypershpere_min_magnitude(min_magnitude, dimensions, rng):
```
hunse: This still has all those pesky `if` statements.

CHANGES.rst
```diff
@@ -66,6 +66,8 @@ Release History
   to specify any logging level, instead of the old binary ``debug`` parameter.
   Cache messages are logged at DEBUG instead of INFO level.
   (`#883 <https://github.com/nengo/nengo/pull/883>`_)
+- Added lower bound option to UniformHypersphere distribution.
```
hunse: This should make it clear that it's a lower bound on the magnitude.

Seanny123:
@hunse I did my best to fix your comments on this issue. Does it look okay now?

nengo/dists.py
```diff
@@ -221,18 +221,32 @@ class UniformHypersphere(Distribution):
     surface : bool, optional (Default: False)
         Whether sample points should be distributed uniformly
         over the surface of the hyperphere (True),
-        or within the hypersphere (False).
+        or within the hypersphere (False). Defaults to False.
+    min_magnitude : Number, optional (Default: 0)
+        Lower bound on the returned vector magnitudes (i.e.
+        magnitude >= min_magnitude). Must be in the range [0, 1]; defaults
+        to 0. Values outside this range will be clipped to this range. Ignored
```
jgosmann: I think this description is not entirely correct. If values outside of the range get clipped, I understand that values get sampled in the range [0, 1] and then everything below `min_magnitude` will set to `min_magnitude` (which will cluster vectors at `min_magnitude`). But as I understand the code values are sampled from `[min_magnitude, 1]` and thus are without clustering at `min_magnitude`.

nengo/dists.py
```diff
@@ -221,18 +221,32 @@ class UniformHypersphere(Distribution):
...
+        if ``surface == True``
     """
 
     surface = BoolParam('surface')
+    min_magnitude = NumberParam('min_magnitude', low=0, high=1)
 
-    def __init__(self, surface=False):
+    def __init__(self, surface=False, min_magnitude=0):
         super(UniformHypersphere, self).__init__()
+        self.min_magnitude = np.clip(min_magnitude, 0, 1)
```
jgosmann: I would prefer to raise an exception when `min_magnitude` is outside of the range (which should happen by just removing the call to `clip` because it should be checked by the NumberParam).

hunse: I agree, and I think this is what caused @jgosmann's confusion with the docstring.

nengo/dists.py
hunse: I think all this docstring is saying is that values for ``min_magnitude`` will be clipped to [0, 1]. I agree that we should have an exception instead of clipping. This is already done by the ``low`` and ``high`` in the parameter definition.

nengo/dists.py
```diff
@@ -221,18 +221,32 @@ class UniformHypersphere(Distribution):
...
+        or within the hypersphere (False). Defaults to False.
+    min_magnitude : Number, optional (Default: 0)
+        Lower bound on the returned vector magnitudes (i.e.
+        magnitude >= min_magnitude). Must be in the range [0, 1]; defaults
+        to 0. Values outside this range will be clipped to this range. Ignored
+        if ``surface == True``
     """
 
     surface = BoolParam('surface')
+    min_magnitude = NumberParam('min_magnitude', low=0, high=1)
```
hunse: Add ``high_open=True`` so that we don't allow ``min_magnitude == 1``, which I think would cause problems because then we'd be sampling from an empty interval.

hunse:
Ok, I've addressed @jgosmann's suggestions.

Seanny123:
Those changes look awesome. Thanks @hunse. I wasn't sure how much of the tests were redundant and how to fix the `__repr__` string code.

Issue #787, opened 2015-07-13
Setting seed for Processes?


arvoelke:
I'd like to be able to control the seed of a white noise process. I thought we could do this before by passing `seed` to the constructor. Is there any way I can reliably get two different simulations to use the exact same signal? 

Should `Node` accept a seed and propagate it to its child process?

tcstewar:
You can set the ```seed``` on the Simulator...  it that what you're looking for?

But I agree it would be nice, I think, to be able to set the seed on a Process....  I think I'd set it there rather than on the Node.

arvoelke:
That won't work for me because they are two different simulations (different ensembles and different connections), and so the effect depends on the build order. :frowning:

tbekolay:
Setting a seed on the node makes sense to me. All of the other objects have seeds, so this seems the most idiomatic way to do it. In fact, we could move the seed to to the `NengoObject` base class.

hunse:
Setting the seed on the node would suggest to me that it should be set implicitly if the network seed is set, like with other objects. In that case, there would be no need for a simulator seed.

tcstewar:
> Setting the seed on the node would suggest to me that it should be set implicitly if the network seed is set, like with other objects. In that case, there would be no need for a simulator seed.

That was my intuition as well -- all the other seeds can get driven by the Network seed, but the Process one can't.  So it feels more consistent to me to set the seed on the Process instead of the Node.

tbekolay:
> it should be set implicitly if the network seed is set

Yeah, I would think so... Would anyone argue that since the `Process` is a runtime thing (rather than a buildtime thing) then it shouldn't be made deterministic with setting the network seed?

> all the other seeds can get driven by the Network seed, but the Process one can't

Currently the process seed is not being set by the build process, but it certainly could be (the builder is what calls `make_step`). The question is whether or not it should be.

hunse:
I had thought it would be useful to be able to change noise just with a reset, but that might just be because I've done a lot of stochastic stuff.

tbekolay:
I think it depends on whether it's part of a node (which to me feels like it should be deterministic) or part of, say, noise injected into an ensemble (which feels like it shouldn't be). I must admit I can't come up with good reasons why either one should / shouldn't be set though, perhaps it should always be deterministic?

tcstewar:
My understanding via a few conversations with @hunse is that the logic for the seed system right now is:

[...]

PR #800, opened 2015-07-27
New processes for convnets

hunse:
I've added a couple new Processes to facilitate running convnets in Nengo, namely one for 2D convolution with a filterbank and one for presenting a series of inputs each for a fixed length of time.

I've also added a seed for random processes, as requested in #787.

This PR is based off the `frozen` branch (see #758).

### DIFF
--- a/nengo/node.py
+++ b/nengo/node.py
@@ -17,6 +17,9 @@ class OutputParam(Parameter):
     def __set__(self, node, output):
         super(OutputParam, self).validate(node, output)
 
+        size_in_set = node.size_in is not None
+        node.size_in = node.size_in if size_in_set else 0
+
         # --- Validate and set the new size_out
         if output is None:
             if node.size_out is not None:
@@ -24,6 +27,8 @@ class OutputParam(Parameter):
                               "'Node.size_in' since 'Node.output=None'")
             node.size_out = node.size_in
         elif isinstance(output, Process):
+            if not size_in_set:
+                node.size_in = output.default_size_in
             if node.size_out is None:
                 node.size_out = output.default_size_out
         elif callable(output):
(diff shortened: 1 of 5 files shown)


Seanny123:
Btw, this pull request is failing tests, but luckily only for Python 3.

hunse:
Yeah it's waiting on dependencies anyway.

hunse:
So we talked about putting this stuff in a separate repository. Any suggestions on whether we should have one repository (`nengo_extra` or something) that has lots of stuff like this, or have a specific one for this stuff?

tbekolay:
Hmm, perhaps these would fit well in [`nengo_deeplearning`](https://github.com/nengo/nengo_deeplearning)? Or a more general `nengo_machinelearning`? Ultimately it's up to you, if you want to have convnet stuff in a specific repo for a while that's cool, it can always be moved to a repo with more other stuff in it later on down the line.

hunse:
I didn't know such a repository existed! That does look like a good place, though ultimately these are a bit more general than just deep learning or machine learning. For example, the `PresentInput` process is useful for any network you want to present a series of inputs to, and `Conv2` could be used with hand-designed Gabor filters as easily as filters that have been learned. But I can put them there for now, as long as people aren't tricked by the name.

tbekolay:
Hmm, `PresentInput` we might want to merge in `nengo`, actually, seems pretty useful. The others maybe in `nengo_deeplearning`. What do you think?

hunse:
Yeah, I'm fine with that. There will have to be a supporting PR for this stuff anyway, since I add a few parameters and some features to Process. So I'll throw PresentInput in there too.

tbekolay:
Great, sounds good! :rainbow:

hunse:
Moved to nengo/nengo_deeplearning#3

Issue #809, opened 2015-08-16
Python 3 Travis-CI errors


xchoo:
A recent commit prompted this error from Travis-CI:

Different tests were collected between gw1 and gw0. The difference is:
```
--- gw1
+++ gw0
@@ -292,86 +292,86 @@
```

According to Trevor, this is because tests run on different cores were returning different results. It's also discussed here:
https://github.com/pytest-dev/pytest/issues/437

Something to do with the way ```parameterize``` uses a python dictionary (which has a non deterministic ordering in Python >= 2.7)

Full job log is here: https://travis-ci.org/nengo/nengo/jobs/75833162

Seanny123:
Did @tbekolay mention a way to fix this?

xchoo:
I think he mentioned switching all of the tests over the ordered lists? Or something of the such.

hunse:
This should be fixed by #830, but if anyone has trouble with this again, we can reopen it.

PR #810, opened 2015-08-16
Spa bg params

xchoo:


xchoo:
Note: Travis-CI is failing because of #809. Actual tests are not failing. :smile:

Seanny123:
#809 was fixed, but this is still breaking with a weirder error that has something to do with miniconda not doing a great job at installing stuff. @tbekolay sorry to ping you again, but have you already seen this error? I seem to remember you mentioning about opening a bug with miniconda?

hunse:
This looks all right to me.

tbekolay:
Needs a changelog entry I think.

xchoo:
@tcstewar Would be the best person to do that. He added some stuff in this PR that I don't use.

jgosmann:
This PR is quite old, marked as needs changes, and misses a description what it does ... :cry: 
So ...
* Is this PR still relevant? (If so, why was there no interest to get this merged sooner?)
* Can someone give me an explanation what this PR does?
* Can someone tell me what changes are required?

Basically, I would like to figure out if I should bring this up-to-date and merge into new-spa or whether I can close it.

tcstewar:
Ah, this PR was a result of wanting to split the GPi and SNr inside the basal ganglia into two separate parts.  The reason to do this was to model Parkinson's (which only affects one, not the other).

As to what we want to do with it....  hmm.  I'd say not to do it.  Instead, this should be a separate basal ganglia available elsewhere (perhaps in nengo_extras), as an example of how people can implement things differently and still use them within a larger spa context.

jgosmann:
ok, I'll then close this as revise and resubmit.

tbekolay:
Note that this is used in a published paper, so it definitely needs to be available somewhere. Please no one delete the branch!

jgosmann:
Would it make sense to create a tag for this branch? To me it seems that a tag is less likely to be accidentally removed ...

tbekolay:
That would make sense! [Tag made](https://github.com/nengo/nengo/releases/tag/spa_bg_params), branch deleted, thanks @jgosmann :)

PR #811, opened 2015-08-19
Updated getting_started.rst

celiasmith:
Put some references to the GUI in getting started docs.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -156,7 +156,9 @@ Probing Nengo objects
 Once you have defined the objects in your model
 and how they're connected,
 you can decide what data you want to collect
-by probing those objects.
+by probing those objects. (In the GUI, you don't define
+probes, but rather right click on network objects and select
+the desired graphs.)
 
 If we wanted to collect data from
 our 2D Ensemble and the Product of those two dimensions::
@@ -175,9 +177,10 @@ Running an experiment
 ---------------------
 
 Once a model has been constructed and we have probed
-certain objects, we can run it to collect data.
+certain objects, we can run it to collect data. (In the GUI, you can
+simply hit the play button.)
 
-To run a model, we must first build a simulator
+To run a model in Python, we must first build a simulator
 based on the model we've defined.
 
 ::


Seanny123:
Looks like a good start until we get around to reviewing the entire documentation in light of the new GUI.

tbekolay:
I think if we do want to add this in, then we should at least put in a link to the GUI's Github page. That said, I'm not sure if it's worth putting these two GUI pointers if they're the only pointers to it... I think ideally I'd prefer we move the Getting Started doc to a more general location that can point to the various subprojects.

Seanny123:
You mean like Nengo.ca?

tbekolay:
Sure, yeah.

Seanny123:
So are you thinking of making it have it's own page? Because right now, it's kind of awkwardly burried in the documentation folder on Nengo.ca. Alternatively, do you want to clean up the Overview page and insert it at the end of that?

tbekolay:
I think the Overview page is right place. I'm not sure we can just copy/paste this in though, as this page wasn't written with the GUI in mind, hence why it feels weird to me to put a few GUI references in there... it'd might need to be rewritten with the GUI in mind. So perhaps that could be the start of the GUI docs, which we could point to instead?

Ideally I'd like to move away from the Drupal site, as it's not very transparent as to what information is there and how to edit it. That'll take a fair bit of time/effort though.

Seanny123:
This discussion seems to escaping the scope of the original issue, so I've marked as discussion and we can deal with it at the next Nengo dev meeting.

tbekolay:
+1 SGTM :)

Seanny123:
I just chatted a bit with @tcstewart and he thinks that switching to Jekyll is worth it, based on the increased security (no more hacks) and stability (no more hosting our own server). I agree with his assessment. 

@tbekolay if you would be so kind as to find a template, I could do the brain dead work of transferring the material 

Actually, maybe before we do that, would you like to have a chat about what functionality you want in the website initially and in the future? We could create a type of road map,  per se.

What do you think?

jgosmann:
The last comment on this PR is more than two years old. Is this still something that should eventually be merged or can we close this PR?

Note that [this page](https://www.nengo.ai/users.html) was created in the meantime.

tbekolay:
Agreed, the new [user guide](https://www.nengo.ai/users.html) obsoletes this PR. Sorry Chris!

Issue #723, opened 2015-05-27
Node API redux


tcstewar:
To follow up on the discussion in #622 now that we have the Process stuff merged in, here's where we currently stand.

You can currently use the Process interface to define a nicely resettable Node that behaves nicely.  Here's what that looks like for making two Nodes: a Delay node and a Steps node that outputs a random sequence of values where each value is kept for a fixed amount of time:

```python
class Steps(nengo.processes.Process):
    def __init__(self, width):
        self.width = width
    def make_step(self, size_in, size_out, dt, rng):
        last_index = np.array([-1])
        last_value = np.zeros(size_out)        
        
        def step(t):
            index = int(t / self.width)
            if index != last_index:
                last_value[:] = rng.uniform(-1, 1, size_out)
                last_index[:] = index
            return last_value
        return step
            
class Delay(nengo.processes.Process):
    def __init__(self, delay):
        self.delay = delay
    def make_step(self, size_in, size_out, dt, rng):
        timesteps = int(self.delay / dt)
        history = np.zeros((timesteps + 1, size_out))
        
        def step(t, x):
            history[:] = np.roll(history, -1)
            history[-1] = x
            return history[0]
        return step
    
model = nengo.Network()
with model:
    stim = nengo.Node(Steps(width=0.1), size_out=1)
    delay = nengo.Node(Delay(delay=0.05), size_in=1, size_out=1)
    nengo.Connection(stim, delay, synapse=None)
    probe1 = nengo.Probe(stim)
    probe2 = nengo.Probe(delay)
```

The question is whether this is the syntax we want people to use to define custom Nodes like this.  The steps with this approach are:
 - you must inherit from ```nengo.processes.Process```
 - you must define a ```make_step(self, size_in, size_out, dt, rng)``` method
 - that method must create state variables and return a function that uses those state variables via the Python closure system (i.e. you are allowed to modify the contents of those variables but you are not allowed to overwrite them)
    - notice that in the Steps case this means doing slightly weird things like ```last_index = np.array([-1])``` instead of ```last_index = -1``` (or the even better ```last_index = None```).

## alternative 1: Use closures, but simplify the syntax a bit

One thing we could do is keep this closure approach, but adjust the syntax so there's fewer things to remember.  The two things that come to mind are:
 - we don't need to inherit from nengo.processes.Process, since we're not using anything from that.  And no one's come up with a good name for what these things should be (```NodeOutput``` is horrible).  
[...]

PR #813, opened 2015-08-19
Run process tests on all backends

hunse:
Use the Simulator fixture in process tests so they run on all
backends. Also flip the signal internal to WhiteSignal, to conform
with our standard orientation (time on the first axis).


tbekolay:
I was going to comment that the `WhiteSignal` change needs a changelog entry, but I realized that we haven't written entries for the `Process` changes yet, which were a pretty big API change. Added a note to #723 so that I make sure to do that.

PR #814, opened 2015-08-19
removed working memory and product config

Seanny123:
As discussed in #790 configs should not be passed as parameters into networks anymore, so I removed them from the Product network and the Working Memory network. 

I would have changed the Action Selection network too, but that depends on #810 .

I'll squash the commits together once everyone agrees on how I proceeded.

### DIFF
--- a/nengo/networks/workingmemory.py
+++ b/nengo/networks/workingmemory.py
@@ -4,26 +4,25 @@ import nengo
 from nengo.networks import EnsembleArray
 
 
-def InputGatedMemory(n_neurons, dimensions, fdbk_scale=1.0, gate_gain=10,
-                     difference_gain=1.0, reset_gain=3,
-                     mem_config=None, net=None):
+def InputGatedMemory(n_neurons, dimensions, feedback=1.0,
+                     difference_gain=1.0, recurrent_synapse=0.1,
+                     difference_synapse=None, net=None):
     """Stores a given vector in memory, with input controlled by a gate."""
     if net is None:
         net = nengo.Network(label="Input Gated Memory")
 
-    if mem_config is None:
-        mem_config = nengo.Config(nengo.Connection)
-        mem_config[nengo.Connection].synapse = nengo.Lowpass(0.1)
+    if difference_synapse is None:
+        difference_synapse = recurrent_synapse
 
     n_total_neurons = n_neurons * dimensions
 
     with net:
         # integrator to store value
-        with mem_config:
-            net.mem = EnsembleArray(n_neurons, dimensions,
-                                    neuron_nodes=True, label="mem")
-            nengo.Connection(net.mem.output, net.mem.input,
-                             transform=fdbk_scale)
+        net.mem = EnsembleArray(n_neurons, dimensions,
+                                neuron_nodes=True, label="mem")
+        nengo.Connection(net.mem.output, net.mem.input,
+                         transform=feedback,
+                         synapse=recurrent_synapse)
 
         # calculate difference between stored value and input
         net.diff = EnsembleArray(n_neurons, dimensions,
@@ -31,21 +30,21 @@ def InputGatedMemory(n_neurons, dimensions, fdbk_scale=1.0, gate_gain=10,
         nengo.Connection(net.mem.output, net.diff.input, transform=-1)
 
         # feed difference into integrator
-        with mem_config:
-            nengo.Connection(net.diff.output, net.mem.input,
-                             transform=difference_gain)
+        nengo.Connection(net.diff.output, net.mem.input,
+                         transform=difference_gain,
+                         synapse=difference_synapse)
 
         # gate difference (if gate==0, update stored value,
         # otherwise retain stored value)
         net.gate = nengo.Node(size_in=1)
         nengo.Connection(net.gate, net.diff.neuron_input,
-                         transform=np.ones((n_total_neurons, 1)) * -gate_gain,
+                         transform=np.ones((n_total_neurons, 1)) * -10,
                          synapse=None)
 
         # reset input (if reset=1, remove all values, and set to 0)
         net.reset = nengo.Node(size_in=1)
         nengo.Connection(net.reset, net.mem.neuron_input,
-                         transform=np.ones((n_total_neurons, 1)) * -reset_gain,
+                         transform=np.ones((n_total_neurons, 1)) * -3,
                          synapse=None)
 
     net.input = net.diff.input
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -57,6 +57,8 @@ Release History
   stand-alone network in ``nengo.networks``. The AssociativeMemory SPA module
   also has an updated argument list.
   (`#702 <https://github.com/nengo/nengo/pull/702>`_)
+- The ``Product`` and ``InputGatedMemory`` networks no longer accept a
+  ``config`` argument. (`#814 <https://github.com/nengo/nengo/pull/814>`_)
 
 **Behavioural changes**
 
--- a/nengo/networks/product.py
+++ b/nengo/networks/product.py
@@ -2,10 +2,9 @@ import numpy as np
 
 import nengo
 from nengo.networks.ensemblearray import EnsembleArray
-from nengo.utils.stdlib import nested
 
 
-def Product(n_neurons, dimensions, input_magnitude=1, config=None, net=None):
+def Product(n_neurons, dimensions, input_magnitude=1, net=None):
     """Computes the element-wise product of two equally sized vectors.
 
     The network used to calculate the product is described in
@@ -18,10 +17,7 @@ def Product(n_neurons, dimensions, input_magnitude=1, config=None, net=None):
     if net is None:
         net = nengo.Network(label="Product")
 
-    if config is None:
-        config = nengo.Config(nengo.Ensemble)
-
-    with nested(net, config):
+    with net:
         net.A = nengo.Node(size_in=dimensions, label="A")
         net.B = nengo.Node(size_in=dimensions, label="B")
         net.output = nengo.Node(size_in=dimensions, label="output")
--- a/nengo/networks/tests/test_product.py
+++ b/nengo/networks/tests/test_product.py
@@ -47,8 +47,9 @@ def test_direct_mode_with_single_neuron(Simulator, plt, seed):
 
     config = nengo.Config(nengo.Ensemble)
     config[nengo.Ensemble].neuron_type = nengo.Direct()
-    product = nengo.networks.Product(
-        1, dim, radius, net=nengo.Network(seed=seed), config=config)
+    with config:
+        product = nengo.networks.Product(
+            1, dim, radius, net=nengo.Network(seed=seed))
 
     func_A = lambda t: np.sqrt(radius)*np.sin(np.arange(1, dim+1)*2*np.pi*t)
     func_B = lambda t: np.sqrt(radius)*np.sin(np.arange(dim, 0, -1)*2*np.pi*t)


nengo/networks/workingmemory.py
```diff
@@ -5,15 +5,13 @@
...
+                     difference_gain=1.0, reset_gain=3, net=None):
     """Stores a given vector in memory, with input controlled by a gate."""
     if net is None:
         net = nengo.Network(label="Input Gated Memory")
 
-    if mem_config is None:
-        mem_config = nengo.Config(nengo.Connection)
-        mem_config[nengo.Connection].synapse = nengo.Lowpass(0.1)
+    mem_config = nengo.Config(nengo.Connection)
+    mem_config[nengo.Connection].synapse = nengo.Lowpass(0.1)
```
tcstewar: If the config parameter is removed (as I completely agree with), this synapse parameters should be exposed.  I'd recommend an argument like ```recurrent_synapse``` and have the default be 0.1.  Hmm, although it looks like the network also uses that same value for the connection from the difference output to the memory input.  Should those be the same value?  Should that be a separate parameter?  If they should be the same value, then is there a better parameter name than ```recurrent_synapse```?

Seanny123: What are common use-cases for changing the `recurrent_synapse` and the `difference_synapse`? Are they usually changed by the same amount?

xchoo: You'd definitely want to be able to chance the synapse of the recurrent connection. 0.1s is a pretty large value, and you can reduce the value for a more responsive (but less accurate) system. 

As for the difference synapse, you'd usually match the two values. But I can see making the difference synapse bigger than the recurrent synapse (you'd want a slower / delayed approach to the input value)

tcstewar: Cool.  How about having ```recurrent_synapse=0.1, difference_synapse=None``` and if ```difference_synapse``` is ```None``` then it defaults to whatever ```recurrent_synapse``` is?

xchoo: Yeah. That works.

Seanny123: Done.

tcstewar: Hmm....  on second thought, another option would be to get rid of ```difference_synapse``` altogether, and just set it to the same as ```recurrent_synapse```.  And maybe call it ```memory_synapse``` instead.

It sounds like the only people who would really want to play with making those parameters different are people mucking with the internals of the network, so they can just do that by changing the object model.  

That said, we don't have a hard and fast rule about where the cutoff of usefulness is, so it's probably fine either way.

xchoo: But because it's a connection property, it's really cumbersome to change the synapse (you'd have to iterate through all of the connections to find the one you want). Unless we change the code to store a reference to the connections.

nengo/networks/workingmemory.py
```diff
@@ -5,15 +5,16 @@
...
     if net is None:
         net = nengo.Network(label="Input Gated Memory")
 
-    if mem_config is None:
-        mem_config = nengo.Config(nengo.Connection)
-        mem_config[nengo.Connection].synapse = nengo.Lowpass(0.1)
+    if difference_synapse is None:
+        difference_synapse = recurrent_synapse
+    mem_config = nengo.Config(nengo.Connection)
+    mem_config[nengo.Connection].synapse = nengo.Lowpass(recurrent_synapse)
```
tcstewar: I don't think this should be done this way.  Get rid of the ```Config``` object completely and just do

```python
        nengo.Connection(net.mem.output, net.mem.input,
                              transform=fdbk_scale, synapse=recurrent_synapse)
```

nengo/networks/workingmemory.py
tcstewar: Hmm, good point... although it's not horrendous to write:

```
for conn in mem.connections:
    if conn.pre == net.diff.output:
        conn.synapse = 0.2
```

xchoo: Blarg. I'd take an additional parameter of an extra for-loop any day. :stuck_out_tongue_closed_eyes:

tcstewar: > Blarg. I'd take an additional parameter over an extra for-loop any day.

It all depends on how often such an additional parameter would be used.  :)

But yeah, I think this is a borderline case.  But we really should remember that we don't need parameters for *everything*, and even something like this isn't that hard to go find in the object model.  So I think it's more of a note that hey, don't worry too much about putting in parameters for everything -- just put in parameters for things that you need to.  After all, we can always add parameters in the future (in a backwards-compatible way) if it turns out we use it a lot.

drasmuss: I'd vote to keep it out of the parameter list, mainly because the
constructor is already pretty bulky. In fact I'd probably take out
reset_gain and gate_gain as well, because those are similarly things that
only a really specific user would want to adjust (I do it very rarely, and
I use input gated memories a lot). But I won't push hard for that.

xchoo: Hmm. That's true. You can effectively modify the reset and gate gain through an external connection anyways. (i.e. if the reset gain is fixed at -5, and you want -10, you just set the connection transform to 2)

Seanny123: So, have we decided to remove the `reset_gain` and  `gate_gain`, but keep the synapse parameters as they current are? Alternatively, should this just be marked as a discussion to continue at the next Nengo-dev meeting?

tcstewar: I'm +1 on removing ```reset_gain``` and ```gate_gain```.  I'm also +1 on removing ```difference_synapse```, but I'm willing to let @xchoo 's vote count stronger than mine. 

Also, speaking of parameters, ```fdbk_scale``` should be changed to be just ``feedback``` (or whatever it's being called in the equivalent case in ```spa.State```), for consistency.

As for discussing at the next Nengo-dev meeting, I think most of the main interested parties are already involved, so hopefully we can sort it out here...  although we might want to move it to the main discussion thread rather than in this comment on an outdated commit....  :)

xchoo: I'd take out ```reset_gain``` and ```gate_gain``` but leave ```difference_synapse``` in there.

tcstewar: Cool... I'm fine with  @xchoo outvoting use 1 to 2....  especially since we're getting fewer parameters anyway, so having one weird one isn't too bad.  :)

drasmuss: :+1:

Seanny123:
Does this look good now? If so, I'll squash up the commits and mark for it to be merged.

nengo/networks/workingmemory.py
```diff
@@ -4,26 +4,25 @@
...
-        with mem_config:
-            net.mem = EnsembleArray(n_neurons, dimensions,
-                                    neuron_nodes=True, label="mem")
-            nengo.Connection(net.mem.output, net.mem.input,
-                             transform=fdbk_scale)
+        net.mem = EnsembleArray(n_neurons, dimensions,
+                                neuron_nodes=True, label="mem")
+        nengo.Connection(net.mem.output, net.mem.input,
+                         transform=fdbk_scale,
+                         synapse=nengo.Lowpass(recurrent_synapse))
```
tcstewar: This should be ```synapse=recurrent_synapse```

nengo/networks/workingmemory.py
```diff
@@ -31,22 +30,20 @@ def InputGatedMemory(n_neurons, dimensions, fdbk_scale=1.0, gate_gain=10,
         nengo.Connection(net.mem.output, net.diff.input, transform=-1)
 
         # feed difference into integrator
-        with mem_config:
-            nengo.Connection(net.diff.output, net.mem.input,
-                             transform=difference_gain)
+        nengo.Connection(net.diff.output, net.mem.input,
+                         transform=difference_gain,
+                         synapse=nengo.Lowpass(difference_synapse))
```
tcstewar: This should be ```synapse=difference_synapse```

nengo/networks/workingmemory.py
```diff
@@ -31,22 +30,20 @@ def InputGatedMemory(n_neurons, dimensions, fdbk_scale=1.0, gate_gain=10,
...
 
         # gate difference (if gate==0, update stored value,
         # otherwise retain stored value)
         net.gate = nengo.Node(size_in=1)
         nengo.Connection(net.gate, net.diff.neuron_input,
-                         transform=np.ones((n_total_neurons, 1)) * -gate_gain,
-                         synapse=None)
+                         transform=np.ones((n_total_neurons, 1)), synapse=None)
```
tcstewar: This should be ```* 10```.  Getting rid of the parameter doesn't mean we should get rid of the effect.

nengo/networks/workingmemory.py
```diff
@@ -31,22 +30,20 @@ def InputGatedMemory(n_neurons, dimensions, fdbk_scale=1.0, gate_gain=10,
...
 
         # reset input (if reset=1, remove all values, and set to 0)
         net.reset = nengo.Node(size_in=1)
         nengo.Connection(net.reset, net.mem.neuron_input,
-                         transform=np.ones((n_total_neurons, 1)) * -reset_gain,
-                         synapse=None)
+                         transform=np.ones((n_total_neurons, 1)), synapse=None)
```
tcstewar: This should be ```* 3```, as above

nengo/networks/workingmemory.py
Seanny123: Just for future reference so I don't make this mistake again, my mental model of the `ptsc` is that the when you pass it a value it turns it into a `nengo.Lowpass` anyways. Looking at the code, I can't find this. What actually happens and where does it happen?

tbekolay: It happens [here in the `SynapseParam`](https://github.com/nengo/nengo/blob/master/nengo/synapses.py#L272). We mostly keep around the ability to pass in a number for convenience and backwards compatibility.

tbekolay: But, it should be noted that you could instead pass in a different `Synapse` class for `difference_synapse` etc, hence why we don't want to unconditionally make it a `Lowpass`.

tcstewar: > But, it should be noted that you could instead pass in a different Synapse class for difference_synapse etc, hence why we don't want to unconditionally make it a Lowpass.

Exactly.  Basically, setting a synapse to a floating point value is a shortcut for one particular type of synapse (Lowpass), but there's no reason to hard-code this to only allow that sort of Synapse.

Seanny123:
Fixed my derps. Thanks @tbekolay and @tcstewar for explaining to me the synapse thing.

nengo/networks/workingmemory.py
```diff
@@ -4,26 +4,25 @@
...
-        with mem_config:
-            net.mem = EnsembleArray(n_neurons, dimensions,
-                                    neuron_nodes=True, label="mem")
-            nengo.Connection(net.mem.output, net.mem.input,
-                             transform=fdbk_scale)
+        net.mem = EnsembleArray(n_neurons, dimensions,
+                                neuron_nodes=True, label="mem")
+        nengo.Connection(net.mem.output, net.mem.input,
+                         transform=feedback,
+                         synapse=nengo.Lowpass(recurrent_synapse))
```
xchoo: This still needs to be changed to ```synapse=recurrent_synapse```.

Seanny123: Okay, I get it now. From now on, whenever the synapse is parameter, you don't cast it as `nengo.Lowpass`.

xchoo: Well, you can. You just don't need to. :smile:

Seanny123:
Squashed and ready to merge.

tbekolay:
Should probably make a note in the changelog about the parameter changes... no need to keep backward compatibility though (I doubt anyone's using these parameters).

Seanny123:
@tbekolay updated the changelog as requested. <3

Seanny123:
@tbekolay I rebased so this is ready to merge

PR #816, opened 2015-08-19
Expose synapse constants for backends

hunse:
This makes it easy for backends to get the coefficients they need to simulate linear synapses.

### DIFF
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -226,12 +226,12 @@ class SlicedCopy(Operator):
         b_slice = self.b_slice
         inc = self.inc
 
-        def step():
+        def step_slicedcopy():
             if inc:
                 b[b_slice] += a[a_slice]
             else:
                 b[b_slice] = a[a_slice]
-        return step
+        return step_slicedcopy
 
 
 class ElementwiseInc(Operator):
--- a/nengo/synapses.py
+++ b/nengo/synapses.py
@@ -1,5 +1,4 @@
 import collections
-import functools
 
 import numpy as np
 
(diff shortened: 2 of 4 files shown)


tbekolay:
Pushed a commit to use callable classes instead of partials. I'm not sure if the base class (`LTIStep`) is actually worthwhile here, we could instead just have each `__init__` store those three things, but probably it's fine either way, so whatever you think think would be easier for Nengo OCL @hunse.

hunse:
Yeah, this is better for sure. I'm fine with the subclassing.

My only concern is that the namespace is a little cluttered, now. We could put the new step classes in LinearFilter, or we could use `__all__` to expose only what we want. Thoughts?

tbekolay:
Hmm... If we put them in `LinearFilter`, I could see renaming them to just `Step`, `NoDen`, `Simple`, `General`, which is still pretty clear to use in other places (e.g., `from nengo.synapses import LinearFilter; my_f = LinearFilter.Genearl(num, den, output)`). So let me try that out.

hunse:
I added a few errors to make sure NoDen and Simple are used correctly.

I think this is good to go. Can we get a second reviewer?

Seanny123:
I could be the second reviewer if you're willing to sit down and explain processes to me. I'm also a bit confused about the "class within a class" pattern you used. I'm not used to that form of OOP.

hunse:
The class within a class idea isn't too complicated. It's just a namespace thing. So rather than those step classes being in the file namespace, they're in the `LinearFilter` class namespace, so you access them as `LinearFilter.NoDen`, etc.

Synapses aren't connected to Processes (yet), so you don't need to understand Processes for this. The way Synapses work right now is in `SimSynapse`, when the `make_step` function is called, it calls out to the `Synapse.make_step` function to get the actual function to call. As part of `Synapse.make_step`, we figure out all the coefficients we need to simulate the synapse. I need those coefficients in Nengo OCL, so this PR makes them available. Rather than `Synapse.make_step` returning a function, it now returns a callable class that has the coefficients (`num` and `den`) available.

Seanny123:
Thanks for explaining that to me. I see now how the "class within a class" thing makes sense. LGTM.

tbekolay:
Cool, I'll merge this then. Think I'll squash it all into the first commit, if that's OK with you @hunse.

hunse:
Sure, fine by me.

PR #817, opened 2015-08-20
Fix SP creation to use correct # of attempts

pblouw:
Fixes a minor bug in the create_pointer method of the Vocabulary class that results in 100 attempts being made to generate an acceptable semantic pointer regardless of the number of attempts specified by the user.

### DIFF
--- a/nengo/spa/vocab.py
+++ b/nengo/spa/vocab.py
@@ -86,18 +86,24 @@ class Vocabulary(object):
         """Create a new semantic pointer.
 
         This will take into account the randomize and max_similarity
-        parameters from self.
+        parameters from self. If a pointer satisfying max_similarity
+        is not generated after the specified number of attempts, the
+        candidate pointer with lowest maximum cosine with all existing
+        pointers is returned.
         """
         if self.randomize:
-            count = 0
-            p = pointer.SemanticPointer(self.dimensions, rng=self.rng)
-            if self.vectors.shape[0] > 0:
-                while count < 100:
-                    similarity = np.dot(self.vectors, p.v)
-                    if max(similarity) < self.max_similarity:
-                        break
-                    p = pointer.SemanticPointer(self.dimensions, rng=self.rng)
-                    count += 1
+            if self.vectors.shape[0] == 0:
+                p = pointer.SemanticPointer(self.dimensions, rng=self.rng)
+            else:
+                p_sim = np.inf
+                for _ in range(attempts):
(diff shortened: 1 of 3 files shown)


nengo/spa/vocab.py
```diff
@@ -92,7 +92,7 @@ def create_pointer(self, attempts=100, unitary=False):
             count = 0
             p = pointer.SemanticPointer(self.dimensions, rng=self.rng)
             if self.vectors.shape[0] > 0:
-                while count < 100:
+                while count < attempts:
```
tcstewar: Damnit!  That bug has been in the code for *years*.   Good find!

jgosmann: Did anyone ever specify a different number of attempts? :stuck_out_tongue:

tcstewar: Not that I know of.

Oh, and while I'm thinking about it, every now and then I realize that this attempt thing really should keep track of the best one that it finds, and use that if it can't find one smaller than max_similarity.  In other words, it could at least do the best it can, rather than just sticking with the last one if it fails.....

pblouw:
I suppose if you're trying keep the dimensionality of your SP's low for computational efficiency but want to use a relatively large number of them, it might be useful to increase the number of attempts. +1 for tracking the using the best candidate SP if the max_similarity constraint is not satisfied though.

nengo/spa/vocab.py
pblouw: Ok, I've updated this pull request with a new commit that includes Terry's suggestion. I checked that the best pointer is saved properly using a few print statements, but feel free to pass along any suggestions for improvements.

hunse:
I fixed up the `create_pointer` function to be a little simpler. I also added a test for the warning. Looks great!

tbekolay:
Agreed, looks good! Should this get a changelog entry?

hunse:
Hmm, I guess it did fix a bug and change behaviour a little. Probably best to get in the habit of adding them, even for little things like this.

hunse:
Okay, added the changelog entry. Travis has gone INSANE!

tbekolay:
> Travis has gone INSANE!

I believe that's because we're depending on `IPython[notebook]` which now no longer gives you the IPython notebok. You have to pip install jupyter. #693 fixes this, which I'm merging shortly, so should be no worries.

hunse:
@tbekolay, do you want to merge this one, since I re-wrote a good chunk of it?

tbekolay:
Sounds good, will look at this now.

PR #819, opened 2015-08-20
Added digital synapses, fixed filter test

hunse:
Added ability to specify linear synapses with digital coefficients
directly, since it can often be easier to design filters directly
in the digital domain.

Fixed the test of general linear filters to use much more stable
(less extreme) coefficients, so that backends with more limited
precision can still pass the test.

### DIFF
--- a/nengo/synapses.py
+++ b/nengo/synapses.py
@@ -34,9 +34,10 @@ class LinearFilter(Synapse):
     .. [1] http://en.wikipedia.org/wiki/Filter_%28signal_processing%29
     """
 
-    def __init__(self, num, den):
+    def __init__(self, num, den, analog=True):
         self.num = num
         self.den = den
+        self.analog = analog
 
     def __repr__(self):
         return "%s(%s, %s)" % (self.__class__.__name__, self.num, self.den)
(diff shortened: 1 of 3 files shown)


tbekolay:
Haha, I have the exact same change in one of my local branches, only with `discrete=False` as the default instead of `analog=True`. I have no real preference once way or the other, though, so this looks good to me!

nengo/synapses.py
```diff
@@ -72,8 +73,13 @@ def general_step(signal, output, x, y, num, den):
...
     def make_step(self, dt, output, method='zoh'):
-        num, den, _ = cont2discrete((self.num, self.den), dt, method=method)
-        num = num.flatten()
+        num, den = self.num, self.den
+        if self.analog:
+            num, den, _ = cont2discrete((num, den), dt, method=method)
+            num = num.flatten()
+
+        if den[0] != 1.:
+            raise ValueError("First element of the denominator must be 1")
```
tbekolay: [`scipy.lifter`](http://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.lfilter.html) handles this by normalizing `num` and `den` by `den[0]` if it isn't 1. Should we adopt that instead? Personally, I spent a buttload of time trying to debug my filter because of this issue, so I'm in favour of the `ValueError`, just wanted to raise it as another possibility.

tbekolay:
LGTM! Just curious about your thoughts on the inline comment.

nengo/synapses.py
hunse: I would just keep the ValueError, because it's nice and explicit if users have to normalize it themselves. Any filters designed using the Scipy functions should already have this property, anyway.

tbekolay: Cool, agreed.

tbekolay:
Added changelog entry (11ad897efd32a28bdd061a3a0). Will merge on your +1 @hunse!

hunse:
Looks good!

PR #821, opened 2015-08-21
Add changelog entry for product improvement

tbekolay:


jgosmann:
LGTM :)

hunse:
I think we should "officially" make it the job of the person submitting the PR to add a changelog entry, and the job of the reviewer to check that there's a changelog entry and ask the author to add one if not. Then the merger can be a second check.

jgosmann:
Can we use the Github API to implement a check whether a PR has a changelog entry? :wrench:

hunse:
I think it would be hard to do with the Github API, because lots of PRs don't require entries, and I'm not sure how to detect the difference. Maybe changes to some core files would make it check for a changelog entry? Anyway, I think it would probably be more trouble than it's worth.

PR #823, opened 2015-08-27
Fixed exception on negative compare and improved error

Seanny123:
Found these bugs while creating the Tower of Hanoi. Also addressed, #759 way more directly.

### DIFF
--- a/nengo/spa/basalganglia.py
+++ b/nengo/spa/basalganglia.py
@@ -113,7 +113,9 @@ class BasalGanglia(Module):
         scale : float
             a scaling factor to be applied to the result
         """
-        raise NotImplementedError('Compare connections not implemented yet')
+        raise NotImplementedError('Compare between two sources will never be '
+                                  'implemented as discussed in '
+                                  'https://github.com/nengo/nengo/issues/759')
 
     def add_dot_input(self, index, source, symbol, scale):
         """Make an input that is the dot product of a Source and a Symbol.
@@ -164,7 +166,7 @@ class BasalGanglia(Module):
                                       "of 1 can be scalar inputs")
 
         try:
-            scale = float(source.transform.symbol)
+            scale = float(eval(source.transform.symbol))
         except ValueError:
             raise ValueError("Transform can only be a scalar value"
                              " the value %s is invalid"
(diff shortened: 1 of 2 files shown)


nengo/spa/basalganglia.py
```diff
@@ -164,7 +166,7 @@ def add_scalar_input(self, index, source):
                                       "of 1 can be scalar inputs")
 
         try:
-            scale = float(source.transform.symbol)
+            scale = float(eval(source.transform.symbol))
```
tcstewar: Ah, right...  because ```source.transform.symbol``` could be ```-1 * 2```.....  good catch.

tcstewar:
LGTM

nengo/spa/tests/test_basalganglia.py
```diff
@@ -22,6 +22,7 @@ def test_basal_ganglia(Simulator, seed, plt):
             'dot(vision,CAT)+0.5-dot(vision,CAT) --> motor=E',
             'dot(vision,PARROT) + compare --> motor=F',
             '0.5*dot(vision,MOUSE) + 0.5*compare --> motor=G',
+            '( dot(vision,MOUSE) - compare) * 0.5 --> motor=H'
```
jgosmann: Does the space after the opening parenthesis have any significance?

Seanny123: Nope, I just thought it made it more legible

jgosmann: Not sure, if I agree ... and even if shouldn't be there a space in front of the closing parenthesis for symmetry reasons too?

And given that are discussing legibility: Why are there no spaces after commas and around operators? (I also don't like to write semantic pointers in all uppercase, it's so 90ies HTML like. But I don't have hope to get this one changed.)

tbekolay: Either way's fine, but if you have a leading space, also have a following space. I.e., `( expr )` rather than `( expr)` as it is now.

jgosmann: But I would be fine with merging the PR as is (there might be more important things than the formatting and most of it was that way already before).

Seanny123: Good point, I'll fix all those things

tcstewar: > Why are there no spaces after commas and around operators?

That's mostly my fault, as all the examples I wrote started off that way.  I've switched over to "spaces after commas, but optional spaces around operators as per readability" which is my current understanding of PEP8.

Seanny123:
Fixed the awkward formatting problems. Ready to merge.

PR #824, opened 2015-08-31
Added seed for random processes

hunse:
This allows processes with a random element to be seeded, so that
results are repeatable from one simulation to the next (though the
same results are not guaranteed for different backends).

Addresses #787.

(Since I'm not sure when #800 will be merged, I brought this out as a separate PR. It's based on #758. Just the last commit is new.)

### DIFF
--- a/nengo/builder/processes.py
+++ b/nengo/builder/processes.py
@@ -31,6 +31,7 @@ class SimProcess(Operator):
         output = signals[self.output] if self.output is not None else None
         size_in = input.size if input is not None else 0
         size_out = output.size if output is not None else 0
+        rng = self.process.get_rng(rng)
         step_f = self.process.make_step(size_in, size_out, dt, rng)
         inc = self.inc
 
--- a/nengo/processes.py
+++ b/nengo/processes.py
@@ -16,19 +16,29 @@ class Process(FrozenObject):
     Attributes
     ----------
     default_size_out : int
-        If `d` isn't specified in `run` or `run_steps`, this will be used.
+        Sets the default size out for nodes running this process. Also,
+        if `d` isn't specified in `run` or `run_steps`, this will be used.
         Default: 1.
     default_dt : float
         If `dt` isn't specified in `run`, `run_steps`, `ntrange`, or `trange`,
         this will be used. Default: 0.001 (1 millisecond).
+    seed : int, optional
+        Random number seed. Ensures noise will be the same each run.
     """
     default_size_out = IntParam(low=0)
     default_dt = NumberParam(low=0, low_open=True)
+    seed = IntParam(low=0, high=npext.maxint, optional=True)
 
-    def __init__(self):
+    def __init__(self, seed=None):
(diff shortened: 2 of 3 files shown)


tbekolay:
Rebased! I have a few questions about this PR.

First, I'm not super clear on the distinction between `get_rng` and `get_sim_rng`; you mention that it's for step order independence, but I'm not sure what that means (likely users won't know either). I would guess that people making subclasses won't know which one to use, so we should do something to clarify that.

Second, should the process seed be included as something that gets seeded when the top-level network seed is set? My instinct says yes.

hunse:
`get_sim_rng` is intended when you need an RNG to use inside a step function. I wanted the state of all the RNGs to be separate, so you'll get the same output from each process whether you call one process's step function 100 times, then another's 100 times, or you call each of them alternately for 100 times. So `get_sim_rng` takes the builder RNG and creates a new RNG.

`get_rng` also takes the builder RNG, but only returns a new RNG if it has to because the seed is set. Otherwise, it will just give you the builder RNG back. This is used if you need an RNG to build part of a process before simulation, see for example `WhiteSignal`.

I agree that this is confusing. I don't see any problem with having one function that's like `get_sim_rng`, since it will never hurt us to make another RNG (just a little overhead).

hunse:
Right now, we still have the ability to have simulator seeds different from builder seeds, so that you can call `sim.reset` and change the seed to change all the processes, while still keeping encoders, decoders, gains and biases, etc., the same.

I would be fine, however, that if the user provides a top-level network seed, and doesn't set a simulator seed, we set the simulator seed to the same thing. But if you have a process in a node in a sub-network, and set the seed on the sub-network, than that wouldn't change the process. This different behaviour could be confusing. Completely getting rid of the simulator seed is the other option, but that's a significant change. Also, I would be more open to getting rid of the simulator seed once we have a way to change/rebuild nodes without rebuilding the whole model.

hunse:
Okay, pushed a change that handles the RNG generation behind the scenes, so users don't have to do it in `make_step`. This should also make it easier if we want to switch to one seed system in the future (i.e. where there is no simulator seed, and process seeds are set during build by parent networks etc.). But switching to that type of system is a big change, so I think we should do that as another PR if that's what we want.

tbekolay:
Yeah, definitely let's not worry about the one seed thing at the moment; with the fixup, this looks great to me! Would like another review on this just in case; @drasmuss perhaps?

drasmuss:
Looks good to me.  The only thing that occurred to me, is there any value in letting people pass a `RandomState` object instead of an int if they want? It'd make it easier to coordinate the rng inside and outside of Nengo, but it's not like it's hard either way.  I think it's fine the way it is, just bringing up the possibility in case someone thinks "yeah I really want that!"

tbekolay:
Hmm, we can think about making it possible to specify the `rng` later; because of the way we assign seeds it's not a one-line change, so I think it'd take a bit more thought. Merging!

PR #828, opened 2015-09-01
fixed exception handling in spa.Module.__exit__

### LINKED ISSUE #827: Adding modules with errors to an SPA network obscures the real error message.
Exceptions thrown within modules are not passed along to the user. Instead, the user just gets a very unhelpful message:
```
ValueError: <ModuleName> was not added as an attribute to of                     the SPA network and won't be detected
```
(Note: the extra spacing is also a bug)

I think this is what is happening:
1. Module is added to the SPA network (nengo registers the add successfully)
2. ```__setattr__``` function is called
3. exception is thrown (and module is not added to the SPA network
4. ```__exit__``` function is called (because of the exception)
5. Which can't find the recently added network in the modules list
6. ValueError is thrown.

Test example:
```
class TestModule(Module):
    def __init__(self, label="Test", seed=None, add_to_container=None):
        super(TestModule, self).__init__(label, seed, add_to_container)
        with self:
            raise Exception("EXCEPTION!")

with spa.SPA() as model:
    model.test = TestModule()

sim = nengo.Simulator(model)
```

Error message:
```
ValueError: <TestModule "Test"> was not added as an attribute to of                     the SPA network and won't be detected
```

tcstewar:
This fixes #827 

Note that the change to ```Input``` is because the last test case in ```test_spa.py::test_spa_verification``` was actually relying on the buggy ValueError throwing that is fixed by this PR.  If ```spa.Input``` couldn't find an object with the name it was looking for, then it would actually throw a TypeError which would then get converted into a ValueError, obscuring the original bug.  So I also fixed ```spa.Input``` to give a useful ValueError in this case.

### DIFF
--- a/nengo/spa/input.py
+++ b/nengo/spa/input.py
@@ -40,7 +40,10 @@ class Input(Module):
         Module.on_add(self, spa)
 
         for name, value in iteritems(self.kwargs):
-            target, vocab = spa.get_module_input(name)
+            try:
+                target, vocab = spa.get_module_input(name)
+            except TypeError:
+                raise ValueError('Could not find input called "%s"' % name)
             if callable(value):
                 val = make_parse_func(value, vocab)
             else:
--- a/nengo/spa/input.py
+++ b/nengo/spa/input.py
@@ -40,10 +40,7 @@ class Input(Module):
         Module.on_add(self, spa)
 
         for name, value in iteritems(self.kwargs):
-            try:
-                target, vocab = spa.get_module_input(name)
-            except TypeError:
-                raise ValueError('Could not find input called "%s"' % name)
+            target, vocab = spa.get_module_input(name)
             if callable(value):
                 val = make_parse_func(value, vocab)
             else:
--- a/nengo/spa/spa.py
+++ b/nengo/spa/spa.py
@@ -104,8 +104,12 @@ class SPA(nengo.Network):
 
             value.on_add(self)
 
-    def __exit__(self, *args):
-        super(SPA, self).__exit__(*args)
+    def __exit__(self, ex_type, ex_value, traceback):
+        super(SPA, self).__exit__(ex_type, ex_value, traceback)
+        if ex_type is not None:
+            # re-raise the exception that triggered this __exit__
+            return False
+
         module_list = frozenset(self._modules.values())
         for net in self.networks:
             # Since there are no attributes to distinguish what's been added
--- a/nengo/spa/spa.py
+++ b/nengo/spa/spa.py
@@ -125,7 +125,9 @@ class SPA(nengo.Network):
             return self._modules[name]
         elif '_' in name:
             module, name = name.rsplit('_', 1)
-            return self._modules[module]
+            if module in self._modules:
+                return self._modules[module]
+        raise KeyError('Could not find module "%s"' % name)
 
     def get_default_vocab(self, dimensions):
         """Return a Vocabulary with the desired dimensions.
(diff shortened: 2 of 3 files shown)


Seanny123:
LGTM, other than the PEP8 error.

BTW, the error is ` ./basalganglia.py:35:1: C901 'BasalGanglia.on_add' is too complex (10)`.

I'm tempted to ignore this, but is this indication that we should open an issue to simplify and/or refactor `basal_ganglia.py`? Admittedly, it does have a lot of if-statements, but I don't really know how to fix that... Any thoughts?

jgosmann:
This is one of the situations where a statically typed language supporting overloading would be nice. I might give it some thought how to refactor. The first thing I would do is take everything in [this if clause](https://github.com/nengo/nengo/blob/master/nengo/spa/basalganglia.py#L56) and put it into it's own function (not sure how to name it because there is already a function `add_dot_input`, maybe rename that one ...). That would reduces the length of the `on_add` function significantly and makes it a lot easier to understand with only one layer of if-clauses (though still two nested for loops). The extracted function would also be not extremely long, but the if nesting in there might still be cleaned up.

nengo/spa/input.py
```diff
@@ -40,7 +40,10 @@ def on_add(self, spa):
         Module.on_add(self, spa)
 
         for name, value in iteritems(self.kwargs):
-            target, vocab = spa.get_module_input(name)
+            try:
+                target, vocab = spa.get_module_input(name)
+            except TypeError:
+                raise ValueError('Could not find input called "%s"' % name)
```
hunse: We do a lot of this in SPA: catching errors and turning them into other ones. It bit me in the butt yesterday, when some code I was writing had a NameError in it, which then got caught and turned into something different. I think we should avoid it when possible. Here, can't we just have `spa.get_module_input` return the correct ValueError if it can't find `name`? That seems much cleaner to me.

jgosmann: And if we do this conversion, we should probably do it in `spa.get_module_input` to have a consistent behaviour.

tcstewar: Yup, I'd be onboard with changing the ```spa.get_*``` stuff so that they raise exceptions rather than returning None (which is what they do now).  Should I add that to this PR or start a new one?

hunse: I'm fine adding it to this. You might want to make those changes first, then rebase this so it comes after.

tcstewar:
> BTW, the error is ./basalganglia.py:35:1: C901 'BasalGanglia.on_add' is too complex (10).
> 
> I'm tempted to ignore this, but is this indication that we should open an issue to simplify and/or refactor basal_ganglia.py? Admittedly, it does have a lot of if-statements, but I don't really know how to fix that... Any thoughts?

Well, seeing as how ```basalganglia.py``` wasn't touched by this PR, I'm not that stressed about it right now.  :)  But I do agree that some refactoring in there along the lines of what @jgosmann said makes sense.  But it'd be a separate PR.

hunse:
Yeah, the PEP8 looks fine to me both locally and on Travis, so I'm not quite sure what @Seanny123 is talking about; something for another PR, anyway.

jgosmann:
It's probably not a PEP8 error but a pylint error.

tcstewar:
The changes to the exception handling turned out to be not too bad, I think.  I think KeyError is the right error in this case, but I'd also be up for something else if anyone has a good suggestion.

hunse:
Yeah, that looks good.

xchoo:
LGTM. Should we wait for the Travis stuff to be sorted out before merging though?

hunse:
Rebased, so hopefully that fixes Travis. I think this is probably fine without a changelog message, since it doesn't fix a bug per se, just makes the exception clearer.

tbekolay:
> I think this is probably fine without a changelog message

Agreed!

hunse:
> It's probably not a PEP8 error but a pylint error.

@jgosmann: when I said PEP8 I really meant `flake8`. I still couldn't find the error...

jgosmann:
Hm, me neither; I only get a complexity rating of 7. But those numbers should be taking with a grain of salt anyways. I would still welcome a PR doing a little bit refactoring on this function.

PR #830, opened 2015-09-02
Sort test_example example lists

### LINKED ISSUE #809: Python 3 Travis-CI errors

tbekolay:
So that test collection is deterministic. Should fix #809.

### DIFF
--- a/nengo/tests/test_examples.py
+++ b/nengo/tests/test_examples.py
@@ -48,6 +48,11 @@ for subdir, _, files in os.walk(examples_dir):
     fast_examples.extend([e for e, f in zip(examples, files)
                           if os.path.splitext(f)[0] not in too_slow])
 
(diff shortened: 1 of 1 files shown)


hunse:
Looks good. This is very minor, so I'll just merge it.

Issue #806, opened 2015-08-12
IPython 4/Jupyter adjustments


jgosmann:
Check the example notebooks in IPython 4/Jupyter. Apparently the markdown parser changed and requires now a space between the hashtags and the heading (that is `# foo` instead of `#foo`). There might be other adjustments necessary too.

tbekolay:
I tested all of our notebooks in Jupyter and didn't see any issues. The only thing that came up was a warning in the console: `Deprecated widget Javascript path /static/widgets/js/*.js was used` However, looking through `nengo/ipynb.py` I can't see anywhere that we're doing this explicitly, so I'm guessing this is an internal notebook thing and not our fault. So, closing this! If we do run into any issues later, we can make new issues for them.

PR #833, opened 2015-09-02
Insert spaces after # of markdown heading/comment.

jgosmann:
To have headings displayed correctly in Jupyter. Looked over all the examples and everything else seems to be rendered fine.
Addresses #806.

### DIFF
--- a/examples/basic/addition.ipynb
+++ b/examples/basic/addition.ipynb
@@ -12,7 +12,7 @@
      "cell_type": "markdown",
      "metadata": {},
      "source": [
-      "#Nengo Example: Addition"
+      "# Nengo Example: Addition"
      ]
     },
     {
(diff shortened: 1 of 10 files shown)


Seanny123:
LGTM

tbekolay:
LGTM too! One minor thing is that these changes should change the notebook's signature. But I think the signatures don't really matter at this point, you just get a minor warning in the console... this might change in the future though, but for now, I'll just merge.

hunse:
> One minor thing is that these changes should change the notebook's signature.

It would be nice if there was some way to build it into our `clearoutputs.py` script, for example. There must be some IPython function somewhere for finding the hash.

jgosmann:
After reading [the documentation on the notebook's signature](https://ipython.org/ipython-doc/dev/notebook/security.html#the-details-of-trust) I don't think that it matters in any way. The value depends on the profile and when loaded in any other profile the notebook will be untrusted anyways (and because the secret key is unknown in that other profile, there is no way to know whether the signature was the actual signature or not).

hunse:
Oh yeah, I didn't realize there's a secret key in there too.

PR #838, opened 2015-09-10
Handle DoctestItems in conftest.py.

tbekolay:
DoctestItems do not have an 'obj' attribute, so our attempts to read it make it impossible to test.

Note that Nengo doesn't use DoctestItems, but `nengo_spinnaker` does. Maybe in the future we'll want to too, so best to make it possible.

### DIFF
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -194,6 +194,8 @@ def pytest_generate_tests(metafunc):
 
 
 def pytest_runtest_setup(item):
+    if not hasattr(item, 'obj'):
+        return
     for mark, option, message in [
             ('example', 'noexamples', "examples not requested"),
             ('slow', 'slow', "slow tests not requested")]:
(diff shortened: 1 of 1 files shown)


hunse:
I have no idea what DoctestItems are, but this looks fine to me. It's just skipping things that would result in an exception anyway, so it seems pretty innocuous.

tbekolay:
DoctestItems are what you get if you tell py.test that you also want to run doctests. Doctests are any code showing up in docstrings, in a particular form. [Python docs](https://docs.python.org/2/library/doctest.html).

hunse:
I see... wouldn't you need full code examples in all of your docstrings, then? We use snippets a lot, and I'm not sure how you could test those, other than doing some simple syntax checks. But maybe that's all they do.

Anyway, this is small, so I don't think we need a second reviewer. I'll just go ahead and merge it.

PR #839, opened 2015-09-12
Corrected regularization and rmse in non-negative solvers

arvoelke:
Previously, the `'rmses`' for the non-negative solvers were off by an order of magnitude, because they were with respect to the Gram system instead of the original system. This has been fixed to make these values now comparable to the other solvers.

This fix has been tested offline using `nengo.utils.connection.eval_point_decoding` on the same set of eval points (now gives an exact match). If desired, I could also add a unit test to check that each solver's rmse is correct with respect to the given system.

Also, the regularization wasn't scaled consistently with the other solvers. This was causing drastically different performance results when using one solver over another (I also wonder if this contributed in part to the performance improvement in #321). I verified the fix (again offline) by comparing the RMSE when decoding a positive constant function -- they were roughly the same across a variety of regularization coefficients. Again, this can be made into a unit test if desired.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -133,7 +133,7 @@ Release History
 - Fix for SPA actions that route to a module with multiple inputs.
   (`#714 <https://github.com/nengo/nengo/pull/714>`_)
 - Corrected the ``rmses`` values in ``BuiltConnection.solver_info`` when using
-  ``NNls`` and ``Nnl2sL2`` solvers.
+  ``NNls`` and ``Nnl2sL2`` solvers, and the ``reg`` argument for ``Nnl2sL2``.
   (`#839 <https://github.com/nengo/nengo/pull/839>`_)
 
 2.0.1 (January 27, 2015)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -132,6 +132,9 @@ Release History
   `#671 <https://github.com/nengo/nengo/pull/671>`_)
 - Fix for SPA actions that route to a module with multiple inputs.
   (`#714 <https://github.com/nengo/nengo/pull/714>`_)
+- Corrected the ``rmses`` values in ``BuiltConnection.solver_info`` when using
+  ``NNls`` and ``Nnl2sL2`` solvers.
+  (`#839 <https://github.com/nengo/nengo/pull/839>`_)
 
 2.0.1 (January 27, 2015)
 ========================
--- a/nengo/solvers.py
+++ b/nengo/solvers.py
@@ -517,7 +517,7 @@ class NnlsL2(Nnls):
         # form Gram matrix so we can add regularization
         GA = np.dot(A.T, A)
         GY = np.dot(A.T, Y)
-        np.fill_diagonal(GA, GA.diagonal() + sigma)
+        np.fill_diagonal(GA, GA.diagonal() + A.shape[0] * sigma**2)
         X, info = super(NnlsL2, self).__call__(GA, GY, rng=rng, E=E)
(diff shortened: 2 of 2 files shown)


jgosmann:
This doesn't effect the standard solver, does it?

arvoelke:
Nope. It just makes the non-negative solvers comparable in regularization/rmse to the standard solver.

nengo/solvers.py
```diff
@@ -19,6 +19,11 @@
 logger = logging.getLogger(__name__)
 
 
+def _rmses(A, Y, X):
```
hunse: Not a big deal, but I would prefer the order of arguments here to be A, X, Y or Y, A, X, probably the first since the system is typically written AX = Y.

nengo/solvers.py
```diff
@@ -515,39 +513,30 @@ def __init__(self, weights=False, reg=0.1):
...
-        G = np.dot(A.T, A)
-        Y = np.dot(A.T, Y)
-        np.fill_diagonal(G, G.diagonal() + sigma)
-        return super(NnlsL2, self).__call__(G, Y, rng=rng, E=E)
+        GA = np.dot(A.T, A)
+        GY = np.dot(A.T, Y)
+        np.fill_diagonal(GA, GA.diagonal() + A.shape[0] * sigma**2)
+        X, info = super(NnlsL2, self).__call__(GA, GY, rng=rng, E=E)
+        # recompute the RMSE in terms of the original matrices
+        info = {'rmses': _rmses(A, Y, X), 'gram_rmses': info['rmses']}
```
hunse: The original `info` could have other values that users would want. I'd suggest doing `info = {'rmses': _rmses(A, Y, X), 'gram_info': info}, just so we keep everything around.

CHANGES.rst
```diff
@@ -132,6 +132,10 @@ Release History
   `#671 <https://github.com/nengo/nengo/pull/671>`_)
 - Fix for SPA actions that route to a module with multiple inputs.
   (`#714 <https://github.com/nengo/nengo/pull/714>`_)
+- Corrected the `'rmses'` values in `BuiltConnection.solver_info` when using
+  `NNls` and `Nnl2sL2` solvers. Previous values were for the Gram matrix, and
```
tbekolay: I would cut this entry after the first sentence, and link to the PR.

The changelog isn't designed to replace what already happens in Github and in commit messages, it's to summarize what's happened as briefly as possible; if people want to know more, then the Github stuff is there.

Also, note that this file is reStructuredText, not markdown, so you need double backticks to render the text as code.

nengo/solvers.py
arvoelke: Done

nengo/solvers.py
arvoelke: Done

CHANGES.rst
arvoelke: If I were to cut it off after the first sentence, then there would be nothing about regularization. So I merged the third sentence into the first sentence and removed the second. Also added double backticks and linked this PR. That said, I wasn't really trying to summarize git; the second sentence was there because there is a separate but connected user-facing change (i.e. `solver_info` now has another key). But I can buy the argument that this is not actually user-facing since nobody yet uses those solvers.

arvoelke:
Made suggested changes.

hunse:
This looks good to me.

tbekolay:
Me too, just needs some squashin'. You want to merge @hunse?

hunse:
sure

PR #841, opened 2015-09-14
Fix unformatted connection signal names

hunse:
We forgot to actually put the connection in, leaving a mysterious `%s`.

This is very minor, so if you're happy @tbekolay, merge away.


tbekolay:
Noticed this yesterday too while debugging. Merging!

hunse:
Looks like I introduced this when I combined the transform and decoders. Whoops!

PR #843, opened 2015-09-16
Ensure learning rule test does not affect others

hunse:
The function `learning_net` in test_learning_rules.py used to set
the learning rate on the PES class itself, which led to changes
affecting other tests when run together.

Everyone should take a look at how this error occurred, since it's an easy way to create problems that are really hard to find. If you change an attribute on a class itself, this will change will remain for **all tests that follow**. So we should never do this.


Seanny123:
LGTM. Will not change attributes on classes in the future.

tbekolay:
This got merged with #758.

PR #844, opened 2015-09-17
Compat fixes

hunse:
A couple minor improvements to `compat.py`. The `as_ascii` function is useful in Nengo OCL, so I thought it might be good to have it here in case we need it in Nengo or other backends down the road.

### DIFF
--- a/nengo/utils/compat.py
+++ b/nengo/utils/compat.py
@@ -80,7 +80,10 @@ def is_integer(obj):
 
 
 def is_iterable(obj):
-    return isinstance(obj, collections.Iterable)
+    if isinstance(obj, np.ndarray):
+        return obj.ndim > 0  # 0-d arrays give error if iterated over
+    else:
+        return isinstance(obj, collections.Iterable)
 
 
 def is_number(obj, check_complex=False):


nengo/utils/compat.py
```diff
@@ -75,12 +75,24 @@ def ensure_bytes(s):
...
+        return string
+
+
 def is_integer(obj):
     return isinstance(obj, int_types + (np.integer,))
 
 
 def is_iterable(obj):
-    return isinstance(obj, collections.Iterable)
+    if isinstance(obj, np.ndarray):
```
tbekolay: Will this actually be true for 0-d arrays? I thought that they were instances of `np.generic` but not `np.ndarray`. Weird that NumPy scalars are still iterable if they can't be iterated over!

hunse: No, there's a difference between Numpy scalars (e.g. `np.int32(0)`) and 0-d Numpy arrays (e.g. `np.array(0, dtype=np.int32)`).

tbekolay: Of course there is ;)

nengo/utils/compat.py
```diff
@@ -75,12 +75,24 @@ def ensure_bytes(s):
 assert TextIO
 
 
+def as_ascii(string):
+    if not PY2 and isinstance(string, bytes):  # Python 3
```
tbekolay: Is the `isinstance` check here necessary? By default, I believe, the `str` type in Python 3 is unicode, so those would not be turned to ascii... I'll write a few tests for this, perhaps.

hunse:
I'm not really convinced anymore that the `as_ascii` function will be useful in other backends, so I'm just going to take it out. So this is just the `is_iterable` change now.

hunse:
Also, this is very small so I don't think it needs a second reviewer (though if someone wants to look at it quick).

PR #846, opened 2015-09-17
Added vocabs as argument to spa

### LINKED ISSUE #774: Specify default vocabularies when create spa.SPA object
Currently the way to specify the default vocabulary for a given dimension of a SPA model is to do:

```
D = 64   # number of dimensions
vocab = spa.Vocabulary(D, ...)

model = spa.SPA(...)
model._default_vocabs[D] = vocab
```

It would be nice to be able to pass the vocabulary into the constructor of the SPA object.

Seanny123:
Resolves #774

### DIFF
--- a/nengo/spa/spa.py
+++ b/nengo/spa/spa.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy as np
 
 import nengo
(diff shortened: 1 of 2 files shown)


tcstewar:
Nice...  this looks good to me and adds some nice functionality.  :)

nengo/spa/spa.py
```diff
@@ -76,12 +78,20 @@ def __init__(self):
...
+    def __init__(self, label=None, seed=None, add_to_container=None,
+                 vocabs=[]):
         super(SPA, self).__init__(label, seed, add_to_container)
         enable_spa_params(self)
         self._modules = {}
         self._default_vocabs = {}
 
+        for vo in vocabs:
+            if vo.dimensions in self._default_vocabs:
+                warnings.warn("Duplicate vocabularies with dimension:%i. "
```
tbekolay: I had no idea `%i` is a thing... we usually use `%d` for integers. What's the difference between the two?

Also I'd also lose the colon there, we don't use it in any other place that we print dimensionality.

tbekolay:
This looks good to me; will merge when the inline comment is addressed. Or I can address it in the merge if that's okay with @Seanny123!

Seanny123:
Go ahead and address it! (:

tbekolay:
Cool, doing so now. Dope tests by the way!

PR #849, opened 2015-09-18
Added wrapper around bind.py

### LINKED ISSUE #847: Create a spa wrapper for the Circular Convolution network
In Nengo GUI, I want to be able to visualise the output of a CircularConvolution network in the same way that I visualise the output of all other SPA modules.
 
- Is this a good idea?
- What should this module be named? `spa.CircularConvolution`?

Seanny123:
Resolves #847 and [#452](https://github.com/nengo/nengo_gui/issues/452).

### DIFF
--- a/nengo/spa/bind.py
+++ b/nengo/spa/bind.py
@@ -1,0 +1,45 @@
+import nengo
+from nengo.spa.module import Module
+
+
+class Bind(Module):
+    """A module for binding together both inputs
+
+    Parameters
+    ----------
+    dimensions : int
+        Number of dimensions for the two vectors to be compared
+    vocab : Vocabulary, optional
+        The vocabulary to use to interpret the vectors
+    n_neurons : int
+        Number of neurons to use in each product computation
+    invert_a, invert_b : bool
+        Whether to reverse the order of elements in either
+        the first input (`invert_a`) or the second input (`invert_b`).
+        Flipping the second input will make the network perform circular
+        correlation instead of circular convolution.
+    input_magnitude : float
+        The expected magnitude (vector norm) of the two input values.
+    """
+    def __init__(self, dimensions, vocab=None, n_neurons=200, invert_a=False,
+                 invert_b=False, input_magnitude=1.0, label=None, seed=None,
+                 add_to_container=None):
+        super(Bind, self).__init__(label, seed, add_to_container)
+        if vocab is None:
+            # use the default vocab for this number of dimensions
+            vocab = dimensions
+        elif vocab.dimensions != dimensions:
+            raise ValueError('Dimensionality of given vocabulary (%d) does not'
+                             'match dimensionality of buffer (%d)' %
+                             (vocab.dimensions, dimensions))
+
+        with self:
+            self.cc = nengo.networks.CircularConvolution(
+                n_neurons, dimensions, invert_a, invert_b,
+                input_magnitude=input_magnitude)
+            self.A = self.cc.A
+            self.B = self.cc.B
+            self.output = self.cc.output
+
+        self.inputs = dict(A=(self.A, vocab), B=(self.B, vocab))
+        self.outputs = dict(default=(self.output, vocab))
--- a/nengo/spa/tests/test_bind.py
+++ b/nengo/spa/tests/test_bind.py
@@ -1,0 +1,52 @@
+import numpy as np
+
+import nengo
+from nengo import spa
+from nengo.utils.numpy import rmse
+
+
+def test_basic():
+    with spa.SPA() as model:
+        model.bind = spa.Bind(dimensions=16)
+
+    inputA = model.get_module_input('bind_A')
+    inputB = model.get_module_input('bind_B')
+    output = model.get_module_output('bind')
+    # all nodes should be acquired correctly
+    assert inputA[0] is model.bind.A
+    assert inputB[0] is model.bind.B
+    assert output[0] is model.bind.output
+    # all inputs and outputs should share the same vocab
+    assert inputA[1] is inputB[1]
+    assert inputA[1].dimensions == 16
+    assert output[1].dimensions == 16
+
+
+def test_run(Simulator, seed):
+    rng = np.random.RandomState(seed)
+    vocab = spa.Vocabulary(16, rng=rng)
+
+    with spa.SPA(seed=seed, vocabs=[vocab]) as model:
+        model.bind = spa.Bind(dimensions=16)
+
+        def inputA(t):
+            if 0 <= t < 0.1:
+                return 'A'
+            else:
+                return 'B'
+
+        model.input = spa.Input(bind_A=inputA, bind_B='A')
+
+    bind, vocab = model.get_module_output('bind')
+
+    with model:
+        p = nengo.Probe(bind, 'output', synapse=0.03)
+
+    sim = Simulator(model)
+    sim.run(0.2)
+
+    error = rmse(vocab.parse("B*A").v, sim.data[p][-1])
+    assert error < 0.1
+
+    error = rmse(vocab.parse("A*A").v, sim.data[p][100])
+    assert error < 0.1
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -133,6 +133,8 @@ Release History
   actions to be added after module has been initialized.
   (`#861 <https://github.com/nengo/nengo/issues/861>`_,
   `#862 <https://github.com/nengo/nengo/pull/862>`_)
+- Added SPA wrapper for circular convolution networks, ``spa.Bind``
+  (`#849 <https://github.com/nengo/nengo/pull/849>`_)
 
 **Bug fixes**
 
--- a/nengo/networks/circularconvolution.py
+++ b/nengo/networks/circularconvolution.py
@@ -88,7 +88,7 @@ def dft_half(n):
 
 
 def CircularConvolution(n_neurons, dimensions, invert_a=False, invert_b=False,
-                        input_magnitude=1, net=None):
+                        input_magnitude=1.0, net=None):
     """Compute the circular convolution of two vectors.
 
     The circular convolution `c` of vectors `a` and `b` is given by
@@ -107,7 +107,7 @@ def CircularConvolution(n_neurons, dimensions, invert_a=False, invert_b=False,
     Parameters
     ----------
     n_neurons : int
-        Number of neurons to be used, in total.
+        Number of neurons to use in each product computation
     dimensions : int
         The number of dimensions of the input and output vectors.
     invert_a, invert_b : bool
--- a/nengo/spa/__init__.py
+++ b/nengo/spa/__init__.py
@@ -1,6 +1,7 @@
 from .actions import Actions
 from .assoc_mem import AssociativeMemory
 from .basalganglia import BasalGanglia
+from .bind import Bind
 from .buffer import Buffer
 from .compare import Compare
 from .cortical import Cortical


nengo/spa/bind.py
```diff
@@ -0,0 +1,41 @@
...
+            vocab = dimensions
+
+        with self:
+            self.cc = nengo.networks.CircularConvolution(
+                n_neurons, dimensions, invert_a, invert_b,
+                input_magnitude=input_magnitude)
+
+            self.output = nengo.Node(size_in=1, label='output')
+
+        self.inputs = dict(A=(self.cc.A, vocab), B=(self.cc.B, vocab))
```
xchoo: There's no guarantee that the ```vocab``` for ```A``` for ```B``` and ```output``` will be the same.

nengo/spa/bind.py
```diff
@@ -0,0 +1,41 @@
...
+        if vocab is None:
+            # use the default vocab for this number of dimensions
+            vocab = dimensions
+
+        with self:
+            self.cc = nengo.networks.CircularConvolution(
+                n_neurons, dimensions, invert_a, invert_b,
+                input_magnitude=input_magnitude)
+
+            self.output = nengo.Node(size_in=1, label='output')
```
xchoo: What's this output for?

nengo/spa/bind.py
Seanny123: Oops. Will delete.

nengo/spa/bind.py
Seanny123: Oops. Will delete.

nengo/spa/bind.py
xchoo: Delete? Um. I think having different vocabs would be a better idea. :smile:

Seanny123: See updated note. I commented on the wrong line.

xchoo: I would have vocabs for each of the inputs and outputs. Although, I can see the argument for having the same vocab for both inputs, and a different vocab for the output.

tcstewar:
Hmm...  I'm not sure there should be an ```input_magnitude``` parameter.  ```spa``` modules are meant to be used with semantic pointers (thus the name), and so all of the currently existing modules assume the ```input_magnitude``` is 1.

tcstewar:
I'm also not sure there should be the ability to set different Vocabularies.  What should the system do if you use different Vocabularies at A, B, and the output?  Should it do a conversion between them?  If not, what's the use case you're envisioning?  If so, what conversion?

I'd just have a single Vocab that can be set for all three....

Seanny123:
>I'm not sure there should be an input_magnitude parameter

I left `input_magnitude` because it exists on the `spa.Compare` network, should it be removed from `spa.Compare` as well?

>I'd just have a single Vocab that can be set for all three....

I don't think I understand what the purpose of a Vocab is anymore... The documentation doesn't really cover translation between Vocabs and the other use cases you folks seem to be talking about. Who should I talk to about this confusion?

Seanny123:
Okay, I talked to Jan about this and now I understand why translation between a lower dimensional Vocab to a higher dimensional Vocab would be desirable. However, I'm uncertain about translation between two Vocabs of the same dimensionality. Can someone give me a scenario where this is desirable and why this translation should happen on a `spa.Bind()` network?

tcstewar:
> I left input_magnitude because it exists on the spa.Compare network, should it be removed from spa.Compare as well?

Hmm...  I see what you mean.  I guess it's not too bad to have it in there, although I really can't think of any situation where I'd want to use the parameter, and I don't find the docstring all that helpful (in both Compare and Bind).  So I'd lean to get rid of them both, but I'm on a bit of a parameter minimization kick lately.  ;)  So, I'm fine either way with having input_magnitude here, but at some point I'd think about a separate PR to get rid of them (but that'd be a longer discussion, I think).

Seanny123:
Given that Terry's already looked over this, I've gone and marked it as "needs merge".

nengo/spa/bind.py
```diff
@@ -0,0 +1,39 @@
...
+    n_neurons : int
+        Number of neurons to use in each product computation
+    invert_a, invert_b : bool
+        Whether to reverse the order of elements in either
+        the first input (`invert_a`) or the second input (`invert_b`).
+        Flipping the second input will make the network perform circular
+        correlation instead of circular convolution.
+    input_magnitude : float
+        Effective input magnitude for the multiplication.
+        The actual input magnitude will be this value times sqrt(2)
```
tcstewar: It's not clear to me what this means, especially given the new product system.  What do you mean by "actual input magnitude"?  What multiplication?  What is this for?  What use case would there be for this parameter?

One possible solution here is just to remove the second line, since as far as I can tell it's both unclear and false.  As I said elsewhere, I'd also be supportive of just removing this parameter altogether, since I can't think of a spa use case for it.  But if the parameter is supposed to be here then it should be clear what it means.

tcstewar:
> Given that Terry's already looked over this, I've gone and marked it as "needs merge".

I'd looked at it briefly, but not given it a full review.  So I changed the marking back.  ;)

nengo/spa/bind.py
Seanny123: >I'd also be supportive of just removing this parameter altogether, since I can't think of a spa >use case for it.
What's the use case outside of SPA?

As for the general comment, I made the mistake of copying the doc string from product when I should have copied it from CircularConvolution. Will fix.

Seanny123:
@tbekolay I'm having the darnest time passing the TravisCI tests because there's so much noise coming out of my module. Is there some way to solve this? Should I try and copy the method to test the neural accuracy of `networks.CircularConvolution`? Also, I seem to be getting different results every time I run the test. Isn't the see parameter supposed to solve this?

nengo/spa/tests/test_bind.py
```diff
@@ -0,0 +1,50 @@
...
+    assert inputB[0] is model.bind.B
+    assert output[0] is model.bind.output
+    # all inputs and outputs should share the same vocab
+    assert inputA[1] is inputB[1]
+    assert inputA[1].dimensions == 16
+    assert output[1].dimensions == 16
+
+
+def test_run(Simulator, seed):
+    vocab = spa.Vocabulary(16)
```
jgosmann: You have to pass a seed or rng to vocabulary to make the semantic pointers reproducible.

nengo/spa/tests/test_bind.py
```diff
@@ -0,0 +1,50 @@
...
+    with spa.SPA(seed=seed) as model:
+        model.bind = spa.Bind(dimensions=32, vocab=vocab)
```
jgosmann: How can you assign a 16-dimensional vocab to something that requires 32 dimensions? Shouldn't this raise an error?

Seanny123: Ah, that was a silly commit. Sorry about that. I got frustrated and didn't
test locally before seeing how it would run on TravisCI. I'll try the seed
thing that you recommended and get back to you.

nengo/spa/tests/test_bind.py
jgosmann: You might not need to specify a vocab at all. This seems like you could use the default vocabs provided by SPA. (You're overwriting the `vocab` variable in line 37 anyways.)

Seanny123:
@s72sue this pull request is important for our tutorial, would you mind taking a shot at reviewing it? If you have any questions about how it works or how SPA works, I would be happy to answer them. For reference, a similar pull request is #850.

Note that the unit tests are failing for this pull request, however this is due to #875 and thus can be safely ignored.

Seanny123:
@s72sue I fixed the tests now, so it's super ready for review now! :smile:

s72sue:
Okay sounds good, I will review it today.

nengo/spa/bind.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    def __init__(self, dimensions, vocab=None, n_neurons=200, invert_a=False,
+                 invert_b=False, input_magnitude=1.0, label=None, seed=None,
+                 add_to_container=None):
+        super(Bind, self).__init__(label, seed, add_to_container)
+        if vocab is None:
+            # use the default vocab for this number of dimensions
+            vocab = dimensions
```
s72sue: I am not sure why you are putting the dimensions in vocab. Shouldn't you use the dimensions to get the default vocab ("vocab = default_vocab" and not "vocab = dimensions")? I am assuming that vocab is supposed to be a Vocabulary object.

s72sue: Also, what happens if the vocab is not none, and the dimensions of the vocab are not equal to the dimensions passed in as the argument?.. Shouldn't we raise an error in that case?

nengo/spa/tests/test_bind.py
```diff
@@ -0,0 +1,52 @@
...
+
+    sim = Simulator(model)
+    sim.run(0.2)
+
+    res = np.dot(vocab.parse("B*A").v, sim.data[p].T)
+
+    assert res[100] < 0.45
+    # significantly lower threshold than usual because of the noise
+    # involved in circular convolution
+    assert res[199] > 0.6
```
s72sue: I am wondering where the numbers 0.45 and 0.6 come from and whether these conditions are guaranteed?
If not, we could even use res[100] < res[199]

nengo/spa/bind.py
tcstewar: The ```vocab = dimensions``` thing is mostly my fault, and is done in all the other ```spa.Module```s.  Basically, if you set the vocab to be an integer, then then ```spa.SPA``` system kicks in and automatically does the looking for the default vocab for that dimensionality.  

So, it's ugly and badly documented, but it's consistent with the other modules.  it might make sense to make a separate issue to take another look at this and either document it better or replace it with something a bit more sane.  Fortunately, it's not a user-facing issue....

s72sue: I see, I just saw the logic in __setattr__ of the spa.SPA system where it
looks for the default vocab for the specified dimensionality. Its not very
obvious and I agree that it needs to be documented better.

nengo/spa/tests/test_bind.py
Seanny123: Aw jeez, you're totally right. Those numbers were picked randomly and aren't informative of the network. I should be using the [same test as the circular convolution network](https://github.com/nengo/nengo/blob/master/nengo/networks/tests/test_circularconv.py). I'll make the switch.

nengo/spa/bind.py
Seanny123: I made an issue to discuss this https://github.com/nengo/nengo/issues/886

nengo/spa/tests/test_bind.py
Seanny123: Fixed. Anything else that needs to be changed?

nengo/spa/bind.py
s72sue: @Seanny123, you might want to address Terry's comments. Its not very clear what the actual magnitude is.

Seanny123: fixed

s72sue:
LGTM

nengo/spa/bind.py
s72sue: @Seanny123, just realized you didn't address my comment above regarding what happens if the vocab is not none, and the dimensions of the vocab are not equal to the dimensions passed in as the argument. Shouldn't we raise an error in that case?

Seanny123: If you pass in a vocabulary that doesn't match the dimensions you're using the code throws an error about a bad connection. This isn't ideal, but since this issue doesn't just affect `bind.py`, but all of the spa modules, I moved it to another issue.

s72sue: I think it makes sense to throw that error in bind.py as well, just as in all other modules. I don't think its too vague, its just missing from bind.py. What do you think would be more ideal?

Seanny123: Aw jeez, I thought it was missing from the other modules, but it's totally not. I'll fix it now. Sorry for misunderstanding you.

Seanny123: Fixed

tcstewar:
This looks good to me too.  There will have to be changes when we make our SPA I/O stuff a bit more consistent and sane, but this looks right to me now, and useful to have.

tcstewar:
@tbekolay I think this is ready for merging into master (and I've forgotten the right way to do that).  Should there be a changelog entry, as this adds a new feature (a new SPA module)?

tbekolay:
Yep, definitely add a changelog entry (under "Improvements")

Seanny123:
Added changelog entry

PR #850, opened 2015-09-18
Removed silly input nodes from compare network

### LINKED ISSUE #848: spa.Compare has silly nodes on it's input
I put input nodes when I could have just used the Product network's input nodes.

Seanny123:
Resolves #848

### DIFF
--- a/nengo/spa/tests/test_compare.py
+++ b/nengo/spa/tests/test_compare.py
@@ -10,8 +10,8 @@ def test_basic():
     inputB = model.get_module_input('compare_B')
     output = model.get_module_output('compare')
     # all nodes should be acquired correctly
-    assert inputA[0] is model.compare.product.A
-    assert inputB[0] is model.compare.product.B
+    assert inputA[0] is model.compare.A
+    assert inputB[0] is model.compare.B
     assert output[0] is model.compare.output
     # all inputs should share the same vocab
     assert inputA[1] is inputB[1]
--- a/nengo/spa/tests/test_compare.py
+++ b/nengo/spa/tests/test_compare.py
@@ -10,8 +10,8 @@ def test_basic():
     inputB = model.get_module_input('compare_B')
     output = model.get_module_output('compare')
     # all nodes should be acquired correctly
-    assert inputA[0] is model.compare.inputA
-    assert inputB[0] is model.compare.inputB
+    assert inputA[0] is model.compare.product.A
+    assert inputB[0] is model.compare.product.B
     assert output[0] is model.compare.output
     # all inputs should share the same vocab
     assert inputA[1] is inputB[1]
--- a/nengo/spa/compare.py
+++ b/nengo/spa/compare.py
@@ -32,15 +32,12 @@ class Compare(Module):
                 neurons_per_multiply, dimensions,
                 input_magnitude=input_magnitude)
 
-            self.inputA = nengo.Node(size_in=dimensions, label='inputA')
-            self.inputB = nengo.Node(size_in=dimensions, label='inputB')
             self.output = nengo.Node(size_in=1, label='output')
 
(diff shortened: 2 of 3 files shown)


tbekolay:
Why closed?

Seanny123:
Because if you remove the input passthrough nodes,  you have to type
`spa.Compare.Product.A`

I somehow missed this while making the pull request

Why closed?

—
Reply to this email directly or view it on GitHub
<https://github.com/nengo/nengo/pull/850#issuecomment-141836021>.

xchoo:
Can you not just do:
```
self.inputA = self.product.A
self.inputB = self.product.B
```
?

xchoo:
With that change mentioned above, the pull request is still valid.

Seanny123:
Yep, that's what I'm going to do in the new pull request

xchoo:
Um... Why not use this pull request? Trevor can squash the commits so that only the last commit matters. (right?)

Seanny123:
Yeah,  I should have just marked it work in progress. Left the office in a
rush and herped when I derped.

Seanny123:
@bjkomer would you mind quickly reviewing this? It's a really simple fix for a really awkward problem.

nengo/spa/tests/test_compare.py
```diff
@@ -10,8 +10,8 @@ def test_basic():
     inputB = model.get_module_input('compare_B')
     output = model.get_module_output('compare')
     # all nodes should be acquired correctly
-    assert inputA[0] is model.compare.inputA
-    assert inputB[0] is model.compare.inputB
+    assert inputA[0] is model.compare.product.A
+    assert inputB[0] is model.compare.product.B
```
hunse: Can't we use `model.compare.A` and `model.compare.B` here? That way, the test doesn't change if we change the internal implementation.

Seanny123: Fixed

Seanny123:
@tbekolay remember that error we had before with the duplicate values in `conftest.py`s? It seems to be happening again in this branch.

Seanny123:
nvm. just realised this isn't totally up to date with master, so I'm going to try that first

bjkomer:
@Seanny123 it looks good to me!

hunse:
This technically changes the API for this network, since `inputA` became `A`, ditto with B. Do we want a changelog entry for that?

@Seanny123: When you update things with master, can you please do a rebase instead of a merge? Don't worry, though, it's easy to fix.

tcstewar:
> This technically changes the API for this network, since inputA became A, ditto with B. Do we want a changelog entry for that?

What's the reasoning for this change?  This API change has nothing to do with "removing silly input nodes"?

(I just have lots of code and models that use the old syntax -- I'm not particularly wedded to the old syntax, but I just don't want to be accidentally changing syntax without a reason).

hunse:
> What's the reasoning for this change?

It makes it the same API as other networks (e.g. product), and for that reason I think it's a good change. But you're right that it has nothing to do with what this PR purports to do, and we should discuss it.

tcstewar:
> It makes it the same API as other networks (e.g. product), and for that reason I think it's a good change. But you're right that it has nothing to do with what this PR purports to do, and we should discuss it.

The consistency with ```Product``` makes sense to me...  still, I'd feel more comfortable doing it via gradual deprecation, rather than directly breaking existing code.  I think something like putting in a ```inputA``` and ```inputB``` property that just returns ```self.A``` or ```self.B```, while also giving a deprecation warning would be appropriate.  But that should also be a separate PR, I think,

Seanny123:
Sorry about that folks, I didn't even realise I made an API change, since the input dict remained the same. I'll make the change back to the old API and [have opened a new issue for further discussion](https://github.com/nengo/nengo/issues/887).

jgosmann:
Superseded by new-spa-overhaul

PR #852, opened 2015-09-18
Remove signatures from notebooks

hunse:
As we realized in #833, notebook signatures are machine-dependent.
We might as well remove them all and keep them out (with a test)
so that they're not showing up in our diffs all the time.

`devscripts/clearoutputs.py` has been changed to remove signatures,
so it should be easy to meet this requirement when changing notebooks.

### DIFF
--- a/examples/advanced/inhibitory_gating.ipynb
+++ b/examples/advanced/inhibitory_gating.ipynb
@@ -1,7 +1,6 @@
 {
  "metadata": {
-  "name": "",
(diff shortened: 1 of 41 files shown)


tbekolay:
I believe that in Jupyter (IPython >4) signature are stored in a local database instead of the notebook, so this seems good to me! We should think about when we want to upgrade to that. Probably not for 2.1, but perhaps 2.2.

Seanny123:
LGTM, obviously.

PR #853, opened 2015-09-21
Give `assoc_mem` ensembles labels with pointer

tbekolay:
Makes it easier to see what these networks are doing in the GUI.

### DIFF
--- a/nengo/spa/assoc_mem.py
+++ b/nengo/spa/assoc_mem.py
@@ -75,6 +75,7 @@ class AssociativeMemory(Module):
         if output_vocab is None:
             output_vocab = input_vocab
             output_vectors = input_vectors
+            output_keys = input_keys
         else:
             if output_keys is None:
                 output_keys = input_keys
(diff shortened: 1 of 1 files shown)


nengo/spa/assoc_mem.py
```diff
@@ -101,6 +102,14 @@ def __init__(self, input_vocab, output_vocab=None,  # noqa: C901
                 label=label, seed=seed,
                 add_to_container=add_to_container,
                 **ens_args)
+
+            for ens, inkey, outkey in zip(
+                    self.am.ensembles, input_keys, output_keys):
```
xchoo: I think you want ```self.am.am_ensembles``` here?

tbekolay: Right you are!

Seanny123:
Where does this clarify what the networks are doing?

tbekolay:
In the GUI when you expand the AssocMem's `am` network. When I did this, I saw that it was just the default number labels that you get with an ensemble array. I wanted to make sure that the AssocMem network was using my vocabulary subset, instead of the full vocabulary; I could have counted it all up for my answer, but I figured being able to see the actual SPs being cleaned up would be better (then I could just visually scan to see if something not in my subset was there).

In general, it's not really easy to see what's going on when looking at big ensemble arrays, but I think there's not really a downside here. It's also convenient if you want to do something specific to one of your cleaned up ensembles, you can iterate through the AssocMem's list of ensemble and look for one with the right label.

Seanny123:
Thanks for explaining it to me. This commit looks good to me. Labels are pretty self-evident in their purpose, so this new block of code doesn't require a comment.

PR #855, opened 2015-09-21
Better handle large cache directories

jgosmann:
So far this contains only a small change and I'm not sure how much it helps.


jgosmann:
This implementation negatively impacts performance. The PR is superseded by #946.

Issue #802, opened 2015-08-05
Use `pytest.warns` and remove `WarningsCatcher`


tbekolay:
pytest will soon contain a `warns` method that works like `raises`, except is for warnings instead of errors. Once it is in a pytest release, we should require that version of pytest, use its implementation of `warns` where we were using `WarningsCatcher` before, and remove warnings-related code in `nengo.utils.testing`.

hunse:
This was done at some point.

PR #856, opened 2015-09-26
Fix recent TravisCI errors

tbekolay:
We were still doing `conda install ipython-notebook` when we should have been installing `jupyter`.

Unfortunately we still have two annoying errors; this is due to a [pytest regression](https://github.com/pytest-dev/pytest/issues/979) which has been [fixed in master](https://github.com/pytest-dev/pytest/pull/983). So, our options are to depends on `pytest==2.7.3` or wait for 2.8.1 to be released with the bugfix. Any preferences?

Also, this PR also removes testing in Python 3.3 (Conda thinks that Jupyter doesn't work in it) and adds testing for Python 3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ matrix:
     - env: PYTHON="2.7" NUMPY="1.7"
     - env: PYTHON="2.7" NUMPY="1.8"
     - env: PYTHON="3.4"
+    - env: PYTHON="3.5"
 
 # Setup Miniconda
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,14 @@ env:
     - NUMPY="1.9"
     - SETUP_CMD="test -a 'nengo -n 2 -v'"
     - EXTRA_CMD=""
-    - CONDA_DEPS="matplotlib ipython-notebook pygments"
+    - CONDA_DEPS="matplotlib jupyter pygments"
     - PIP_DEPS="mistune pytest==2.7.3 pytest-xdist"
 
 matrix:
   include:
     - env: >
         PYTHON="2.6"
-        CONDA_DEPS="$CONDA_DEPS ordereddict pygments"
+        CONDA_DEPS="matplotlib ipython-notebook ordereddict pygments"
         PIP_DEPS="$PIP_DEPS counter importlib IPython==1.2.1"
     - env: >
         PYTHON="2.7"
@@ -33,7 +33,6 @@ matrix:
     - env: PYTHON="2.7" NUMPY="1.6"
     - env: PYTHON="2.7" NUMPY="1.7"
     - env: PYTHON="2.7" NUMPY="1.8"
-    - env: PYTHON="3.3"
     - env: PYTHON="3.4"
 
 # Setup Miniconda
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ notifications:
 env:
   global:
     - NUMPY="1.9"
-    - SETUP_CMD="test -a 'nengo -n 2 -v'"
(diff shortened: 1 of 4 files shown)


hunse:
How long do you think it will be for the Pytest release? They still have a few issues on their milestone. Maybe switch to `pytest==2.7.3` for the time being.

tbekolay:
No idea. Agreed on switching it; done, tests should pass now.

Seanny123:
LGTM. BTW, `2.8.1` is supposed to be released tomorrow.

hunse:
Has anyone tried `pytest` master to see if it actually fixes our problem? I just tried it, and it still seems to have the error for me, even though the commit that purportedly fixes it is in there.

hunse:
It only has problems with the `nl` and `nl_nodirect` fixtures. If I make them using `@pytest.fixture` like we do our others, then it works, but it collects only 540 items instead of 653.

tbekolay:
From what I understand of pytest, you can't make these typical `fixtures`; we call `parametrize` on tests that have `nl` or `nl_nodirect`, a fixture can only return one thing. The thing I don't understand is why it's only a problem for those tests; there are tests in `utils/tests` that use `nl` and `nl_nodirect`, but they don't fail like these...

hunse:
It seems to depend how you load them. If you just do `py.test nengo/tests/test_solvers.py`, it works, but if you do `py.test nengo` then there's an error in `test_solvers.py`. I'm trying to create a minimal example, but no success yet. Could it be because we have multiple `conftest.py` files that load from the main one?

tbekolay:
I initially tried removing the other `conftest` files, but then nothing works. In the past, my understanding of the conftest loading what that it would traverse parent directories looking for `conftest` files, but since `tests/conftest` is in a subdirectory of a parent directory, I didn't think it got loaded. Maybe something's changed in terms of where it looks for `conftest` files?

Seanny123:
I have a silly question. What's the motivation of staying with the most
recent py.test? Are there bugs resolved or features added in the newer
version that benefit us?

hunse:
So, it seems like if you do `py.test nengo`, it crawls all directories and loads all conftest files, which is a problem for the `nl` and `nl_nodirect` fixtures (because of how they're declared). If we delete all the `conftest.py` files except the main one, it's fine when we run all the tests, but if you want to run just one file (or one directory) without a `conftest.py` in it then it's a problem (e.g. `py.test nengo/utils/tests`).

hunse:
> What's the motivation of staying with the most
recent py.test? Are there bugs resolved or features added in the newer
version that benefit us?

Yes, e.g. #802.

tbekolay:
@Seanny123: Because of the way package managers are, you can never really guarantee that the old version of something will be available to you in the forms that you want. So, if you want to stick with an old version of something semi-permanently, you have to "vendorize" it, meaning that you copy the source of the whole package into your package, which adds clutter. It also means that if they do release something you want, then you have to update your vendorized copy, which is a pain. Generally, it's more effort to decide to stick with the older version of something than it is to make things work with the newer version.

hunse:
Back to the subject at hand, even my explanation doesn't seem to fully capture the problem, because I can't reproduce it in my own little example. Something weird is going on in our project.

hunse:
The best solution I've come up with so far is having one `conftest.py` at `nengo/conftest.py`, since this is in a parent directory to everything. Not ideal, but it works.

tbekolay:
Might have to be the solution, then. All of our backends will have to update accordingly as well.

hunse:
Okay, I've found a minimal example, which I've submitted to `pytest` at pytest-dev/pytest#1082.

hunse:
When the problem I pointed out above is fixed in `pytest` master, then we should change our dependency back to the current pytest.

PR #858, opened 2015-09-29
Simplified signal logic again

hunse:
This is a new version of #681. I took @drasmuss's commit to remove the SignalView stuff and just use Numpy for all our slicing needs, and modified it a bit. I also took out the controversial stuff about moving the simulation values from the simulator's SignalDict to the signals themselves. We can always do that later if we want, since it's not directly related to these changes.

### DIFF
--- a/LICENSE.rst
+++ b/LICENSE.rst
@@ -33,7 +33,7 @@ Nengo also includes code modified from other libraries.
 
 ------------
 
-The file ``nengo/builder.py`` contains code with the following
+The file ``nengo/builder/operator.py`` contains code with the following
 license:
 
 Copyright (c) 2014, James Bergstra
--- a/nengo/builder/builder.py
+++ b/nengo/builder/builder.py
@@ -3,7 +3,6 @@ import warnings
 
 import numpy as np
 
-import nengo.utils.numpy as npext
 from nengo.builder.signal import Signal, SignalDict
 from nengo.cache import NoDecoderCache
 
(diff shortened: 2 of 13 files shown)


tbekolay:
I'm curious to do a bit of speed / memory profiling of this PR, but looking through the code I'm quite happy with how much it's simplified things!

hunse:
The code I used for memory profiling is in the `memory-use2` branch. It's a bit messy since I just kept tacking stuff on to get more detail into certain parts, but it works pretty well (it helped me find some of those needless copies, anyway).

hunse:
Okay, got this working with Numpy 1.6, so I think it's all good now (fingers crossed).

hunse:
I think it would be good to get this in 2.1 if it's not too much work. It makes things a lot cleaner and it will help me in Nengo OCL too.

tbekolay:
I'll take a look today!

hunse:
Thanks! I just rebased it.

hunse:
I fixed up `LICENCE.rst`.

tbekolay:
Awesome, this looks great to me. @drasmuss, since you started off this process, could you give this a review as well?

drasmuss:
I might be inclined to rename `signal.value` to `signal.init_value`, just to be clear what its role is.  That's a minor nit though, other than that looks good to me.

tbekolay:
> rename signal.value to signal.init_value

That is definitely more clear! I was a bit worried because I thought that we used `.value` a lot in the code, but actually we only use it in about 10 or so lines (if my `grep`ping is right). So my vote would be for `init_value` or `init_val` or `initial_val` or something like that too.

hunse:
Okay, I renamed it to `initial_value`, i.e. none of the things you suggested!

drasmuss:
Looks good to me.

tbekolay:
I totally implied `initial_value` though! It was there in spirit.

I'll clean up the history and merge!

Issue #857, opened 2015-09-29
Simulator should have close() method


jgosmann:
I think the `Simulator` class should have a `close` method because some backends require closing (e.g. SpiNNaker). This would make it easier to write code independent of the backend (which comes up in my spaopt branch).

In addition to that I would vote to allow a the simulator in a `with` statement like so:
```python
with nengo.Simulator(model) as sim:
    sim.run(1.)
```
(I think we had an unresolved discussion on this in some other issue already.)

hunse:
Another useful thing I found recently is the `atexit` module. It allows you to register functions to run when the program is exiting. We could use this, combined with a global weakref list of simulators (or something), to make sure that all simulators are closed on exit. This saves users having to use the `with` statement (though we can still have that functionality for users who want their simulator to close earlier).

jgosmann:
Doesn't solve my problem at hand because I want to create another simulator after closing the first one in the same script. But maybe it makes sense to implement this in general. (I believe nengo_spinnaker might already be doing this?)

hunse:
As I said, we'd still have the option to manually close. The `atexit` is just to make sure that if users don't close, the board still gets cleaned up.

e2crawfo:
See #739, I (roughly) followed the recommendations there in nengo_mpi. https://github.com/e2crawfo/nengo_mpi/blob/operational/nengo_mpi/simulator.py

hunse:
Yes, it seems that we just repeated that conversation, just with different people :smile: Good that we're all on the same page now.

tcstewar:
:)

One thing that comes to mind that didn't come up in the previous conversation:  should ```.close()``` on the reference Simulator be a no-op? Or should it set a flag so that subsequent calls to ```.run()``` raise an Exception?

And can you ```.close()``` more than once?

hunse:
I would say a) `.close()` on the reference simulator should stop subsequent calls to `.run` and b) you can close more than once (the subsequent calls will just do nothing).

jgosmann:
That seems sensible to me too.

tbekolay:
I was the main dissenting voice on the `.close()` stuff, but yeah, the consensus is pretty overwhelming so this all sounds good to me.

Makes sense for consistency that `close()` in the reference sim should stop subsequent calls to `.run`. I think in the future we will have some reasons to us `close()`

Closing more than once seems fine to me. I just tried this with file pointers in Python; you can call close multiple time, so there's precedent for that.

[...]

Issue #739, opened 2015-06-14
A standard way for Simulators to free resources


tcstewar:
Some backends need ways to indicate that they are finished and their resources are no longer needed.  For example, ```nengo_spinnaker``` needs to put the SpiNNaker machine back into a clean state so that new programs can be loaded on (without rebooting the board).

Right now, the ```nengo_spinnaker``` implementation does this with a ```sim.close()``` call.  You can call ```sim.run(time)``` multiple times, and then do a ```sim.close()``` when you are finished with the simulator.  You cannot call ```sim.run(time)``` after calling ```sim.close()```.

It also supports using the ```Simulator``` as a context manager, with an empty ```__enter__``` and a ```__exit__``` that just calls ```self.close()```.  Using the context manager is strictly optional.

Right now, you must call ```close``` after running a Simulation.  Otherwise, the board gets left in a state that means it needs to be physically reset before loading a new program.  It might be possible to add an ```atexit``` handler to do this, at the cost of a global list of un-closed Simulators (and this would be in addition to the ```close``` system, since we need to support running one Simulator and then another within the same program, such as ```nengo_gui```).

As was discussed a bit in #737 , it's not clear whether we should add a ```sim.close()``` to all Simulators (and have it do nothing if the Simulator doesn't need to be closed), or whether this should be something that each backend that needs it can implement in their own way.  I guess the first question is, which backends could use something like this?  Would it be useful in ```nengo_ocl```?  What about ```nengo_mpi``` or ```nengo_bluegene```?  

The second question is, if we do make a ```close()``` method standard, should we also include the context manager version?

On the first question, I think this is something that will be needed with hardware backends if they are to run in ```nengo_gui```.  I'd like to add an ```atexit``` handler to ```nengo_spinnaker``` so that if people forget the ```close``` when writing scripts it will still get closed, but something explicit like this is still needed if people want to run multiple simulations in the same script.  I can foresee a similar thing happening with ```nengo_brainstorm``` and ```nengo_mpi```.  So I think it'd be nice to just add a standard ```.close()``` method to all Simulators.  Of course, we can always just do a ```if hasattr(sim, 'close'): sim.close()``` instead.

On the second question, I'd vote that if we do support a ```.close()``` method we should also support being a context manager.  It's only 4 lines of code, and this is exactly the sort of situation context managers are for.

mundya:
[...]

PR #859, opened 2015-09-30
Add close method to Simulator.

jgosmann:
Adresses #739, #857.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -70,7 +70,8 @@ Release History
   parameters from ``nengo.networks.assoc_mem.AssociativeMemory`` and modifying
   the defaults of others.
   (`#797 <https://github.com/nengo/nengo/pull/797>`_)
-- Add ``close`` method to ``Simulator``.
+- Add ``close`` method to ``Simulator``. ``Simulator`` can now be used
+  used as a context manager.
   (`#857 <https://github.com/nengo/nengo/issues/857`_,
   `#739 <https://github.com/nengo/nengo/issues/739>`_,
   `#859 <https://github.com/nengo/nengo/pull/859>`_)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -70,6 +70,10 @@ Release History
   parameters from ``nengo.networks.assoc_mem.AssociativeMemory`` and modifying
   the defaults of others.
   (`#797 <https://github.com/nengo/nengo/pull/797>`_)
+- Add ``close`` method to ``Simulator``.
+  (`#857 <https://github.com/nengo/nengo/issues/857`_,
+  `#739 <https://github.com/nengo/nengo/issues/739>`_,
+  `#859 <https://github.com/nengo/nengo/pull/859>`_)
 
 **Behavioural changes**
 
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -130,6 +130,12 @@ class Simulator(object):
         seed = np.random.randint(npext.maxint) if seed is None else seed
         self.reset(seed=seed)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
     @property
     def dt(self):
         """The time step of the simulator"""
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -95,6 +95,8 @@ class Simulator(object):
             build artifacts in the Model before building the network,
             then you can pass in a ``nengo.builder.Model`` instance.
         """
+        self.closed = False
+
         if model is None:
             dt = float(dt)  # make sure it's a float (for division purposes)
             self.model = Model(dt=dt,
@@ -172,6 +174,9 @@ class Simulator(object):
     def step(self):
         """Advance the simulator by `self.dt` seconds.
         """
+        if self.closed:
+            raise ValueError("Simulator cannot run because it is closed.")
+
         self.n_steps += 1
         self.signals['__time__'][...] = self.n_steps * self.dt
 
@@ -246,6 +251,9 @@ class Simulator(object):
             or inputs (e.g. from Processes), but not the built objects
             (e.g. ensembles, connections).
         """
+        if self.closed:
+            raise ValueError("Simulator closed.")
+
         if seed is not None:
             self.seed = seed
 
@@ -265,3 +273,11 @@ class Simulator(object):
         # clear probe data
         for probe in self.model.probes:
             self._probe_outputs[probe] = []
+
+    def close(self):
+        """Closes the simulator.
+
+        Any call to ``run``, ``run_steps``, ``step``, and ``reset`` on a closed
+        simulator will raise a ``ValueError``.
+        """
+        self.closed = True
--- a/nengo/tests/test_simulator.py
+++ b/nengo/tests/test_simulator.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
 
 import nengo
 import nengo.simulator
@@ -52,3 +53,20 @@ def test_probedict():
     probedict = nengo.simulator.ProbeDict(raw)
     assert np.all(probedict["scalar"] == np.asarray(raw["scalar"]))
     assert np.all(probedict.get("list") == np.asarray(raw.get("list")))
+
+
+def test_close(Simulator):
+    m = nengo.Network()
+    with m:
+        nengo.Ensemble(10, 1)
+
+    sim = Simulator(m)
+    sim.close()
+    with pytest.raises(ValueError):
+        sim.run(1.)
+    with pytest.raises(ValueError):
+        sim.run_steps(1)
+    with pytest.raises(ValueError):
+        sim.step()
+    with pytest.raises(ValueError):
+        sim.reset()
--- a/nengo/tests/test_simulator.py
+++ b/nengo/tests/test_simulator.py
@@ -70,3 +70,21 @@ def test_close(Simulator):
         sim.step()
     with pytest.raises(ValueError):
         sim.reset()
+
+
+def test_usage_in_with_statement(Simulator):
+    m = nengo.Network()
+    with m:
+        nengo.Ensemble(10, 1)
+
+    with Simulator(m) as sim:
+        sim.run(0.01)
+
+    with pytest.raises(ValueError):
+        sim.run(1.)
+    with pytest.raises(ValueError):
+        sim.run_steps(1)
+    with pytest.raises(ValueError):
+        sim.step()
+    with pytest.raises(ValueError):
+        sim.reset()


jgosmann:
This does not add any global tracking of open simulators or automatic closing ``atexit``. But it seems that we might discuss that functionality further and it can be implemented in a separate PR.

mundya:
[Mega nit-pick!] Any reason why `ValueError`? I tend to read `ValueError` as "the thing you want to do would work with a different parameter value" rather than "from the state you're in, this thing will never work".

Also, briefly glancing at `nengo_spinnaker`s error throwing for this, I think the error could be more explanatory, e.g, "Simulator has been closed and can't be used to run further simulations."

tbekolay:
Yeah, I think there should be a specific exception subclass for this, similar to `StopIterationError`; maybe `SimulatorClosedError` or something. That could be done in #781 though.

hunse:
For now, I would go with `RuntimeError`.

hunse:
Specifically `RuntimeError("Simulator cannot run since it has already been closed.")` or `RuntimeError("Simulator has been closed, so it cannot run anymore.")`

jgosmann:
Closed file object raise a `ValueError` when you try to read them. 
```
ValueError                                Traceback (most recent call last)
<ipython-input-26-bacd0e0f09a3> in <module>()
----> 1 f.read()

ValueError: I/O operation on closed fil
```

jgosmann:
Which seems somewhat reasonbale considering that `f.read()` could also be interpreted as `read(f)` and in that case it is indeed the value that's invalid.

hunse:
Yeah, that makes some sense I guess. Trying to write to a read-only array in Numpy used to give a RuntimeError, I think, but now gives a ValueError; it's a bit similar to our situation here. But in `nengo/network.py`, we use RuntimeError a lot if the state of things isn't the way we expect. So I don't see a clear winner either way, and I'm not sure if it matters too much, as long as the message makes it clear why there's a problem.

tbekolay:
Rebased this. Given that #781 will make a specific exception for this, I'm cool with keeping the ValueErrror (though I made a slight wording change in 316f5c95b0cf10a). Other than that, looks good to me, so will merge if someone can give this a quick review.

drasmuss:
This may be a TODO for another PR, but just a pedagogical comment.  Right now `close` doesn't really do anything, and it isn't used anywhere, so there's no reason a user would put it in their code.  Which kind of defeats the purpose, that this was a backend-agnostic way to shut down a simulator.  In other words, users are unlikely to put this into their code until they read up on Spinnaker and realize that there is a `close` method there that does useful things (which is how things worked before this PR anyway).  

So it'd just be good if we could get users in the habit of gracefully closing their simulators as they learn Nengo, to make the transition between backends smoother.

If we really wanted to encourage this, we could do something like
```python
def __del__(self):
    if not self.closed:
        warnings.warn("Simulator deleted without being closed")
```
That might be overkill, we could also just switch to using the context manager syntax in our examples, so that new users just learn that and don't question it.

Or if we wanted to be more magicy, we could call `self.close()` automatically inside `__del__`

hunse:
I wasn't thinking about putting the call in `__del__`, which might be a good idea, but in #857 we were talking about having `close()` called automatically for all open simulators when the program ends. So basically the same idea. I'm in favour of something like that.

tcstewar:
> So it'd just be good if we could get users in the habit of gracefully closing their simulators as they learn Nengo, to make the transition between backends smoother.

Hmm..  I don't think the added complexity (from the perspective of a user) is worth it.  Having all those sim.close() lines cluttering up the code feels wrong to me.  I'd much rather go with what @hunse mentioned about having ```close()``` be automatically called when the program ends.

The main reason that having the ```close()``` function was suggested in the first place was just so that people who *are* using simulators that require a ```close()``` don't have to write code like this:

```python
if isinstance(sim, nengo_spinnaker.Simulator):
    sim.close()
```

or

```python
if hasattr(sim, 'close'):
    sim.close()
```

or

```python
try:
    sim.close()
except AttributeError:
    pass
```

Having a ```close()``` function on all Simulators is meant to simplify things in this sort of scenario, but I don't think it should mean that everyone should always have to manually call ```close()```.

tbekolay:
I don't think there's any way around forcing people to call `close`. If you don't call `close` on an open file descriptor, you will eventually get problems, and the same will eventually be true of `Simulator.close`. We either teach people to always do this, or we make it automatic. In previous threads about this, I was trying to advocate for it being automatic, but people wanted more control (which in the end I agree with). But with control comes responsibility, and it's on us to teach people the right way to do things. Python doesn't do this with `open`, but I wish it did because it would probably help debug a lot of weird issues. So I agree that we should warn and then close all open simulators when the Python process ends, if we're able to (note that we are not able to when the process is killed, only when it ends naturally or is terminated).

Only nitpick is that we should do this with `atexit` instead of `__del__`, as I am pretty sure `__del__` is not guaranteed to be called in as many situations as `atexit`... though that's just off the top of my head, we should look into it more.

drasmuss:
Yeah, and when we keep in mind the context manager syntax, I don't think we'll have a bunch of `sim.close()` "cluttering up the code".  It's just

```python
    sim.run(1)
```
instead of 
```python
sim = nengo.Simulator(net)
sim.run(1)
```

The nice thing about `__del__` is that it'd handle things properly if the user does something like
```python
for i in range(10):
   sim = nengo.Simulator(net[i])
   sim.run(1)
```
(i.e. it handles the simulator getting deleted mid-script).  But I'll admit that I've never used `atexit` so I don't really know what it can do.

tbekolay:
Right, I see what you're saying Dan. So it should probably be both; have the check in the `__del__` and then at exit, we do `for sim in nengo.simulator.open_sims: del sim`.

tcstewar:
Hmm...  I think I see what you're saying about consistency...  my problem might be is that I'm mentally expecting it to behave like how file handles in Python behave -- no one forces me to manually call ```close()``` and I just rely on it being closed when Python quits.  If I'm in a situation where I want control over closing, then I can call ```close()```.  But Python doesn't give me a warning on exit when a file hasn't been closed.

So I was expecting this to behave in the same way.  Are there other examples of things in Python where you have to call ```close()``` on them, otherwise they complain, rather than silently closing themselves in their ```__del__``` or ```atexit```?

hunse:
> In previous threads about this, I was trying to advocate for it being automatic, but people wanted more control (which in the end I agree with).

Why can't we have both? It will be done automatically at exit or `__del__`, whichever comes first, but users can also force premature close by calling `close()` themselves.

> no one forces me to manually call ```close()``` and I just rely on it being closed when Python quits.  If I'm in a situation where I want control over closing, then I can call ```close()```.

This is exactly what I had in mind. Is there anyone making an argument against this? It seems to me we're all pretty much on the same page here, and I'm not quite sure what we're actually discussing.

drasmuss:
> no one forces me to manually call close() and I just rely on it being closed when Python quits

:scream:

tbekolay:
This isn't really a situation where we should look to Python. Python doesn't do any validation of function inputs either, but we dedicate a lot of our codebase to validation in order to make Nengo easier for our users.

I'm envisioning a `warnings.warn` if you don't manually close the simulator. You can easily suppress these warnings with `warnings.filter` if you don't like them. But it's pretty similar to us printing out that you have zombie threads when closing `nengo_gui`, IMO.

hunse:
What is the problem with automatically closing simulators? I just don't see why we want to encourage manual closing. Certainly some backends might require it if you want to make more than one simulator (that's why this issue came up in the first place), but that's something those backends can deal with (they could even throw an exception if a second simulator is made once one already exists, or something).

drasmuss:
I think it's based on the reliability of `__del__` and `atexit`.  To use `file.close` as an example, that does work OK most of the time if you just let it be closed automatically.  But sometimes it doesn't work, which is why every guide to file IO says that you should really close the file yourself (https://docs.python.org/2/tutorial/inputoutput.html).  So the question is just, should we let users do things wrong if they really want to? Or should we gently encourage them to use best practices?

tbekolay:
I find it hard to dissociate the discussion of manually closing with having a `close` method in `nengo.Simulator` the first place. On the one hand, you want Nengo code to look the same regardless of the backend used, so you add a `close` method to the simulator. On the other hand, you still need to care about your backend to actually call the `close` method? The two seem to go hand-in-hand to me...

Sidenote: whoo, Nengo discussions! It's been too long :octopus: :rainbow:

hunse:
> On the one hand, you want Nengo code to look the same regardless of the backend used, so you add a `close` method to the simulator.

Fair enough. I would do it context manager style then, and have `run` throw an error if it's called outside that context. Also, we'd need an error if a second simulator is made inside the context of another? Or while another still exists? I guess that depends if we expect simulators to allocate resources on creation or upon entering the context manager.

tcstewar:
Hmm, it feels to me like we're veering a lot from the original purpose here.  The intent of the PR was to add a dummy close() method so that people wouldn't have to do ```if hasattr(sim, 'close'): sim.close()``` whenever they're working with multiple backends.  I think it's a separate discussion to decide that we're going to throw up a warning if people don't call ```close()```.  Adding in a new warning and changing how we expect people to call a Simulator (i.e. using the context approach) is a big change that affects a lot of code, and I don't see that being worth it.

tcstewar:
> Sidenote: whoo, Nengo discussions! It's been too long

Agreed!  :) :) :)

tcstewar:
> On the other hand, you still need to care about your backend to actually call the close method? 

When you're doing backend-specific things, yes, you need to know about your backend.  So, for example, nengo_spinnaker has a limitation that only one Simulator can be open at a time.  So if I try doing more than one, I get an error saying "hey, you should close the old one".  Right now, that requires doing one of these ```if hasattr(sim, 'close'): sim.close()```, if I want my code to still work with the original backend.

The intent here was to get rid of that bit of annoying code, and, since it's likely that some other backends might have a similar ```close()``` requirement, we may as well do it in a consistent manner.   But I don't think that solving this problem means forcing everyone to always call ```close```, or forcing everyone to always use a context manager.  Indeed, even the nengo_spinnaker backend doesn't give any sort of warning when it goes ahead and cleans up the Simulator on ```atexit```.

hunse:
> Adding in a new warning and changing how we expect people to call a Simulator (i.e. using the context approach) is a big change that affects a lot of code, and I don't see that being worth it.  

But creating a best-practice that users always call `close()` in their scripts is also a big change, IMO. If that's the route that we might be going down, then I think that warrants some discussion, and we should definitely consider the context manager approach in that.

I do agree, though, that I think we can merge this PR as-is, and push that discussion until later. I think this is an improvement no matter what direction we take down the road.

tcstewar:
> But creating a best-practice that users always call close() in their scripts is also a big change, IMO

I completely agree with you.  This PR was not meant to introduce such a best-practice.

>  If that's the route that we might be going down, then I think that warrants some discussion, and we should definitely consider the context manager approach in that.  I do agree, though, that I think we can merge this PR as-is, and push that discussion until later. I think this is an improvement no matter what direction we take down the road.

Yup, I'd definitely be interested in that sort of discussion, long-term (in amongst all the other design discussions -- yay!).

hunse:
> Indeed, even the nengo_spinnaker backend doesn't give any sort of warning when it goes ahead and cleans up the Simulator on ```atexit```.

That reminds me, the other reason this discussion came up is because we were talking about standardizing that `atexit` call so it's handled by core Nengo, so every backend doesn't have to go do that itself. But again, that can wait.

hunse:
Anyway, in the course of this discussion, I looked over the PR, and it looks good to me.

drasmuss:
Yeah I didn't take any of this as blocking this PR, this all started with "This may be a TODO for another PR...".  I can make a separate issue if you want to keep the discussion out of this thread.

PR #862, opened 2015-09-30
Added ability to Actions to add actions after init

### LINKED ISSUE #861: allow spa.Actions to add more arguments after being initialised
For the TOH, I have a build process where I add more production rules if I have more disks. I can't do that very elegantly right now because `spa.Actions` can't be modified after it's initialised. I would suggest adding a function `spa.Actions.add()` that would accept arguments in the same way as the `spa.Actions` builder.

Is this a good/bad idea?

Seanny123:
Resolves #861

### DIFF
--- a/nengo/spa/actions.py
+++ b/nengo/spa/actions.py
@@ -1,6 +1,7 @@
 """Expressions and Effects used to define all Actions."""
 
 import re
+import warnings
 
 from nengo.spa.action_objects import Symbol, Source, DotProduct, Summation
 from nengo.utils.compat import iteritems, OrderedDict
@@ -148,13 +149,23 @@ class Actions(object):
     The *args and **kwargs are treated as unnamed and named Actions,
     respectively.  The list of actions are only generated once process()
     is called, since it needs access to the list of module inputs and
-    outputs from the SPA object.
+    outputs from the SPA object. The **kwargs are sorted alphabetically before
+    being processed.
     """
 
     def __init__(self, *args, **kwargs):
         self.actions = None
         self.args = args
         self.kwargs = kwargs
+
+    def add(self, *args, **kwargs):
+        if self.actions is not None:
+            warnings.warn("The actions currently being added must be processed"
+                          " either by spa.BasalGanglia or spa.Cortical"
+                          " to be added to the model.")
+
+        self.args += args
+        self.kwargs.update(kwargs)
 
     @property
     def count(self):
@@ -168,7 +179,9 @@ class Actions(object):
         sources = list(spa.get_module_outputs())
         sinks = list(spa.get_module_inputs())
 
+        sorted_kwargs = sorted(self.kwargs.items())
+
         for action in self.args:
             self.actions.append(Action(sources, sinks, action, name=None))
-        for name, action in iteritems(self.kwargs):
+        for name, action in sorted_kwargs:
             self.actions.append(Action(sources, sinks, action, name=name))
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -128,6 +128,10 @@ Release History
   (`#754 <https://github.com/nengo/nengo/pull/754>`_)
 - ``Config.configures`` now accepts multiple classes rather than
   just one. (`#842 <https://github.com/nengo/nengo/pull/842>`_)
+- Added ``add`` method to ``spa.Actions``, which allows
+  actions to be added after module has been initialized.
+  (`#861 <https://github.com/nengo/nengo/issues/861>`_,
+  `#862 <https://github.com/nengo/nengo/pull/862>`_)
 
 **Bug fixes**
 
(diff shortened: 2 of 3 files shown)


nengo/spa/actions.py
```diff
@@ -154,7 +154,11 @@ class Actions(object):
     def __init__(self, *args, **kwargs):
         self.actions = None
         self.args = args
-        self.kwargs = kwargs
+        self.kwargs = OrderedDict(kwargs)
```
arvoelke: What's the reason for the `OrderedDict`? It's perhaps a little misleading because `**kwargs` is not guaranteed to maintain order, so I'm not sure how you expect to use the ordering.

nengo/spa/actions.py
```diff
@@ -154,7 +154,11 @@ class Actions(object):
...
+
+    def add(self, *args, **kwargs):
+        self.args += args
+        self.kwargs.update(OrderedDict(kwargs))
```
arvoelke: One thing that's a little disconcerting is that in theory this could be called after `process`, and then a user might be confused why it doesn't just "work". My suggestion is to check if `self.actions is None`, otherwise raise a warning that says the added action will not be in the model until the next 'whenever'.

nengo/spa/actions.py
Seanny123: Because dicts are handled differently in Python 2 and Python 3, I used an OrderedDict so that I could easily write a tests that would pass for both and so that the outcome would be the same regardless of Python version

hunse: It's probably good to remember what order people added actions in, so I support the use of `OrderedDict`. If that's something that's going to effect the model that's built, then we definitely want to ensure it's the same between Python versions.

arvoelke: Ah, looking at the last commit, I see now. One thing worth noting is it doesn't have as much to do with the Python version as it has to do with the nature of undefined behaviour. You could reorganize the code slightly (maybe change the name of the keyword arg) and then suddenly you could get the flip behaviour of it working in the opposite version but not the other (or both, or neither), or even failing on different machines.

At any rate, the comment about it being misleading is still a concern. To be clear about what I mean...

```
actions.add(a=x, b=y)
```

could have the order `('a', 'b')` in some cases, and the order `('b', 'a')` in other cases, due to the undefined ordering of `**kwargs`. The order will only be preserved under the condition that `len(kwargs) <= 1`.

IMO, making it a list again might be the way to go, to avoid having a user rely on the behaviour only to inevitably encounter a weird issue like the above example. To make the tests work, you could do something like sort the actions in the test?

hunse: Yes, very good point. This is a problem in `add`, too.

hunse: That said, this was a problem before, too. At least this change makes things a *bit* better, but we should definitely put a comment in that things are not as ordered as they look. There's a PEP request for [ordered kwargs](https://www.python.org/dev/peps/pep-0468/), but without something like that, I don't think we can get the ordering without changing the syntax. The other option is to sort the keys (either as we add them, or when we're building the model), so that at least it's deterministic, if not in the same order that the user added them.

Seanny123: I think I'm going to got with @hunse suggestion of sorting it alphabetically before processing. That way, we have deterministic behaviour can tested and if for some bizarre reason the user does want an oddly specific order, they can just prepend their `kwargs` with an index.

arvoelke: Yeah that sounds good. The main idea is it should be all-or-nothing. So either clearly unordered or clearly ordered, otherwise people come to rely on something that seems defined, only to have things confusingly fall apart later when things get more complex and difficult to debug. :+1:

Seanny123:
Fixed all the aforementioned issues.

nengo/spa/actions.py
```diff
@@ -168,7 +179,9 @@ def process(self, spa):
         sources = list(spa.get_module_outputs())
         sinks = list(spa.get_module_inputs())
 
+        sorted_kwargs = OrderedDict(sorted(self.kwargs.items()))
```
arvoelke: Nit: you should be able to just do `sorted(self.kwargs.items())` and then remove the `iteritems` on line 186. The `OrderedDict` part is redundant.

nengo/spa/actions.py
```diff
@@ -156,6 +158,15 @@ def __init__(self, *args, **kwargs):
         self.args = args
         self.kwargs = kwargs
 
+    def add(self, *args, **kwargs):
+        if self.actions != None:
```
arvoelke: Convention is to use `is not` rather than `!=` when comparing to `None`, even in situations where it doesn't really make a difference. Good habit for when it can make a difference.

arvoelke:
LGTM!

arvoelke:
Try rebasing to master and pushing again to fix the Travis issue?

tbekolay:
I'm in the process of merging, so I'll rebase shortly.

tbekolay:
Merge halted: realized this should have a changelog entry too. Added one in 9f1db3f92df123b; can someone look it over?

Seanny123:
Changelog looks good.

arvoelke:
:+1:

PR #864, opened 2015-10-01
Faster TravisCI builds with NumPy 1.9.2

tbekolay:
1.9.3 is much slower in TravisCI for some reason.

This also shows the NumPy configuration before running tests,
so that we can see when Conda changes some aspect of the NumPy
environment.


Seanny123:
LGTM.

PR #866, opened 2015-10-02
Connections cannot connect to connections

### LINKED ISSUE #865: PES silently fails when connecting to a connection
If you accidentally connect to the connection, rather than the connection's `learning_rule` attribute, then PES silently fails to work. I'm wondering why this is allowed, or if it should throw an error / warning?

Can observe this by running the below and then setting `do_bug = True`, which will output a value close to `initial = -0.2`, indicating that no learning occurred.

```python
import nengo

do_bug = False
initial = -0.2
desired = 0.7

    node_x = nengo.Node(output=initial)
    node_ystar = nengo.Node(output=desired)
    node_y = nengo.Node(size_in=1)
    
    ens_x = nengo.Ensemble(200, 1)
    error = nengo.Node(size_in=1)
    nengo.Connection(node_ystar, error, transform=-1)
    nengo.Connection(node_y, error)
    
    nengo.Connection(node_x, ens_x)
    conn = nengo.Connection(ens_x, node_y, learning_rule_type=nengo.PES(learning_rate=1e-4))
    nengo.Connection(error, conn if do_bug else conn.learning_rule)
    
    p = nengo.Probe(node_y, synapse=0.005)

    sim = nengo.Simulator(model)
    sim.run(5.0)
    print sim.data[p][-1]
```

hunse:
Fixes #865.

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -18,6 +18,15 @@ from nengo.utils.compat import is_iterable, iteritems
 logger = logging.getLogger(__name__)
 
 
+class PrePostParam(NengoObjectParam):
+    def validate(self, conn, nengo_obj):
+        super(PrePostParam, self).validate(conn, nengo_obj)
+        if isinstance(nengo_obj, Connection):
+            raise ValueError(
+                "Cannot connect to or from connections. "
+                "Did you mean to connect to the connection's learning rule?")
+
+
 class ConnectionLearningRuleTypeParam(LearningRuleTypeParam):
     """Connection-specific validation for learning rules."""
 
@@ -227,8 +236,8 @@ class Connection(NengoObject):
         The seed used for random number generation.
     """
 
-    pre = NengoObjectParam(nonzero_size_out=True)
-    post = NengoObjectParam(nonzero_size_in=True)
+    pre = PrePostParam(nonzero_size_out=True)
+    post = PrePostParam(nonzero_size_in=True)
     synapse = SynapseParam(default=Lowpass(0.005))
     transform = TransformParam(default=np.array(1.0))
     solver = ConnectionSolverParam(default=LstsqL2())
--- a/nengo/tests/test_connection.py
+++ b/nengo/tests/test_connection.py
@@ -687,22 +687,39 @@ def test_solverparam():
         inst.sp = 'a'
 
 
-def test_nonexistant_prepost(Simulator):
+def test_prepost_errors(Simulator):
     with nengo.Network():
-        a = nengo.Ensemble(100, 1)
+        x = nengo.Ensemble(10, 1)
+        y = nengo.Ensemble(10, 1)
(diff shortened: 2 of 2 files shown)


nengo/connection.py
```diff
@@ -18,6 +18,15 @@
 logger = logging.getLogger(__name__)
 
 
+class PrePostParam(NengoObjectParam):
+    def validate(self, conn, nengo_obj):
+        super(PrePostParam, self).validate(conn, nengo_obj)
+        if isinstance(nengo_obj, Connection):
+            raise ValueError(
+                "Connections cannot connect to or from connections. "
```
tbekolay: I'd maybe drop "connections" at the start here... seem redundant; there's no other way to connect, so probably just `"Cannot connect to or from connections."` would be sufficient.

hunse: Good point. What I have is a bit redundant.

tbekolay:
This LGTM. @arvoelke, does this look good and handle #865?

nengo/tests/test_connection.py
```diff
@@ -687,22 +687,35 @@ class Test(object):
...
+def test_prepost_errors(Simulator):
     with nengo.Network():
-        a = nengo.Ensemble(100, 1)
+        x = nengo.Ensemble(10, 1)
+        y = nengo.Ensemble(10, 1)
+        c = nengo.Connection(x, y)
+        with pytest.raises(ValueError):
+            nengo.Connection(3, y)
+        with pytest.raises(ValueError):
+            nengo.Connection(x, c)
```
arvoelke: Should also test that you can't connect _from_ `c`?

hunse: Yeah, always a question of how many things you want to test. I could also test connecting to a not-NengoObject above, and many other things. In this case, connecting to and from are running through the same validation code, so it's checked either way. I mean, someone could still mess it up, but they'd almost have to be trying. But if you're worried about that case for some reason, I could add it. Also, my general philosophy when fixing a bug is to add minimal tests to make sure that we don't revert, and cover the added code, since by that point the bug is already identified so the tests are less useful. For example, I'd rather add an additional test making sure we can't connect to Probes or some other NengoObject that's not connectable than beat a dead horse with the bug we've already solved.

hunse: Okay, I added it anyway, as well as adding an error for Probes.

hunse:
Okay, ditched that probe idea since I forgot there will still be validation when we make connections to probes in the builder. It still might be nice at some point to stop people from making multiple connections to probes, since if they do want to probe the sum of something I'd suggest making a node first then probing that, and it could help stop some easy mistakes. But that can be for later.

arvoelke:
LGTM. And yeah it's not clear what the best way forward is to handle that case. Good to address it elsewhere.

PR #867, opened 2015-10-03
Utility for setting PES learning_rate

arvoelke:
This makes the results from some recent research easy to use. Just specify the error that you want PES to be able to achieve within some specified time, and `pes_learning_rate` will tell you the exact `learning_rate` that accomplishes this (for constant input, no error filtering, and no noise). At the very least, this should make the task of choosing learning rates better motivated. :sweat_smile: 

Details: http://compneuro.uwaterloo.ca/publications/voelker2015.html

### DIFF
--- a/nengo/utils/connection.py
+++ b/nengo/utils/connection.py
@@ -100,3 +100,51 @@ def eval_point_decoding(conn, sim, eval_points=None):
     decoded = np.dot(activities, weights.T)
     targets = get_targets(sim.model, conn, eval_points)
     return eval_points, targets, decoded
+
+
+def pes_learning_rate(epsilon, activities, t, dt=0.001):
+    """Determine the ideal learning rate for PES without noise or filtering.
+
+    This function returns a `learning_rate` for use in the PES rule, such that
+    after `t` seconds (with a simulator timestep of `dt`) a constant input
+    will have error equal to `epsilon` times the initial error. [1]_
+
+    Parameters
+    ----------
+    epsilon : float
+        The desired approximation factor. The resulting error will be `epsilon`
+        times the initial error. If you want the error to be at most some
+        constant, then divide `epsilon` by the largest possible initial error
+        (usually no more than 2, when the radius is 1).
+    activities : array_like (N,)
+        An array of N activity rates. Less activity (small ||a||) need a higher
+        learning rate. Pick the activities with the smallest ||a|| that you
+        want to learn within epsilon, or make it the average firing rate of
+        each neuron.
+    t : float
+        The amount of simulation time (in seconds) required to obtain the
+        desired error.
+    dt : float (optional)
+        The simulation timestep, defaults to 1 ms.
+
+    Returns
+    -------
+    learning_rate : float
+        The learning rate to provide to the PES rule.
+    gamma : float
+        The rate of convergence, such that the error is the initial error
+        multiplied by `gamma ** k` on the k'th timestep.
+
+    References
+    ----------
+    .. [1] http://compneuro.uwaterloo.ca/publications/voelker2015.html
+    """
+    activities = np.asarray(activities)
+    if activities.ndim != 1:
+        raise ValueError("activities must be a one-dimensional array")
+    n, = activities.shape  # number of neurons
+    a_sq = np.dot(activities.T, activities) * dt  # ||a||^2
+    k = (t - dt) / dt  # number of simulation timesteps
+    gamma = epsilon**(1.0 / k)  # rate of convergence
+    kappa = (1 - gamma) / a_sq  # rearrange equation from theorem
+    return kappa * n, gamma
--- a/nengo/utils/tests/test_connection.py
+++ b/nengo/utils/tests/test_connection.py
@@ -3,8 +3,10 @@ import pytest
 from matplotlib.mlab import griddata
 
 import nengo
+from nengo.builder.ensemble import get_activities
 from nengo.dists import UniformHypersphere
-from nengo.utils.connection import target_function, eval_point_decoding
+from nengo.utils.connection import (
+    target_function, eval_point_decoding, pes_learning_rate)
 from nengo.utils.numpy import rms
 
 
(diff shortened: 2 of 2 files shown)


Seanny123:
Your code and (what I can understand of) your technical report seems fine. However, I have a silly question. Why do you never consider filtering in your technical report?

arvoelke:
Mm, good question... not silly at all. Short answer is I don't know how, and that it shouldn't make much of a difference. I found that for tau < 0.01, the decoders and error are well-approximated by just filtering them by tau. But for larger tau, the added dynamics begin to dominate and cause oscillations (since the feedback dynamics are being changed, not simply the output).

The longer answer is I think it can be done by modifying the equations from the beginning of the proof to use a filtered error, and then taking the Laplace transform to see how the new transfer function differs from the original. Then the idea is you see how this changes the eigenvalues, which will give you the new gamma. The problem is that filtering the error is not equivalent to simply filtering the decoders. It's equivalent to filtering the delta, while separately integrating them into the decoders, and so you have two different control systems that mutually interact. I ran into some problems when trying to combine these transfer functions to find the new eigenvalues... I might just have to try harder.

I also have some intuitions about how to extend this to handle noise, but it only complicates the story. I think you can make a connection to random walks and null-recurrence to show that the decoders basically "wander" off in the hyperplane orthogonal to a (making PES asymptotically unstable for constant inputs), which seems to agree with what I've observed from some tests.

Seanny123:
Thanks for the detailed reply Aaron. As long as this is documented as part of the discussion, then I think this is ready to merge.

tbekolay:
Very nice to have something principled for setting the learning rate!

I have a few concerns, though. The first is the interface that you're expecting from users. The arguments are `epsilon, activities, t, dt=0.001`. The idea is that, given constant input resulting in activities `activities`, the error will be reduced by `epsilon` (where `epsilon` is "The desired approximation factor [...] The resulting error will be `epsilon` times the initial error."). Even after looking at the tech report, I don't have any intuitions as to what a good value of epsilon should be, meaning that we've just moved the confusion from setting the `learning_rate` to setting `epsilon`. Can we just assume some reasonable value for epsilon? Or is it possible to rename epsilon such that people might have some idea of how to set it?

Similarly, `activities` is going to be an annoying quantity for most users to specify. If you're in the situation where you want to set a learning rate, you're rarely going to also be in a situation where you have an existing set of neural activities. However, you do have at that point a set of `max_rates`, as you had to specify the ensembles that are going to be used. Could we instead ask for `pre_max_rates`, or a `pre` ensemble, which we'd use to inspect the `max_rates`?

Finally, somewhat relating to `t`, but also a more fundamental question, how does the time taken to settle to a certain error for constant input relate to different, more realistic learning situations? I worry that people will immediately think "oh, I want this to learn as fast as possible, so we'll set a very low `t`." But while that may work for constant input, my experience is that nonlinear functions require lower learning rates to learn as weird transients in input or feedback delivery can easily send you in the wrong direction, which is hard to overcome if a high learning rate sends you far away from a good solution. While I'm not sure if this is something that we can really answer in a nice quantitative way as you've done for constant input, I feel like rather than providing this tempting parameter that people can crank all the way down, it might be more prudent to, again, provide a reasonable default that should work in most situations, and include the caveat that this function can get people in the right ballpark, but tweaking is still likely to be necessary for the specific learning situation.

arvoelke:
> Even after looking at the tech report, I don't have any intuitions as to what a good value of epsilon should be, meaning that we've just moved the confusion from setting the learning_rate to setting epsilon. Can we just assume some reasonable value for epsilon? Or is it possible to rename epsilon such that people might have some idea of how to set it?

It depends on your desired level of tolerance, which will depend on the application (the requirements of wherever the output is headed). I can imagine in some situations that an error of 0.1 is tolerable, while in others 0.001 is needed. I don't think this quantity is just as confusing as the learning rate, because I find it much more natural to think about the goal in terms of "what is the longest time that is acceptable to get within some prescribed error?", versus "how much do I want to scale my deltas?". The "maximum time" part is `t`, while the "prescribed error" part is `epsilon`. 

A reasonable default for `epsilon` might be something like `1e-3`, since that puts you within 0.1%. If it's more intuitive to think of it as a constant error, rather than a percentage, than `0.5e-3` would put you within an absolute error of at most 0.001 assuming the worst case initial error of 2 (i.e. radius of 1). In the former case we could name it `rel_prescribed_error`, and the latter case we could name it `abs_prescribed_error` (and then use the `radius` to determine this value).

> Similarly, activities is going to be an annoying quantity for most users to specify. Situation where you want to set a learning rate are not situations where you have an existing set of neural activities. However, you will have at that point a set of max_rates, as you had to specify the ensembles that are going to be used. Could we instead ask for pre_max_rates, or a pre ensemble, which we'd use to inspect the max_rates?

I would recommend using the average firing rates (as stated in the docstring). 

I think the above two comments can be resolved by making a wrapper that uses the `radius` to set an absolute error, and the `max_rates` distribution to find the average activity vector. But, since it will be coupled to Nengo, it may be more appropriate to implement this in a later PR, somehow within Nengo (maybe the builder, which would allow it to know the outcome of the distribution).

Also, this isn't really something I wanted to get into right now, but this is a good example of why it makes sense to more easily support incremental building in Nengo. I've ran into a few examples so far, including this one, where you need to know decoders and/or tuning curves before setting up some other ensemble or connection. Eric C ran into this for knowledge representation, I ran into this for some temporal representation stuff, and for my touch classifier from the summer school (done in three incremental builds). Something to think about might be a way of adding custom hooks/callbacks/dependencies into the build process to allow networks to build themselves incrementally once certain data becomes known. This has been a recurring pattern for me, and I expect this to be needed more often when people start building networks that need to exploit the results of distributions and/or decoders and/or tuning curves in models that are more 'machine-learny'. I recall overhearing @hunse suggest this too as a potential difficulty with the iterative nature of building deep networks in Nengo.

> Finally, somewhat relating to t, but also a more fundamental question, how does the time taken to settle to a certain error for constant input relate to different, more realistic learning situations? I worry that people will immediately think "oh, I want this to learn as fast as possible, so we'll set a very low t." 

Well, there are a few things that can be said here. It depends on how quickly your input is changing relative to how quickly you want to learn that input. I believe this is quantifiable by extending these results (and in some respects it already is if you approximate your input signal by a k-step function for some resolution of k that gives you activity vectors within a small distance of one another within each bucket), but let's not try to think about solving everything at once. The intuition should already help for slowly changing inputs, and situations where the examples come in discretized bursts (many SPA examples seem to do this), in which case the `t` is simply the rate of change (and then `epsilon` is how much progress you want to make between each change). These ideas can be built upon once people become comfortable with the basic concepts.

It should be no more confusing that cranking these parameters will be detrimental than it is for the case of the learning rate. An important difference is here you can actually check the returned `gamma` to see the stability (i.e. make sure it's "close" to 1 -- we could also throw a warning if it's below something like 0.95 that we find to be reasonable after using this in practice for a while). But this is somewhat missing the point, because the purpose here is to put these parameters _just above the lower limit of what your application can tolerate_. Not the complete opposite of trying to maximize performance by pushing them to the point where learning could break.

> it might be more prudent to, again, provide a reasonable default that should work in most situations

Maybe something like the time that it takes the basal ganglia to switch from one action to another?

But, again, I think this is trying to solve too much at once. This isn't something a causal user is going to be calling, and there are too many situations where PES can be applied. I think that as a _utility_, its purpose is to provide a tool that makes it easy enough to understand the role of each knob. I am welcome to reorganizing these knobs and documenting them to make their effects more clear, but it seems like coming up with reasonable defaults should be the role of the build-mechanisms/sub-networks where this utility ends up being used.

jgosmann:
> Also, this isn't really something I wanted to get into right now, but this is a good example of why it makes sense to more easily support incremental building in Nengo.

I haven't followed the whole discussion, but incremental building could also be helpful for `spaopt`.

arvoelke:
Moved to an [external library](https://github.com/arvoelke/nengolib).

arvoelke:
> Your code and (what I can understand of) your technical report seems fine. However, I have a silly question. Why do you never consider filtering in your technical report?

FYI, @Seanny123, we now have a [much more detailed analysis of PES](http://compneuro.uwaterloo.ca/publications/voelker2017c.html) that accounts for arbitrary linear filtering. This provides a new understanding/characterization of the dynamics of PES, as a transfer function: $$F(s) = \frac{-s}{s H(s)^{-1} + \kappa \|\vec{a}\|^2}$$ where $H(s)$ is the synaptic filter, $\kappa$ is the learning rate, and $\vec{a}$ is the activity vector. This understanding can be exploited to, for instance, build a decaying oscillator using only a single spiking LIF neuron.

arvoelke:
This also suggests a learning rate of `n / (4*tau*c)`, where `n` is the number of neurons, `tau` is the lowpass time-constant, and `c` is the largest diagonal of the gamma matrix from the decoder solver.

Issue #822, opened 2015-08-26
Things to change in 3.0


tbekolay:
Figured we'd better track these, or else we're unlikely to actually do it.

Should be completed before switch-over:
- [ ] Builder refactor
- [ ] Nengo GUI re-write

Remove Python 2.6 support (should list out specific things):
- [x] Remove support note from the README
- [x] Remove a bunch of the utility functions (`nested` comes to mind)

Remove obsolete things:
- [ ] Remove all `ObsoleteParams`, like `ObsoleteError` for `Connection.modulatory`
- [ ] Remove all things attached to `DeprecationError`, like  `EnsembleArray.neuron_nodes`
- [ ] Remove all `self.tau` attributes from the learning rules

Switch over to `nengo_spa`:
- [ ] remove support for underscore notation in SPA
- [ ] remove deprecation warning for `spa.SPA()`
- [ ] `spa.Memory` and `spa.Buffer`, which means changing the SPA test 
- [ ] if `nengo_gui` supports `nengo_spa` and `nengo_spa` reaches a `1.0` release, then remove `nengo.spa` entirely

Deprecate:
- [ ] [`ProbeDict` should be renamed](https://github.com/nengo/nengo/issues/1388)
- [ ] Change `nengo.utils.connection.target_function` to raise `DeprecationError`

jgosmann:
Does nothing in 2.1 break Python 2.6 compatibility?

tbekolay:
Not yet, I don't think... I don't know if anyone's actively tried, but the unit tests run in 2.6 at least.

Seanny123:
Are we planning on completely removing `nengo.spa` from `nengo` or are we leaving some sort of basic skeleton? Because the current state of this issue seems to be the latter.

tbekolay:
> Are we planning on completely removing nengo.spa from nengo or are we leaving some sort of basic skeleton?

Hm... I think I was originally planning to do this, but I hadn't anticipated that nengo_spa would change as rapidly as it's currently changing. So, my current feeling is to leave `nengo.spa` as is for 3.0 but add deprecation messages everywhere. We'd remove it fully the first major release after `nengo_spa` has a stable (1.0) release.

Seanny123:
Alternatively, if `nengo_spa` reaches 1.0 before `nengo` reaches 3.0 would you be comfortable with removing it?

tbekolay:
Could be convinced yep.

tcstewar:
[...]

PR #868, opened 2015-10-04
Add EnsembleArray.add_neuron_input/output.

tbekolay:
Previously, these both happened whenever `neuron_nodes=True` was passed in. However, there are occasions when you only want a neuron input (as occurred in `InputGatedMemory`) or only want a neuron output (for plotting a raster, for example). Additionally, adding these functions makes the `__init__` function a bit more readable.

For the time being, I've kept `neuron_nodes` around as a shortcut to calling `add_neuron_input` and `add_neuron_output`. However, I've added a DeprecationWarning when it's set to True; we should  remember to remove `neuron_nodes` in Nengo 2.2.

I also redid the unit tests appropriately, and moved the EA.dimensions property up in the file, as I tend to expect all properties to be organized before all methods.

### DIFF
--- a/nengo/networks/ensemblearray.py
+++ b/nengo/networks/ensemblearray.py
@@ -29,8 +29,10 @@ class EnsembleArray(nengo.Network):
         Whether to create a node that provides each access to each individual
         neuron, typically for the purpose of inibiting the entire
         EnsembleArray. Default: False.
+        *Note: this parameter is deprecated. Please call add_neuron_input
+        or add_neuron_output instead.*
     label : str, optional
-        A name to assign this EnsmbleArray.
+        A name to assign this EnsembleArray.
         Used for visualization and debugging.
     seed : int, optional
         Random number seed that will be used in the build step.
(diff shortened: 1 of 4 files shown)


hunse:
What if we just had `neuron_input` and `neuron_output` be properties. When they're called the first time, they'll create a new node, otherwise they'll just return the existing one. That way, if someone uses `neuron_input` or `neuron_output` in the model, they'll be created, otherwise they won't.

tbekolay:
Could do it that way. Just seems a bit magical, but maybe in a good way?

tcstewar:
> What if we just had neuron_input and neuron_output be properties. When they're called the first time, they'll create a new node, otherwise they'll just return the existing one. That way, if someone uses neuron_input or neuron_output in the model, they'll be created, otherwise they won't.

I'll vote against that approach. I did it that way in ```spa``` for some bias nodes, and the problem is that when the GUI system starts looking inside an object for properties, it references those, causing them to be created.  So at some point I'll be going in and fixing the bias system to not do that.  But I think doing it as a side effect of just accessing the variable is a bit too magical, especially given all the introspection stuff that's possible/encouraged in Python.

nengo/networks/ensemblearray.py
```diff
@@ -60,44 +60,73 @@ def __init__(self, n_neurons, n_ensembles, ens_dimensions=1,
...
-            if neuron_nodes and isinstance(e.neuron_type, nengo.Direct):
-                warnings.warn("Creating neuron nodes in an EnsembleArray"
-                              " with Direct neurons")
+        if neuron_nodes:
+            self.add_neuron_input()
+            self.add_neuron_output()
+            warnings.warn(
+                "'neuron_nodes' argument will be removed in Nengo 2.2. Use "
+                "'add_neuron_input' and 'add_neuron_output' methods instead.",
+                DeprecationWarning)
```
arvoelke: Add to #822.

arvoelke:
My only concern is that `add_output` can be called multiple times, while `add_neuron_input` and `add_neuron_output` can only be called once. This limitation seems a little arbitrary/inconsistent to the end-user. I'd be more comfortable if they could both be called multiple times, or moved into `__init__`.

hunse:
I really don't think we want `add_neuron_input/output` being callable multiple times. Then you'd end up with multiple nodes doing exactly the same thing, slowing stuff down.

arvoelke:
Yeah, I was mainly trying to motivate the logic for moving it into `__init__`. :wink:

tbekolay:
How would it work in `__init__`? A separate `neuron_input` and `neuron_output` boolean flag? That wouldn't work for my use case. I have an ensemble array that normally doesn't need neuron input or output. However, for debugging, it was convenient to add in a neuron output in order to look at the spike raster. The ensemble array was already created (in a function called from another function) so it would have been cumbersome to pass along a flag across a bunch of functions. Having a function that I could call after the network's been created allowed me to keep all that network creation code the same, and just call a function and add a probe when I needed to do that debugging.

arvoelke:
Ah, I see... In that case I guess it's easier to make it consistent with `add_output`. It can technically be called multiple times already, so that's fine. The only part that's not technically consistent is the `None` return value, so just make it return `self.neuron_input` / `self.neuron_output`?

hunse:
The thing I don't like about that is people might think they have to call it each time to get the neuron output, and not realize it's adding lots of nodes (though the warning should help).

jgosmann:
> The thing I don't like about that is people might think they have to call it each time to get the neuron output, and not realize it's adding lots of nodes (though the warning should help).

Wouldn't that be clear enough from the name saying `add_...` (and not `get_...`)?

hunse:
I guess the way @tbekolay has it now, it won't actually add a second neuron input or output, it will just give the warning and return. So we could just have it return the existing one there, and it'd be fine. @jgosmann, you're right that `add_` should make it clear, but what's not clear is how to access the thing that's been added. It might be good to add docstrings to the add functions saying where they put the added nodes.

tbekolay:
Added a fixup commit to address comments.

nengo/networks/workingmemory.py
```diff
@@ -37,12 +35,14 @@ def InputGatedMemory(n_neurons, dimensions, feedback=1.0,
...
         net.gate = nengo.Node(size_in=1)
+        net.diff.add_neuron_input()
         nengo.Connection(net.gate, net.diff.neuron_input,
                          transform=np.ones((n_total_neurons, 1)) * -10,
                          synapse=None)
 
         # reset input (if reset=1, remove all values, and set to 0)
         net.reset = nengo.Node(size_in=1)
+        net.mem.add_neuron_input()
         nengo.Connection(net.reset, net.mem.neuron_input,
```
arvoelke: Nit: could demonstrate the alternative of moving the above line `net.mem.add_neuron_input()` to replace `net.mem.neuron_input`.

arvoelke:
LGTM

hunse:
LGTM too. When Travis passes I'll merge.

arvoelke:
Seems like something might be wrong with Travis. Job 2014 was terminated after hanging for 10 minutes?

tbekolay:
Looks like that job wasn't rebased to master, so it was still using a really slow NumPy.

tbekolay:
Also, warning to @hunse, pushed another commit to address the nit, and fix a typo I noticed in ensemblearray's docstring.

hunse:
Should this have a changelog?

tbekolay:
Oh dag, you right. Writing one now!

tbekolay:
Done.

PR #870, opened 2015-10-07
`max_rate` and `intercepts` only set if used

hunse:
When building an ensemble, `max_rate` and `intercepts` were computed
and stored in the `BuiltEnsemble` even if the gains and biases were
provided, in which case they were not used. This was both extra
computation and meant that the max rates and intercepts in the
`BuiltEnsemble` were meaningless and misleading.


Seanny123:
LGTM

tbekolay:
Me too!

PR #871, opened 2015-10-07
Ensemble utility for optimizing parameters

hunse:
I added a utility for doing a simple random search of ensemble parameters to best decode a given function.

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -8,12 +8,12 @@ from nengo.builder.node import SimPyFunc
 from nengo.builder.operator import (
     DotInc, ElementwiseInc, PreserveValue, Reset, SlicedCopy)
 from nengo.connection import Connection
-from nengo.dists import Distribution
 from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError, ObsoleteError
 from nengo.neurons import Direct
 from nengo.node import Node
 from nengo.utils.compat import is_iterable, itervalues
+from nengo.utils.builder import sample
 
 built_attrs = ['eval_points', 'solver_info', 'weights', 'transform']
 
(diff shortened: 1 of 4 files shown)


drasmuss:
This makes me think, it should be possible to write some fairly easy wrappers to convert our distributions into the distributions hyperopt uses (just a matter of mapping our uniform to their uniform, etc.).  Then you could use basically this same function, but with the fancy search routines of hyperopt rather than random search.  Not suggesting that for this PR, just noting it down while I think of it.

hunse:
Yeah, that's a good idea, and should be pretty straightforward (though UniformHypersphere might be a bit tricky, maybe?). I'd still want to have this basic random one, in case people don't have hyperopt installed or don't want to get into it, but it would be great to have that option. I was also wondering whether any more traditional optimization techniques would be applicable (e.g. stochastic gradient descent), but probably not without applying backprop to the whole encoders->neurons->decoders chain.

Seanny123:
LGTM.

drasmuss:
I noticed we were redoing that `sample` function a few places in the code, so I factored it out into `utils`.  I made it a separate commit, because it isn't directly related to this PR, just inspired by it.

Only significant difference is that I used `np.array` instead of `np.asarray`.  We could add a `copy` flag to set which behaviour we want if someone thinks that's important, but as a default I figured I'd just go for the safer option.

drasmuss:
There might be an argument for putting this hyperparameter optimization in `nengo_extras` instead.  No strong feelings either way.

Seanny123:
FYI @drasmuss this is failing flake8

jgosmann:
Might this be better placed in `nengo_extras`?

drasmuss:
That's 2.5 votes for moving this to `nengo_extras` instead, any objections @hunse?

hunse:
Yeah, sounds good to me.

jgosmann:
This mostly LGTM. I think we should separate @drasmuss's factoring out of the `sample` function as this has to be done on the Nengo side. The rest should go into `nengo_extras`.

Only thing that makes me a little bit worried is that this depends on a number of functions from the builder code and I'm not sure how public that API is supposed to be.

Would it make sense to cache decoders in this function? Or isn't it worth the added complexity? (Probably one should save the result of this function somewhere instead of doing this optimization in every build.)

tbekolay:
I've now split this into #1181 and https://github.com/nengo/nengo_extras/pull/26. Closing!

PR #875, opened 2015-10-14
Fix setup.py

tbekolay:
Our Travis build was failing on Python 3 due to a recent `setuptools` release. This fixes it.


Seanny123:
LGTM

hunse:
No idea how any of this stuff works, but your description makes sense!

PR #880, opened 2015-10-22
Fixes error empty vocabs for assoc_mem

### LINKED ISSUE #879: spa.AssociativeMemory give obscure memory when empty vocab given
To replicate run:
```
import nengo
from nengo import spa

D = 32
vo = spa.Vocabulary(D)
model = spa.SPA()

with model:
    model.ass = spa.AssociativeMemory(vo)
```

A warning that the problem is that the vocab is empty would be appreciated

xchoo:
- Creating an associative memory with an empty vocabulary now throws an error.
- Added empty vocabulary test.

Resolves #879

### DIFF
--- a/nengo/networks/assoc_mem.py
+++ b/nengo/networks/assoc_mem.py
@@ -60,6 +60,19 @@ class AssociativeMemory(nengo.Network):
             input_vectors = np.array(input_vectors, ndmin=2)
         if is_iterable(output_vectors):
             output_vectors = np.array(output_vectors, ndmin=2)
+
+        if input_vectors.shape[0] == 0:
+            raise ValueError('Number of input vectors cannot be 0.')
+        elif input_vectors.shape[0] != output_vectors.shape[0]:
+            # Fail if number of input items and number of output items don't
+            # match
+            raise ValueError(
+                'Number of input vectors does not match number of output '
+                'vectors. %d != %d'
+                % (input_vectors.shape[0], output_vectors.shape[0]))
+
+        # Handle possible different threshold / input_scale values for each
+        # element in the associative memory
         if not is_iterable(threshold):
             threshold = threshold * np.ones(input_vectors.shape[0])
         else:
--- a/nengo/networks/tests/test_assoc_mem.py
+++ b/nengo/networks/tests/test_assoc_mem.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
 
 import nengo
 from nengo.networks.assoc_mem import AssociativeMemory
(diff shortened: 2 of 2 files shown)


Seanny123:
LGTM

nengo/networks/assoc_mem.py
```diff
@@ -75,9 +75,12 @@ def __init__(self, input_vectors, output_vectors=None,  # noqa: C901
         if is_iterable(output_vectors):
             output_vectors = np.matrix(output_vectors)
 
-        # Fail if number of input items and number of output items don't
-        # match
-        if input_vectors.shape[0] != output_vectors.shape[0]:
+        if input_vectors.shape[0] <= 0:
+            # Fail if number of input items <= 0
+            raise ValueError('Number of input vectors cannot be <= 0.')
```
hunse: A shape can never be less than zero, so I would change the check above to be `== 0` and this message to say "cannot be zero".

hunse: Also, maybe put a note like "If using a Vocabulary, make sure it is not empty". In fact, I might prefer to have a separate check for Vocabularies, so that we can give a tailored error message.

tbekolay: It might also help to add an `is_array` check. And, if you really want it to be a `np.matrix`, you should make it a matrix if you get an array; right now, you might get different behavior if you pass in a list versus an equivalent np.array of that list.

hunse: That's a good point. I would recommend we never use matrices in Nengo because a) they're never necessary, and b) basic operators like `*` perform different operations for arrays and matrices, making things very confusing.

hunse:
I looked at rebasing this, but so much has changed in this code that I'm not comfortable doing it.

Seanny123:
I'll rebase it next week.

Seanny123:
I've rebased this with master and changed code based on previous comments. It is now ready to be reviewed.

nengo/networks/tests/test_assoc_mem.py
```diff
@@ -21,6 +22,12 @@ def test_am_basic(Simulator, plt, seed, rng):
     """Basic associative memory test."""
 
     D = 64
+    vocab = np.array([])
+
+    with pytest.raises(ValueError):
+        with nengo.Network('model_error', seed=seed):
```
tbekolay: Label and seed are unnecessary here.

tbekolay:
Looks good to me! @drasmuss can you give a quick look? It's pretty small.

PR #881, opened 2015-10-22
Move timer

hunse:
In Nengo OCL, I've been using `Timer` to time the build process. @xchoo was running this on a machine without `pytest` installed, and running into problems because `Timer` was in `utils/testing.py` which imports `pytest` globally. Both these commits help address that. Really, either one of them would fix the problem, but I think they're both good changes.

### DIFF
--- a/nengo/tests/test_probe.py
+++ b/nengo/tests/test_probe.py
@@ -2,7 +2,7 @@ import numpy as np
 
 import nengo
 from nengo.utils.compat import range
-from nengo.utils.testing import Timer
+from nengo.utils.stdlib import Timer
 
 
 def test_multirun(Simulator, rng):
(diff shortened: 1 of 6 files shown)


Seanny123:
LGTM other than then `flake8 errors`.

tbekolay:
I've been using `Timer` to time stuff too; `stdlib` makes sense to me as a place to put it. My only merge hesitation is whether we should add a changelog message... I highly suspect that only Nengo devs are going to know that `Timer` exists, so I think it's okay to go forward without adding a changelog entry, and our changelog is already pretty long, so omitting minor changes makes sense to me also. So yeah, if that makes sense to you @hunse I'll merge!

hunse:
I agree, I don't think we need a changelog entry.

PR #883, opened 2015-10-22
Change cache logging level to debug

hunse:
Large models can have lots of cache interaction, and logging at
'info' level can flood the log. Also made the `log` utility function
able to set arbitrary levels.


jgosmann:
Some tests need adjustment. Or maybe it is best to keep the debug argument for backwards compatibility?
Apart for that it looks good to me.

hunse:
I think very few people are using this (if any), so I'm not worried about backwards compatibility. Fixed the tests.

jgosmann:
This should probably have a changelog entry.

hunse:
Good point!

PR #893, opened 2015-11-10
Explicitly check that output is array_like

### LINKED ISSUE #892: Building a model with a node that has a node as a parameter causes unbounded memory usage
Discovered this after making a typo in my code. Minimal example:

```
    node1 = nengo.Node(output=lambda t: t)
    node2 = nengo.Node(node1)

sim = nengo.Simulator(model)
```

The code is meant to be ```node2 = nengo.Probe(node1)```, but my brain farted. :smile:

hunse:
We assumed that node outputs are array_like if not None, Process,
or callable. This is not necessarily the case, and trying to cast
some objects (e.g. Node) to an array can cause infinite memory
usage. Fixes #892.

### DIFF
--- a/nengo/node.py
+++ b/nengo/node.py
@@ -6,6 +6,7 @@ import nengo.utils.numpy as npext
 from nengo.base import NengoObject, ObjView
 from nengo.params import Default, IntParam, Parameter, StringParam
 from nengo.processes import Process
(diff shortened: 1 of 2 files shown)


Seanny123:
LGTM

Issue #891, opened 2015-11-04
Hetero-associative relationships in spa.AssociativeMemory


ikajic:
From the current `spa.AssociativeMemory` documentation:

```
input_vocab: list of numpy.array, spa.Vocabulary
    The vocabulary (or list of vectors) to match.
output_vocab: list of numpy.array, spa.Vocabulary, optional
    The vocabulary (or list of vectors) to be produced for each match. If
    not given, the associative memory will act like an auto-associative
    memory (cleanup memory).
```

one could conclude that when the `output_vocab` is passed the memory should act like a hetero-associative memory, fairly standard use-case. By definition, this means that the network outputs a different stored pattern from the presented one, and this relationship is somehow stored in the network weights.
Now, when I try to use `spa.AssociativeMemory` in this way, this mapping doesn't happen.
Here a simple example:

``` python
import nengo
from nengo import spa
import matplotlib.pyplot as plt

vocab_input = spa.Vocabulary(dimensions=16)
vocab_output = spa.Vocabulary(dimensions=16)

vocab_input.parse('ONE + TWO')
vocab_output.parse('TWO + ONE')

with nengo.Network('model', seed=1) as model:
    assoc_mem = spa.AssociativeMemory(input_vocab=vocab_input, output_vocab=vocab_output)

    input_node = nengo.Node(output=vocab_input['ONE'].v, label='input')
    nengo.Connection(input_node, assoc_mem.input)

    input_probe = nengo.Probe(input_node)
    output_probe = nengo.Probe(assoc_mem.output, synapse=0.03)

sim = nengo.Simulator(model)
sim.run(.5)
t = sim.trange()

input_net = sim.data[input_probe]
output_net = sim.data[output_probe]

plt.figure(figsize=(16, 4))
plt.subplot(1, 2, 1)
plt.plot(t, nengo.spa.similarity(input_net, vocab_input))
plt.title("Input similarity")
plt.xlabel('Time')
plt.ylim(top=1.1)
plt.legend(vocab_input.keys, loc='best')
plt.subplot(1, 2, 2)
plt.plot(t, nengo.spa.similarity(output_net, vocab_output))
plt.title("Output similarity")
plt.xlabel('Time')
plt.legend(vocab_output.keys, loc='best')

plt.show()
```

For a hetero-associative memory I would expect to see "TWO" in the second plot.

Now, is what is currently happening really the intended behaviour? If it really is, what is the idea behind it?

xchoo:
[...]

PR #895, opened 2015-11-10
Added tutorial on spa.AssociativeMemory

ikajic:
Here are a few simple examples for the `spa.AssociativeMemory` module. 
The basic idea was to show how to use it and to explain how different parameters influence memory behavior. Let me know what you think of the examples and/or if you'd like me to add something :)

I was also thinking of adding a final section mentioning more advanced use-cases such as @arvoelke [Learning new associations](https://github.com/nengo/nengo/blob/voja2/examples/learning/learn_associations.ipynb), but wasn't sure how this would fit here.

P.S: I'll do another PR for fixing the docstrings in `spa.AssociativeMemory`

### DIFF
--- a/examples/spa/associative_memory.ipynb
+++ b/examples/spa/associative_memory.ipynb
@@ -1,0 +1,501 @@
+{
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.4.2"
+  },
+  "name": ""
+ },
+ "nbformat": 3,
+ "nbformat_minor": 0,
+ "worksheets": [
+  {
+   "cells": [
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "# Associative Memory\n",
+      "\n",
+      "This tutorial introduces the Associative Memory (AM) module in the SPA.\n",
+      "\n",
+      "An associative memory is a neural network used to store and recall patterns.\n",
+      "When the network receives a partial or noisy pattern at the input, it can either recover the same pattern or recall another stored pattern. \n",
+      "If the recovered pattern is the same as the input pattern, the memory is said to be autoassociative or a *clean-up* memory. Otherwise, if the recovered pattern is different from the presented one, the network is heteroassociative. \n",
+      "\n",
+      "Patterns stored by the AM module in the SPA are semantic pointers organised in a SPA vocabulary.\n",
+      "The examples in this tutorial demonstrate how to use the AM module to store and recall patterns.\n",
+      "Advanced functionality of the module, such as the recall of multiple memories similar to the input, is also presented.\n",
+      "\n",
+      "A theoretical explanation on how the associative memory is implemented in NEF is available in [Stewart et al. 2010](http://compneuro.uwaterloo.ca/files/publications/stewart.2011.pdf)."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "import numpy as np\n",
+      "import matplotlib.pyplot as plt\n",
+      "%matplotlib inline\n",
+      "\n",
+      "import nengo\n",
+      "from nengo import spa\n",
+      "%load_ext nengo.ipynb"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "## Step 1: Creating an Associative Memory\n",
+      "\n",
+      "We start by creating a set of patterns the AM will store. \n",
+      "The vocabulary in this example contains five words: `ORANGE, APRICOT, CHERRY, STRAWBERRY` and `APPLE`.\n",
+      "Each word is represented as a semantic pointer, an $n$-dimensional vector.\n",
+      "When creating a vocabulary, we specify the number of dimensions for all semantic pointers.\n",
+      "Then, we add the words to the vocabulary with `parse`:"
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "dim = 32\n",
+      "vocab = spa.Vocabulary(dimensions=dim)\n",
+      "\n",
+      "words = ['ORANGE', 'APRICOT', 'CHERRY', 'STRAWBERRY', 'APPLE']\n",
+      "\n",
+      "for word in words:\n",
+      "    vocab.parse(word)"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "Next, we create an autoassociative memory with a corresponding set of stored patterns.\n",
+      "To check the functionality of the memory module, we want to test whether the pattern at the input has been successfully retrieved at the output.\n",
+      "\n",
+      "The memory is created within a `spa.SPA` module.\n",
+      "Using the `spa.Input` module we set the name of the semantic pointer `APPLE` as the input to the associative memory. \n",
+      "`nengo.Probe`s have been added to record inputs and outputs of the module, which we plot for visual inspection.\n",
+      "Finally, we run the simulation for the 0.2 seconds simulation time."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "with spa.SPA('AssociativeMemory', seed=1) as model_1:\n",
+      "    # create the AM module\n",
+      "    model_1.assoc_mem = spa.AssociativeMemory(input_vocab=vocab)\n",
+      "\n",
+      "    # present input to the AM\n",
+      "    model_1.am_input = spa.Input(assoc_mem='APPLE')\n",
+      "\n",
+      "    # record the inputs and outputs during the simulation\n",
+      "    input_probe = nengo.Probe(model_1.assoc_mem.input)\n",
+      "    output_probe = nengo.Probe(model_1.assoc_mem.output, synapse=0.03)\n",
+      "\n",
+      "sim = nengo.Simulator(model_1)\n",
+      "sim.run(0.2)\n",
+      "t = sim.trange()"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "We gather the simulation data:"
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "input_data = sim.data[input_probe]\n",
+      "output_data = sim.data[output_probe]\n",
+      "print(\"Input dimensions: (%d, %d)\" % input_data.shape)\n",
+      "print(\"Output dimensions: (%d, %d)\" % output_data.shape)"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "We use a dot product (implemented in `nengo.spa.similarity`) to compare the similarity of inputs and outputs to all patterns in the vocabulary at every time step. Recall that the first dimension corresponds to the number of time steps and the second dimension to the 32-dimensional pattern represented in the associative memory.\n",
+      "If the similarity between the output vector and the input vector is close to one, we can say that the associative memory successfully retrieved the pattern (or cleaned up the input)."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": true,
+     "input": [
+      "def plot_similarities(input_data, output_data, vocab1, vocab2=None, autoscale=False):\n",
+      "    if vocab2 is None:\n",
+      "        vocab2 = vocab1\n",
+      "    \n",
+      "    ymin, ymax = -1.2, 1.2    \n",
+      "    plt.figure(figsize=(12, 4))\n",
+      "    plt.subplot(1, 2, 1)\n",
+      "    plt.ylim(ymin, ymax)\n",
+      "    if autoscale:\n",
+      "        plt.autoscale(autoscale, axis='y')\n",
+      "    plt.grid(True)\n",
+      "    plt.plot(t, spa.similarity(input_data, vocab1))\n",
+      "    plt.title(\"Input similarity\")\n",
+      "    plt.xlabel(\"Time\")\n",
+      "    plt.legend(vocab1.keys, loc='best')\n",
+      "    \n",
+      "    plt.subplot(1, 2, 2)\n",
+      "    plt.plot(t, spa.similarity(output_data, vocab2))\n",
+      "    plt.title(\"Output similarity\")\n",
+      "    plt.xlabel(\"Time\")\n",
+      "    plt.ylim(ymin, ymax)\n",
+      "    if autoscale:\n",
+      "        plt.autoscale(autoscale, axis='y')\n",
+      "    plt.grid(True)\n",
+      "    plt.legend(vocab2.keys, loc='best')"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "plot_similarities(input_data, output_data, vocab)"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {
+      "collapsed": true
+     },
+     "source": [
+      "We see that the network input has been constant throughout the simulation (`APPLE=1`). Notice that there might be some small similarity between the pattern `APPLE` and some other patterns. This is because the semantic pointers are not perfectly orthogonal and the dot product will amount to some value different from zero. This can be improved by increasing the dimensionality of vectors, yielding more orthogonal representations.\n",
+      "At the output, the similarity of the represented semantic pointer with the semantic pointer `APPLE` increases until it reaches the maximal value (`=1`). This means that the associative memory successfully retrieved the input pattern. The exponential increase is due to the synaptic filtering in `nengo.Probe`. This is used to show how the input to another group of neurons connected to the output of this particular AM module would look like.\n"
+     ]
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "## Step 2: Clean-up Memory\n",
+      "\n",
+      "\n",
+      "In the next example, we show the ability of the associative memory to retrieve a clean pattern from a noisy input. \n",
+      "This is a very common use of autoassociative memories, also called *clean-up* as the AM removes the noise in the input.\n",
+      "\n",
+      "Now, we construct a noisy semantic pointer using a combination of several semantic pointers. This is done by passing the expression `0.9*APPLE+0.5*CHERRY+0.4*APRICOT` as the input to the the network.\n",
+      "The resulting input vector will still be mostly similar to one semantic pointer (`APPLE`) but is also somewhat similar to other semantic pointers in the vocabulary. Therefore, the task of the network will be to clean up the vector `APPLE`.\n",
+      "One way to achieve this in SPA is by using the thresholding mechanism in the AM module. By default, the threshold in `spa.AssociativeMemory` is set to 0.3, assuming that random noise fluctuations in the input should not be contained at the output of the network. \n",
+      "By specifying the threshold, the output will be similar only to those inputs whose similarity with a vocabulary item is above a certain value. In this example we will increase the selectivity of the AM by setting the threshold to to a higher value: "
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "with spa.SPA('CleanupThreshold', seed=1) as model_3:\n",
+      "    model_3.assoc_mem = spa.AssociativeMemory(input_vocab=vocab, threshold=0.7)\n",
+      "    \n",
+      "    input_expr = '0.9*APPLE + 0.5*CHERRY + 0.4*APRICOT'\n",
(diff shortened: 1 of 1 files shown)


jgosmann:
Related to #891 and potentially #797 (those might require changes to the tutorial).

jgosmann:
The tests are failing because the notebook cells have not been cleared as they are supposed to. Hint: There's a useful script to do that in [ctn-waterloo/devscripts](https://github.com/ctn-waterloo/devscripts).

Which IPython notebook version is this using? We should make sure that all examples use the same version (and we should agree on which version that should be).

jgosmann:
Step 3 comes before step 2.

ikajic:
> Which IPython notebook version is this using? 
4.0.6

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "After the network has been created, we can test its functionality by presenting a semantic pointer at the input and comparing it with a semantic pointer at the output after running a simulation. The comparison is done by calculating the dot product. If the similarity between the output vector and the input vector is close to one, we can say that the associative memory successfully retrieved the pattern.\n",
```
Seanny123: Maybe remove "after running a simulation". The user should be familiar with the idea of simulation at this point.

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+      "To achieve this task, we need to introduce two additional objects: `nengo.Node` and `nengo.Probe`. The `nengo.Node` object provides input to the associative memory. \n",
```
Seanny123: Hmmm... Once again, this comes down to what we assume the user is familiar with at this point. I consider it unlikely that they randomly jumped into the Associative Memory example without going through the basic examples, but I'm willing to discuss this further.

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     "metadata": {},
+     "source": [
+      "## Step 2: Memory clean-up\n",
+      "\n",
+      "\n",
+      "In the next example we show the ability of the associative memory to retrieve a noisy pattern. \n",
+      "This is a very common use of autoassociative memories, also called \"clean-up\" as the AM removes the noise in the input.\n",
+      "\n",
+      "Now, instead of presenting just one semantic pointer from a vocabulary, we present a combination of several semantic pointers. This is still one single vector, which is mostly similar to one semantic pointer but is also somewhat similar to other semantic pointers. Therefore, as the input is not \"clean\" we can say that we are presenting a noisy pattern.\n",
+      "The effect would have been similar if we had chosen randomly distributed numbers using, for example, `numpy`'s `random.rand` function. To indicate that the output of the AM should be similar to just one stored vector, we set the parameter `wta_output` to `True`. `WTA` is a computational principle called winner-take-all, stating that one, mostly active element should be regarded as the winner among possible, less similar alternatives."
```
Seanny123: I'm not sure if I understand this line. As in, the effect would have been similar to adding a vector of random numbers to the vector that we want to de-noise?

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     "metadata": {
+      "scrolled": true
+     },
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "We can see that the input presented to the network is mostly similar to the semantic pointer `CHERRY`, but it is also similar to other semantic pointers in the vocabulary (e.g. `APRICOT` and `APPLE`). At the output, the network successfully represents a pattern similar to `CHERRY`. If we wanted to get a pattern at the output whose similarity with the most similar vocabulary item is one, we could have used the `threshold_output` parameter when creating the AM module and setting it to `True`."
```
Seanny123: Maybe instead of "As we have specified the threshold of 0.4 the resulting output vector will be similar to APPLE and CHERRY as their similarity with the input vector is greater than the threshold." write "Since we have specified the threshold of 0.4, and APPLE and CHERRY are both above this threshold, both semantic pointers will be output at their original similarity values."

examples/spa/associative_memory.ipynb
ikajic: It should mean "the noisy vector can also be created by adding random numbers to it". 
It's sort of obvious, so I don't know whether this sentence adds anything new. I will probably remove it.

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "The only input our model receives is `ONE` at the beginning of the simulation. After that, it produces a sequence of consecutive numbers up to five."
```
Seanny123: Why does the similarity of the vector seem to grow over time?

Seanny123:
Very useful tutorial! I've added some comments when you have the time to look at them. (:

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "Next, we create an autoassociative memory with a corresponding set of stored patterns.\n",
+      "The memory is created within a `nengo.Network` module."
```
jgosmann: Because the associative memory is a SPA module `spa.SPA` should be used instead of `nengo.Network`. Not sure if using `nengo.Network` might have unintended consequences, but it might lead to confusing problems in user code. Thus, I think we should always use `spa.SPA` when dealing with SPA stuff.

jgosmann: Note, the PR #797 introduces a basic associative memory implementation that is a normal network (not part of SPA) and a SPA implementation based on this.

examples/spa/associative_memory.ipynb
jgosmann: I agree with @Seanny123. Also, is there a reason to not use `spa.Input`? I think, we talked about that shortly, but I forgot what you said. :frowning: As this example is using a SPA module, I would stick to the SPA methods of connecting things where possible.

ikajic: Wait, the PR #797 is using associative memory just in tests, or am I not
seeing non-test related examples?

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+      "In this example, the input to the network is the semantic pointer `APPLE`:"
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "with model1:\n",
+      "    input_vector = vocab['APPLE'].v\n",
+      "    input_node = nengo.Node(output=input_vector, label='input')\n",
```
jgosmann: `output=input_...` seems strange/contradictory, maybe another reason for using `spa.Input`. Or maybe just leave the `output=` keyword out?

examples/spa/associative_memory.ipynb
ikajic: Wait, the PR #797 is using associative memory just in tests, or am I not
seeing non-test related examples?

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "sim = nengo.Simulator(model1)\n",
+      "sim.run(0.2)\n",
+      "t = sim.trange()\n",
+      "print('Number of simulation steps: %d' % len(t))"
```
jgosmann: I would remove this line as I don't see relevance of the number of simulation steps here.

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "We see that for each simulation step (rows of the `input_net` and the `output_net` arrays) there is a 64-dimensional pattern represented in the associative memory. \n",
```
jgosmann: 32-dimensional
Also it is explicitely set before to 32, so I am not sure if it is necessary to reiterate this fact here and if so, it might be worth explaining that the other array dimension is the number of time steps.

Furthermore, I might name `*_net` `*_data` because it could get confused with a network.

examples/spa/associative_memory.ipynb
ikajic: I actually don't recall any discussion about `spa.Input`...  but I will
look into that!
I just used `spa.Node`s as they were also used in the associative memory
tests...

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {
+      "collapsed": true
+     },
+     "source": [
+      "We see that the network input has been constant throughout the simulation (`APPLE=1`). Notice that there might be some small similarity between the pattern `APPLE` and some other patterns. This is because the semantic pointers are not orthogonal and the dot product will amount to some value different from zero. This can be changed by increasing the dimensionality of vectors, yielding more dissimilar representations.\n",
+      "At the output, the similarity of the represented semantic pointer with the semantic pointer `APPLE` increases until it reaches the maximal value (`=1`). This means that the associative memory successfully retrieved a pattern similar to the input pattern. The exponential increase is due to the synaptic filtering in `nengo.Probe`. This is used to show how the input to another group of neurons connected to the output of this particular AM module would look like.\n"
```
jgosmann: `a pattern similar to the input pattern`
Isn't it almost exactly the input pattern?

examples/spa/associative_memory.ipynb
jgosmann: +1 for removing

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,462 @@
...
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "Now we introduce a Nengo's feature that allows us to pass a function as the input to a `Node`. \n",
```
jgosmann: I think, users should already be familiar with passing functions to nodes at this point.

examples/spa/associative_memory.ipynb
jgosmann: Look at the added file `nengo/networks/assoc_mem.py` (and the changes to `nengo/spa/assoc_mem.py`).

jgosmann:
I like the tutorial and it is definitely a good addition!

I made some minor edits (mostly language even though I am not a native speaker as well ...) apart from the comments I made (note that some of the comments are now on outdated diffs, but still relevant).

Some further general comments:
* I am not sure about the word "retrieve" (might prefer "clean up" in some places). It certainly makes sense when thinking of the network as a memory, but to me (maybe not to others?) it has a connotation of a process which takes some time (like the time a Hopfield net needs to converge), whereas here the process is pretty much instantaneous (except for synaptic delays or if the WTA connections are used).
* The implications of using WTA are somewhat glossed over and I think it might be worth going more into detail about why and when or why not to use WTA. If you want, I can take a stab at adding this myself, but I don't know when I will have time for that (or want to procrastinate other things). The basic spiel is: No WTA is fast because it is purely feed-forward (and I think what is mostly used in our models) and will output only one vector if all the noise is below the threshold. However, it is possible that more than one vector exceed the threshold and the cleaned up version will be a mixture of all of these vectors. This might be desired (e.g. it is not implausible that some recall process recalls two things), but sometimes only one item should win (despite above threshold activation for multiple items). That is where the WTA comes in. But it is not without caveats: It can be slower (especially when switching from one output to another, I believe) and balancing the inhibition might be difficult.
* The tutorial seems to imply that the associative memory does no thresholding per default, but that is not right. Actually, it is our primary method to have only one output vector active (as long as the noise is small enough).
* The tutorial does not explain how the associative memory works. This, however, might be quite useful as it is different from most other approaches I have seen. Or maybe this could be a second tutorial (and this one would only show how to use opposed to how it works internally). Or if we decide that the internals are not important for most people it might be good to at least reference [this paper](http://compneuro.uwaterloo.ca/files/publications/stewart.2011.pdf).

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,495 @@
...
+      "Patterns stored by the AM module in the SPA are semantic pointers organised in a SPA vocabulary.\n",
+      "The examples in this tutorial demonstrate how to use the AM module, store and recall patterns.\n",
+      "Advanced functionality of the module, such as the recall of multiple memories similar to the input, is also presented."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "# python libraries and plotting tools\n",
```
jgosmann: I consider this comment as unnecessary.

examples/spa/associative_memory.ipynb
jgosmann: Just realized that I am not sure whether I understood your question correctly. Yes, PR #797 is not giving examples outside of tests. But my point is that it add an associative memory which does not depend on SPA.

ikajic: Ok a few points here: in the examples there, the AM network is created within `nengo.Network()` and not `spa.SPA()`. If I do this:
```python
In [8]: with spa.SPA():
   ...:     am = spa.AssociativeMemory(v)
```

I get:
```python
ValueError: <AssociativeMemory (unlabeled) at 0x7f20375f2a50> was not added as an attribute of the SPA network and won't be detected
```

Is that expected?

ikajic: Ok, I created the network in a wrong way, this works just fine:
```python

In [11]: with spa.SPA() as model:
    model.am = spa.AssociativeMemory(v)
```

The question remains whether to create it in `nengo.Network` or `spa.SPA`

jgosmann: Yeah, the error is on purpose. Previously, it was not thrown and that could easily lead to very hard to debug problems.

Which examples are you speaking of?

I'm strongly in favor of `spa.SPA`. If there someone disagrees, we should probably bring it up at the dev meeting.

examples/spa/associative_memory.ipynb
ikajic: If you ask for similarity of the vector with output vectors, it's because we use synaptic delay in probes to simulate how this input will look like for another module that connects to associative memory. And the same holds then for the inputs in this examples, which are just rewired outputs.

jgosmann:
1. I think, I would swap the order of introducing thresholding and WTA because it is the default operation and more commonly used in NEF/SPA models.
2. To me `thresholded_output` the explanation of thresholded output could be improved. Especially, as the target vector in your input has already unit length. In that case `thresholded_output` does not really make any difference.

In my opinion the following could be a natural order for introducing the different aspects (potentially switch point two and three):

1. We want to clean up a vector, how do we do that? -> Thresholding
2. Target vector in the input is <1, how to we get a unit length output? -> `thresholded_output`
3. Two vectors in the input are above threshold, but we only want one. -> WTA 

> Nope, but that was not the goal. The goal was to see how to use it and to see what some of the important parameters do. I wouldn't say we should be explaining how the AM works here, but I did include the reference for folks who want to know more.
Fair enough. I think, the reason why I would be interested in a tutorial explaining the AM implementation is that it shows you a lot of advanced tricks, e.g. how to chose neuron tuning curves to perform a specific operation. I have to check, but I believe this does not really get covered in any of the examples. But it is probably best to write another tutorial for that and focus on the usage of the AM module here.

jgosmann:
Also, added a few minor edits (spelling/grammar).

examples/spa/associative_memory.ipynb
tbekolay: I think @Seanny123 is referring to the fact that the initial similarity is around 0.6, but then grows to around 1.2 by pointer FIVE. This is because the transform on the associative memory's recurrent connection is 5. It needs to be somewhat high, as the next pointer needs some momentum or else the associative memory's representation will fade to nothing (the associative memory is not like a working memory; it's not usually recurrently connected).

I found that changing the transform from 5 to 3.3 gives the output that one would expect. I think that a more robust solution would require some intermediate ensembles, or a gating scheme as Xuan did in [working memory](https://github.com/ctn-waterloo/summerschool2015/blob/master/tutorials/memory/Notebooks/4.UsingWM.ipynb), but this is a tutorial and a more complicated network would be more confusing than instructive. But yeah, I think using a transform of 3.3 might raise fewer eyebrows ;)

tbekolay:
Just read through this. Very clear and useful! I haven't seen these used for counting before, so the example at the end was quite cool to see.

Only thing that comes to mind as needing to change is the notebook format. For now, all our notebooks are made with IPython < 3. We'll upgrade them all at some point, but we want all the notebook to be consistent. I can convert the notebook when merging, if you'd like.

jgosmann:
I like the new order. :)

Not sure about changing the transform in the last example. Now I get decaying similarities:
index

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,499 @@
...
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "with spa.SPA('CleanupThreshold', seed=1) as model_3:\n",
+      "    model_3.assoc_mem = spa.AssociativeMemory(input_vocab=vocab, wta_output=True, threshold_output=True)\n",
+      "    \n",
+      "    input_expr = '0.89*APPLE + 0.85*CHERRY + 0.8*APRICOT'\n",
```
jgosmann: Why is this `0.89*APPLE` and not `0.9*APPLE` like in the other examples?

jgosmann:
And one more thing: Would it make sense to have the input and output plot share the y-axis? I feel in some cases it would help.

ikajic:
I think in depends on the generated vectors in the vocabulary.
Have you tried running it several times?

examples/spa/associative_memory.ipynb
ikajic: Because c/p... I'll fix it.

jgosmann:
oh, I only reran the simulation, but with different vocabularies it seems to work mostly ...

jgosmann:
@tbekolay Should a changelog entry be added?

ikajic:
Yup, so if the run looks sloppy we can just advise people to re-run again. Actually, I can add this comment at the end? 
As Trevor pointed out, it's somewhat difficult to get stability in this example without introducing a lot of complexity, the point is to show that this works.

jgosmann:
Yeah, but the comment has to make clear that it is not sufficient to just rerun the last cell because the vocab generation is a few cells further up.
Alternatively we could use fixed seeds (for the vocab and network).

ikajic:
So, with respect to the y-axis... sometimes the lines on the left will go below 0 and above 1, depends again on the generated vectors and how orthogonal they are, and on the right there might be a slightly different range. I would prefer not to change it at this point.

ikajic:
This is what I was planning to add: "If the sequence of numbers at the output decays to quickly, try re-running all the cells in Step 3."

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,499 @@
...
+      "    plt.plot(t, spa.similarity(input_data, vocab1))\n",
+      "    plt.title(\"Input similarity\")\n",
+      "    plt.xlabel(\"Time\")\n",
+      "    plt.legend(vocab1.keys, loc='best')\n",
+      "    \n",
+      "    plt.subplot(1, 2, 2)\n",
+      "    plt.plot(t, spa.similarity(output_data, vocab2))\n",
+      "    plt.title(\"Output similarity\")\n",
+      "    plt.xlabel(\"Time\")\n",
+      "    plt.legend(vocab2.keys, loc='best')"
```
xchoo: The plots have different axis ranges. Makes comparing them less clear.

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,499 @@
...
+     "collapsed": false,
+     "input": [
+      "sim = nengo.Simulator(model_2)\n",
+      "sim.run(0.2)\n",
+      "t = sim.trange()\n",
+      "\n",
+      "input_data = sim.data[input_probe]\n",
+      "output_data = sim.data[output_probe]\n",
+      "\n",
+      "plot_similarities(input_data, output_data, vocab)"
```
xchoo: This plot seemed to cut off the ```APPLE``` entry in the left figure. Is this intentional?

examples/spa/associative_memory.ipynb
xchoo: Reducing the threshold of the cleanup will also help with the decay on the output (i.e. lower threshold = smaller decay). For me, since i like round-ish numbers, I found that a synapse of 0.2 and transform of 3.5 works (although, for those values there is an increase in the output similarity for the first 2 numbers.

examples/spa/associative_memory.ipynb
ikajic: No, not really. It depends on generated vectors. I guess this is a good argument for setting a fixed axis range.

Seanny123:
I think all the issues have been resolved. Is this ready to merge?

examples/spa/associative_memory.ipynb
```diff
@@ -0,0 +1,501 @@
...
+      "\n",
+      "with spa.SPA('Counting', seed=1) as model_5:\n",
+      "    model_5.assoc_mem = spa.AssociativeMemory(input_vocab=vocab_numbers, output_vocab=vocab_numbers,\n",
+      "                                              input_keys=input_keys, output_keys=output_keys,\n",
+      "                                              wta_output=True)\n",
+      "\n",
+      "    model_5.am_input = spa.Input(assoc_mem=input_fun)\n",
+      "\n",
+      "    # added feedback connection\n",
+      "    nengo.Connection(model_5.assoc_mem.output, model_5.assoc_mem.input, synapse=.18, transform=3.3)\n",
```
Seanny123: A quick comment on where the synapse of .18 came from would be appreciated.

Seanny123:
Ready to merge now!

tbekolay:
Cool, looks good to me. I'll add a changelog entry on the merge.

Only thing that I need feedback on is how much of the history of the branch is important. I think my preference is to squash this all into one commit (with shoutouts to @jgosmann and @Seanny123 in the commit message) but I may be missing some changes that would useful to keep for posterity. Any thoughts?

jgosmann:
I'm pretty sure it should be squashed into one commit.

Issue #894, opened 2015-11-10
Move SPA gate and channel to target module


celiasmith:
Currently the gates and channels created for actions that require them are placed in the thalamus itself. This seems to suggest a higher connectivity/bandwidth between cortex/thalamus that I think is expected.  These could instead go in the target modules of those connections, having things depend more on cortical-cortical connections that are mediated by thalamus.

This does not change any models, it just places elements in different places.

Seanny123:
Does this imply just taking these things out of the thalamus or putting them into somewhere else as well?

celiasmith:
Yes, moving them into cortex. @tcstewar had some ideas about how to accomplish this, but basically making them go into the containing network of the target of the action was the plan.  this would make 'Cortical' unnecessary, use that network structure, and allow functions on State to take care of the connections.

s72sue:
I am wondering what the purpose of the bias node in the thalamus is, and where does this node get created?

s72sue:
Figured it out, bias node basically assists in routing and gets created in nengo.networks.Thalamus. 
@celiasmith, I agree that the channels should be moved to the target module, however does moving the gates make much difference?

(a) Following is what moving only the channels looks like in the GUI:
![channels shifted](https://cloud.githubusercontent.com/assets/12284191/11194660/64a125b8-8c7b-11e5-97cb-fe33c9ce1826.png)

(b) and following is what moving both channel and the gates looks like:
![channels and gate shifted](https://cloud.githubusercontent.com/assets/12284191/11194664/6b0d7e4c-8c7b-11e5-84d8-5642300a5a2c.png)
(b)  looks a bit confusing because there are two connections from thalamus to each of the components in the cortex. Keeping the gates in thalamus (as shown in (a) ) makes it look a bit more clear.

xchoo:
The thalamus bias node can be spilt up and relocated into each of the target networks.

s72sue:
Okay, I moved the gate, channel and the bias nodes to the target module. 
However, the connections from thalamus to cortex show up only when I double click the thalamus. @tcstewar, can this be fixed on the nengo_gui side, or should I be doing the restructuring in a different way? 
![image](https://cloud.githubusercontent.com/assets/12284191/11328641/6c589daa-915c-11e5-91a4-db36a7486d30.png)

Seanny123:
[...]

Issue #905, opened 2015-11-24
For convolution effects in the Cortical, the connections do not appear in the nengo_gui


s72sue:
The connections for convolution effects do not appear in the nengo_gui since the convolution network is being created in the 'cortical' context. Since cortical is not represented by a separate entity in nengo_gui, the convolution network as well as the connections do not show up. I think it would make sense to create the convolution network in the context of the target module.

Seanny123:
What network did you try this out with?

s72sue:
The network in the cortical tests (`test_convolution` method) in nengo/spa/test/test_cortical.py

Seanny123:
After discussing during the meeting and with Terry, the convolution networks should be moved from the `cortical` network into the target network, thus enabling us to eliminate the `cortical` network

Seanny123:
Resolved via #906

PR #906, opened 2015-11-24
Moved gate, channel, convolution network to target module

s72sue:
Fix for #894

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -89,6 +89,10 @@ Release History
   vector).
   (`#775 <https://github.com/nengo/nengo/issues/775>`_,
   `#782 <https://github.com/nengo/nengo/pull/782>`_)
+- Bias nodes in ``spa.Cortical``, and gate ensembles and connections in
+  ``spa.Thalamus`` are now stored in the target modules.
+  (`#894 <https://github.com/nengo/nengo/issues/894>`_,
+  `#906 <https://github.com/nengo/nengo/pull/906>`_)
 
 **Improvements**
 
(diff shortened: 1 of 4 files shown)


s72sue:
Does anyone know what this error means?

ikajic:
If you click on "Details" in box, you'll see a little red cross in a build job that failed. Now, you can click on that job and see that there's this line:
```
nengo/spa/thalamus.py:153:30: E128 continuation line under-indented for visual indent
```

so the error simply means that the code-style is not conforming to PEP standards and flake package checks for that. If you just indent ` synapse=self.synapse_to_gate, transform=-1)` in `thalamus.py` like it was before (so it is right below the bracket `(` in the line before), my guess is that this should pass.

Seanny123:
After you fix the syntax error, this pull request looks good to me. Thanks for tackling the problem Su!

s72sue:
Thanks Ivana and Sean!
@Seanny123 , You might want to take another look since I made some changes to the cortical. Since the conv network was moved to the target, it didn't make sense to keep the bias node in cortical, so I have moved that to the target (as required by #905). 

However, the cortical still shows up. Is that because nengo_gui creates a component for each spa module? If yes, should it be checking whether the module is empty?

Seanny123:
Yeah, that's a bug for Nengo GUI to deal with.

Seanny123:
@tbekolay are we still supposed to ping you for merges?

tbekolay:
Eric and Dan can merge as well; anyone on the maintainers team. But yeah, I
can merge this tomorrow morning!

tbekolay:
This makes some API changes, but they're unlikely affect many people, so I'm going to add a short changelog entry. Aside from that, looks great!

PR #915, opened 2015-12-07
Only add PreserveValue for encoders once

### LINKED ISSUE #911: Unable to learn on sliced connections
The following code throws an error in the builder:
```
import nengo
from nengo import spa

import numpy as np
from collections import OrderedDict
import itertools
import ipdb

sim_mode = "gui"
D = 32
rng = np.random.RandomState(0)
vocab = spa.Vocabulary(D, unitary=["ONE"], rng=rng)

vocab.parse("ZERO+ONE+TWO+THREE")

model = spa.SPA(vocabs=[vocab], label="Count Net", seed=0)
n_neurons = 500

with model:
    model.in_1 = spa.State(D)
    model.in_2 = spa.State(D)

    voja1 = nengo.Voja(post_tau=None, learning_rate=5e-2)
    voja2 = nengo.Voja(post_tau=None, learning_rate=5e-2)
    adder = nengo.Ensemble(n_neurons, D*2)
    learning = nengo.Node([0])

    conn_in1 = nengo.Connection(model.in_1.output, adder[D:], learning_rule_type=voja1)
    conn_in2 = nengo.Connection(model.in_2.output, adder[:D], learning_rule_type=voja2)
    nengo.Connection(learning, conn_in1.learning_rule, synapse=None)
    nengo.Connection(learning, conn_in2.learning_rule, synapse=None)

sim = nengo.Simulator(model)
sim.run(1)
```

The error it throws:
```
AssertionError                            Traceback (most recent call last)
/home/sean/nengo_test/sliced_learning.py in <module>()
     31     nengo.Connection(learning, conn_in2.learning_rule, synapse=None)
     32 
---> 33 sim = nengo.Simulator(model)
     34 sim.run(1)

/home/sean/git/nengo/nengo/simulator.pyc in __init__(self, network, dt, seed, model)
    116 
    117         # Order the steps (they are made in `Simulator.reset`)
--> 118         self.dg = operator_depencency_graph(self.model.operators)
    119         self._step_order = [op for op in toposort(self.dg)
    120                             if hasattr(op, 'make_step')]

/home/sean/git/nengo/nengo/utils/simulator.pyc in operator_depencency_graph(operators)
     45             ups[node].append(op)
     46 
---> 47     validate_ops(sets, ups, incs)
     48 
     49     # -- Scheduling algorithm for serial evaluation:

/home/sean/git/nengo/nengo/utils/simulator.pyc in validate_ops(sets, ups, incs)
     82     # -- assert that only one op sets any particular view
     83     for node in sets:
---> 84         assert len(sets[node]) == 1, (node, sets[node])
     85 
     86     # -- assert that only one op updates any particular view

AssertionError: (Signal(<Ensemble (unlabeled) at 0x7f0aa1531390>.scaled_encoders, shape=(500, 64)), [<nengo.builder.operator.PreserveValue object at 0x7f0aa14ff6d0>, <nengo.builder.operator.PreserveValue object at 0x7f0aa14ffdd0>])

```

Is this a bug or should this be disallowed?

hunse:
If multiple VOja (or other encoder learning) rules affect the same
ensemble, the PreserveValue op for the encoders should only be added
once. Fixes #911.

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -233,8 +233,11 @@ def build_connection(model, conn):
             targets.append(r.modifies)
 
         if 'encoders' in targets:
-            model.sig[conn.post_obj]['encoders'].readonly = False
-            model.add_op(PreserveValue(model.sig[conn.post_obj]['encoders']))
+            encoder_sig = model.sig[conn.post_obj]['encoders']
+            if not any(isinstance(op, PreserveValue) and op.dst is encoder_sig
+                       for op in model.operators):
+                encoder_sig.readonly = False
+                model.add_op(PreserveValue(encoder_sig))
         if 'decoders' in targets or 'weights' in targets:
             if weights.ndim < 2:
                 raise BuildError(
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -69,6 +69,10 @@ class Operator(object):
     def __init__(self, tag=None):
         self.tag = tag
 
+    def __repr__(self):
+        return "<%s%s at 0x%x>" % (
+            self.__class__.__name__, self._tagstr, id(self))
+
     @property
     def _tagstr(self):
         return ' "%s"' % self.tag if self.tag is not None else ''


hunse:
This one should be quick to review, @drasmuss or @tbekolay.

nengo/builder/connection.py
```diff
@@ -219,8 +219,11 @@ def get_prepost_signal(is_pre):
             targets.append(r.modifies)
 
         if 'encoders' in targets:
-            model.sig[conn.post_obj]['encoders'].readonly = False
-            model.add_op(PreserveValue(model.sig[conn.post_obj]['encoders']))
+            encoder_sig = model.sig[conn.post_obj]['encoders']
+            if not any(isinstance(op, PreserveValue) and op.dst is encoder_sig
+                       for op in model.operators):
```
tbekolay: Hmm... it's probably not that expensive, but it feels like we shouldn't have to look through all the ops here. Feels like we should have a data structure keeping track of ops like we do for signals in `model.sig`... but for the time being, we do a quick hack of something like a `model.preservevalues` dict that is keyed by the encoder signal which we look at here. Does that make sense or is it overkill for how long this is likely to take? We could also defer that to a later time; if so, make a quick issue so we don't forget and then this looks good to me :)

hunse: It's overkill. Even for a huge model, this is completely negligible compared with solving for decoders and things like that.

But having better storage of operators is something I've wanted in the past, too, particularly when I was looking at optimizers. Do you think it would make sense to store them based on the object that creates them, like we do with signals? That's what I had thought in the past, and I think it would work fairly well. It's not quite as clear-cut as with signals, since many operators copy things from signals belonging to one object to signals belonging to another object. But each operator is unambiguously created in some object's builder, so I think it makes sense to assign ownership that way. I'll make an issue and note that this is one place it would help.

tbekolay: Sounds good!

hunse: See #986

tbekolay:
LGTM. @drasmuss, can you review and merge?

PR #922, opened 2015-12-15
Changed PyPI badges to shields.io

drasmuss:
The PyPI badges were broken, and pypip.in seems to be down indefinitely (https://github.com/badges/pypipins/issues/37).  So changed the badges to use shields.io instead.


hunse:
Seems like PyTest is using shields.io too, so we're not the only ones. Weird that PyPI isn't hosting its own badges, but whatever. I don't pretend to understand this stuff.

PR #924, opened 2015-12-17
Made UniformHypersphere documentation clearer

### LINKED ISSUE #923: Hypersphere teminology
Currently, the `UniformHypersphere` distribution accepts a `surface=False` argument. According to both Wolfram and Wikipedia, a [Hypersphere](http://mathworld.wolfram.com/Hypersphere.html) is _always_ the surface, while a [Ball](http://mathworld.wolfram.com/Ball.html) (a.k.a. n-Disk) is what we have erroneously been calling a sphere (and is the default). It may be too late to change this, so perhaps we should make this clear in the documentation? Or, at the very least, keep this note in mind for any time we're talking about these things to people.

hunse:
It is now clear that we sample from the volume of the sphere by default,
but can optionally sample from the surface. Fixes #923.

### DIFF
--- a/nengo/dists.py
+++ b/nengo/dists.py
@@ -197,7 +197,10 @@ class Exponential(Distribution):
 
 
 class UniformHypersphere(Distribution):
-    """Distributions over an n-dimensional unit hypersphere.
+    """Uniform distribution on or in an n-dimensional unit hypersphere.
+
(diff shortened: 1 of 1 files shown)


tbekolay:
Definitely clearer!

PR #926, opened 2015-12-29
Add SimVoja operator using add_op.

e2crawfo:
Need this change because nengo_mpi overrides add_op.

### DIFF
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -255,7 +255,7 @@ def build_voja(model, voja, rule):
     encoder_scale = model.params[post].gain / post.radius
     assert post_filtered.shape == encoder_scale.shape
 
-    model.operators.append(
+    model.add_op(
         SimVoja(pre_decoded=model.sig[conn]['out'],
                 post_filtered=post_filtered,
                 scaled_encoders=scaled_encoders,


Seanny123:
LGTM. Is there some way we can make sure this mistake won't happen again?

e2crawfo:
We could rename it to _operators to discourage direct manipulation, and then make a property called operators which returns a copy of _operators. Not sure how people feel about that.

Also, can I merge as soon as its been reviewed? Been a while since I've done this :8ball:

tbekolay:
> Also, can I merge as soon as its been reviewed? Been a while since I've done this

Me or @hunse will merge; I'm pretty free ATM so I'll merge shortly.

tbekolay:
And I like the `_operators` change! Maybe in a separate PR though.

hunse:
The `operators` property would have to return the list of operators as a tuple, so that people couldn't append to it. This would mean a bit more copying of stuff. Of course, people can always access `_operators` directly if they really need speed and know what they're doing. [This post](http://stackoverflow.com/questions/23474648/python-read-only-lists-using-the-property-decorator) talks a bit more about it, and the top answer says that copying a list to a tuple *often* has negligible overhead, whatever that means.

PR #928, opened 2016-01-08
Return better error for invalid node return val

Seanny123:
Usually when I accidentally return something from a Node function that can't be turned into a Numpy array, I get a really vague error that doesn't tell me what I did wrong. In this PR, I've changed the error to return the function that returned the invalid value and what the invalid value is.

### DIFF
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -383,9 +383,12 @@ class SimPyFunc(Operator):
             args = () if self.x is None else (np.copy(signals[self.x]),)
             y = fn(t_sig.item(), *args) if t_in else fn(*args)
             if output is not None:
-                if y is None:
-                    raise ValueError(
-                        "Function '%s' returned invalid value" % fn.__name__)
-                output[...] = y
+                if y is None:  # required since Numpy turns None into NaN
+                    raise ValueError("Function %r returned None" % fn.__name__)
+                try:
+                    output[...] = y
+                except ValueError:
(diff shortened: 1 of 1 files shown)


nengo/builder/operator.py
```diff
@@ -386,6 +386,10 @@ def step_simpyfunc():
                 if y is None:
                     raise ValueError(
                         "Function '%s' returned invalid value" % fn.__name__)
```
hunse: Maybe we should change this message too, to say that the invalid value here is `None`. I was trying to think if we could get away without this block, but we still need it because Numpy casts None to NaN, annoyingly.

hunse:
We do check that the function returns a Numpy array at time 0, right? So your functions originally return an array, then at some later time don't?

In Python, I don't think try-catch blocks slow anything down, so I don't see any reason not to have this change.

Seanny123:
> We do check that the function returns a Numpy array at time 0, right? So your functions originally return > an array, then at some later time don't?

Yeah, they'll be returning an array and then I'll mess up one of the cases and then get the vague error.

nengo/builder/operator.py
Seanny123: I updated the message and added a comment

hunse:
What value was actually being returned that was not `None` but causes the error that's caught? Is it some object or something you're returning?

Seanny123:
It was some object or something.

hunse:
Okay, I slightly reworded the messages. LGTM.

Issue #931, opened 2016-01-14
How to handle Simulator.close


drasmuss:
Re: discussion in #859

I think we are all in relative agreement that automatically calling `sim.close` (via `__del__` and `atexit`) would be a good thing.

The only debate is, when we do that, would it be good to throw a warning (saying that the user should probably call close manually, rather than letting it be handled automatically).

I think most(?) people would agree that calling close manually is best practices (for the same reason that calling `file.close` is best practices, even though it can be done automatically).  So the question is, do we want to warn users when they are not following best practices?

I think providing a warning in this case is consistent with our general approach in Nengo.  For example, we throw a warning whenever users create a new attribute on a Nengo object, which isn't even really a bad thing to do.  But that warning helps users avoid potentially hard to find bugs, just like the kind you get when an automatic close doesn't happen as you expected.

drasmuss:
(to be clear, I think it is also reasonable to say that we should not give a warning in this case.  I think there are arguments to be made either way, but I lean like 70/30 towards giving a warning)

tcstewar:
> But that warning helps users avoid potentially hard to find bugs, just like the kind you get when an automatic close doesn't happen as you expected.

I think it'd help to have an example in mind of what sort of bug this could be.  I think that may be part of my problem here -- I can't think of any surprising bugs here.  In nengo_spinnaker, it's just that a nice clear Exception gets thrown telling you that there's still a nengo_spinnaker.Simulator around that hasn't been closed yet.  Are there other sorts of bugs people are thinking of?

drasmuss:
Well it's difficult to predict what will happen when a simulator isn't closed properly, across a range of different backends.  Memory leaks, resource conflicts, it will all depend on what that simulator is supposed to be doing when it closes.  Basically if you have a `close` method, we would assume that is because it is important that the simulator be shut down properly.  And if you want the simulator to be shut down properly, something like the context manager syntax is the best way to do that (which is why Python adopted that format for file IO).

tbekolay:
My preference is toward giving a warning too. If this were code I was writing for myself I wouldn't bother, but we've made lots of decisions toward being more verbose about best practices in the past, so it feels like it's in keeping with the rest of Nengo. It's also not really a big ask of our users, as it doesn't add any additional lines of code, and they are already familiar with the `with` syntax from building their model. And like I mentioned before, people who are annoyed can always filter the warning out with `warnings.filter` (I do this for some of the warnings Brian spits out that are irrelevant to me).

[...]

PR #932, opened 2016-01-15
Close simulator in tests and examples

tbekolay:
Still a work in progress, but thought it might be useful as something concrete for the #931 discussion.

This uses the simulator as a context manager in Nengo tests, and closes open simulators when the process ends. It raises a warning if it does this (which is controversial; we'll leave discussion of that to #931?).

It doesn't do `__del__`, because it's impossible to delete a simulator without closing it, since we keep track of open simulators in `nengo.Simulator.open_sims`.

Note that this is necessary for `nengo_spinnaker` to run unit tests.

Still to do:

- [x] Use sim as context manager in examples (this is causing the new `test_simulator` tests to fail at the moment)
- [x] Change things as a result of discussion in #931


tbekolay:
This PR's been updated to close simulators in examples also. I must be missing something though, as the test of `close_open_sims` ([L97](https://github.com/nengo/nengo/commit/f140fc1c87ed11b5d90dceefc4e500ee5be1d2bf#diff-cf86547944b52ff847e0d9456cc3e7beR97) has 13 open sims if I don't close them all at the start of the test.

In any case, this passes tests locally, so will hopefully also pass on TravisCI. This PR also automatically closes all simualtors at exit; it currently warns if this happens, but that's still up in the air depending on discussion in #931. These changes are in f140fc1c87ed11b.

tbekolay:
I've rebased this to master and pulled out the warning into a separate PR, as per the meeting (see #962). Ready for review!

hunse:
LGTM. I think it's as advertised, and just changing the examples and tests to use `with` blocks, right? There's nothing else I should be looking for?

On a related note, I tried out the `with` block with different types of errors, and it works as we expected I think. If there's an error during build, the simulator `with` block never gets entered, so the simulator is not closed. If there's an error during running (e.g. a node returns `None` unexpectedly), then the simulator does get closed. So that's a good reason to use `with` blocks over the `.close()` method whenever possible.

@tcstewar, what happens on `nengo_spinnaker` when there's an error during build? Does all the memory get properly freed?

tbekolay:
> just changing the examples and tests to use with blocks, right?

Yes, the only other changes are several instances where we either had `Simulator` in the args list but didn't make a `Simulator` instance, or one instance of using `nengo.Simulator` directly. They were obvious here since I was grepping all files for instances of `Simulator`.

hunse:
> the only other changes are several instances where we either had `Simulator` in the args list but didn't make a `Simulator` instance, and one instance of using `nengo.Simulator` directly. They were obvious here since I was grepping all files for instances of `Simulator`.

Would it be possible to add static checks for this? Does `flake8` allow that somehow? The latter wouldn't be too hard to do with our own script, since `nengo.Simulator` should never appear in any of the test files, but checking that all arguments to test functions are used would be difficult.

tbekolay:
> Would it be possible to add static checks for this?

Hmm... it's definitely possible; I am surprised PyFlakes doesn't already do this, as it will complain for other locally scoped variables that don't get used. But not arguments, for some reason. I believe that PyLint would pick up this, but PyLint emits a lot of issues for Nengo ATM so updating the code to pass PyLint would take some time (I tried to do it at one point).

If we want to just add this, or any other minor style thing, we can actually write our own tool that integrates with `flake8`, as `flake8` allows plugins to emit error codes through a `setuptools` entry point. That's something I've had in the back of my mind for a while but haven't put the time into doing it yet. See the [extending flake8 document](http://flake8.readthedocs.org/en/latest/extensions.html).

hunse:
I added an issue for that static checking, and found two more instances of `nengo.Simulator` in the tests, and a few Simulator instances that weren't closed. So I think this is good to go now.

Can we get a second reviewer and/or merger? We've kind of gotten out of the random reviewer selection system. @tbekolay, would it help you out if we picked some random reviewers to get some of this stuff reviewed?

tbekolay:
I stopped doing random people as it's an unnecessary burden for lab members who are busy with other things, so I've been favouring ABR reviewers whenever no one volunteers. Which means @drasmuss!

drasmuss:
Looks good to me

PR #934, opened 2016-01-25
Added PresentInput process

hunse:
This process can present any set of signals in a regular piecewise manner, where each member of the set is presented for the same length of time. For example, this could be a series of images that are each presented for a certain length of time.

I also added two parameters (TupleParam and EnumParam) and made some changes to the Process interface, to support the Conv2d and Pool2d processes that will reside in another respository.

### DIFF
--- a/nengo/processes.py
+++ b/nengo/processes.py
@@ -29,9 +29,9 @@ class Process(FrozenObject):
     default_dt = NumberParam(low=0, low_open=True)
     seed = IntParam(low=0, high=npext.maxint, optional=True)
 
-    def __init__(self, seed=None):
+    def __init__(self, default_size_out=1, seed=None):
         super(Process, self).__init__()
-        self.default_size_out = 1
+        self.default_size_out = default_size_out
         self.default_dt = 0.001
         self.seed = seed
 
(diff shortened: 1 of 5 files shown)


tbekolay:
Aside from my comment on the `TupleParam`, this looks good to me. @drasmuss could you give this a quick review as well?

nengo/processes.py
```diff
@@ -264,6 +278,40 @@ def step(t):
...
+
+    def make_step(self, size_in, size_out, dt, rng):
+        assert size_in == 0
+        assert size_out == self.inputs[0].size
+
+        n = len(self.inputs)
+        inputs = self.inputs.reshape(n, -1)
+        presentation_time = float(self.presentation_time)
+
+        def step_image_input(t):
```
drasmuss: Maybe just `step_input`?

drasmuss:
Just the one very minor comment, other than that looks good to me.

nengo/processes.py
hunse: The names of these functions don't really matter, though they might appear in debugging sometimes. I'm wondering if we should always have them be `step_<class name>`, so in this case `step_presentinput`.

drasmuss: Yeah the only way it really matters is for someone reading through the code and trying to understand what each function does.  In this case it's just kind of odd to mention `image` here, since none of this is image specific.

hunse: That's true. I think that's a remnant of an earlier implementation.

PR #940, opened 2016-01-28
Simultaneous ops

hunse:
This is some work I did a little while ago to allow ops that modify different parts of the same base array to not be dependent on one another in the op graph. This change allows these ops to all be scheduled simultaneously in Nengo OCL.

One good example of this is in EnsembleArray, where many connections modify different (non-overlapping) slices of the output node.


tbekolay:
Asked a clarification question, but the rest looks great! Does this have a big performance impact on OCL?

hunse:
Yeah, this is pretty big for OCL.

tbekolay:
@drasmuss could you be the second reviewer on this one too?

drasmuss:
Looks good to me

Issue #869, opened 2015-10-05
Making allowances for an incremental build process


Seanny123:
As discussed in #867 adding some sort of ability for an incremental build process would be useful.

@jgosmann wants it for SpaOpt
@arvoelke used it for his touch classification paper
@hunse maybe mentioned it for deep belief networks?

I thought I would open a new thread for discussion so this point doesn't get lost to history.

arvoelke:
Thanks. :smile: Disclaimer: it's not at all clear to me how such a process should look from the user's perspective and how this might affect the bigger picture (e.g. nengo_gui, other backends). I have at least two ideas in mind that are very different. I'm happy to brainstorm a bit if we want to take this seriously for the future.

Seanny123:
This seems like a pretty big topic with a lot of things to consider. Maybe we should postpone discussion until the next Nengo meeting (tomorrow)?

hunse:
I would postpone for a while. Getting 2.1 out is more of a priority. If people want to meet in small groups to discuss this, that's fine, and actually I think that kind of brainstorming should happen before we bring this to the big group.

Seanny123:
Sounds good to me.

arvoelke:
Agreed. This is a tricky topic that will take some time to think over.

tcstewar:
Another use case for this: Parisien transforms   #921

tcstewar:
As a quick update, I finally got around to implementing the Parisien transform and so I at least know now what information is needed from the build process for this use case.

The code's in #921 but the basic information needed is:
 - the pre-synaptic decoders
 - the post-synaptic encoders
 - the activity matrices and eval_points for the pre-synaptic neurons
 - the synapse used by the Connection
 - the number of neurons in the pre-synaptic ensemble

Given all of that, it creates three new Connections and a new Ensemble (the interneurons)

Anyway, not sure how all that fits into people's thinking, but I thought I'd post that list here for posterity.

jgosmann:
I finally got spaopt based on the deferrables into a working state (#941) and I still favour this approach over a helper function that modifies the model after constructing it.

[...]

PR #941, opened 2016-02-03
Spaopt v4

jgosmann:
I started putting together another spaopt branch which will hopefully be merged in the future. This implementaton is still WIP, but I would like some comments on how I implement some stuff.

The main issue with spaopt is that it requires the radius to be determined at build time. For that I defined a new type of parameter `BuildTimeParam` which can be set to a function with two arguments: The building simulator and the Nengo object being build (in that order). The builder is supposed to call that function and use the return value as a parameter value.

Because we want to set the radius at build time for spaopt, but not in other cases, I introduce [another type of parameter called `StaticOrBuildTimeParam`](https://github.com/nengo/nengo/compare/spaopt-v4?expand=1#diff-455ef842e1c051dcd59972ebcb821c46R366) that allows for both. There are a few things I would like opinions on:

* `StaticOrBuildTimeParam` gets instantiated with another `Parameter` instance as argument which specifies the permitted parameter values for non-build-time (static) parameters. The `default`, `optional`, and `readonly` values of that instance are ignored. Not sure if that might cause confusion
* `StaticOrBuildTimeParam` derives from `BuildTimeParam`, but not from the class of the static parameter type. That leads to slightly different implementations depending on what value is set (calling `super` vs accessing `self.static...`). But it is pretty equivalent to what is done in DistOrArrayParam.
* It is not possible to have a `StaticOrBuildTimeParam` where the static parameter variant accepts a function because setting it to a function is taken as in indicator that it should be calculated at build time.

A related discussion point:
* Changes to the builder are needed. It has to check whether the parameter is a function (and call that function) or otherwise just use whatever value is given. This will require all backends to implement that. Not sure if there would be any way around this.

[I hard coded the use of the reference `Simulator` in the reference builder.](https://github.com/nengo/nengo/compare/spaopt-v4?expand=1#diff-03ffa76bf6ad121e7cfb21534a08961dR59) One reason is that we have no way to access the Simulator in the build functions at the moment (and I am not sure whether we should add that argument to all the build functions or the model, though maybe some kind of `build_with` argument on the (builder) model would be appropriate). Also, I am not sure how much of an issue the hard coded `Simulator` is. Potentially it could be the wrong one if other backends reuse the builder code, but if they do it might matter because they are using the reference builder code anyways? However, it is ugly that this requires a circular import at the moment (which only works when putting the import in the line before the Simulator is used).

One more question: Could it be an issue to build a model within the build process of another model with some backends? (i.e. When does SpiNNaker acquire the hardware resources in the build/run process?)


jgosmann:
Just remembered one more thing in this PR, I might not be happy with: [The way I copy an ensemble object.](https://github.com/nengo/nengo/pull/941/files#diff-eb1aa957b1efe60550163ce67b82d2bfR34) It seems that it might be desirable to have an easier way to duplicate Nengo objects. (`copy.copy` does not work because they can't be pickled).

tbekolay:
Reminder from the meeting: people who are interested in the build-time parameter setting should read the code in this PR and comment as to what they think of it, and whether it will be a clean solution for their particular use case.

/cc @tcstewar, @arvoelke (?)

mundya:
> Potentially it could be the wrong one if other backends reuse the builder code, but if they do it might matter because they are using the reference builder code anyways? 

This is fine for SpiNNaker at the moment, we have very few calls into the Nengo builder (and I think they're all decoder-solving related).

> One more question: Could it be an issue to build a model within the build process of another model with some backends? (i.e. When does SpiNNaker acquire the hardware resources in the build/run process?)

I think this is fine for us as well. We acquire hardware resources *just* before running becomes possible (i.e., we build a netlist, then acquire/allocate hardware resources, then run).

arvoelke:
Note this PR is a special case of #869.

I don't think there's any way to circumvent the need to integrate this with the builder, and that's probably for the best if we want to keep this general enough to handle a good class of use cases.

How difficult would it be to make `RadiusForUnitVector` have the same interface as the builder methods? That is, take `model` rather than `Simulator`. This might solve the circular dependency problem that you have.

Without yet discussing the potential scope here, I think my only other concern here is the case-handling that was added to determine where the radius should come from. Might be good to have this abstracted away (and as a nitpick, checking the instance is some base class, rather than callable).

I'll have to do some more thinking to determine how close this is to addressing my use cases. But even if it doesn't, I think this potentially taps into an good chunk of them in a fairly straightforward way.

jgosmann:
> How difficult would it be to make RadiusForUnitVector.__call__ have the same interface as the builder methods? That is, take model rather than Simulator. This might solve the circular dependency problem that you have. This would also make it more semantically clear that this method has the same capabilities/scope as their respective/parent builder methods.

I need the `Simulator` class corresponding to the builder. That being said it could certainly make sense to add this as an attribute to the model (the one created in the builder, not the one passed in) and pass that to the function. I think, that could improve a number of things (circular dependencies, hard coded use of `Simulator`).

> Without yet discussing the potential scope here, I think my only other concern here is the case-handling that was added to determine where the radius should come from. Might be good to have this abstracted away

I agree. I am not sure at the moment whether there was reason that made this difficult ...

> and as a nitpick, checking the instance is some base class, rather than callable -- this would also solve your third bullet point

That adds more boiler plate, but might be for the better in this case.

jgosmann:
> I think my only other concern here is the case-handling that was added to determine where the radius should come from. Might be good to have this abstracted away

This would require to get the parameter value one always has to call a function even in the case of a static value. Thus, it would not easily possible to read out the config value if it is a static value. For example when setting `ens.radius = 0.5` and than accessing `ens.radius` it would return a function instead of `0.5`.

jgosmann:
Also note, that function is not easy to evaluate outside of a builder because the (builder) `model` or `Simulator` class will be needed.

jgosmann:
One possibility I see is to treat all parameters as potential build time parameters and add a function to `Parameter` that is called in the builder to access the value. For most parameters it would just return the static value, but `BuildTimeParam` would overwrite it and evaluate the corresponding function if needed. Than the user-facing code would stay the same, but all access to parameters in the builder should be changed (though it would still work if only radius is changed).

Comments? Opinions?

jgosmann:
> Just remembered one more thing in this PR, I might not be happy with: The way I copy an ensemble object. It seems that it might be desirable to have an easier way to duplicate Nengo objects. (copy.copy does not work because they can't be pickled).

@tbekolay We should discuss this at the next dev meeting.

tbekolay:
Sounds good! FYI I have a branch in which I pickle some Config stuff and `LearningRule`s (I used `dill` but I think it works with `pickle`) at [`pickle-config`](https://github.com/nengo/nengo/commits/pickle-config); I think it uses the same general approach that you use.

tbekolay:
> treat all parameters as potential build time parameters

I think this would be my inclination, as I see no reason why some parameters should have this capability but not others, so I would think that it's something we change in either the `Parameter` or config / build system somehow. It's a bit difficult with the `Parameter` itself though since it is used in non-NengoObject subclasses..

jgosmann:
Next todos:

1. [ ] Add `simulator` attribute to builder `model`.
2. [ ] Change `BuildTimeParam` function signature to match other builder functions
3. [ ] Make all Nengo parameters potential build time parameters. (How to handle parameters for non-NengoObject subclasses?)
4. [ ] Be explicit when a function is to be evaluated at build time or is a static parameter?
5. [ ] Figure out proper copying of Nengo objects.
6. [ ] spaopt stuff

jgosmann:
I renamed `StaticOrBuildTimeParam` to `Deferrable`. This class is non-specific to the Nengo build process, it could be used for any kind of parameter that allows for a deferred initialization. Functions to be evaluated at this later point have to be a callable object deriving from `Deferral` (that allows to have deferrable parameters that take functions). A function `func` is easily wrapped into such an object with `Deferral(func)`.

To get easy access to the calculated parameters it is possible to use an `Undeferred` object that evaluates and caches the built time params on access.

The `Deferrable` constructor has changed slightly to produce less boiler plate. In most cases it just takes the parameter class instance as an argument. Or in other words to make an existing parameter deferrable, just add `Deferrable(...)` around it. The only exception is if the default value is supposed to be deferred. In that case the `Deferrable(..., default=Deferral(fn))` argument has to be added (because the original parameter won't allow the assignment because of the type validation).

I wrapped most parameters in the Nengo objects, but some are still missing because they make the tests fail. I still have to look into the reason for that. Some parameters refer to other Nengo objects (e.g. target of a probe). I will probably not wrap those because I don't think we can insert new Nengo objects into the model during the build process and even if we could some function signatures would have to change to actually do that. Apart from that there are certainly parameters where it doesn't seem to very useful to have them deferred to build time, but it shouldn't break anything and so I lean toward allowing it. Who know what use cases people come up with ...

jgosmann:
Two questions we might have to discuss:

1. Do we need to have the parameters of FrozenObjects (e.g. synapses, learning rules) deferrable? It seems to me that these are always assigned to a parameter on a NengoObject which could be deferred and generate the appropriate FrozenObject at build time.
2. Parameter values calculated at build time should be stored for later access. Not sure what the best method is. Adding all parameters (even when they are static) to the named tuples generated by the builder seems to be not the best option and would produce a lot of boilerplate. Maybe it makes sense to just add the parameters that were calculated at build time? Then, however, we would have to change the named tuples to a dict or something similar.

jgosmann:
There are some things I would like to discuss at the next dev meeting (tomorrow?).

tbekolay:
Yep, tomorrow. Anything we can read to prep?

jgosmann:
It's mostly about question 2 a few comments up. There is some (partial) implementation right now in this branch if you want to look at that.

jgosmann:
Oh and the other thing: Now that we have deferrable parameters, it would probably be possible to implement sampling of distributions and setting gains + biases as deferrables. Not sure if that is something we want to do though ...

tbekolay:
Probably here's as good a place as any to organize it a meeting for this week! I made a doodle poll (below). Sign up if you're interesting in talking about the deferred parameter stuff, and are able to take a look at the implementation in this PR by that time. Specifically, you should read (at least) the following commits: 5d9d532a4daa035435f799, b890e80f8b7928c4f46e4cd, bbf51fa885028aa6d7f7b165747c5, 46f305154286ed2c

And if you have time and a use case, it would be great to see if you can add your use case to https://github.com/nengo/nengo/blob/bbf51fa885028aa6d7f7b165747c591afbb61c9f/nengo/utils/optimization.py

Doodle poll: http://doodle.com/poll/ydrcgd2gewbthng4

jgosmann:
Seems like we're meeting tomorrow afternoon?

tbekolay:
Yep, looks like the four of us will meet today (Thursday) at 1pm. Anyone who's interested but didn't get to the doodle poll is welcome to join!

tbekolay:
Meeting starting now! Can someone skype / google hangouts me in?

jgosmann:
I just got an idea, that it should be fairly easy to combine the current approach of `Derral`s with the possibility to register backend specific functions for a `Deferral`.

todos:
- [ ] Separate deferral stuff and spaopt into two PRs.
- [ ] Add helper function to have non supported backends fail.

tbekolay:
Nice :D I will still make a discussion issue as I feel like it's good to keep discussion about the deferral system in general separate from a specific implementation (in case multiple implementations are proposed). Sound good?

jgosmann:
Sounds good.

jgosmann:
arg, I would like to have this PR to stay the deferrable PR (because that's what's mostly discussed here), but I want to change the branch name because it won't be `spaopt-v4` anymore ... But renaming the branch in a PR is not possible, so I guess I have to make this the spaopt PR and a new PR for deferrables ... :angry:

jgosmann:
Recreated this branch by merging 'copy' and 'deferrables' (which are required for spaopt) and reintroducing the spaopt commit. When rebasing this later one should easily be able to rebase everything after the merge commits.

jgosmann:
This should be in a usable state now. The easiest way to activate spaopt is to call `module.use_spaopt()` on the SPA root module. It would be good to get some opinions on the implementation. Currently it defines a `radius_method` parameter on appropriate SPA modules that can be configured. This gives some slight code duplication, but I'm not sure whether the config system handles it correctly if we introduce a common base class for that. Also, the default for circular convolution is slightly different (it uses `2. / np.sqrt(dimensions)` instead of `3.5 / np.sqrt(dimensions)`). Unifying this would make for a little bit less code duplication, but would break some tests that would need readjustment in the error bounds.

Also, note that this PR is somewhat tied to `no-magic-transform` (PR #1014). The spaopt part could potentially be separated out, but I think the implementation would not be as straight forward.

List of further things to do:
- [ ] How to store parameter values computed at build time for later access? Note that all those things will already be cached by the `Undeferred` object for the life time of the `Simulator` object. In theory on could obtain the values by passing a Nengo object through the `Undeferred` constructor as long as the reference to the simulator still exists. Maybe a helper function on the simulator to do this would be the right approach? (Right now everything is also added to a `deferrables` dictionary on the `Built*` objects.)
- [ ] Add helper function to have non supported backends fail.
- [ ] Documentation (new parameters, constructor args, `optimization.py`)
- [ ] Skip tests with spaopt if SciPy is not installed.
- [ ] Use Compare module for dot product
- [ ] Create product module and get rid of the `product_neurons` parameter on `Module`

jgosmann:
The Travis CI build is failing because there is no `scipy.optimize`. Do we have ScipPy at all on Travis?

jgosmann:
#### TODO

Spaopt caches the computed distortion, but it should also cache the optimal radius because the scipy optimize also takes quite some time.

jgosmann:
[Given the decision regarding deferrables](https://github.com/nengo/nengo/pull/995#issuecomment-241622867) in #995 this needs to be reimplemented as a helper function.

Please keep the branch around for reference until this has happened.

PR #943, opened 2016-02-09
Check transform shape for weight learning rules

### LINKED ISSUE #639: Unsupervised learning without initial transform breaks
```python
import nengo

    a = nengo.Ensemble(100, 1)
    b = nengo.Ensemble(100, 1)
    conn = nengo.Connection(a.neurons, b.neurons)
    conn.learning_rule_type = nengo.BCM()

sim = nengo.Simulator(model)
```
This breaks at build time because we try to cheat with the transform, since in the case of a normal connection we would just leave the transform as `np.array(1)` and take advantage of Numpy's broadcasting to apply it properly. In the case of a learned connection, though, the transform ends up too small, and the builder can't apply the learning rule update.

hunse:
Learning rules that modify 'weights' need a 2D transform or weight
solver. This checks that the transform is the right shape. Fixes #639.

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -59,12 +59,25 @@ class ConnectionLearningRuleTypeParam(LearningRuleTypeParam):
                     "(got %r) for learning rule '%s'"
                     % (conn.post_obj.__class__.__name__, rule))
 
-        # If the rule modifies 'weights', then it must have full weights
-        if rule.modifies == 'weights' and conn.is_decoded:
-            raise ValueError(
-                "Learning rule '%s' can not be applied to decoded "
-                "connections. Try setting solver.weights to True or "
-                "connecting between two Neurons objects." % rule)
+        if rule.modifies == 'weights':
+            # If the rule modifies 'weights', then it must have full weights
+            if conn.is_decoded:
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


tbekolay:
This looks great to me! @drasmuss can you review?

nengo/connection.py
```diff
@@ -59,12 +59,25 @@ def validate_rule(self, conn, rule):
...
+                raise ValueError(
+                    "Learning rule '%s' can not be applied to decoded "
+                    "connections. Try setting solver.weights to True or "
+                    "connecting between two Neurons objects." % rule)
+
+            # transform matrix must be 2D
+            pre_size = (
+                conn.pre_obj.n_neurons if isinstance(conn.pre_obj, Ensemble)
+                else conn.pre.size_out)
+            post_size = conn.post.size_out
```
drasmuss: Should this be `conn.post.size_in`?

drasmuss:
Other than that one thing, looks good!

nengo/connection.py
hunse: Hmm, yes, very good point!

hunse:
Ok, fixed that. Thanks for catching it, @drasmuss!

PR #944, opened 2016-02-09
Test function names allow multiple 'parametrize'

hunse:
The `parametrize_function_name` helper in `conftest.py` now supports
multiple 'parametrize' marks on one function.

We only have one function that uses multiple parametrize marks (`test_ensemble.test_eval_points_scaling`), and it doesn't use `plt` or anything else that requires function names, so this wasn't currently breaking anything. But it's nice to have the ability.

### DIFF
--- a/nengo/tests/conftest.py
+++ b/nengo/tests/conftest.py
@@ -67,15 +67,14 @@ def recorder_dirname(request, name):
 def parametrize_function_name(request, function_name):
     suffixes = []
     if 'parametrize' in request.keywords:
-        argnames = [
-            x.strip()
-            for x in request.keywords['parametrize'].args[0].split(',')]
+        argnames = request.keywords['parametrize'].args[::2]
+        argnames = [x.strip() for names in argnames for x in names.split(',')]
         for name in argnames:
             value = request.getfuncargvalue(name)
             if inspect.isclass(value):
                 value = value.__name__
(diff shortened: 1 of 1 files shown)


nengo/tests/conftest.py
```diff
@@ -67,15 +67,14 @@ def recorder_dirname(request, name):
...
-            for x in request.keywords['parametrize'].args[0].split(',')]
+        argnames = request.keywords['parametrize'].args[::2]
+        argnames = [x.strip() for names in argnames for x in names.split(',')]
         for name in argnames:
             value = request.getfuncargvalue(name)
             if inspect.isclass(value):
                 value = value.__name__
             suffixes.append('{0}={1}'.format(name, value))
-    return '_'.join([function_name] + suffixes)
+    return '+'.join([function_name] + suffixes)
```
hunse: I've also changed the separator between parameters to be a `+`. This was the only character I could find that was different from `_`, which we regularly use in function names and thus I'd like to avoid here, but still fit the [cross-platform file name best practices](https://support.apple.com/en-ca/HT202808).

jgosmann: We could also use `-` which is another common separation character in filenames which cannot occur in function names (but as part of numbers behind the `=`).

I don't really have a preference either way.

jgosmann:
LGTM :cake:

PR #945, opened 2016-02-09
Made Synapses a subclass of Processes

hunse:
This unifies the Synapse and Process interfaces.

The big changes happen with moving Process to a separate file to avoid circular imports. The next commit actually makes Synapses a subclass of Process. This breaks some things, which is why I've left it marked as WIP with a couple TODO items in the message. The next two commits go about fixing these problems. I'm not sure how best to arrange this for merge, so I've just left it like this for now. But I think it should be ready for review and discussion.


hunse:
@drasmuss, could you give this a look-over?

hunse:
Okay, I've got everything passing here, and addressed @drasmuss's comments. Specifically, I've put Process in `base.py` instead of its own file, and used the name `mode` for the way that the `SimProcess` op writes its output (rather than `kind`).

drasmuss:
Looks good!

hunse:
Oh, one other thing to draw attention to: In the last commit, I make the default for `filt` to use the first value of the input being filtered to initialize the filter, rather than zero. This means if you're filtering data that doesn't start at zero, your filtered output will not have that annoying dip at the beginning. For example, under this new system, if you have a signal that's a constant value, the filtered version will be that same constant value (i.e. filtering will do nothing), whereas under the old system, you would get a signal that starts at zero and then approaches the constant value exponentially (assuming you're using a Lowpass filter). I think this new behaviour is better.

The only downside is that this behaviour is no longer the same as we use in our models. In our models, synapses are always initialized to zero. So in some of the tests, when calling `filt`, I have to set `y0=0` so that those synapses start at zero like the models.

For the most part, I think `filt` is going to be used by people to filter their own data, in which case I think they'll want the new behaviour. I just wanted to point it out because it is different, and will definitely warrant a changelog entry (as will other parts of this PR).

tbekolay:
I'll look at this PR as soon as I can!

tbekolay:
OK, took a look at this. I think in general it all looks good to me, but probably needs one or two changelog entries. Some of the arg names and whatnot I'm not a fan of (e.g., `x0`, `y0`) but I realize that that's what is used in signal processing literature, so I'm ok with it. Using the first value for `filt` is fine by me also.

The deprecation of `nengo.synapses.filt` and `filtfilt` should be in the changelog as well... and probably some other stuff! But yeah, I think with the changelog entry this is good to go.

Re: moving `Process` to `base.py`, it might make sense (in the future, definitely not this PR) to move other subclasses to `base.py` as well (e.g., `NeuronType`, `LearningRuleType`).

hunse:
Ok, added some changelog entries!

PR #946, opened 2016-02-10
Cache improvements

jgosmann:
The cache was always enabled, even without a seed (as long as not explicitely disabled). This PR changes this to use it only if the seed is actually set.

Currently, I am looking a bit more into other ways to improve the performance. This will likely include not caching results for small ensembles.

### DIFF
--- a/nengo/builder/builder.py
+++ b/nengo/builder/builder.py
@@ -10,23 +10,65 @@ from nengo.exceptions import BuildError
 
 
 class Model(object):
-    """Output of the Builder, used by the Simulator."""
+    """Output of the Builder, used by the Simulator.
+
+    Attributes
+    ----------
+    config : Config instance or None
+        Build functions can set a config object here to affect sub-builders.
+    decoder_cache : DecoderCache instance
+        An object that will be used to cache decoders for faster builds.
+    dt : float
+        The length of each timestep, in seconds.
+    label : str or None
+        A name or description to differentiate different models.
+    operators : list of Operator instances
+        List of all operators created in the build process.
+        All operators must be added to this list, as it is used by Simulator.
+    params : dict (NengoObject -> namedtuple)
+        Mapping from objects to namedtuples containing parameters generated
+        in the build process.
+    probes : list of Probe
+        List of all probes. Probes must be added to this list in the build
+        process, as this list is used by Simulator.
+    seeded : dict (NengoObject -> bool)
+        All objects are assigned a seed, whether the user defined the seed
+        or it was automatically generated. 'seeded' keeps track of whether
+        the seed is user-defined. We consider the seed to be user-defined
+        if it was set directly on the object, or if a seed was set on the
+        network in which the object resides, or if a seed was set on any
+        ancestor network of the network in which the object resides.
+    seeds : dict (NengoObject -> int)
+        Mapping from objects to the integer seed assigned to that object.
+    sig : dict (str -> dict (object -> Signal))
+        A dictionary of dictionaries that organizes all of the signals
+        created in the build process, as build functions often need to
+        access signals created by other build functions.
+    step : Signal(dtype=int64)
+        The current step (i.e., how many timesteps have occurred thus far).
+    time : Signal(dtype=float64)
+        The current point in time.
+    toplevel : Network instance
+        The top-level network being built.
+        This is sometimes useful for accessing network elements after build,
+        or for the network builder to determine if it is the top-level network.
+    """
 
     def __init__(self, dt=0.001, label=None, decoder_cache=NoDecoderCache()):
         self.dt = dt
         self.label = label
         self.decoder_cache = decoder_cache
 
-        # We want to keep track of the toplevel network
+        # Will be filled in by the network builder
         self.toplevel = None
-        # Builders can set a config object to affect sub-builders
         self.config = None
 
-        # Resources used by the build process.
+        # Resources used by the build process
         self.operators = []
         self.params = {}
-        self.seeds = {}
         self.probes = []
+        self.seeds = {}
+        self.seeded = {}
 
         self.sig = collections.defaultdict(dict)
         self.sig['common'][0] = Signal(0., readonly=True, name='ZERO')
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -80,7 +80,8 @@ def build_decoders(model, conn, rng, transform):
         targets = multiply(targets, transform.T)
 
     try:
-        wrapped_solver = model.decoder_cache.wrap_solver(solve_for_decoders)
+        wrapped_solver = (model.decoder_cache.wrap_solver(solve_for_decoders)
+                          if model.seeded[conn] else solve_for_decoders)
         decoders, solver_info = wrapped_solver(
             conn.solver, conn.pre_obj.neuron_type, gain, bias, x, targets,
             rng=rng, E=E)
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -295,17 +295,21 @@ class DecoderCache(object):
         up to date. Once legacy files are removed, a legacy file will be
         written to avoid a costly ``os.listdir`` after calling this.
         """
-        if self._check_legacy_file():
-            return
+        lock_filename = 'legacy.lock'
+        with FileLock(os.path.join(self.cache_dir, lock_filename)):
+            if self._check_legacy_file():
+                return
 
-        for f in os.listdir(self.cache_dir):
-            path = os.path.join(self.cache_dir, f)
-            if os.path.isdir(path):
-                shutil.rmtree(path)
-            else:
-                os.remove(path)
+            for f in os.listdir(self.cache_dir):
+                if f == lock_filename:
+                    continue
+                path = os.path.join(self.cache_dir, f)
+                if os.path.isdir(path):
+                    shutil.rmtree(path)
+                else:
+                    os.remove(path)
 
-        self._write_legacy_file()
+            self._write_legacy_file()
 
     @staticmethod
     def get_default_dir():
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -46,6 +46,14 @@ def safe_remove(path):
         logger.warning("OSError during safe_remove: %s", err)
 
 
+def safe_makedirs(path):
+    if not os.path.exists(path):
+        try:
+            os.makedirs(path)
+        except OSError as err:
+            logger.warning("OSError during safe_makedirs: %s", err)
+
+
 class Fingerprint(object):
     """Fingerprint of an object instance.
 
@@ -166,8 +174,7 @@ class DecoderCache(object):
         if cache_dir is None:
             cache_dir = self.get_default_dir()
         self.cache_dir = cache_dir
-        if not os.path.exists(self.cache_dir):
-            os.makedirs(self.cache_dir)
+        safe_makedirs(self.cache_dir)
         self._fragment_size = get_fragment_size(self.cache_dir)
         try:
             self._remove_legacy_files()
@@ -405,8 +412,7 @@ class DecoderCache(object):
         prefix = key[:2]
         suffix = key[2:]
         directory = os.path.join(self.cache_dir, prefix)
-        if not os.path.exists(directory):
-            os.makedirs(directory)
+        safe_makedirs(directory)
         return os.path.join(directory, suffix + self._CACHE_EXT)
 
 
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -1,18 +1,23 @@
 """Caching capabilities for a faster build process."""
 
+import errno
 import hashlib
 import inspect
 import logging
 import os
+import shutil
 import struct
+from uuid import uuid1
+import warnings
 
 import numpy as np
 
-from nengo.exceptions import FingerprintError
+from nengo.exceptions import FingerprintError, TimeoutError
 from nengo.rc import rc
+from nengo.utils import nco
 from nengo.utils.cache import byte_align, bytes2human, human2bytes
 from nengo.utils.compat import is_string, pickle, PY2
-from nengo.utils import nco
+from nengo.utils.lock import FileLock
 
 logger = logging.getLogger(__name__)
 
@@ -68,6 +73,68 @@ class Fingerprint(object):
         return self.fingerprint.hexdigest()
 
 
+class CacheIndex(object):
+    def __init__(self, filename):
+        self.filename = filename
+        self._lock = FileLock(self.filename + '.lock')
+        with self._lock:
+            self._index = self._load_index()
+        self._updates = {}
+        self._deletes = set()
+        self._removed_files = set()
+
+    def __getitem__(self, key):
+        if key in self._updates:
+            return self._updates[key]
+        else:
+            return self._index[key]
+
+    def __setitem__(self, key, value):
+        self._updates[key] = value
+
+    def __delitem__(self, key):
+        self._deletes.add(key)
+
+    def remove_file_entry(self, filename):
+        self._removed_files.add(filename)
+
+    def __del__(self):
+        self.sync()
+
+    def _load_index(self):
+        assert self._lock.acquired
+        try:
+            with open(self.filename, 'rb') as f:
+                return pickle.load(f)
+        except IOError as err:
+            if err.errno == errno.ENOENT:
+                return {}
+            else:
+                raise
+
+    def sync(self):
+        try:
+            with self._lock:
+                self._index = self._load_index()
+                self._index.update(self._updates)
+
+                for key in self._deletes:
+                    del self._index[key]
+
+                with open(self.filename, 'wb') as f:
+                    pickle.dump(
+                        {k: v for k, v in self._index.items()
+                         if v[0] not in self._removed_files},
+                        f, pickle.HIGHEST_PROTOCOL)
+        except TimeoutError:
+            warnings.warn(
+                "Decoder cache index could not acquire lock. "
+                "Cache index was not synced.")
+
+        self._updates.clear()
+        self._deletes.clear()
+
+
 class DecoderCache(object):
     """Cache for decoders.
 
@@ -90,8 +157,9 @@ class DecoderCache(object):
     """
 
     _CACHE_EXT = '.nco'
+    _INDEX = 'index'
     _LEGACY = 'legacy.txt'
-    _LEGACY_VERSION = 0
+    _LEGACY_VERSION = 1
 
     def __init__(self, readonly=False, cache_dir=None):
         self.readonly = readonly
@@ -101,7 +169,28 @@ class DecoderCache(object):
         if not os.path.exists(self.cache_dir):
             os.makedirs(self.cache_dir)
         self._fragment_size = get_fragment_size(self.cache_dir)
-        self._remove_legacy_files()
+        try:
+            self._remove_legacy_files()
+            self._index = CacheIndex(os.path.join(self.cache_dir, self._INDEX))
+        except TimeoutError:
+            warnings.warn(
+                "Decoder cache could not acquire lock and was deactivated.")
+            self._index = {}
+            self.readonly = True
+        self._fd = None
+
+    def _get_fd(self):
+        if self._fd is None:
+            self._fd = open(self._key2path(str(uuid1())), 'wb')
+        return self._fd
+
+    def _close_fd(self):
+        if self._fd is not None:
+            self._fd.close()
+            self._fd = None
+
+    def __del__(self):
+        self._close_fd()
 
     def get_files(self):
         """Returns all of the files in the cache.
@@ -145,10 +234,15 @@ class DecoderCache(object):
         limit : int, optional
             Maximum size of the cache in bytes.
         """
+        if self.readonly:
+            return
+
         if limit is None:
             limit = rc.get('decoder_cache', 'size')
         if is_string(limit):
             limit = human2bytes(limit)
+
+        self._close_fd()
 
         fileinfo = []
         excess = -limit
@@ -167,10 +261,14 @@ class DecoderCache(object):
                 break
 
             excess -= size
+            self._index.remove_file_entry(path)
             safe_remove(path)
+
+        self._index.sync()
 
     def invalidate(self):
         """Invalidates the cache (i.e. removes all cache files)."""
+        self._close_fd()
         for path in self.get_files():
             safe_remove(path)
 
@@ -202,8 +300,10 @@ class DecoderCache(object):
 
         for f in os.listdir(self.cache_dir):
             path = os.path.join(self.cache_dir, f)
-            if not os.path.isdir(path):
-                safe_remove(path)
+            if os.path.isdir(path):
+                shutil.rmtree(path)
+            else:
+                os.remove(path)
 
         self._write_legacy_file()
 
@@ -244,20 +344,25 @@ class DecoderCache(object):
 
             key = self._get_cache_key(
                 solver_fn, solver, neuron_type, gain, bias, x, targets, rng, E)
-            path = self._key2path(key)
             try:
+                path, start, end = self._index[key]
+                if self._fd is not None:
+                    self._fd.flush()
                 with open(path, 'rb') as f:
+                    f.seek(start)
                     solver_info, decoders = nco.read(f)
             except:
-                logger.debug("Cache miss [{}].".format(key))
+                logger.debug("Cache miss [%s].", key)
                 decoders, solver_info = solver_fn(
                     solver, neuron_type, gain, bias, x, targets, rng=rng, E=E)
                 if not self.readonly:
-                    with open(path, 'wb') as f:
-                        nco.write(f, solver_info, decoders)
+                    fd = self._get_fd()
+                    start = fd.tell()
+                    nco.write(fd, solver_info, decoders)
+                    end = fd.tell()
+                    self._index[key] = (fd.name, start, end)
             else:
-                logger.debug(
-                    "Cache hit [{}]: Loaded stored decoders.".format(key))
+                logger.debug("Cache hit [%s]: Loaded stored decoders.", key)
             return decoders, solver_info
         return cached_solver
 
--- a/nengo/tests/test_cache.py
+++ b/nengo/tests/test_cache.py
@@ -271,6 +271,19 @@ def test_cache_works(tmpdir, RefSimulator, seed):
         assert len(os.listdir(cache_dir)) == 2  # legacy.txt and *.nco
 
 
+def test_cache_not_used_without_seed(tmpdir, Simulator):
+    cache_dir = str(tmpdir)
+
+    model = nengo.Network()
+    with model:
+        nengo.Connection(nengo.Ensemble(10, 1), nengo.Ensemble(10, 1))
+
+    assert len(os.listdir(cache_dir)) == 0
+    Simulator(model, model=nengo.builder.Model(
+        dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir)))
+    assert len(os.listdir(cache_dir)) == 1  # legacy.txt
+
+
 def calc_relative_timer_diff(t1, t2):
     return (t2.duration - t1.duration) / (t2.duration + t1.duration)
 
--- a/nengo/tests/test_cache.py
+++ b/nengo/tests/test_cache.py
@@ -151,6 +151,7 @@ def test_decoder_cache_shrinking(tmpdir):
         timestamp -= 60 * 60 * 24 * 2  # 2 days
         os.utime(path, (timestamp, timestamp))
 
+    cache = DecoderCache(cache_dir=cache_dir)
     cache.wrap_solver(another_solver)(**get_solver_test_args())
 
     cache_size = cache.get_size_in_bytes()
@@ -189,13 +190,17 @@ def test_decoder_cache_shrink_threadsafe(monkeypatch, tmpdir):
     cache_size = cache.get_size_in_bytes()
     assert cache_size > 0
 
-    def raise_file_not_found(*args, **kwargs):
-        raise OSError(errno.ENOENT, "File not found.")
+    def raise_file_not_found(orig_fn):
+        def fn(filename, *args, **kwargs):
+            if filename.endswith('.lock'):
+                return orig_fn(filename, *args, **kwargs)
+            raise OSError(errno.ENOENT, "File not found.")
+        return fn
 
     monkeypatch.setattr(cache, 'get_size_in_bytes', lambda: cache_size)
-    monkeypatch.setattr('os.stat', raise_file_not_found)
-    monkeypatch.setattr('os.remove', raise_file_not_found)
-    monkeypatch.setattr('os.unlink', raise_file_not_found)
+    monkeypatch.setattr('os.stat', raise_file_not_found(os.stat))
+    monkeypatch.setattr('os.remove', raise_file_not_found(os.remove))
+    monkeypatch.setattr('os.unlink', raise_file_not_found(os.unlink))
 
     cache.shrink(limit)
 
@@ -268,7 +273,7 @@ def test_cache_works(tmpdir, RefSimulator, seed):
     assert len(os.listdir(cache_dir)) == 0
     with RefSimulator(model, model=nengo.builder.Model(
             dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):
-        assert len(os.listdir(cache_dir)) == 2  # legacy.txt and *.nco
+        assert len(os.listdir(cache_dir)) == 3  # legacy.txt, index, and *.nco
 
 
 def test_cache_not_used_without_seed(tmpdir, Simulator):
@@ -281,11 +286,13 @@ def test_cache_not_used_without_seed(tmpdir, Simulator):
     assert len(os.listdir(cache_dir)) == 0
     Simulator(model, model=nengo.builder.Model(
         dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir)))
-    assert len(os.listdir(cache_dir)) == 1  # legacy.txt
+    assert len(os.listdir(cache_dir)) == 2  # legacy.txt and index
 
 
-def calc_relative_timer_diff(t1, t2):
-    return (t2.duration - t1.duration) / (t2.duration + t1.duration)
+def reject_outliers(data):
+    med = np.median(data)
+    limits = 1.5 * (np.percentile(data, [25, 75]) - med) + med
+    return np.asarray(data)[np.logical_and(data > limits[0], data < limits[1])]
 
 
 class TestCacheBenchmark(object):
@@ -392,12 +399,6 @@ sim = nengo.Simulator(model)
         plt.ylabel("Build time (s)")
         plt.legend(loc='best')
 
-    @staticmethod
-    def reject_outliers(data):
-        med = np.median(data)
-        limits = 1.5 * (np.percentile(data, [25, 75]) - med) + med
-        return data[np.logical_and(data > limits[0], data < limits[1])]
-
     @pytest.mark.compare
     @pytest.mark.parametrize('varying_param', ['D', 'N', 'M'])
     def test_compare_cache_benchmark(self, varying_param, analytics_data, plt):
@@ -410,8 +411,8 @@ sim = nengo.Simulator(model)
 
         print("Cache, varying {0}:".format(axis_label))
         for label, key in zip(self.labels, self.keys):
-            clean_d1 = [self.reject_outliers(d) for d in d1[key]]
-            clean_d2 = [self.reject_outliers(d) for d in d2[key]]
+            clean_d1 = [reject_outliers(d) for d in d1[key]]
+            clean_d2 = [reject_outliers(d) for d in d2[key]]
             diff = [np.median(b) - np.median(a)
                     for a, b in zip(clean_d1, clean_d2)]
 
@@ -430,3 +431,64 @@ sim = nengo.Simulator(model)
         plt.xlabel("Number of %s" % axis_label)
         plt.ylabel("Difference in build time (s)")
         plt.legend(loc='best')
+
+
+class TestCacheShrinkBenchmark(object):
+    n_trials = 50
+
+    setup = '''
+import numpy as np
+import nengo
+import nengo.cache
+from nengo.rc import rc
+
+rc.set("decoder_cache", "path", {tmpdir!r})
+
+for i in range(10):
+    model = nengo.Network(seed=i)
+    with model:
+        a = nengo.networks.EnsembleArray(10, 128, 1)
+        b = nengo.networks.EnsembleArray(10, 128, 1)
+        conn = nengo.Connection(a.output, b.input)
+    with nengo.Simulator(model):
+        pass
+
+rc.set("decoder_cache", "size", "0KB")
+cache = nengo.cache.DecoderCache()
+    '''
+
+    stmt = 'cache.shrink()'
+
+    @pytest.mark.slow
+    @pytest.mark.noassertions
+    def test_cache_shrink_benchmark(self, tmpdir, analytics, logger):
+        times = timeit.repeat(
+            stmt=self.stmt, setup=self.setup.format(tmpdir=str(tmpdir)),
+            number=1, repeat=self.n_trials)
+        logger.info("Shrink took a minimum of %f seconds.", np.min(times))
+        logger.info("Shrink took a %f seconds on average.", np.mean(times))
+        logger.info(
+            "Shrink took a %f seconds on average with outliers rejected.",
+            np.mean(reject_outliers(times)))
+        analytics.add_data('times', times)
+
+    @pytest.mark.compare
+    def test_compare_cache_shrink_benchmark(self, analytics_data, plt):
+        stats = pytest.importorskip('scipy.stats')
+
+        d1, d2 = (x['times'] for x in analytics_data)
+        clean_d1 = reject_outliers(d1)
+        clean_d2 = reject_outliers(d2)
+
+        diff = np.median(clean_d2) - np.median(clean_d1)
+
+        p_value = 2. * stats.mannwhitneyu(clean_d1, clean_d2)[1]
+        if p_value < .05:
+            print("Significant change of {d} seconds (p <= {p:.3f}).".format(
+                d=diff, p=np.ceil(p_value * 1000.) / 1000.))
+        else:
+            print("No significant change ({d}).".format(d=diff))
+        print("Speed up:", np.median(clean_d1) / np.median(clean_d2))
+
+        plt.scatter(np.ones_like(d1), d1, c='b')
+        plt.scatter(2 * np.ones_like(d2), d2, c='g')
--- a/nengo/tests/test_cache.py
+++ b/nengo/tests/test_cache.py
@@ -313,22 +313,22 @@ with model:
 
     with_cache_miss_ro = {
         'rc': '''
+nengo.cache.DecoderCache().invalidate()
 rc.set("decoder_cache", "enabled", "True")
 rc.set("decoder_cache", "readonly", "True")
 ''',
         'stmt': '''
-nengo.cache.DecoderCache().invalidate()
 sim = nengo.Simulator(model)
 '''
     }
 
     with_cache_miss = {
         'rc': '''
+nengo.cache.DecoderCache().invalidate()
 rc.set("decoder_cache", "enabled", "True")
 rc.set("decoder_cache", "readonly", "False")
 ''',
         'stmt': '''
-nengo.cache.DecoderCache().invalidate()
 sim = nengo.Simulator(model)
 '''
     }
--- a/nengo/tests/test_cache.py
+++ b/nengo/tests/test_cache.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import errno
+import multiprocessing
 import os
 import timeit
 
@@ -276,7 +277,7 @@ def test_cache_works(tmpdir, RefSimulator, seed):
         assert len(os.listdir(cache_dir)) == 3  # legacy.txt, index, and *.nco
 
 
-def test_cache_not_used_without_seed(tmpdir, Simulator):
+def test_cache_not_used_without_seed(tmpdir, RefSimulator):
     cache_dir = str(tmpdir)
 
     model = nengo.Network()
@@ -284,9 +285,36 @@ def test_cache_not_used_without_seed(tmpdir, Simulator):
         nengo.Connection(nengo.Ensemble(10, 1), nengo.Ensemble(10, 1))
 
     assert len(os.listdir(cache_dir)) == 0
-    Simulator(model, model=nengo.builder.Model(
-        dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir)))
-    assert len(os.listdir(cache_dir)) == 2  # legacy.txt and index
+    with RefSimulator(model, model=nengo.builder.Model(
+            dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):
+        assert len(os.listdir(cache_dir)) == 2  # legacy.txt and index
+
+
+def build_many_ensembles(cache_dir, RefSimulator):
+    with nengo.Network(seed=1) as model:
+        for _ in range(100):
+            nengo.Connection(nengo.Ensemble(10, 1), nengo.Ensemble(10, 1))
+
+    with RefSimulator(model, model=nengo.builder.Model(
+            dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):
+        pass
+
+
+@pytest.mark.slow
+def test_cache_concurrency(tmpdir, RefSimulator):
+    cache_dir = str(tmpdir)
+
+    n_processes = 100
+    processes = [
+        multiprocessing.Process(
+            target=build_many_ensembles, args=(cache_dir, RefSimulator))
+        for _ in range(n_processes)]
+    for p in processes:
+        p.start()
+    for p in processes:
+        p.join(60)
+    for p in processes:
+        assert p.exitcode == 0
 
 
 def reject_outliers(data):
@@ -325,7 +353,8 @@ rc.set("decoder_cache", "enabled", "True")
 rc.set("decoder_cache", "readonly", "True")
 ''',
         'stmt': '''
-sim = nengo.Simulator(model)
+with nengo.Simulator(model):
+    pass
 '''
     }
 
@@ -336,7 +365,8 @@ rc.set("decoder_cache", "enabled", "True")
 rc.set("decoder_cache", "readonly", "False")
 ''',
         'stmt': '''
-sim = nengo.Simulator(model)
+with nengo.Simulator(model):
+    pass
 '''
     }
 
@@ -344,9 +374,13 @@ sim = nengo.Simulator(model)
         'rc': '''
 rc.set("decoder_cache", "enabled", "True")
 rc.set("decoder_cache", "readonly", "False")
-sim = nengo.Simulator(model)
+with nengo.Simulator(model):
+    pass
 ''',
-        'stmt': 'sim = nengo.Simulator(model)'
+        'stmt': '''
+with nengo.Simulator(model):
+    pass
+'''
     }
 
     labels = ["no cache", "cache miss", "cache miss ro", "cache hit"]
--- a/nengo/utils/lock.py
+++ b/nengo/utils/lock.py
@@ -1,0 +1,51 @@
+import errno
+import os
+import os.path
+import time
+
+from nengo.exceptions import TimeoutError
+
+
+class FileLock(object):
+    def __init__(self, filename, timeout=10., poll=0.1):
+        self.filename = filename
+        self.timeout = timeout
+        self.poll = poll
+        self._fd = None
+
+    def acquire(self):
+        start = time.time()
+        while True:
+            try:
+                self._fd = os.open(
+                    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)
+                return
+            except OSError as err:
+                if err.errno not in (errno.EEXIST, errno.EACCES):
+                    raise
+                elif time.time() - start >= self.timeout:
+                    raise TimeoutError(
+                        "Could not acquire lock '{filename}'.".format(
+                            filename=self.filename))
+                else:
+                    time.sleep(self.poll)
+
+    def release(self):
+        if self._fd is not None:
+            os.close(self._fd)
+            os.unlink(self.filename)
+            self._fd = None
+
+    @property
+    def acquired(self):
+        return self._fd is not None
+
+    def __enter__(self):
+        self.acquire()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.release()
+
+    def __del__(self):
+        self.release()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -214,6 +214,8 @@ Release History
   `#974 <https://github.com/nengo/nengo/pull/974>`_)
 - Fixed thread-safety of using networks and config in ``with`` statements.
   (`#989 <https://github.com/nengo/nengo/pull/989>`_)
+- The decoder cache will only be used when a seed is specified.
+  (`#946 <https://github.com/nengo/nengo/pull/946>`_)
 
 2.0.3 (December 7, 2015)
 ========================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -186,6 +186,9 @@ Release History
   can be used to disallow adding new semantic pointers. Vocabulary subsets
   are read-only by default.
   (`#699 <https://github.com/nengo/nengo/pull/699>`_)
+- Improved performance of the decoder cache by writing all decoders
+  of a network into a single file.
+  (`#946 <https://github.com/nengo/nengo/pull/946>`_)
 
 **Bug fixes**
 
--- a/nengo/builder/network.py
+++ b/nengo/builder/network.py
@@ -34,6 +34,7 @@ def build_network(model, network):
     if model.toplevel is None:
         model.toplevel = network
         model.seeds[network] = get_seed(network, np.random)
+        model.seeded[network] = getattr(network, 'seed', None) is not None
 
     # Set config
     old_config = model.config
@@ -44,6 +45,8 @@ def build_network(model, network):
     sorted_types = sorted(network.objects, key=lambda t: t.__name__)
     for obj_type in sorted_types:
         for obj in network.objects[obj_type]:
+            model.seeded[obj] = (model.seeded[network] or
+                                 getattr(obj, 'seed', None) is not None)
             model.seeds[obj] = get_seed(obj, rng)
 
     logger.debug("Network step 1: Building ensembles and nodes")
--- a/nengo/builder/probe.py
+++ b/nengo/builder/probe.py
@@ -20,6 +20,7 @@ def conn_probe(model, probe):
                       solver=probe.solver, add_to_container=False)
 
     # Set connection's seed to probe's (which isn't used elsewhere)
+    model.seeded[conn] = model.seeded[probe]
     model.seeds[conn] = model.seeds[probe]
 
     # Make a sink signal for the connection
--- a/nengo/exceptions.py
+++ b/nengo/exceptions.py
@@ -94,3 +94,7 @@ class Unconvertible(NengoException, ValueError):
 
 class CacheIOError(NengoException, IOError):
     """An IO error in reading from or writing to the decoder cache."""
+
+
+class TimeoutError(NengoException):
+    """A timeout occurred while waiting for a resource."""
--- a/nengo/utils/nco.py
+++ b/nengo/utils/nco.py
@@ -25,6 +25,12 @@ array data an alignment of 16 bytes.
 
 The Numpy NPY format is documented here:
 https://github.com/numpy/numpy/blob/master/doc/neps/npy-format.rst
+
+As of legacy version 1 of the cache, multiple NCO files will be concatenated
+into one file. The start and end of each subfile will be stored in a cache
+index, but can also be recovered from reading the headers of the NCO files in
+order as each one gives the start of the next header (corresponding to the
+end of the array data).
 """
 
 from __future__ import absolute_import
@@ -42,7 +48,7 @@ from ..exceptions import CacheIOError
 class Subfile(object):
     """A file-like object for limiting reads to a subrange of a file.
 
-    This clss only supports reading and seeking. Writing is not supported.
+    This class only supports reading and seeking. Writing is not supported.
 
     Parameters
     ----------
@@ -85,6 +91,9 @@ class Subfile(object):
         self.max_size = self.end - offset
         self.fileobj.seek(offset)
 
+    def tell(self):
+        return self.fileobj.tell() - self.start
+
 
 MAGIC_STRING = ensure_bytes('NCO')
 SUPPORTED_PROTOCOLS = [0]
@@ -105,7 +114,8 @@ def write(fileobj, metadata, array):
     array : ndarray
         Numpy array with the actual data to store.
     """
-    pickle_start = byte_align(HEADER_SIZE, ALIGNMENT)
+    start = fileobj.tell()
+    pickle_start = byte_align(start + HEADER_SIZE, ALIGNMENT)
     fileobj.seek(pickle_start)
     pickle.dump(metadata, fileobj, pickle.HIGHEST_PROTOCOL)
     pickle_end = fileobj.tell()
@@ -118,8 +128,9 @@ def write(fileobj, metadata, array):
     header = struct.pack(
         HEADER_FORMAT, MAGIC_STRING, 0, pickle_start, pickle_end,
         array_start, array_end)
-    fileobj.seek(0)
+    fileobj.seek(start)
     fileobj.write(header)
+    fileobj.seek(array_end)
 
 
 def read(fileobj):
--- a/nengo/utils/tests/test_lock.py
+++ b/nengo/utils/tests/test_lock.py
@@ -1,0 +1,51 @@
+import os.path
+
+import pytest
+
+from nengo.utils.lock import FileLock, TimeoutError
+
+
+def test_can_acquire_filelock_at_most_once(tmpdir):
+    filename = os.path.join(str(tmpdir), 'lock')
+    lock = FileLock(filename, timeout=0.01, poll=0.01)
+    lock.acquire()
+    with pytest.raises(TimeoutError):
+        lock.acquire()
+    lock2 = FileLock(filename, timeout=0.01, poll=0.01)
+    with pytest.raises(TimeoutError):
+        lock2.acquire()
+    lock.release()
+
+
+def test_released_filelock_can_be_reacquired(tmpdir):
+    filename = os.path.join(str(tmpdir), 'lock')
+    lock = FileLock(filename, timeout=0.01, poll=0.01)
+    lock.acquire()
+    lock.release()
+    lock.acquire()
+    lock.release()
+
+
+def test_can_release_filelock_multiple_times(tmpdir):
+    filename = os.path.join(str(tmpdir), 'lock')
+    lock = FileLock(filename, timeout=0.01, poll=0.01)
+    lock.release()
+    lock.acquire()
+    lock.release()
+    lock.release()
+
+
+def test_filelock_supports_with_statement(tmpdir):
+    filename = os.path.join(str(tmpdir), 'lock')
+    with FileLock(filename):
+        pass
+
+
+def test_filelock_gets_released_on_lock_deletion(tmpdir):
+    filename = os.path.join(str(tmpdir), 'lock')
+    lock = FileLock(filename, timeout=0.01, poll=0.01)
+    lock.acquire()
+    del lock
+    lock = FileLock(filename, timeout=0.01, poll=0.01)
+    lock.acquire()
+    lock.release()


jgosmann:
I don't see why these changes broke a whole bunch of tests ... :confused:

jgosmann:
I added a commit that should significantly reduce the number of files and overhead due to block alignment in the cache. That in turn should improve the performance of the `shrink()` function.

All decoders from a single build will be written to the same file with this change (instead of individual files). The implementation of doing this per build is somewhat implicit right now: It happens when shrink is called because that requires the current file to be closed and it gets called after the build process. Potentially we might want to explicitely say that a model build is complete and a new cache file should be started.

To be able to reuse decoders from previous builds, this requires an index file. For that file it is important to ensure correct handling of multiple processes accessing this file which is done with a file lock. That prevents multiple processes from reading at the same time. But this should not be too bad as for each model build that file is only accessed twice. Once before the build process and once after.

Because the `shrink()` can now only delete complete cached models, we might end up keeping some unneccessary decoders around if such a file contains other decoders that have been accessed recently. But that's probably fine too. Most likely we gain more by reducing the overhead of having a lot of small files.

It might be worth to mention, that the data format for those files is just a concatenation of multiple NCO files.

I haven't tested this on Windows yet. Potential problems might be that the file lock does not work as it should and that the implementation might open the same file twice.

Another interesting observation: Computing the decoders makes use of my CPU cores, but loading the cached decoders makes only use of one. For a test network with a lot of small ensembles this resulted in pretty much the same time for both cases, but of courses in the former case the system load is much higher.


With these changes it does not seem worth to me to not cache small ensembles. Cache hits seem to be always faster than using no cache except when you have more dimensions than neurons in the ensemble. The extra time in case of a cache miss seems to be about the same time saved with a cache hit. Thus building a model twice with the cache means you break even and the third time you have saved time overall. Furthermore, the problem of a large number of files on the filesystem should be solved with these changes.

jgosmann:
Are we still going to support Python 2.6 with Nengo 2.1? It's complaining about some syntax I used (I could fix it ...).

jgosmann:
One more comment: This implementation probably makes better use of I/O buffering as there are probably many instances where the decoders can now be read from a single file in a linear fashion.

jgosmann:
And I should mention: All my tests were SSD based so far.

Seanny123:
As of 2.1, we are no longer supporting Python 2.6, as discussed in #954.

jgosmann:
Another thing which might be good to test is concurrent access to the cache.

jgosmann:
Rebased onto `master`. Added a concurrency test. Changed tests to use `RefSimulator` instead of `Simulator`. Changed tests to use `with` style closing of simulator.

jgosmann:
Rebased onto `master`.

jgosmann:
I'm in the process of testing this with Windows in a Vagrant Virtual Box and there might be some issues. But I am not sure whether it is just that the disk image has not enought space.

jgosmann:
If anyone could run the cache tests (with the `--slow` flag) on a real Windows machine, this would be awesome.

Seanny123:
I ran the cache tests with the `--slow` flag on my Windows machine. You can find the results here.

```
============================= test session starts =============================
platform win32 -- Python 3.5.1, pytest-2.8.1, py-1.4.30, pluggy-0.3.1
rootdir: c:\Users\Sean Aubin\git\nengo, inifile: setup.cfg
collected 33 items

test_cache.py ........................Fssss

================================== FAILURES ===================================
___________________________ test_cache_concurrency ____________________________

tmpdir = local('C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0')
RefSimulator = <class 'nengo.simulator.Simulator'>

    @pytest.mark.slow
    def test_cache_concurrency(tmpdir, RefSimulator):
        cache_dir = str(tmpdir)
    
        n_processes = 100
        processes = [
            multiprocessing.Process(
                target=build_many_ensembles, args=(cache_dir, RefSimulator))
            for _ in range(n_processes)]
        for p in processes:
            p.start()
        for p in processes:
            p.join(60)
        for p in processes:
>           assert p.exitcode == 0
E           assert 1 == 0
E            +  where 1 = <Process(Process-1, stopped[1])>.exitcode

test_cache.py:316: AssertionError
---------------------------- Captured stderr call -----------------------------
Process Process-60:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 167, in __init__

    self._index = CacheIndex(os.path.join(self.cache_dir, self._INDEX))

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 79, in __init__

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000A70251DF28>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 117, in sync

    self._index.update(self._updates)

AttributeError: 'CacheIndex' object has no attribute '_updates'

Exception ignored in: <bound method DecoderCache.__del__ of <nengo.cache.DecoderCache object at 0x000000A70251DE48>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 181, in __del__

    self._close_fd()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 176, in _close_fd

    if self._fd is not None:

AttributeError: 'DecoderCache' object has no attribute '_fd'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x0000008F2160DF98>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-80:

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 333, in cached_solver

    path, start, end = self._index[key]

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 89, in __getitem__

    return self._index[key]

KeyError: '0681421c0087ff1d416c18301f1de7d8ea9fbcb2'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 110, in __init__

    self.model.build(network)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\network.py", line 70, in build_network

    model.build(conn)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 177, in build_connection

    eval_points, decoders, solver_info = build_decoders(model, conn, rng)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 83, in build_decoders

    rng=rng, E=E)

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 344, in cached_solver

    fd = self._get_fd()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 172, in _get_fd

    self._fd = open(self._key2path(str(uuid1())), 'wb')

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 390, in _key2path

    os.makedirs(directory)

  File "c:\Users\Sean Aubin\Anaconda3\lib\os.py", line 241, in makedirs

    mkdir(name, mode)

FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\59'

Process Process-66:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-98:

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 333, in cached_solver

    path, start, end = self._index[key]

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 89, in __getitem__

    return self._index[key]

KeyError: '28f00ad597b4084c14f110ce5c437ddb223ac846'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 110, in __init__

    self.model.build(network)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\network.py", line 70, in build_network

    model.build(conn)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 177, in build_connection

    eval_points, decoders, solver_info = build_decoders(model, conn, rng)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 83, in build_decoders

    rng=rng, E=E)

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 344, in cached_solver

    fd = self._get_fd()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 172, in _get_fd

    self._fd = open(self._key2path(str(uuid1())), 'wb')

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 390, in _key2path

    os.makedirs(directory)

  File "c:\Users\Sean Aubin\Anaconda3\lib\os.py", line 241, in makedirs

    mkdir(name, mode)

FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\59'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000BFFDA3CFD0>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-39:

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 333, in cached_solver

    path, start, end = self._index[key]

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 89, in __getitem__

    return self._index[key]

KeyError: '415f719c25a2748936bac9e962fc36202ca33f9c'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 110, in __init__

    self.model.build(network)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\network.py", line 70, in build_network

    model.build(conn)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 177, in build_connection

    eval_points, decoders, solver_info = build_decoders(model, conn, rng)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 83, in build_decoders

    rng=rng, E=E)

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 344, in cached_solver

    fd = self._get_fd()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 172, in _get_fd

    self._fd = open(self._key2path(str(uuid1())), 'wb')

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 390, in _key2path

    os.makedirs(directory)

  File "c:\Users\Sean Aubin\Anaconda3\lib\os.py", line 241, in makedirs

    mkdir(name, mode)

FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\59'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000133074DF28>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-1:

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 333, in cached_solver

    path, start, end = self._index[key]

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 89, in __getitem__

    return self._index[key]

KeyError: '555535a4e3ca400399e1f7159bd4bd21cd77301a'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 110, in __init__

    self.model.build(network)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\network.py", line 70, in build_network

    model.build(conn)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 177, in build_connection

    eval_points, decoders, solver_info = build_decoders(model, conn, rng)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 83, in build_decoders

    rng=rng, E=E)

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 344, in cached_solver

    fd = self._get_fd()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 172, in _get_fd

    self._fd = open(self._key2path(str(uuid1())), 'wb')

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 390, in _key2path

    os.makedirs(directory)

  File "c:\Users\Sean Aubin\Anaconda3\lib\os.py", line 241, in makedirs

    mkdir(name, mode)

FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\59'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x0000001F8828DF98>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-78:

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 333, in cached_solver

    path, start, end = self._index[key]

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 89, in __getitem__

    return self._index[key]

KeyError: 'f555f8ecdae655697ddf3681172e9f5b4d503271'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 110, in __init__

    self.model.build(network)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\network.py", line 70, in build_network

    model.build(conn)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 38, in build

    return Builder.build(self, obj, *args, **kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\builder.py", line 83, in build

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 177, in build_connection

    eval_points, decoders, solver_info = build_decoders(model, conn, rng)

  File "c:\Users\Sean Aubin\git\nengo\nengo\builder\connection.py", line 83, in build_decoders

    rng=rng, E=E)

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 344, in cached_solver

    fd = self._get_fd()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 172, in _get_fd

    self._fd = open(self._key2path(str(uuid1())), 'wb')

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 390, in _key2path

    os.makedirs(directory)

  File "c:\Users\Sean Aubin\Anaconda3\lib\os.py", line 241, in makedirs

    mkdir(name, mode)

FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\59'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x0000004858F1CFD0>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-25:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-94:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-5:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 167, in __init__

    self._index = CacheIndex(os.path.join(self.cache_dir, self._INDEX))

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 79, in __init__

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000F2114FDF98>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 117, in sync

    self._index.update(self._updates)

AttributeError: 'CacheIndex' object has no attribute '_updates'

Exception ignored in: <bound method DecoderCache.__del__ of <nengo.cache.DecoderCache object at 0x000000F2114FDEB8>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 181, in __del__

    self._close_fd()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 176, in _close_fd

    if self._fd is not None:

AttributeError: 'DecoderCache' object has no attribute '_fd'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000248248DF98>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

PermissionError: [Errno 13] Permission denied: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x00000056B16ADF60>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

PermissionError: [Errno 13] Permission denied: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'

Process Process-84:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-32:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 167, in __init__

    self._index = CacheIndex(os.path.join(self.cache_dir, self._INDEX))

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 79, in __init__

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x0000005C0091CFD0>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 117, in sync

    self._index.update(self._updates)

AttributeError: 'CacheIndex' object has no attribute '_updates'

Exception ignored in: <bound method DecoderCache.__del__ of <nengo.cache.DecoderCache object at 0x0000005C0091CEF0>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 181, in __del__

    self._close_fd()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 176, in _close_fd

    if self._fd is not None:

AttributeError: 'DecoderCache' object has no attribute '_fd'

Process Process-13:

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

PermissionError: [Errno 13] Permission denied: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'

Process Process-91:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x0000005ADB8BDF98>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

PermissionError: [Errno 13] Permission denied: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000E71773DF28>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

PermissionError: [Errno 13] Permission denied: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000B9AC74DF60>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

PermissionError: [Errno 13] Permission denied: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'

Process Process-52:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-6:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000FEF104DF60>>

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 101, in __del__

    self.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

PermissionError: [Errno 13] Permission denied: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'

Process Process-99:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-16:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

Process Process-83:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 23, in acquire

    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)

FileExistsError: [Errno 17] File exists: 'C:\\Users\\SEANAU~1\\AppData\\Local\\Temp\\pytest-of-Sean Aubin\\pytest-1\\test_cache_concurrency0\\index.lock'



During handling of the above exception, another exception occurred:



  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 254, in _bootstrap

    self.run()

  File "c:\Users\Sean Aubin\Anaconda3\lib\multiprocessing\process.py", line 93, in run

    self._target(*self._args, **self._kwargs)

  File "c:\Users\Sean Aubin\git\nengo\nengo\tests\test_cache.py", line 298, in build_many_ensembles

    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):

  File "c:\Users\Sean Aubin\git\nengo\nengo\simulator.py", line 112, in __init__

    self.model.decoder_cache.shrink()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 252, in shrink

    self._index.sync()

  File "c:\Users\Sean Aubin\git\nengo\nengo\cache.py", line 115, in sync

    with self._lock:

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 44, in __enter__

    self.acquire()

  File "c:\Users\Sean Aubin\git\nengo\nengo\utils\lock.py", line 31, in acquire

    filename=self.filename))

nengo.utils.lock.TimeoutError: Could not acquire lock 'C:\Users\SEANAU~1\AppData\Local\Temp\pytest-of-Sean Aubin\pytest-1\test_cache_concurrency0\index.lock'.

============== 1 failed, 24 passed, 4 skipped in 103.42 seconds ===============
```

jgosmann:
Seems like there is a problem with the `FileLock` implementation :weary:

jgosmann:
I found a fixed bug which is probably unrelated, but can you try again?

Seanny123:
It works now!

jgosmann:
Awesome. :)
Not su

jgosmann:
Still failing in my vagrant bow, though ...

Seanny123:
My laptop is Windows 8, what's the OS of your Vagrant Box? Have you tried the office laptop? I think it has Windows 10.

jgosmann:
I believe it's Windows 7. I'm not trying to find a Windows version it works on, but I wan't to solve the problems on any reasonably recent Windows version and setup.

jgosmann:
It seems Windows is scheduling the processes in the worst possible order and the first started process is never able to acquire the file lock. I'm experimenting now with random sleep intervals.

jgosmann:
I see two possibilities fixing this problem:

1. **Don't use a timeout on the file lock.** The reason for the timeout is that I am afraid that in some circumstances the file lock might not be released correctly and then cannot be acquired again without manual intervention. Without the timeout this would result in an infinite loop which can be hard to detect because it is hard to tell whether the build process just takes so long, the process just has to wait that long to acquire the lock, or an actual deadlock occured.
2. **Continue without caching if the file lock cannot be acquired within the time limit.** In this case we might waste time on writing cache files that cannot be used afterwards (due to some implementation details). But if it takes a long time to get that lock it might be better to just disable the cache and continue the build/not make the decoders accessible in the cache. This should probably produce a warning.

I might have a slight preference towards 2, but I am not completely decided. Any comments on the issue?

jgosmann:
(btw: With option 1 the tests succeed in my Vagrant box.)

@Seanny123 What Python version did you use? I was using Python 3.4 so far, because there seem to be some other issues with 2.7 and multiprocessing on Windows. Not sure if that is fixable or whether I have to mark the test as failing for that environment.

Seanny123:
I was using Python 2.7.

jgosmann:
You didn't get any
```
  File "<string>", line 1, in <module>
  File "C:\Python27\lib\multiprocessing\forking.py", line 380, in main
    prepare(preparation_data)
  File "C:\Python27\lib\multiprocessing\forking.py", line 488, in prepare
    assert main_name not in sys.modules, main_name
AssertionError: __main__
```
errors?

Seanny123:
Nope.

jgosmann:
What's your exact Python version?

jgosmann:
Further todos:
* [ ] `_check_legacy_file` does not handle concurrency safely
* [ ] Trying to acquire the lock can lead to a permission denied error for unknown reasons

Complete test output related to these points:
```
============================= test session starts =============================
platform win32 -- Python 2.7.11, pytest-2.9.0, py-1.4.31, pluggy-0.3.1
rootdir: C:\cygwin64\vagrant, inifile: setup.cfg
collected 34 items

nengo\tests\test_cache.py .........................Fssss

================================== FAILURES ===================================
___________________________ test_cache_concurrency ____________________________

tmpdir = local('c:\\cygwin64\\tmp\\pytest-of-vagrant\\pytest-0\\test_cache_concurrency0')
RefSimulator = <class 'nengo.simulator.Simulator'>

    @pytest.mark.slow
    def test_cache_concurrency(tmpdir, RefSimulator):
        cache_dir = str(tmpdir)
    
        n_processes = 100
        processes = [
            multiprocessing.Process(
                target=build_many_ensembles, args=(cache_dir, RefSimulator))
            for _ in range(n_processes)]
        for p in processes:
            p.start()
        for p in processes:
            p.join(60)
        for p in processes:
>           assert p.exitcode == 0
E           assert 1 == 0
E            +  where 1 = <Process(Process-11, stopped[1])>.exitcode

nengo\tests\test_cache.py:316: AssertionError
---------------------------- Captured stderr call -----------------------------
Process Process-11:
  File "c:\python27\lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "c:\python27\lib\multiprocessing\process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "C:\cygwin64\vagrant\nengo\tests\test_cache.py", line 298, in build_many_ensembles
    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):
  File "C:\cygwin64\vagrant\nengo\cache.py", line 166, in __init__
    self._remove_legacy_files()
  File "C:\cygwin64\vagrant\nengo\cache.py", line 283, in _remove_legacy_files
    if self._check_legacy_file():
  File "C:\cygwin64\vagrant\nengo\cache.py", line 265, in _check_legacy_file
    version = int(lf.read().strip())
ValueError: invalid literal for int() with base 10: ''
Exception AttributeError: "'DecoderCache' object has no attribute '_fd'" in <bound method DecoderCache.__del__ of <nengo.cache.DecoderCache object at 0x0000000002325DA0>> ignored
Process Process-67:
  File "c:\python27\lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "c:\python27\lib\multiprocessing\process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "C:\cygwin64\vagrant\nengo\tests\test_cache.py", line 298, in build_many_ensembles
    dt=0.001, decoder_cache=DecoderCache(cache_dir=cache_dir))):
  File "C:\cygwin64\vagrant\nengo\cache.py", line 167, in __init__
    self._index = CacheIndex(os.path.join(self.cache_dir, self._INDEX))
  File "C:\cygwin64\vagrant\nengo\cache.py", line 79, in __init__
    with self._lock:
  File "C:\cygwin64\vagrant\nengo\utils\lock.py", line 43, in __enter__
    self.acquire()
  File "C:\cygwin64\vagrant\nengo\utils\lock.py", line 23, in acquire
    self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)
OSError: [Errno 13] Permission denied: 'c:\\cygwin64\\tmp\\pytest-of-vagrant\\pytest-0\\test_cache_concurrency0\\index.lock'
Exception AttributeError: AttributeError("'CacheIndex' object has no attribute '_updates'",) in <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x000000000292E278>> ignored
Exception AttributeError: "'DecoderCache' object has no attribute '_fd'" in <bound method DecoderCache.__del__ of <nengo.cache.DecoderCache object at 0x000000000230BB70>> ignored
============== 1 failed, 25 passed, 4 skipped in 783.67 seconds ===============
```

jgosmann:
I think I might have fixed all the problems. But I'm still in the process of running unit tests in different environments. I decided to leave a timeout on the lock. If it cannot be acquired it will disable the cache (or won't sync the index if happens after the build; this is the worst case because we write all the data to the disk, but can't include it in the cache) and generate a warning.

To do the testing I updated the Windows Vagrant Box. It has now 2.7.11 (which fixes a problem with mulitprocessing), 3.4 and 3.5. There is no easy way to install Numpy/Scipy on Windows for 3.3 (at least not anymore). So I decided that it is not worth the effort. (Maybe we could drop 3.3 support in favor of 3.5 in Nengo 2.1 anyways?)

Seanny123:
+1 for supporting 3.5 instead of 3.3

tbekolay:
Hmm, it's mostly an oversight that we're not testing on Python 3.3, but if it's hard to install Numpy/Scipy on it then I'm cool dropping 3.3 support!

jgosmann:
> but if it's hard to install Numpy/Scipy on it then I'm cool dropping 3.3 support!

At least [this (unofficial) site](http://www.lfd.uci.edu/~gohlke/pythonlibs/) doesn't provide any 3.3 packages anymore. Not sure how it is with Anaconda etc, but for the Vagrant Box I didn't want to use it because it might do things I am not aware off (making it more difficult to debug the Python installations and their potential interaction).

jgosmann:
It's astonishing into what kind of obscure bugs you can run: pytest-dev/pytest#1453
Anyways, I think (or hope) that the cache problems on Windows are fixed. The remaining issues are with the test scaffolding itself and not related to the cache. Unfortunately, I don't see an easy workaround to get tox working for us on Windows. :/ Suggestions welcome.

nengo/tests/test_cache.py
```diff
@@ -425,3 +426,58 @@ def test_compare_cache_benchmark(self, varying_param, analytics_data, plt):
...
+    @pytest.mark.noassertions
+    def test_cache_shrink_benchmark(self, analytics, logger):
+        times = timeit.repeat(
+            stmt=self.stmt, setup=self.setup, number=1, repeat=self.n_trials)
+        logger.info("Shrink took a minimum of %f seconds.", np.min(times))
+        logger.info("Shrink took a %f seconds on average.", np.mean(times))
+        logger.info(
+            "Shrink took a %f seconds on average with outliers rejected.",
+            np.mean(reject_outliers(times)))
+        analytics.add_data('times', times)
```
tbekolay: Should call `rc.set("decoder_cache", "size", default")` here to undo the effects of `rc.set` in the setup.

nengo/utils/lock.py
```diff
@@ -0,0 +1,51 @@
+import errno
+import os
+import os.path
+import time
+
+
+class TimeoutError(Exception):
```
tbekolay: Note to whoever rebases this to master: this exception should be moved to `nengo/exceptions.py`.

nengo/tests/test_cache.py
jgosmann: Should not be required [because of `pytest_configure`](https://github.com/nengo/nengo/blob/master/nengo/conftest.py#L21). But I wonder whether the decoder cache needs to be enabled first ...?

jgosmann: Actually, `pytest_configure` gets only called once. So yes the config should be reset ... but I would like to do that after automatically after every test (then it is less likely to be forgotten).

Another potential problem: This is using default cache paths, but it should use the temporary pytest directory.

tbekolay: I do this in [`test_examples`](https://github.com/nengo/nengo/blob/master/nengo/tests/test_examples.py#L58); I'm not sure if we can have that teardown affect all test or not? If not, we could put it instead in `pytest_runtest_setup` in `conftest.py`.

nengo/utils/lock.py
```diff
@@ -23,7 +23,7 @@ def acquire(self):
                     self.filename, os.O_CREAT | os.O_RDWR | os.O_EXCL)
                 return
             except OSError as err:
-                if err.errno != errno.EEXIST:
+                if err.errno not in [errno.EEXIST, errno.EACCES]:
```
tbekolay: Very esoteric nitpick, but I usually use a tuple in these situations.

nengo/cache.py
```diff
@@ -45,6 +45,14 @@ def safe_remove(path):
         logger.warning("OSError during safe_remove: %s", err)
 
 
+def safe_makedirs(path):
+    if not os.path.exists(path):
+        try:
+            os.makedirs(path)
+        except OSError as err:
```
tbekolay: Should we make sure that the error is that the file exists here? `if not 'File exists' in str(err): raise` or something like that?

jgosmann: I was thinking about that. I did not do it to keep it equivalent to the other `safe_<foo>` functions. If we do that it probably would be better to check the error code instead of the string, but Windows might give a weird error number that is not in `errno` in that case I believe.

nengo/tests/test_cache.py
jgosmann: Created PR #983.

nengo/utils/lock.py
jgosmann: Should this be a `NengoException`? At least in the cache this kind of exception should always be caught and handled.

jgosmann: Though, maybe it is useful to backends that need to acquire some hardware resource.

tbekolay: For simplicity I would think all exceptions in Nengo should be a subclass of `NengoException`, rather than having to figure out when it should be and when it shouldn't

tbekolay:
I have a Windows virtualbox, so I'll test this out there. But just looking through the code, aside from the nitpicky inline comments I made, I had two thoughts:

1. If I'm reading this correctly, the cache index pickles a dictionary that contains paths, and start and end indices for those paths (for where the decoders are in that file). I'm wondering if there's any real benefit to using pickled dictionaries here instead of something human-readable? Could be some simple and custom like each line is an entry, or it could be JSON or CSV. Not sure how often it would be read, but it at least would make sense if people are curious and try to open up the file. It also would be more robust to pickle issues (e.g., upgrading Python versions with different pickle priorities).
2. Again if I'm reading correctly, we use the default timeout for getting the cache index, which is 10 seconds. This should only really matter if we are building two models at the same time (right?) but I wonder in how many cases it would take more than 10 seconds to build the model. Maybe we should shorten the timeout?

jgosmann:
Rebased to `test-config`, moved `TimeoutError`.

jgosmann:
> I'm wondering if there's any real benefit to using pickled dictionaries here instead of something human-readable?

* Binary format and potentially faster to read and write?
* Easier implementation

> if people are curious and try to open up the file

Loading a pickle in an interactive Python session isn't too hard either I would say. And the information isn't that useful either for a human because you don't know what networks the hashes correspond to.

> It also would be more robust to pickle issues (e.g., upgrading Python versions with different pickle priorities).

I don't think pickle issues are a big issue here. If a pickle cannot be read it results in a cache miss and newly calculated decoders will be written in a pickle format that is readable. The cache is also not intended to be shared across machines. Newer versions of the pickle module are usually also able to read older versions of the pickle protocol.

> I wonder in how many cases it would take more than 10 seconds to build the model.

That's not really relevant. The relevant time is how long it takes to read and write the index. The index will be read at the beginning of the build and for that duration the lock will be acquired. During the build the lock is released and at the end of the build it is acquired again to write the updates to the index file.

tbekolay:
> Loading a pickle in an interactive Python session isn't too hard either I would say

True, but people have to know it's pickled and not some other binary format. In any case, the easier implementation is reason enough I think ;) and you're right that with the hashes it's not too helpful regardless.

> The index will be read at the beginning of the build and for that duration the lock will be acquired.

If another builder starts at the same time, then does it wait for 10 seconds trying to acquire the lock? I'm thinking that if the second builder just didn't wait and proceeded with the build without the cache, it might finish faster than if it waits to acquire the lock, then proceeds with the build with the cache.

jgosmann:
> If another builder starts at the same time, then does it wait for 10 seconds trying to acquire the lock?

Only if the first builder needs more than 10 seconds to read the index file.

> I'm thinking that if the second builder just didn't wait and proceeded with the build without the cache, it might finish faster than if it waits to acquire the lock, then proceeds with the build with the cache.

Possible, but there is no way to know. Also, if we don't wait at all and lot of builds are started at the same time, none (except) one will use the cache.

tbekolay:
> none (except) one will use the cache

Right, so I'm wondering whether this case is worse than waiting for 10 s. It really depends on your application of course, so I'm not sure there's a good heuristic to determine the right timeout... in conclusion ¯\_(ツ)_/¯

jgosmann:
Does this need a changelog entry?

hunse:
Yeah, I think it should have one, either under "bug fixes" or "improvements".

tbekolay:
I rebased this and tested it (`py.test nengo\tests\test_cache.py --slow --plots`) on both Python 2 and Python 3, and they worked for me! Any lingering issues with this branch or is it good to go?

jgosmann:
Good to go! :cake:

nengo/builder/connection.py
```diff
@@ -80,7 +80,11 @@ def build_decoders(model, conn, rng, transform):
         targets = multiply(targets, transform.T)
 
     try:
-        wrapped_solver = model.decoder_cache.wrap_solver(solve_for_decoders)
+        if model.seeded[conn]:
```
hunse: I'm trying to figure out how `model.seeded` works. Doesn't it mean that if only the top level network is seeded, connections in sub-networks won't be cached?

jgosmann: The relevant code is in `builder/network.py`. A network sets `seeded` for all it's sub-objects (including networks). Thus, a sub-network of a seeded network will also be seeded. It's analogous to how we set the network seeds.

hunse: Oh yeah, I see now. I thought the one line in `builder/network.py` was for all networks, but it's just for the top-level. Then all networks act like objects. Okay, this makes sense.

nengo/builder/builder.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, dt=0.001, label=None, decoder_cache=NoDecoderCache()):
         self.operators = []
         self.params = {}
         self.seeds = {}
+        self.seeded = {}
```
hunse: Should we have a brief description here? It took me a bit to figure out what this is actually doing. A few of the others could probably use descriptions, too (of course they weren't part of this PR, so they're not necessarily your job).

jgosmann: Sure, but I'm busy with other stuff right now.

tbekolay: No worries, I'll do it :)

tbekolay:
@hunse added docs to the `Model` docstring in 98b0af549fe5ca3367f85.

tbekolay:
More `seeded` deets in 240d70c141808d. Look OK @hunse?

hunse:
Yeah, I think that's nice and clear.

PR #952, opened 2016-02-17
Removed extra synapse in Product network

tcstewar:
The reorganized Product network was introducing an extra 0.005s synapse on its inputs!  This PR removes this synapse.


Seanny123:
Good catch! LGTM.

PR #953, opened 2016-02-17
Modifying semantic pointers in a vocabulary.

pblouw:
Currently, vocabularies store info about SPs in two separate places: the `.pointers` dictionary that maps vocabulary keys to SemanticPointer objects, and the `.vectors` array whose rows correspond to the vector attributes of each SP in the vocabulary.  It is possible to modify an SP in one of these places and not the other, which can lead to confusing errors of the sort reported in nengo/nengo_gui#438. (e.g. one might want to change the SP associated with a particular key to be structured in some way, randomly perturbed, or whatever). So, this PR introduces a `set_value` method that takes a key and vector as input and updates the vocabulary such that `.pointers` and `.vectors` are consistent with one another. 

I've flagged this pull request as a discussion, since it is not obvious to me that this is the best solution to this problem. Another option would be to get rid of the pointers dictionary altogether and create new SemanticPointer objects each time `__getitem__` is called by indexing into the `.vectors` array.

### DIFF
--- a/nengo/networks/assoc_mem.py
+++ b/nengo/networks/assoc_mem.py
@@ -60,6 +60,19 @@ class AssociativeMemory(nengo.Network):
             input_vectors = np.array(input_vectors, ndmin=2)
         if is_iterable(output_vectors):
             output_vectors = np.array(output_vectors, ndmin=2)
+
+        if input_vectors.shape[0] == 0:
+            raise ValueError('Number of input vectors cannot be 0.')
+        elif input_vectors.shape[0] != output_vectors.shape[0]:
+            # Fail if number of input items and number of output items don't
+            # match
+            raise ValueError(
+                'Number of input vectors does not match number of output '
+                'vectors. %d != %d'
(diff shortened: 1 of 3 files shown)


Seanny123:
SPA is in a bit of a state of tumultuous change, so it is my opinion that this should be revisited after #1014 is merged. Actually, is it possible that the functionality of the new `VocabularySet` mapping satisfies your requirements?

pblouw:
Sure, I'm fine with revisiting this later. Though I'm not sure that the new `VocabularyMap` will solve the problem, since as far I can tell vocab objects will still retain info about each SP in both `.vectors` and `.pointers`, and it's still easy to modify one without modifying the other.

jgosmann:
Yes, this issue won't be solved by #1014 and related PRs.

jgosmann:
Semantic pointers and the `Vocabulary.vectors` matrix will be readonly in new-spa to prevent accidental modification. If there is a valid use case, we could add a method to `Vocabulary` to change an item in the vocabulary, but I would prefer not to.

PR #954, opened 2016-02-19
Stop supporting Python 2.6

tbekolay:
This is still a work in progress, but making the PR now in advance of the feature freeze on Feb. 23.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,6 @@ env:
 matrix:
   include:
     - env: >
-        PYTHON="2.6"
-        CONDA_DEPS="matplotlib ipython-notebook ordereddict pygments"
-        PIP_DEPS="$PIP_DEPS counter importlib IPython==1.2.1"
-    - env: >
         PYTHON="2.7"
         SETUP_CMD="test -a '--cov-config .coveragerc --cov nengo nengo -n 2 -v --durations 20'"
         EXTRA_CMD="coveralls"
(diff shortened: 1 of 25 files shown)


jgosmann:
I added a commit to disable testing with Python 2.6.

Do you intend to remove all Python 2.6 specific code in this PR? Or are we just leaving it as it is and fix it in case we touch that code in the future?

tbekolay:
> Do you intend to remove all Python 2.6 specific code in this PR?

I'm hoping to, but I'll inevitably miss some. But yeah, this is still a WIP in my mind.

tbekolay:
I think I got all of the 2.6 -> 2.7 changes now, so this is no longer a work in progress. Note that it is based on #972 as both of these touch `.travis.yml`.

hunse:
I have no idea if you got everything, but what's there LGTM.

jgosmann:
> I have no idea if you got everything, but what's there LGTM.

Same here.

I would suggest swapping the commits 2f957a1 and 819652e so that none of the commits in the history has failing tests. (Or squashing all of it together, but to me disabling Py 2.6 testing and doing the 2.6 -> 2.7 changes seem separate enough to keep it as two commits.)

hunse:
I was thinking of squashing it all. Any preference @tbekolay?

tbekolay:
Hmm squashing would be my preference, but I'm also ok with 819652e048d67fcd235c040426663ef2a234160a going first!

tbekolay:
Also, in squashing, keeping the squashed commit messages in the body of the main commit would be good too.

PR #955, opened 2016-02-19
Promote processes as a core part of Nengo

tbekolay:
This is another WIP PR to register that these things should be part of 2.1

Still to do in this PR:

- [x] Make an example notebook using processes
- [x] Make an example notebook making a new custom process
- [x] Add documentation for processes
  - [x] Make sure `SimProcess` is documented / added to API docs
- [x] Update the changelog to introduce processes and point to docs
- [x] Ensure everything from #622 and #723 is done

### DIFF
--- a/examples/advanced/processes.ipynb
+++ b/examples/advanced/processes.ipynb
@@ -1,0 +1,571 @@
+{
+ "metadata": {
+  "name": ""
+ },
+ "nbformat": 3,
+ "nbformat_minor": 0,
+ "worksheets": [
+  {
+   "cells": [
+    {
+     "cell_type": "heading",
+     "level": 1,
+     "metadata": {},
+     "source": [
+      "Processes and how to use them"
+     ]
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "Processes in Nengo can be used to describe general functions or dynamical systems,\n",
+      "including those with randomness.\n",
+      "They can be useful if you want a `Node` output\n",
+      "that has a state (like a dynamical system),\n",
+      "and they're also used for things like \n",
+      "injecting noise into Ensembles\n",
+      "so that you can not only have \"white\" noise that samples from a distribution,\n",
+      "but can also have \"colored\" noise\n",
+      "where subsequent samples are correlated with past samples.\n",
+      "\n",
+      "This notebook will first present the basic process interface,\n",
+      "then demonstrate some of the built-in Nengo processes\n",
+      "and how they can be used in your code.\n",
+      "It will also describe how to create your own custom process."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "import numpy as np\n",
+      "\n",
+      "import matplotlib.pyplot as plt\n",
+      "%matplotlib inline\n",
+      "\n",
+      "import nengo\n",
+      "%load_ext nengo.ipynb"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "heading",
+     "level": 2,
+     "metadata": {},
+     "source": [
+      "Interface"
+     ]
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "We will begin by looking at how to run an existing process instance.\n",
+      "\n",
+      "The key functions for running processes are `run`, `run_steps`, and ``apply``.\n",
+      "The first two are for running without an input,\n",
+      "and the third is for applying the process to an input.\n",
+      "\n",
+      "There are also two helper functions,\n",
+      "`trange` and `ntrange`,\n",
+      "which return the time points corresponding to a process output,\n",
+      "given either a length of time or a number of steps, respectively."
+     ]
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "### `run`: running a process for a length of time\n",
+      "\n",
+      "The `run` function runs a process for a given length of time, without any input.\n",
+      "Many of the random processes in `nengo.processes` will be run this way,\n",
+      "since they do not require an input signal."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "# Create a process (details on the FilteredNoise process below)\n",
+      "process = nengo.processes.FilteredNoise(\n",
+      "    synapse=nengo.synapses.Alpha(0.1), seed=0)\n",
+      "\n",
+      "# run the process for two seconds\n",
+      "y = process.run(2.0)\n",
+      "\n",
+      "# get a corresponding two-second time range\n",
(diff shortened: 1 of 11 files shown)


hunse:
We might want to do this after #945, since that moves this stuff around a bit.

nengo/__init__.py
```diff
@@ -26,11 +26,13 @@
 from .learning_rules import PES, BCM, Oja, Voja
 from .params import Default
 from .probe import Probe
+from .processes import (BrownNoise, FilteredNoise, PresentInput, Process,
+                        WhiteNoise, WhiteSignal)
```
hunse: This is a general question: when do we want things to become a "core" part of nengo, by which I mean be accessible simply with `nengo.<class>` rather than by `nengo.<module>.<class>`? Currently we don't have any of the distributions as core things; I think Gaussian and Uniform are so common that they could be core. With respect to processes, I would have `Process` and `WhiteSignal` for sure, `WhiteNoise` and `FilteredNoise` very probably, `PresentInput` probably, and `BrownNoise` maybe.

One thing to remember is once we promote something, it's hard to demote it without breaking code, whereas promoting things later is always fine because it will never break anything.

jgosmann: Not sure how much sense it makes to have only parts of a module as core part. If I have to think about whether I can use a specific class with `nengo.Foo` or it has to be `nengo.dists.Foo`/`nengo.processes.Foo`, I'd rather always type `nengo.dists.Foo`/`nengo.processes.Foo`.

hunse: > Not sure how much sense it makes to have only parts of a module as core part

We do currently have only parts in the core, either intentionally or unintentionally. For example `nengo.synapses.Triangle`. I guess since we're only going to be having a limited set of things in this repository anyway (e.g. new neuron types will mostly go in user repositories or some "extras" repo), that's another reason to import everything in a module. I'm thinking particularly with `dists` though, that there are some I'd like to be easily accessible, but I'm worried adding them all will clutter up the core namespace. It is a good point, though, that users don't want to worry about what may be imported and what's not. Furthermore, if something isn't imported users might not realize it exists if they don't know e.g. only some synapses are imported.

One argument for importing less is that it might get confusing what is the purpose of some of the base objects. For example, is `nengo.Gaussian` a distribution, process, or neuron type?

tbekolay: Hmm... yeah, looking back at this, I'm leaning towards just keeping the `processes` module import and having people use the full name. The `dists` name is intentionally short so that it isn't too bad to do `nengo.dists.whatever`. But yeah, we should maybe be more consistent and import either none or all of a particular module, so I'll make sure we're doing that.

tbekolay:
I've finished this PR now, so it's ready for review! With #959 merged, this is the last PR in the 2.1.0 release! Where has the time gone, am I right all? aaaaaanyway this doesn't change much code, mostly documentation stuff, but fortunately far less than #959 so it shouldn't be too bad to review. Tagging @drasmuss and @hunse for review, or anyone else that has time!

Also @hunse I kept 7f313787f29ec06ed6b10862 as a separate commit so you can see the changes I'm proposing to the process notebook.

nengo/processes.py
```diff
@@ -66,17 +65,16 @@ class FilteredNoise(Process):
...
-        Default: Gaussian(mean=0, std=1)
-    scale : bool, optional
+    scale : bool, optional (Default: True)
         Whether to scale the white noise for integration, making the output
-        signal invariant to `dt`. Defaults to True.
-    synapse_kwargs : dict, optional
-        Arguments to pass to `synapse.make_step`.
-    seed : int, optional
+        signal invariant to ``dt``.
+    synapse_kwargs : dict, optional (Default: None)
```
hunse: Maybe say the default is `{}`? Or do you think this is clearer?

examples/advanced/processes.ipynb
```diff
@@ -0,0 +1,571 @@
...
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "process = nengo.processes.FilteredNoise(\n",
+      "    synapse=nengo.synapses.Alpha(0.1), seed=0)\n",
+      "\n",
+      "# run the process for 1000 steps\n",
+      "y = process.run_steps(1000)\n",
+      "\n",
+      "# get a corresponding 500-step time range\n",
```
drasmuss: 1000 step

hunse:
LGTM. Just that one minor comment.

nengo/processes.py
tbekolay: Not sure, this is what I've used more often in other docstrings... usually I say something like `If None, ...` but in this case it's pretty obvious that no arguments will be passed if you pass `None`, I think.

examples/advanced/processes.ipynb
```diff
@@ -0,0 +1,571 @@
...
+      "The `FilteredNoise` process with an `Alpha` synapse (blue)\n",
+      "has significantly lower high-frequency components\n",
+      "than a similar process with a `Lowpass` synapse (green)."
+     ]
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "## `PresentInput`\n",
```
drasmuss: Unrelated, but this example just made me think that `nengo.utils.functions.piecewise` and `PresentInput` are doing very similar things.  If we made `PresentInput` a little bit more general so that it could have different presentation times, we could merge them into one.

examples/advanced/processes.ipynb
```diff
@@ -0,0 +1,571 @@
...
+      "that can implement arbitrary linear dynamical systems,\n",
+      "given `A` and `B` matrices.\n",
+      "We will overload the `__init__` method\n",
+      "to take and store these matrices,\n",
+      "as well as check the matrix shapes\n",
+      "and set the default size in and out.\n",
+      "The advantage of using the default sizes\n",
+      "is that when we then create a node using the process,\n",
+      "or run the process using `apply`,\n",
+      "we do not need to specify the sizes."
```
drasmuss: Might be nice to just throw in a callout here saying "or we could make arbitrary linear dynamical systems in neurons! see notebook x"

examples/advanced/processes.ipynb
hunse: It would be great to have a process for `utils.functions.piecewise`. I'd keep it separate from `PresentInput`, though, or at least have some version of it with fixed times (maybe if you pass a constant it could have fixed times, or a list of times it could have varying?). It's just easier to do in OCL if the times are fixed.

Also with piecewise it would be great to have linear interpolation, too.

examples/advanced/processes.ipynb
hunse: This is an "advanced" notebook, so hopefully users already know that :) I guess a callout wouldn't hurt, but I don't want to distract from what's going on here, so I'd put it at the end of this example if at all. I'd probably prefer to keep it out for now, but it's not a big deal to me.

tbekolay:
Made issue #1036 for the piecewise process and fixed the 1000-step typo, but left the arbitrary LTI thing as is; can always modify it later when we see how people are using the docs.

Issue #837, opened 2015-09-09
`test_cache.test_cache_works` should use `RefSimulator` instead of `Simulator`


tbekolay:
Only backends using our builder (which is what `RefSimulator` gives you) should have the cache tests done on them.

hunse:
Fixed by #956.

PR #956, opened 2016-02-20
Use RefSimulator for tests using 'step'

### LINKED ISSUE #836: `sim.step` in tests
In [test_unconnected_node, L220 and L222](https://github.com/nengo/nengo/blob/master/nengo/tests/test_node.py#L220) we use `sim.step` to ensure that we're only doing one simulator timestep rather than multiple. That's fine for testing Nengo, but since we use the `Simulator` we're saying that this should be a backend-agnostic tests. Some backends, like SpiNNaker, don't implement `Simulator.step`, nor should they have to.

We should either 1) replace `sim.step` with `sim.run(dt)` or 2) use `RefSimulator` here instead of `Simulator` to denote that not all backends have to pass this test.

tbekolay:
Simulator.step is used to run the reference backend, but we don't require a step method for all backends. Tests that use step, therefore, should use the RefSimulator fixture, or run for `dt` time.

Fixes #836 and #837.

### DIFF
--- a/nengo/tests/test_simulator.py
+++ b/nengo/tests/test_simulator.py
@@ -55,7 +55,7 @@ def test_probedict():
     assert np.all(probedict.get("list") == np.asarray(raw.get("list")))
 
 
-def test_close(Simulator):
+def test_close_function(Simulator):
     m = nengo.Network()
     with m:
         nengo.Ensemble(10, 1)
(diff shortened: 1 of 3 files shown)


nengo/tests/test_simulator.py
```diff
@@ -83,8 +79,25 @@ def test_usage_in_with_statement(Simulator):
     with pytest.raises(ValueError):
         sim.run(1.)
     with pytest.raises(ValueError):
+        sim.reset()
+
+
+def test_close_steps(RefSimulator):
+    m = nengo.Network()
```
Seanny123: Maybe add a docstring explaining that this test is checking the two ways to close a simulator (implicitly with `with` and explicitly with `close()`)?

hunse:
I made the test names a bit more descriptive and added the docstring @Seanny123 asked for. LGTM now.

tbekolay:
Thanks @hunse! LGTM. @Seanny123, does this look good to you now?

Seanny123:
Yep! LGTM.

tbekolay:
This needed a rebase because of #932. I rebased it, so it should be good for merge @hunse.

Issue #772, opened 2015-07-07
Can't connect an error (size_out=1) to a learning rule object with size_in > 1


s72sue:
There seems to be no way to connect an error (with size_out=1) to a learning rule object with size_in=3.  It seems like you need to specify a (3x1) transform:
![image](https://cloud.githubusercontent.com/assets/12284191/8555525/a37d9eda-24be-11e5-803b-8dcdaa945e93.png)
Or a function:
![test2](https://cloud.githubusercontent.com/assets/12284191/8556502/632a46d8-24c5-11e5-83b9-223c231ca245.png)

However, both of the above approaches give a run time erorr in PES:
![image](https://cloud.githubusercontent.com/assets/12284191/8555466/531cbd04-24be-11e5-9802-c8af5667d21c.png)

hunse:
Can you post a whole minimal example?

s72sue:
I have created a simple example shown below  ( you can also download the iPython notebook from here if you want to play with it: https://drive.google.com/file/d/0B90f2TExXnVlU2xxUkFhc0VHWXc/view?usp=sharing ). Also note the change in the title (basically it doesn't work if the learning rule has size_in>1).

![test1](https://cloud.githubusercontent.com/assets/12284191/8571924/8f07d59e-2557-11e5-9c54-99c2f07bfe38.png)

arvoelke:
If you set `weights=False` then it works. I think the issue is that when you use a weight solver, you no longer have access to the vector space (i.e. the decoders), and so PES updates the entire weight matrix instead. Still not quite sure why I couldn't get it to work. Try looking at the example here: https://github.com/nengo/nengo/blob/master/nengo/tests/test_learning_rules.py#L77

Specifically the unit test `test_pes_weight_solver`. Switching to Eric's branch `monolithic-transform` may also help in some respects. Also going to the lines in `nengo` where the exceptions are being thrown can be helpful.

drasmuss:
I don't think the PES rule is set up to work on transformation matrices right now, it only works on decoders (`PES.error_type == 'decoders'`).  And it looks like the `LearningRule.size_in` is set up to only work with `decoders` as well.  That's what setting `weights=True` does, it (in theory) tells the PES rule that it should be learning a transformation matrix, not decoders.  But while the PES rule itself is set up to support that, it doesn't look like the general learning rule code is.

arvoelke:
[...]

PR #957, opened 2016-02-21
Add unit test for learning with a transform.

tbekolay:
This adds the unit test mentioned in #772. It tests that when you learning on a connection with a transform mapping from an N to an M dimensional ensemble, the error signal should be M-dimensional.

Having implemented it though, I'm not sure if this is actually what is supposed to happen. The minimal example that @drasmuss posted in #772 fails because we apply the transform and then do learning (since #729); this is enforced partly by `test_pes_transform` But, this seems to be in opposition to how we [normally describe](http://pythonhosted.org/nengo/examples/learn_communication_channel.html) PES, which is that the error signal modifies the decoders. Now, we're essentially learning on the transformed decoders, rather than the decoders themselves, which means that when N > M, you're not able to learn in the full vector space, and when N < M, you're learning in a higher dimensional space than your input vector, which kind of feels like cheating?

Anyway, I think either way could be argued for, but this is probably something we should be more explicit about and update the learning notebook accordingly.

### DIFF
--- a/nengo/tests/test_learning_rules.py
+++ b/nengo/tests/test_learning_rules.py
@@ -138,6 +138,30 @@ def test_pes_transform(Simulator, seed):
     assert np.allclose(sim.data[p_b][tend], [1, -1], atol=1e-2)
 
 
+@pytest.mark.parametrize('weights', [True, False])
+def test_pes_multidim_error(Simulator, rng, weights):
+    """Test that PES works on error connections mapping from N to 1 dims.
+
+    Note that the transform is applied before the learning rule, so the error
+    signal should be 1-dimensional.
+    """
+
+    with nengo.Network() as net:
+        err = nengo.Node(output=[0])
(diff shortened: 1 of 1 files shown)


hunse:
This only affects ensemble->ensemble connections, right? In neuron->ensemble connections (as they are now), the "transform" always specifies the initial decoders, and then learning happens on these decoders; there's no transform in the traditional sense. So it makes sense to have ensemble->ensemble connections work similarly, in that the transform is combined with the solved decoders to get the overall initial decoders, and then learning happens on that whole thing.

Also, we'd have to make some significant changes to the connection builder if we want to have decoders->transform->encoders, since right now we just have weights->encoders so we have to do the learning on the transformed decoders.

All this by way of saying this PR looks good to me.

tbekolay:
Just rebased this; @drasmuss can you give this a look and merge if it looks good to you?

Issue #851, opened 2015-09-18
Update docs


hunse:
A TODO list of things we need to update in the docs for the 2.1 release

- [x] Fix docstring "parameters" and "arguments". Many of these have been removed (see e.g. Connection). Also, it would be nice to have a consistent ordering (I would prefer the same order as they appear in the constructor).

arvoelke:
My initial feeling is to order the parameters by how they appear in the constructor, and the attributes alphabetically (since it may include additional variables that need to be visually searched).

hunse:
> the parameters by how they appear in the constructor, and the attributes alphabetically

Agreed.

tbekolay:
Done in #959.

PR #958, opened 2016-02-22
Move label and seed to base NengoObject

tbekolay:
While making the docstring changes in #851, this seemed like a natural thing to do, so I made it its own PR as it should be considered in more depth than documentation changes.

Moving the label and seed to the base NengoObject simplifies the subclass code a little bit, and makes their interfaces more obvious and consistent. This adds a label to `Connection` and a seed to `Node`; the node seed doesn't do anything right now (I think?) but we could later on hook it up to the process if the node has a process.

I made the ordering of label and seed consistent -- label first, then seed, but could switch it around if there are arguments for having label last.

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -9,9 +9,8 @@ from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import ValidationError
 from nengo.learning_rules import LearningRuleType, LearningRuleTypeParam
 from nengo.node import Node
-from nengo.params import (
-    Default, Unconfigurable, ObsoleteParam, BoolParam, FunctionParam,
-    IntParam, NdarrayParam)
+from nengo.params import (Default, Unconfigurable, ObsoleteParam,
+                          BoolParam, FunctionParam, NdarrayParam)
 from nengo.solvers import LstsqL2, SolverParam
 from nengo.synapses import Lowpass, SynapseParam
 from nengo.utils.compat import is_iterable, iteritems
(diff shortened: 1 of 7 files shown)


hunse:
I'm worried right now that without the Node seed hooked up to processes, users might try to set Node seed intending to affect the process and be surprised when it doesn't. Can we error on setting Node seed until we have it do something?

tbekolay:
Fair point; I'll try hooking it up, as that shouldn't take long. If it's taking long then I'll add an error.

tbekolay:
I tried hooking it up but it was non-trivial. Mostly, it is not clear what the correct behavior is; I made issue #980 to discuss that. For now, I'll add the error.

nengo/connection.py
```diff
@@ -347,6 +346,9 @@ def size_out(self):
 
     @property
     def _label(self):
```
hunse: Maybe we should call this something other than `_label` now? like `_str`, or `_tag`, or something.

tbekolay: Yeah, makes sense! It's just here as a helper for `__str__` and `__repr__` anyhow. Think I'll go with `_str`.

tbekolay:
Added the error and renamed `_label` to `_str`. Ready for re-review from @hunse and review from @drasmuss!

nengo/connection.py
```diff
@@ -319,9 +317,10 @@ def __init__(self, pre, post, synapse=Default, transform=Default,
         self.function_info = function  # Must be set after transform
         self.solver = solver  # Must be set before learning rule
         self.learning_rule_type = learning_rule_type  # set after transform
-        self.seed = seed
         self.modulatory = modulatory
 
+        super(Connection, self).__init__(label=label, seed=seed)
```
hunse: Is there some reason this is called at the end? Usually I put it at the beginning if it doesn't matter, but that's just personal preference.

hunse:
> I think I'm putting myself on the other side of the fence now, just because up until now `NetworkMember` has been pretty invisible/undocumented from the user's perspective, so it seems like less of a burden to just keep it that way.

I agree. For some things I might like the `**kwargs` approach, but here it's good to have all the documentation in each class, I think.

hunse:
This all LGTM.

nengo/connection.py
tbekolay: No particular reason. The start probably makes more sense in case something goes wrong (as then the label will be set for easier debugging). I'll change!

tbekolay:
Yeah, my thought is the same as @hunse, in this case we expect users to read the `__init__` signature relatively frequently, so having it all there is better I think.

tbekolay:
Pushed a commit moving the `super().__init__` calls. Should be ready to merge for whoever gets to it first!

Issue #929, opened 2016-01-11
spa.AssociativeMemory doc string is out of date.


e2crawfo:

hunse:
Before the next release, we should go through all the SPA documentation if possible. There have been a lot of changes there so I'm guessing a number of things are out of date.

tbekolay:
Done in #959.

PR #959, opened 2016-02-22
Update docstrings and other documentation

tbekolay:
This is a WIP PR to track the documentation work from #851. I've only done Connection and Ensemble so far.

Note that this PR is based on #958, which also touches the `__init__` functions and orders of things. This PR is also going to be a tough one to manage in terms of merging things, so it might be a good idea to keep this as a WIP until we've merged a bunch of the other things in the 2.1 queue.

Todo:

- [x] Finish updating docstrings for other objects
- [x] Ensure object members are well ordered (e.g., properties first, then `__x__` methods, etc). Add a note to dev docs about what the order should be
- [x] Globally replace all references to links to what they're referencing (e.g., `:class:thing` instead of `thing`)
- [x] Make a note about parameter / attribute ordering in dev docs
- [x] Ensure SPA documentation is up to date (see #929)
- [x] Add SPA docs to the `docs` folder


Seanny123:
May I recommend that this pull request also fix #929?

tbekolay:
Sounds good! It wasn't in the milestone, so it wasn't in my radar, but I've added it now.

hunse:
This is a crazy idea, but could we have documentation attached to Parameters? I'm doing a bit of documentation for the LIF neuron types, and realizing I'm doing a lot of copy-pasting. Not for 2.1 probably, but maybe down the road? It's not clear to me how we'd then get that documentation into docstrings, but there should be some way, right?

Seanny123:
This was definitely possible when I was programming in Ruby, but for some
reason I'm having a hard time figuring out how to do it in Sphinx.
http://stackoverflow.com/questions/8320909/repeated-parameters-with-python-sphinx

tbekolay:
Yeah, I definitely want to do that, but probably not for 2.1. I think it'll take some metaclass stuff.

Seanny123:
Does documenting "other things" include `docstrings` for tests? If yes, can I include some documentation I've made for the EnsembleArray tests?

tbekolay:
Doc changes are safe (i.e., shouldn't affect anyone not looking at documentation) so if you feel like documenting something, feel free! Just add a comment here when you push the commit so people can pull the updated version of the branch.

hunse:
> Ensure object members are well ordered (e.g., properties first, then `__x__` methods, etc). Add a note to dev docs about what the order should be

What should the order be? I would say:

1. Class-level member variables (e.g. `nengo.params.Parameter.equatable`; or `error_type`, `modifies`, and `probeable` on `LearningRuleType`)
2. Parameters (i.e. derived from `nengo.params.Parameter`) with the parameters in `__init__` going first in that order, then parameters that don't appear in `__init__` in alphabetical order. All these parameters should appear in the docstring in the same order.
3. `__init__`
4. other `__x__` methods alphabetically
5. `@property` properties in alphabetical order
6. member functions in alphabetical order.
7. hidden properties and member functions (i.e. `_x` names) I'm not sure about. Currently, we seem to put them right above the function(s) they're used in. An alternative would be right after the function(s) they're used in, since you typically want to read the main function first and then the helper functions. A third option is just put them all at the end. A fourth option is to ignore the underscore and just put them alphabetically with everything else. I'd probably lean towards a hybrid of right after where they're used and at the end, leaning towards right after unless they're really long or are used in more than one place or in a subclass, in which case at the end.

tbekolay:
I think my order is exactly your order, though I am also not 100% sure about hidden things. Right after those functions makes sense to me! I often put them right before because in my mind you have to define what you use before you use it, but after is probably easier to read, so that works for me.

hunse:
Okay, I got most things obeying the order. I don't think we should follow it to a T, though, because there are some things that naturally group together (e.g. `__getstate__` and `__setstate__`), and there's no point in splitting those just to be alphabetical.

hunse:
Any preferences on how to list defaults for parameters in docstrings? I've been using "<name> : <type>, optional (Default: <default>)\n <description>" sometimes, but in the past we've put the default in the description.

jgosmann:
Shoudn't the default be visible from the function signature?

hunse:
> Shoudn't the default be visible from the function signature?

For some things (i.e. `FrozenObjects`), probably. For things that use the config system, we typically put the default default in the docstring. There are also cases where the default is `None`, usually meaning that the value is somehow chosen based on other values (e.g. `dt` in `Synapse.filt`). In those cases, it's nice to have a brief description of how that works.

tbekolay:
I'm +1 on putting the default next to the optional marker. Will go through and normalize that.

tbekolay:
Something that I'm realizing we didn't talk about, but should, in terms of how things should be ordered in objects... should we group `staticmethod`s and `classmethod`s and put them before other methods? I feel like sometimes they are a very different thing than other methods, but sometimes they're glorified helper functions that should stay close to the method they're used in... should we leave this as a case-by-case thing, or add it to the ordered list above?

hunse:
Yeah, I realized going through stuff that a lot of stuff is case-by-case. As a general rule, though, I would maybe try to put them before other methods, after properties? I'm pretty sure I'd separate them out in some way, I don't think we'd usually want them mixed in alphabetically with other methods.

tbekolay:
For some classmethods (e.g., factories like `dict.from_keys`) I might even want them before `__init__`... or at least before `__x__` methods. But I think in the general case I agree with before other methods, after properties. Could you update your list above? I don't trust myself to remember this when documenting it later!

hunse:
@tbekolay, when you have a chance, could you take a look at 06fd6d5 and let me know what you think? I know moving stuff around is a bit beyond the mandate of this PR, but it seemed easier to document and understand this stuff if it was separated up. Also, since the LeastSquaresSolvers have their own Parameter, it makes it clearer in the other solvers what is expected (a number of people in the lab had asked me about that before). If you think that looks good, for the most part, I can add some documentation to those.

tbekolay:
Already looked at it, seems like a good move to me! I think I would not worry too much about adding docs to the LeastSquaresSolvers at the moment. But if you have some time in the next little bit feel free! ... but don't push to this branch, I'm likely to be force pushing a bunch soon.

tbekolay:
OK, so I _finally_ got through all of the docstrings and referencing and otherwise got the documentation back up and looking nice.

I haven't made the commits for them yet as it's about 3000 lines of changes and I don't want to just dump it all in one commit. So I'll go back through it and try to organize the changes.

In the meantime, the newly rendered docs can be seen [here](http://www.bekolay.org/nengodocs-rtd/). In particular, please check out the [frontend API docs](http://www.bekolay.org/nengodocs-rtd/frontend_api.html), [network docs](http://www.bekolay.org/nengodocs-rtd/networks.html), and the [backend API docs](http://www.bekolay.org/nengodocs-rtd/backend_api.html), as these are where most of the changes show up.

I'm posting this now so that people can start reading through these pages (in particular @drasmuss and @hunse) to make the reviewing process faster as it would be nice to get the release out this week!

A few things that I didn't get to here are the process docs (which I will do in #955) and the config and cache docs (I made issue #1017 to track that; I think that can wait for 2.1.1).

tbekolay:
Ah, I also forgot to add the SPA docs somewhere. I'll do that tomorrow also.

nengo/builder/connection.py
```diff
@@ -129,12 +157,40 @@ def slice_signal(model, signal, sl):
     else:
         size = np.arange(signal.size)[sl].size
         sliced_signal = Signal(np.zeros(size), name="%s.sliced" % signal.name)
-        model.add_op(SlicedCopy(signal, sliced_signal, a_slice=sl))
+        model.add_op(SlicedCopy(sliced_signal, signal, src_slice=sl))
```
drasmuss: Was this just broken before?

nengo/builder/operator.py
```diff
@@ -247,50 +388,124 @@ def step_copy():
...
 
 class ElementwiseInc(Operator):
-    """Increment signal Y by A * X (with broadcasting)"""
+    """Increment signal ``Y`` by ``A * X`` (with broadcasting).
+
+    Implements ``Y[...] += A * X``.
+
+    Parameters
+    ----------
+    A : Signal
```
drasmuss: Can `A` be a vector, or are we restricting it to be a scalar?  Obviously works either way here, just wondering for other backends like nengo_ocl.

tbekolay: Good question, would be good to make explicit. @hunse, does nengo OCL have this restriction or does it work either way?

hunse: `A` can definitely be a vector (PES uses this). Nengo OCL even supports it being a matrix, but I'm not sure if we ever use this functionality in Nengo.

nengo/builder/operator.py
```diff
@@ -356,10 +571,41 @@ def reshape_dot(A, X, Y, tag=None):
...
+
+    Implements ``Y[...] += np.dot(A, X)``.
+
+    .. note:: Currently, this only supports matrix-vector multiplies
+              for compatibility with Nengo OCL.
+
+    Parameters
+    ----------
+    A : Signal
+        The first signal to be multiplied.
```
hunse: I'm pretty sure A has to be the matrix and X has to be the vector.

tbekolay:
Thanks for the edits @drasmuss! I'm afraid I've added a few new commits to look at though: d6dacee77d1fabde3308953 a52c223f632afb 5dba0f804a195fc

The diffs shouldn't be as bad here. Mostly, it's minor changes to [network docs](http://www.bekolay.org/nengodocs-rtd/networks.html) and the newly rendered [SPA docs](http://www.bekolay.org/nengodocs-rtd/spa.html).

With this, the content's all done IMO, but I'm not happy with the history right now so I'd like to fix that up before merge.

drasmuss:
Just that one comment on the new commits, looks good :+1:

tbekolay:
OK, I've gone through and fixed up the history so it should be much easier to review now. All of the code moves and all of the docstrings are in two big commits, so everything that's not one of those things should be in a clear commit now. @hunse you review? Then you or @drasmuss can merge.

nengo/builder/operator.py
```diff
@@ -247,45 +247,87 @@ def step_copy():
...
+
+    Notes
+    -----
+    1. sets ``[] if inc else [dst]``
+    2. incs ``[dst] if inc else []``
+    3. reads ``[src]``
+    4. updates ``[]``
+    """
+
+    def __init__(self, dst, src, dst_slice=Ellipsis, src_slice=Ellipsis,
```
hunse: I would prefer `src, dst`, and change `Copy` to be the same. That seems more similar to me to things like `DotInc`, `SimProcess`, etc., with the inputs first then the outputs.

hunse:
I didn't read through all the docstrings, but I'm sure they're fine :) Of the code stuff, my only suggestion is switching things on `Copy` and `SlicedCopy`. Otherwise LGTM.

tbekolay:
I made the copy / slicedcopy order changes in a94a4c3cf5df2bd4. I also added the associative memory example in f87de180fa8a, so I think this is all done and good to go! @hunse or @drasmuss, please merge if you also think it's good to go!

tbekolay:
Oh, and I made issue #1034 to possibly remove the Copy operator (but that would for sure be after the release).

PR #960, opened 2016-02-22
Make human2bytes robust to no space before unit.

### LINKED ISSUE #950: human2bytes function cannot parse bytes
Numbers given in bytes are not parsed correctly because `'B'` is one instead of two characters like all other unit suffixes.

tbekolay:
Previously, `human2bytes` assumed that there was a space between the digit string and the unit. It now allows there to be no space between the two. `bytes2human` remains unchanged, and will return a string with a single space between the digit string and the unit.

Fixes #950.

### DIFF
--- a/nengo/utils/tests/test_cache.py
+++ b/nengo/utils/tests/test_cache.py
@@ -11,6 +11,8 @@ def test_human2bytes():
     assert human2bytes('1 MB') == 1048576
     assert human2bytes('1.5 GB') == 1610612736
     assert human2bytes('14 B') == 14
+    assert human2bytes('1B') == 1
+    assert human2bytes('1   B') == 1
 
 
 def test_byte_align():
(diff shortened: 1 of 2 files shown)


nengo/utils/tests/test_cache.py
```diff
@@ -11,6 +11,8 @@ def test_human2bytes():
     assert human2bytes('1 MB') == 1048576
     assert human2bytes('1.5 GB') == 1610612736
     assert human2bytes('14 B') == 14
+    assert human2bytes('1.5GB') == 1610612736
```
jgosmann: I believe this case was already working. There should be a test like
`assert human2bytes('1B') == 1`

tbekolay: Whoops! Added a fixup commit testing the real bug.

hunse:
LGTM

jgosmann:
LGTM too :cake:

PR #962, opened 2016-02-24
Close open simulators at exit

tbekolay:
Right now, this raises a warning when a simulator is closed at exit so that users will learn to close open simulators manually.

This was broken out from #932 into a separate PR as it is somewhat controversial whether we should raise a warning. #932 should be merged before this PR.

In the meantime, I would appreciate if people could vote as to whether they support raising a warning or not. I think everything else is pretty much agreed upon.

I am +1 on raising a warning.

### DIFF
--- a/nengo/tests/test_simulator.py
+++ b/nengo/tests/test_simulator.py
@@ -1,9 +1,13 @@
+import gc
+
 import numpy as np
 import pytest
 
 import nengo
 import nengo.simulator
 from nengo.exceptions import SimulatorClosed
+from nengo.utils.compat import ResourceWarning
+from nengo.utils.testing import warns
 
 
 def test_steps(RefSimulator):
@@ -99,3 +103,15 @@ def test_close_steps(RefSimulator):
         sim.run_steps(1)
     with pytest.raises(SimulatorClosed):
         sim.step()
+
+
+def test_warn_on_opensim_gc(Simulator):
+    with nengo.Network() as net:
+        nengo.Ensemble(10, 1)
+
+    sim = Simulator(net)
+    assert sim
+
+    with warns(ResourceWarning):
+        sim = None
+        gc.collect()
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -2,8 +2,9 @@
 
 from __future__ import print_function
 
-from collections import Mapping
 import logging
+import warnings
+from collections import Mapping
 
 import numpy as np
 
@@ -12,7 +13,7 @@ from nengo.builder import Model
 from nengo.builder.signal import SignalDict
 from nengo.cache import get_default_decoder_cache
 from nengo.exceptions import ReadonlyError, SimulatorClosed
-from nengo.utils.compat import range
+from nengo.utils.compat import range, ResourceWarning
 from nengo.utils.graphs import toposort
 from nengo.utils.progress import ProgressTracker
 from nengo.utils.simulator import operator_depencency_graph
@@ -135,6 +136,14 @@ class Simulator(object):
 
         seed = np.random.randint(npext.maxint) if seed is None else seed
         self.reset(seed=seed)
+
+    def __del__(self):
+        """Raise a ResourceWarning if we are deallocated while open."""
+        if not self.closed:
+            warnings.warn(
+                "Simulator with model=%s was deallocated while open. Please "
+                "close simulators manually to ensure resources are properly "
+                "freed." % self.model, ResourceWarning)
 
     def __enter__(self):
         return self
--- a/nengo/utils/compat.py
+++ b/nengo/utils/compat.py
@@ -38,6 +38,17 @@ if PY2:
                                'replace')
             StringIO.write(self, data)
 
+    class ResourceWarning(DeprecationWarning):
+        """A warning about resource usage.
+
+        Note that we subclass from DeprecationWarning here solely because
+        DeprecationWarnings are filtered out by default in Python 2.7,
+        while in Python 3.2+ both DeprecationWarnings and ResourceWarnings
+        are filtered out. Subclassing from DeprecationWarning gives
+        the same (or at least very similar) behavior in Python 2 and 3
+        without having to modify filters in the warnings module.
+        """
+
 else:
     import pickle
     import configparser
@@ -46,6 +57,7 @@ else:
     string_types = (str,)
     int_types = (int,)
     range = range
+    ResourceWarning = ResourceWarning
 
     # No iterkeys; use ``for key in dict:`` instead
     iteritems = lambda d: iter(d.items())


hunse:
+1

drasmuss:
+1 on warning

tcstewar:
I very much do not think we should be raising a warning if people do not close the Simulator (either explicitly via .close() or via the with statement).  This is a rather far-reaching change and I think we need more experience with this change before imposing something like this.  In particular, I want feedback on how it changes Chris' extensive examples in his course notes.  During the last meeting one complication came up (splitting the ```with``` across jupyter cells), and I'm not convinced that there aren't any others.

jgosmann:
-1 on warning (I would be ok with backends that actually need to be closed, like nengo_spinnaker, giving a warning)

hunse:
For people who do not support a warning, do you think that we should be promoting always closing simulators as the best practice (preferrably with `with` but with `.close()` when that's not an option)? I guess to me, these things go hand-in-hand. If we're saying that users should be closing their own simulators because we can't or don't want to do it for them, then the warning makes sense to me because it helps encourage this behaviour. In this situation, we're basically saying `atexit` will work some of the time to close, but not all of the time, so close your own. The warning would be saying "we caught the unclosed simulator this time, but we can't always promise we'll do that, so please close yourself in the future."

The other alternative that I would support is a system where the user doesn't have to deal with closing simulators at all. We don't have manual closing in our examples, and we always do our best to close simulators at exit. Ideally, this would be with a robust solution that makes sure they will get closed even if an exception is thrown (kind of like a `with` block around the whole script), but if that's not possible with `atexit`. Manual closing would still exist for situations where users want more control (due to memory constrains or a backend that doesn't allow multiple simulators, or some place internal to something like Nengo GUI where we want to make sure that simulators are closed even under exception-al circumstances).

I'm less keen on the options in between. I know that files in Python are one example. Files don't get closed if the user doesn't do it. But would we want an `atexit` hook to close simulators when possible, like this PR has now (i.e. just remove the warning from this PR)? If we're doing that, then why should users bother doing it themselves? Why not have a fully automatic system? I know I'm being a bit of a devil's-advocate here; there are some arguments for this in-between option, but I'll let the proponents of it make them :). I'm just trying to summarize what I think is the thinking of many of us in favour of a warning. It seems like it's kind of waffley. If we want a manual system, why not do everything in our power to encourage users to do things right? And if we want an automatic system, why not just go all out and do that? If a user is using a backend that only supports one simulator at a time, they still might run into problems if they're using `.close()`, which they could have learned from notebooks that have simulations spread across multiple cells.

I don't want to bring up the original debate of manual versus automatic closing necessarily, though this is kind of what I've done. I really just want to give an idea of where I think the warning people are coming from. Truth be told, I think this is a pretty minor issue.

I think that **no matter what**, we should have a notebook explaining the whole system so that users understand things like: how to manually close their simulators; why manually closing with a `with` block is best, because it ensures things are closed; how some backends may not support multiple open simulators, while others do; how to close with `.close()` and why this might be useful for notebooks or multiple open simulators.

Sorry for all that writing. I know a lot of it reiterates what we already talked about, and I could probably make it much more concise, but I've written it now and I'm too lazy...

jgosmann:
I suppose my argument (or at least part of it) is that it is good to be sparse with warnings because otherwise the really important warnings get overlooked and ignored in the not-so-important warnings.

Also, I would like to try out how this warning works in Jupyter notebooks where I might not use the `with` style and might not close the simulator before halting or restarting the kernel. (Which is perfectly fine to do with the reference simulator.)

[Also the equivalent discussion about warning about unclosed files might be of interest.](https://bugs.python.org/issue10093) Seems like they settled for a warning that can optionally be activated?

Could we use `Simulator.__del__` instead of `atexit` for automatic closing? Or does this have the same problems as `atexit`?

drasmuss:
I think it also helps to look at the other types of warnings we throw.  Two that come to mind are that we warn when adding an attribute to a Nengo object, and we warn when setting a high learning rate.  Both of those are cases where there isn't even anything really wrong with what the user is doing, we're just using warnings to help the user avoid potential errors.  And those warnings are certainly more intrusive than the warning we're proposing here, since the user might receive a warning for doing something they totally intend to do, and one that they cannot easily fix without making changes to the way they're actually building their model.

Compared to that, the warning here seems pretty mild.  If users want their code to be able to run across back-ends, which is one of the main design features of Nengo, then they really need to be closing the Simulator.  This warning helps them avoid potentially tricky errors (whether by the Simulator not being closed at all, or the automatic `atexit`/`del` cleanup failing), which is consistent with our previous design philosophy.

tbekolay:
> Could we use Simulator.__del__ instead of atexit for automatic closing? Or does this have the same problems as atexit?

Yes, I believe `__del__` has the same issues as `atexit`; in general, my recollection is that it is less likely to fire than `atexit`. It is also impossible with the changes in this PR, because there would be a reference to an open simulator in `Simulator.open_sims` so it would never be garbage collected except in cases where you've already closed it.

I didn't know that there as a patch to Python for warning with open files -- if anything, that (to me) removes one of the main criticisms of throwing a warning, which was that Python doesn't do it for files (which are somewhat analogous). As for the criticism of having too much warning noise, it should be noted that the Python warnings module by default will only raise each warning once, so it should never be that annoying, and if a user is annoyed by it (as I was with some Brian warnings) you can filter individual warnings in your own scripts with `warnings.filter` (which is what I did with those Brian warnings).

jgosmann:
I tried this out in different settings:

1. `jupyter notebook`: You never see the warning (no matter whether you restart, interrupt, or halt the kernel).
2. `jupyter qtconsole`: You don't see the warning.
3. `jupyter console`: You don't see the warning.
4. `ipython`: You see the warning.
5. `python <some script>`: You see the warning.

Basically everything that connects to a kernel won't show the warning. (Note `ipython` is different from `ipython console`/`jupyter console`.)

> because there would be a reference to an open simulator in Simulator.open_sims

We could make that a weak reference?

[This statement](https://docs.python.org/3/whatsnew/3.2.html) sounds to me that file objects are closed with the `__del__` method (and that'ts where they do the warning apparently):

> While the deallocator for such object ensures it closes the underlying operating system resource (usually, a file descriptor), the delay in deallocating the object could produce various issues, especially under Windows.

.
> if anything, that (to me) removes one of the main criticisms of throwing a warning, which was that Python doesn't do it for files

But I don't think it is done by default. You have to explicitely activate it. (Which for some reason is done with `-Wdefault` :confounded: )

> you can filter individual warnings in your own scripts with warnings.filter

The situation I'm thinking of is some quick and dirty (potentially interactive) hacking where I don't won't to bother with the `close` boiler plate (e.g., I just want to close the interpreter session with Ctrl+D) and `with` might not be a possibility either. If I'm too lazy to type `sim.close()` in that case, I definitely won't care to filter the warning.

I'm also not opposed to be able to optionally enable this warning (as for files) as it can be helpful for debugging.

jgosmann:
If we do the warning, should it be a `ResourceWarning` instead of a `UserWarning`? To me `ResourceWarning` seems like a better match, but has implications on whether the warning gets displayed. By default it won't, but it will with the `-Wdefault` flag apparently or if Python was built in debug mode.

jgosmann:
> It is also impossible with the changes in this PR, because there would be a reference to an open simulator in Simulator.open_sims so it would never be garbage collected except in cases where you've already closed it.

Couldn't we get rid of that list anyways, if we'd do the `close` and warning in `__del__`?

tbekolay:
Yeah, `ResourceWarning` is a good idea. It's worth pointing out that our `DeprecationWarning`s also don't get shown without the `-Wdefault` flag.

> Couldn't we get rid of that list anyways, if we'd do the close and warning in __del__?

Part of the reason for the list is so that you can close all open simulators at once with `nengo.simulator.close_open_sims`. But yeah, if we move things to `__del__` it might be best to get rid of that stuff anyhow, as global state causes problems.

jgosmann:
> Part of the reason for the list is so that you can close all open simulators at once with nengo.simulator.close_open_sims

Could still be done with weak references. I think, that would preferably anyways because it would still allow automatic garbage collection.

tbekolay:
> weak references

Yes for sure, I'll change that if we decide to keep the list.

jgosmann:
I also can't find any reliable information that `__del__` is less likely to be called than `atexit`. To me it seems that it should be called whenever `atexit` would also be called. The only thing to be careful about is that other globals might not be available anymore in the `__del__` method.

tbekolay:
This PR's now been reworked based on discussion in this thread and at the dev meeting.

Now, we warn in `__del__` rather than using the `atexit` module (which happily removes some global state), and we raise a `ResourceWarning` rather than a `UserWarning`, so that the warning won't be shown in most cases.

Note that `ResourceWarning` doesn't exist in Python 2.7, so I [added it to `compat.py`](https://github.com/nengo/nengo/pull/962/commits/1f3d2bfb0671c6789be472046d5149f3c7163e06#diff-65b08fafb2509cf98b5405a3eb21768eR41).

Could we get reviews this afternoon? Pinging @drasmuss and whoever else wants to step up (yeah I'm calling y'all out)

nengo/tests/test_simulator.py
```diff
@@ -99,3 +103,15 @@ def test_close_steps(RefSimulator):
...
+def test_warn_on_opensim_gc(Simulator):
+    with nengo.Network() as net:
+        nengo.Ensemble(10, 1)
+
+    sim = Simulator(net)
+    assert sim
+
+    with warns(ResourceWarning):
+        sim = None
+        gc.collect()
```
hunse: Would this be easier to test with `del sim`? Doesn't really matter much to me, just wondering.

hunse:
That GIF really disturbs me. I think we need to close this issue immediately and never look at it again.

hunse:
Seriously though, this is fine by me.

nengo/tests/test_simulator.py
tbekolay: Python's test suite does this in the patch where they added the warning for open files, so I figured it was for a reason. Might not be! It works though, so ¯\_(ツ)_/¯

hunse: Okay, that's as good a reason as any.

jgosmann: I think even with `del sim` the call to `gc.collect()` is necessary because `del` doesn't call `__del__` directly.

drasmuss:
It feels like a `ResourceWarning` is essentially no warning, since the point of the warning is to teach novice users (who are pretty unlikely to have these warnings enabled).  So I would still be in favour of something more obvious, for the same reasons outlined above.  But it doesn't seem like we're going to get a consensus on that, so I'm fine with this just so we can close the PR.

tbekolay:
Yeah, I'm with you @drasmuss; we can revisit later for sure. Also, anyone who +1'd this PR early on has "told you so" rights when we get issues rolling in that are solved with closing sims ;)

@drasmuss can you merge?

PR #964, opened 2016-02-29
Have SlicedCopy keep track of the original slices.

e2crawfo:
In the constructor for SlicedCopy, if the passed-in slices are indeed
slices (as opposed to lists of integers), then the slicing is done
right in the constructor, a sliced view of the original array is
stored, and the original slices are thrown away. This commit
changes the constructor so that the original slices are also stored.
This lets us write a __str__ method in terms of the arguments that were
passed to the constructor. Also, this information is required by nengo_mpi.

### DIFF
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -215,6 +215,9 @@ class SlicedCopy(Operator):
     """Copy from `a` to `b` with slicing: `b[b_slice] = a[a_slice]`"""
     def __init__(self, a, b, a_slice=Ellipsis, b_slice=Ellipsis,
                  inc=False, tag=None):
+        self.a_base_slice = a_slice
+        self.b_base_slice = b_slice
+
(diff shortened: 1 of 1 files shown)


hunse:
Why is this required my nengo_mpi?

e2crawfo:
I need to be able to access how the sliced signal is related to the base signal, which doesn't seem to be available anywhere else. Thinking about it a bit more though, this PR is probably not the right solution; the new "__str__" method doesn't do the right thing when either of the passed in signals are views, for instance, so I'm going to close this for now.

hunse:
That relation should all be contained in the sliced signal. The sliced signal will have a stride (signal.elemstrides) and an offset (signal.offset), as well as a shape (signal.shape), which should be all you need to work with it. That's what I use in nengo_ocl anyway.

e2crawfo:
OK thanks I'll try that out.

PR #966, opened 2016-02-29
Added signals for simulator time and step

hunse:
This makes things like resetting the simulator cleaner, since
time and step are no longer treated specially.

Sorry for the late addition to 2.1, but this will make some stuff in Nengo OCL easier. If it's contentious, it's not imperative.

### DIFF
--- a/examples/usage/strings.ipynb
+++ b/examples/usage/strings.ipynb
@@ -275,8 +275,8 @@
       "print(nengo.builder.learning_rules.SimOja(sig, sig, sig, sig, 0.1, 1.0, tag=\"tag\"))\n",
       "print(nengo.builder.neurons.SimNeurons(nengo.LIF(), sig, sig, [sig]))\n",
       "print(nengo.builder.neurons.SimNeurons(nengo.LIF(), sig, sig, [sig], tag=\"tag\"))\n",
-      "print(nengo.builder.processes.SimProcess(nengo.processes.Process(), sig, sig))\n",
-      "print(nengo.builder.processes.SimProcess(nengo.processes.Process(), sig, sig, tag=\"tag\"))\n",
+      "print(nengo.builder.processes.SimProcess(nengo.processes.Process(), sig, sig, sig))\n",
+      "print(nengo.builder.processes.SimProcess(nengo.processes.Process(), sig, sig, sig, tag=\"tag\"))\n",
       "print(nengo.builder.synapses.SimSynapse(sig, sig, nengo.Lowpass(0.005)))\n",
       "print(nengo.builder.synapses.SimSynapse(sig, sig, nengo.Lowpass(0.005), tag=\"tag\"))"
      ],
--- a/nengo/builder/builder.py
+++ b/nengo/builder/builder.py
@@ -4,6 +4,7 @@ import warnings
 import numpy as np
 
 from nengo.builder.signal import Signal, SignalDict
+from nengo.builder.operator import TimeUpdate
 from nengo.cache import NoDecoderCache
 from nengo.exceptions import BuildError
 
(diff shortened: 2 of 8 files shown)


tbekolay:
No worries, these look like good changes to have in 2.1 IMO! Reading through the code it looks good to me. @drasmuss can you look this over as well?

hunse:
So one problem with this whole thing: `Simulator.time`, as well as `n_steps` and `trange`, all depend on signals, which aren't guaranteed to be around when the simulator closes (I get rid of them in Nengo OCL).

tbekolay:
I think `Simulator.time` and `Simulator.n_steps` makes sense to remove on close (they feel like simulation state) but `trange` is used at the same time as `data` for plotting and whatnot, and we use `n_steps` in `trange`. Perhaps `n_steps` should be stored as `Simulator._n_steps` on close? Or whenever probe data is pulled off the device?

hunse:
`time` and `n_steps` can both be useful for dealing with the data, I think, so I'd like them to still work on close, along with `t_range`.

I think storing both on close makes the most sense. Doing it when probe data is stored would result in extra device accesses for no benefit that I can see.

tbekolay:
What happens when someone asks for `trange` or `data` before closing the sim?

hunse:
Then we get what we need from the signal

tbekolay:
Okey dokes. Then the benefit to storing it when probe data is stored would be to avoid device accesses when calculating `trange`; either way should be fine though.

hunse:
Okay, so we can cache the values to minimize device reads, but it does mean that we have to clear the cached values on step and reset. Does this seem okay?

hunse:
What do people think about this? I think this is one of the last things I'm waiting on for nengo_ocl.

tbekolay:
I like `TimeUpdate`; I'm not sure about f9e5eede8ad278dc8a. Seems like it's not really useful for Nengo, just for Nengo OCL; does OCL not override these?

hunse:
No, with f9e5eed, nengo_ocl doesn't have to override them.

hunse:
@tbekolay, would you prefer if we just copied these off the "device" when we do probes, and put them in `_n_steps` and `_time`? That might make the code a little cleaner. I've tried that out in [`time-step-signals2`](https://github.com/nengo/nengo/tree/time-step-signals2).

tbekolay:
Yeah, I like the approach in `time-step-signals2` better. I'd be good to merge with that and @drasmuss's +1!

drasmuss:
If we made a function like

``` python
class Set(Operator):
    def __init__(self, dst, src=None, func=None, tag=None):
        self.dst = dst
        self.src = src
        self.func = func or lambda x: x
        self.tag = tag

        self.sets = [dst]
        self.incs = []
        self.reads = [] if src is None else [src]
        self.updates = []

        dst = signals[self.dst]
        
        if self.src is not None:
            src = signals[self.src]
            def step_set():
                dst[...] = self.func(src)
        else:
            def step_set():
                dst[...] = self.func()
        return step_set
```

then `Copy`, `Reset`, and `TimeUpdate` could all be implemented just as special cases of that.  I can imagine that that could cause some problems for `nengo_ocl` though, since it is a more general operation and therefore less optimizable?

hunse:
Yeah, that would make things really annoying for nengo_ocl. I would basically have to go through and re-separate out all those ops, which would involve somehow figuring out what the function is doing (whether it's returning a constant, etc.). If you'd like to investigate that, we can always change it down the road.

drasmuss:
Alright looks ready to go to me then.

hunse:
Ok, rebased this and squashed it all together, bringing over the additional commit from the other branch. Should be good to go.

PR #968, opened 2016-02-29
Can read sliced signals from SignalDict

### LINKED ISSUE #967: Sliced voltage Probes don't work
This used to work but does not any more:

```python
import nengo

model = nengo.Network()
with model:
    ens = nengo.Ensemble(100, 1)
    p = nengo.Probe(ens.neurons[:2], 'voltage')
    
sim = nengo.Simulator(model)
with sim:
    sim.run(1)
```

This seems to be broken as of commit 82fa822 and means that Nengo GUI voltage probes do not work with the latest nengo.

hunse:
This allows things like slices on voltage probes to work.
Fixes #967.

### DIFF
--- a/nengo/builder/signal.py
+++ b/nengo/builder/signal.py
@@ -1,5 +1,6 @@
 """Signals represent values that will be used in the simulation.
 """
+from __future__ import division
 
 import numpy as np
 
@@ -46,8 +47,14 @@ class Signal(object):
         return self.initial_value.dtype
 
     @property
+    def elemoffset(self):
+        """Offset of data from base in elements"""
+        return self.offset // self.itemsize
+
+    @property
     def elemstrides(self):
-        return tuple(s / self.itemsize for s in self.initial_value.strides)
+        """Strides of data in elements"""
+        return tuple(s // self.itemsize for s in self.strides)
 
     @property
     def initial_value(self):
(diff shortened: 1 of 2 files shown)


tcstewar:
Cool!  Thank you for the fast response on this!

Should a test case be added?  I'm not sure what the scope of the problem was....  :)

hunse:
It was easy enough to modify a test to check the basic functionality of this. It doesn't test that all the slicing is correct for all different types and permutations of slices, but I don't think that's necessary.

Seanny123:
FYI, the test is failing on Python 3 stuff.

hunse:
Okay, this should fix Python3. I changed the Signal interface a bit to what I think is much clearer (it now has `strides` and `offset` both in bytes, and `elemstrides` and `elemoffset` in elements), but this will break any backends that use `offset`, because it's now in terms of bytes, not strides. @e2crawfo, do you use this on nengo_mpi? This is probably related to dealing with views in things like SlicedCopy, which I think you were working on recently, so maybe now is not a bad time to change this.

e2crawfo:
Yeah, I'm in the process of improving how nengo_mpi deals with views, I'll make sure to use elemoffset instead of offset.

hunse:
Great! This should be ready for re-review, then.

drasmuss:
Looks good

hunse:
@tbekolay can you merge? It will make some nengo_ocl stuff easier for me right now.

tbekolay:
Sure, doing so now!

PR #969, opened 2016-03-01
Up TravisCI NumPy version to 1.10

tbekolay:
This speed up our Travis builds significantly; compare [before](https://travis-ci.org/nengo/nengo/builds/112884369) to [after](https://travis-ci.org/nengo/nengo/builds/112892842). For the lazy, our 2.7, 3.4 and 3.5 builds go from ~18 minutes to ~5 minutes. And that includes time to download ~120 MB MKL binary.

NumPy 1.10 isn't packaged for Python 2.6, so I'll work on #954 next, after which I think this makes sense to merge next, as it'll make testing the rest of the 2.1 PRs faster.


drasmuss:
Looks good (assuming the failing tests will be fixed after #954).  We could also consider dropping numpy 1.6, since we're adding a new test for 1.9, just to keep our test time down (because really who is still using 1.6).  But meh.

Seanny123:
FYI, [Ubuntu 12.04 comes with Numpy 1.6 by default](http://stackoverflow.com/questions/19839488/upgrade-to-numpy-1-8-0-on-ubuntu-12-04), which is technically [supported until 2017](https://wiki.ubuntu.com/LTS).

requirements.txt
```diff
@@ -1 +1 @@
-numpy>=1.5
```
hunse: Was this just a typo? Did we ever support 1.5?

hunse:
LGTM. This is ready to go now, right?

tbekolay:
Not quite, I was experimenting a bit with the number of threads to use on TravisCI. Looks like 2 was optimal, so I'll push that shortly.

requirements.txt
tbekolay: This was actually here from literally the first commit (https://github.com/nengo/nengo/commit/a21c503940d4331f1876e2295554cea06816d89c#diff-b4ef698db8ca845e5845c4618278f29a) so I think it was a pretty arbitrary.

hunse: Haha, wow, it's crazy looking back on that first commit. I was just wondering if this is something that deserves a changelog entry (saying we no longer support 1.5). I think I'm fine without it, since  `setup.py` has listed 1.6 as the requirement for a while now (maybe forever).

tbekolay: Right, good point. I think given that `setup.py` would have borked on version 1.5 it's probably fine (I wonder if we do actually run on 1.5...)

tbekolay:
OK should be ready to go now

PR #970, opened 2016-03-01
Return self in `Config.__enter__`

drasmuss:
This is needed so that `with nengo.Config() as x:` works properly.


tbekolay:
This is a bug IMO so adding it to 2.1.0. LGTM! Will merge if someone else can give at a review.

hunse:
LGTM

PR #972, opened 2016-03-02
Update to newer pytest

tbekolay:
We talked about this in #856; pytest still hasn't fixed the bug (https://github.com/pytest-dev/pytest/issues/1082) but just released 2.9.0, so I'm going to assume that they're unlikely to fix it anytime soon. So, this PR implements @hunse's recommended fix of moving `conftest.py` to the root `nengo` directory and removing all of the conftests from test directories. I think this is an okay solution, and does reduce the number of files in the repo, so that's a plus at least.

This came up as I updated to pytest 2.9.0 locally since it was just released, and noticed that this was still not fixed. I think it makes sense to do this though, and merge it first as I'm making other changes to `.travis.yml` in #954 and #969.


hunse:
Yep, seems to work well! Even if pytest does fix in the future, it'll be nice to work with any pytest version. LGTM

tbekolay:
@drasmuss can you review / merge?

jgosmann:
I would like to look at this later today (can't do so right now). It might be related to another PR by me.

tbekolay:
Cool, disregard @drasmuss!

jgosmann:
LGTM :cake: 
will merge

PR #974, opened 2016-03-02
Make individual connection output probeable

### LINKED ISSUE #973: Probing Connections
It appears that a probe on a connection returns the output of the post object rather than just the signal from that connection. So if you have multiple inputs going into an object, the connection probes on each return the sum of all the inputs, not the signal going through each connection.

In the following example all the probes show the same value.

```
import nengo

    input = nengo.Node(output=1)
    ens1 = nengo.Ensemble(dimensions=1, n_neurons=100)
    ens2 = nengo.Ensemble(dimensions=1, n_neurons=100)
    ens3 = nengo.Ensemble(dimensions=1, n_neurons=100, radius=1)
    nengo.Connection(input, ens1)
    nengo.Connection(input, ens2)
    c1 = nengo.Connection(ens1, ens3, transform=2)
    c2 = nengo.Connection(ens2, ens3, transform=-1)
    p1 = nengo.Probe(c1, synapse=0.05)
    p2 = nengo.Probe(c2, synapse=0.05)
    p3 = nengo.Probe(ens3, synapse=0.05)

s = nengo.Simulator(model)
s.run(1)
plt.plot(s.trange(), s.data[p1], label='conn1')
plt.plot(s.trange(), s.data[p2], label='conn2')
plt.plot(s.trange(), s.data[p3], label='ens')
plt.legend(loc='best')
plt.ylim(-2, 2)
plt.show()
```

hunse:
Fixes #973.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -184,6 +184,10 @@ Release History
   creation attempts, and returns the most dissimilar pointer if none can be
   found below the similarity threshold.
   (`#817 <https://github.com/nengo/nengo/pull/817>`_)
+- Probing a Connection's output now returns the output of that individual
+  Connection, rather than the input to the Connection's post Ensemble.
+  (`#973 <https://github.com/nengo/nengo/issues/973>`_,
+  `#974 <https://github.com/nengo/nengo/pull/974>`_)
 
 2.0.3 (December 7, 2015)
 ========================
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -215,6 +215,9 @@ def build_connection(model, conn):
     if conn.synapse is not None:
         signal = model.build(conn.synapse, signal)
 
+    # Store the weighted-filtered output in case we want to probe it
+    model.sig[conn]['weighted'] = signal
+
     # Copy to the proper slice
     model.add_op(SlicedCopy(
         signal, model.sig[conn]['out'], b_slice=post_slice,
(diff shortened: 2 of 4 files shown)


tbekolay:
This LGTM and as a bugfix makes sense for 2.1. Can you add a changelog entry?

tbekolay:
A test would also be good; I can add that if you want.

hunse:
Sure, if you have a good idea for a test go for it. I'll add a changelog entry.

tbekolay:
Rebased to master and added a test. @drasmuss can you review?

drasmuss:
I think it might make sense to include the slice in the probed output.  I.e., if I do
``` python
conn = nengo.Connection(a, b[0])
probe = nengo.Probe(conn)
```
I think I would expect the probed values to be 1D.  Something like this instead:
``` python
      # Store the weighted-filtered output in case we want to probe it
      model.sig[conn]['weighted'] = signal[post_slice]
```

hunse:
The slice is on where the function is going, though, not on the function itself. So I'm fine with leaving off the output slice. In your example, the probed values HAVE to be 1D, because your connection code is saying "compute this function and put it into the first dimension of 'b'", and since the first dimension of 'b' is 1D then the probed thing has to be 1D. So the output slice isn't being applied to the output of the connection, it's being applied to where the connection's output is going. Does that make sense?

drasmuss:
Ah yes of course you're right, I just glanced at the 
``` python
model.add_op(SlicedCopy(
          signal, model.sig[conn]['out'], b_slice=post_slice,
         inc=True, tag="%s.gain" % conn))
```
right beneath and was thinking "we should have that in", but of course the slicing is being done by the weights already.  All looks good then!

PR #978, opened 2016-03-08
PresentInput shows first input for equal time

### LINKED ISSUE #976: PresentInput skips first input
I was using the `PresentInput` process to present a series of inputs one timestep at a time, and ran into this issue.  Basically because the first time step in Nengo is `t=dt` (not `t=0`), `PresentInput` starts with the second item in the list instead of the first.  Here's a minimal example:

``` python
import nengo
import numpy as np

inputs = np.arange(10)
dt = 1e-3
with nengo.Network() as net:
    x = nengo.Node(output=nengo.processes.PresentInput(inputs, dt))
    p = nengo.Probe(x)

sim = nengo.Simulator(net, dt=dt)
sim.run_steps(5)

print(sim.data[p])
print(sim.trange())
```
prints
```
[[ 1.] [ 2.] [ 3.] [ 4.] [ 5.]]
[ 0.001  0.002  0.003  0.004  0.005]
```
(note, we never see 0).

My proposed fix would be to change the step function to:
``` python
    def step_presentinput(t):
            i = int((t-dt) / presentation_time + 1e-7)
            return inputs[i % n]
```
Would that cause any problems (or seem wrong) to you @hunse?

hunse:
Fixes #976.

### DIFF
--- a/nengo/processes.py
+++ b/nengo/processes.py
@@ -306,7 +306,7 @@ class PresentInput(Process):
         presentation_time = float(self.presentation_time)
 
         def step_presentinput(t):
-            i = int(t / presentation_time + 1e-7)
+            i = int((t-dt) / presentation_time + 1e-7)
             return inputs[i % n]
 
         return step_presentinput
--- a/nengo/tests/test_processes.py
+++ b/nengo/tests/test_processes.py
@@ -297,7 +297,7 @@ def test_present_input(Simulator, rng):
         sim.run(1.0)
 
     t = sim.trange()
-    i = np.floor(t / pres_time + 1e-7) % n
+    i = np.floor((t - sim.dt) / pres_time + 1e-7) % n
     y = sim.data[up].reshape(len(t), c, ni, nj)
     for k, [ii, image] in enumerate(zip(i, y)):
         assert np.allclose(image, images[ii], rtol=1e-4, atol=1e-7), (k, ii)


hunse:
I don't think this needs a changelog because we haven't published a release with PresentInput, yet.

nengo/processes.py
```diff
@@ -306,7 +306,7 @@ def make_step(self, size_in, size_out, dt, rng):
         presentation_time = float(self.presentation_time)
 
         def step_presentinput(t):
-            i = int(t / presentation_time + 1e-7)
+            i = int((t-dt) / presentation_time + 1e-7)
```
arvoelke: Nit: I've had uses for this too. Can we add a constant for epsilon to a utility file like `nengo.utils.compat`? [One possibility](http://stackoverflow.com/questions/19141432/python-numpy-machine-epsilon) is `np.finfo(float).eps`?

arvoelke:
LGTM (agreed about changelog)

nengo/processes.py
hunse: One problem is what you need to add to get to the next floating point number depends on the current floating point number. You can figure it out exactly with `lambda x: np.nextafter(x, np.inf) - x`. `f(1) == np.finfo(float).eps == 2.22e-16` (this must be where they get it from), but `f(10) == 1.77e-15`, etc.

PR #979, opened 2016-03-09
Compatibility for nengo_lasagne

drasmuss:
Some changes I made for compatibility with the `nengo_lasagne` back end.

The first commit just separates out the code for computing gains and biases into its own function (so that I can reuse that code in `nengo_lasagne`).  Doesn't change anything in reference Nengo.

The second commit is more significant, it adds the option to pass a Distribution for the transform parameter.  This isn't a necessary feature (you could always just manually sample a correctly sized matrix from the desired distribution, and pass that instead), but it is useful (and has been desired in other cases as well #639/#942).

I would say that this is a post-2.1 change, unless someone feels strongly about having `nengo_lasagne` be compatible with 2.1.

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -9,6 +9,7 @@ from nengo.builder.operator import (
     DotInc, ElementwiseInc, PreserveValue, Reset, SlicedCopy)
 from nengo.builder.signal import Signal
 from nengo.connection import Connection
+from nengo.dists import Distribution
 from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError, ObsoleteError
 from nengo.neurons import Direct
@@ -165,6 +166,11 @@ def build_connection(model, conn):
     signal_size = conn.size_out
     post_slice = conn.post_slice
 
+    # Sample transform if given a distribution
+    transform = (conn.transform.sample(conn.size_out, conn.size_mid, rng=rng)
+                 if isinstance(conn.transform, Distribution) else
+                 conn.transform)
+
     # Figure out the signal going across this connection
     in_signal = model.sig[conn]['in']
     if (isinstance(conn.pre_obj, Node) or
@@ -189,13 +195,13 @@ def build_connection(model, conn):
             post_slice = Ellipsis  # don't apply slice later
             weights = decoders.T
         else:
-            weights = multiply(conn.transform, decoders.T)
+            weights = multiply(transform, decoders.T)
     else:
         in_signal = slice_signal(model, in_signal, conn.pre_slice)
 
     # Add operator for applying weights
     if weights is None:
-        weights = np.array(conn.transform)
+        weights = np.array(transform)
 
     if isinstance(conn.post_obj, Neurons):
         gain = model.params[conn.post_obj.ensemble].gain[post_slice]
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -70,7 +70,7 @@ def build_linear_system(model, conn, rng):
     return eval_points, activities, targets
 
 
-def build_decoders(model, conn, rng):
+def build_decoders(model, conn, rng, transform):
     encoders = model.params[conn.pre_obj].encoders
     gain = model.params[conn.pre_obj].gain
     bias = model.params[conn.pre_obj].bias
@@ -83,7 +83,7 @@ def build_decoders(model, conn, rng):
     if conn.solver.weights:
         E = model.params[conn.post_obj].scaled_encoders.T[conn.post_slice]
         # include transform in solved weights
-        targets = multiply(targets, conn.transform.T)
+        targets = multiply(targets, transform.T)
 
     try:
         wrapped_solver = model.decoder_cache.wrap_solver(solve_for_decoders)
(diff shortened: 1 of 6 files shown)


nengo/connection.py
```diff
@@ -128,44 +128,47 @@ def validate(self, conn, function_info):
         transform = conn.transform
         size_mid = conn.size_in if function is None else size
 
-        if transform.ndim < 2 and size_mid != conn.size_out:
-            raise ValueError("function output size is incorrect; should "
-                             "return a vector of size %d" % conn.size_out)
+        if isinstance(transform, np.ndarray):
```
arvoelke: Is it right that `DistOrArrayParam` will typecast the `transform` to an `ndarray` if it's something like a 2D-list?

arvoelke:
The code looks good to me (assuming my question is right), but it's missing unit tests!

nengo/connection.py
drasmuss: Yeah, it happens in `NdarrayParam` (the parent of `DistOrArrayParam`).

Seanny123:
I have a vested interest in getting this merged to master because I'll soon be using Nengo with Lasagne. Are unit tests, formatting and rebasing all this requires?

drasmuss:
Assuming no one has any objections to the transform distribution initialization part.  That part hasn't really been battle tested, I just threw it in and tried a few examples.

nengo/builder/ensemble.py
```diff
@@ -49,6 +49,23 @@ def get_activities(model, ens, eval_points):
         x, model.params[ens].gain, model.params[ens].bias)
 
 
+def get_gain_bias(ens, rng=np.random):
```
Seanny123: Why bother extracting this into a function? This refactoring doesn't seem to be being used anywhere else in the pull request. Maybe extract it to a separate one?

arvoelke: According to the OP,
> (so that I can reuse that code in `nengo_lasagne`)

As long as there's a unit test for this new function, and for the new use case in general, I'm happy with the rest.

Seanny123: Derp. Don't know how I missed that.

drasmuss:
Rebased and added some unit tests.  I also added a changelog entry in the 2.1 section, but as I said I don't think this is critical to have in that release.

Seanny123:
LGTM

hunse:
So currently I think this will break if the transform is a distribution and you use a weight solver. That code was already pretty messy, and the distribution stuff makes it even more so, so I'm going to look at if it can be refactored easily.

hunse:
Okay, I fixed this so that transform distributions work with weight solvers.

I also added a commit to make the transform accessible in BuiltConnection. This makes sense to me because now `transform` is something that might be sampled, so users might want to know what it is independent of the weights. I think the meaning of `transform` in `BuiltConnection` is the same as it was in 2.0, so we don't have any problem there, though we might want to modify the change log since now we've removed `transform` from `BuiltConnection` and then added it back in the course of 2.1 development.

nengo/builder/connection.py
```diff
@@ -169,7 +171,7 @@ def get_prepost_signal(is_pre):
     # Sample transform if given a distribution
     transform = (conn.transform.sample(conn.size_out, conn.size_mid, rng=rng)
                  if isinstance(conn.transform, Distribution) else
-                 conn.transform)
+                 np.array(conn.transform))
```
drasmuss: The only reason to make a copy here is so that if someone accesses `BuiltConnection.transform` it isn't aliased with the `Connection.transform` right?  Just wondering if we could mark it as readonly or something instead, to cut down on memory usage (since these could be large arrays).

hunse: So that's the side benefit. The main reason we want the copy is so that if we're doing something that doesn't have an ensemble as the pre, the transform simply becomes `weights`, and then we don't want our weights signal value to be a view on `conn.transform`.

hunse: The code used to try to be super smart about this, and only made the copy if we were not in that case where we were solving for decoders. But it made things a bit more complicated to read, and I also wanted to avoid making a copy of the sampled distribution, since that's extra computation (though arguably memory savings are more important). And then I realized that if we're storing it in BuiltConnection, it would be nice to have a copy anyway, and so that convinced me that it's fine to just make a copy right off the bat. A readonly view is a possibility, though the thing I always worry about is a user creates an array, sets it to be the transform, then modifies the original array. There's no way that we can set the user's view on their array to be readonly, so they'll always be able to do this. But maybe that's their own damn fault? There are definitely arguments for saving memory, especially when these transforms are in neuron->neuron connections and therefore possibly very large. Could we have an rc setting that copies by default, but the user can advise to not copy if they need the memory?

drasmuss: Ah yeah I see, you just moved the copy from line 204 to here.  So we'll end up with an extra copy of the transform in the `pre==Ensemble` case, but in that case the transform is probably pretty small so it doesn't really matter.

drasmuss:
Those changes all look good to me.

nengo/builder/connection.py
drasmuss: Sorry commented on your first post before reading the second.  Yeah, I think it's easiest/clearest to just do the copy here.  But that'd be a nice wishlist item, a config setting that'd act like a compiler flag to optimize for memory usage.

hunse:
All looks good to me, too. I checked the changelog, and even though we've added `transform` back into BuiltConnection, that wasn't referenced specifically so it reads fine as-is.

@tbekolay, do you want to do the honours, since my hands are on this now too? I figured we would get this into 2.1, because it might be nice to have it and the changelog entry is already under 2.1 :)

tbekolay:
Yep, I'll take a last look and then merge :)

Issue #885, opened 2015-10-22
Fix spa.AssociativeMemory documentation (and add a few examples)


Seanny123:
I was trying to use the `spa.AssociativeMemory` as a basic dictionary and was having the darnest time. The code to finally get this scenario to work is:

The reasons I couldn't figure this out on my own is:
- There are no associative memory examples in the docs
- The arguments `input_keys` and `output_keys` are undocumented

ikajic:
I'm just going through this. First, I'm getting an `AttributeError` which I think I shouldn't be getting:
```python
In [4]: spa.AssociativeMemory(np.arange(10))
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-e7f05435fee3> in <module>()
----> 1 spa.AssociativeMemory(np.arange(10))

/home/ivana/phd/workspace/nengo/nengo/spa/assoc_mem.pyc in __init__(self, input_vocab, output_vocab, input_keys, output_keys, default_output_key, threshold, input_scale, inhibitable, inhibit_scale, wta_output, wta_inhibit_scale, wta_synapse, threshold_output, label, seed, add_to_container, **ens_args)
     66 
     67         if input_keys is None:
---> 68             input_keys = input_vocab.keys
     69             input_vectors = input_vocab.vectors
     70         else:

AttributeError: 'numpy.ndarray' object has no attribute 'keys'
```

If the input is `spa.Vocabulary` it works, so it smells of a bug. What is the rationale behind keys when using a numpy.array?

xchoo:
The ```spa.AssociativeMemory``` network is expecting a vocabulary as an input. If you want to use vectors (or numpy arrays), try the ```networks.AssociativeMemory``` network. (The spa network is just a wrapper for the networks network)

ikajic:
Hm, the documentation for `spa.AssociativeMemory`says something different:
```
input_vocab: list of numpy.array, spa.Vocabulary
    The vocabulary (or list of vectors) to match.
```
So, even if I put a list of numpy arrays it throws the same error.

xchoo:
Oh. Hah! Whoops. That's an error in the documentation.

s72sue:
+1 for adding some associative memory examples to the docs.

ikajic:
Alright. I'll take care of fixing the docstrings and put a few examples in there.

ikajic:
What is the purpose of `add_to_container` parameter in `spa.AssociativeMemory`?

[...]

PR #982, opened 2016-03-14
Provide SP expressions as keys to the SPA associative memory

ikajic:
I needed this scenario for my model, so I decided to submit a PR in case it might be of interest for others. Currently, only single keys can be passed as the input and output keys to the associative memory module. This PR slightly changes the existing code so that it also supports expressions in both input or output keys. Then something like this can be done:

```python
import nengo
from nengo import spa

with spa.SPA() as model:
    D = 16
    
    voc1 = spa.Vocabulary(D)
    voc2 = spa.Vocabulary(D)
    
    voc1.parse('A+B+C')
    voc2.parse('D+E+F')
    
    in_keys = ['A', 'B+C']
    out_keys = ['D-E', '0.4*F']
    
    model.am = spa.AssociativeMemory(input_vocab=voc1,
                                     output_vocab=voc2,
                                     input_keys=in_keys,
                                     output_keys=out_keys
                                    )
    model.inp = spa.State(D, vocab=voc1)
    nengo.Connection(model.inp.output, model.am.input)
```

### DIFF
--- a/nengo/spa/assoc_mem.py
+++ b/nengo/spa/assoc_mem.py
@@ -60,7 +60,9 @@ class AssociativeMemory(Module):
             input_keys = input_vocab.keys
             input_vectors = input_vocab.vectors
         else:
-            input_vectors = input_vocab.create_subset(input_keys).vectors
+            # parse keys (allows expressions as keys)
+            input_vectors = [input_vocab.parse(key).v for key in
+                             input_keys]
 
         # If output vocabulary is not specified, use input vocabulary
         # (i.e autoassociative memory)
@@ -70,7 +72,10 @@ class AssociativeMemory(Module):
         else:
             if output_keys is None:
                 output_keys = input_keys
-            output_vectors = output_vocab.create_subset(output_keys).vectors
+
+            # parse keys (allows expressions as keys)
+            output_vectors = [output_vocab.parse(key).v for key in
+                              output_keys]
 
         if default_output_key is None:
             default_output_vector = None
--- a/nengo/spa/assoc_mem.py
+++ b/nengo/spa/assoc_mem.py
@@ -60,9 +60,7 @@ class AssociativeMemory(Module):
             input_keys = input_vocab.keys
             input_vectors = input_vocab.vectors
         else:
-            # parse keys (allows expressions as keys)
-            input_vectors = [input_vocab.parse(key).v for key in
-                             input_keys]
+            input_vectors = [input_vocab.parse(key).v for key in input_keys]
 
         # If output vocabulary is not specified, use input vocabulary
         # (i.e autoassociative memory)
@@ -73,9 +71,7 @@ class AssociativeMemory(Module):
             if output_keys is None:
                 output_keys = input_keys
 
-            # parse keys (allows expressions as keys)
-            output_vectors = [output_vocab.parse(key).v for key in
-                              output_keys]
+            output_vectors = [output_vocab.parse(key).v for key in output_keys]
 
         if default_output_key is None:
             default_output_vector = None
--- a/nengo/spa/tests/test_assoc_mem.py
+++ b/nengo/spa/tests/test_assoc_mem.py
@@ -1,5 +1,7 @@
 import nengo
-from nengo.spa import Vocabulary
+import numpy as np
+from nengo.spa import Vocabulary, Input
+from nengo.spa.utils import similarity
 from nengo.spa.assoc_mem import AssociativeMemory
 
 
@@ -35,3 +37,60 @@ def test_am_spa_interaction(Simulator, seed, rng):
     # Check to see if model builds properly. No functionality test needed
     with Simulator(m):
         pass
+
+
+def test_am_spa_keys_as_expressions(Simulator, plt, seed, rng):
+    """Provide semantic pointer expressions as input and output keys."""
+    D = 64
+
+    vocab_in = Vocabulary(D, rng=rng)
+    vocab_out = Vocabulary(D, rng=rng)
+
+    vocab_in.parse('A+B')
+    vocab_out.parse('C+D')
+
+    in_keys = ['A', 'A*B']
+    out_keys = ['C*D', 'C+D']
+
+    with nengo.spa.SPA(seed=seed) as model:
+        model.am = AssociativeMemory(input_vocab=vocab_in,
+                                     output_vocab=vocab_out,
+                                     input_keys=in_keys,
+                                     output_keys=out_keys)
+
+        model.inp = Input(am=lambda t: 'A' if t < 0.1 else 'A*B')
+
+        in_p = nengo.Probe(model.am.input)
+        out_p = nengo.Probe(model.am.output, synapse=0.03)
+
+    with nengo.Simulator(model) as sim:
+        sim.run(0.2)
+
+    # Specify t ranges
+    t = sim.trange()
+    t_item1 = (t > 0.075) & (t < 0.1)
+    t_item2 = (t > 0.175) & (t < 0.2)
+
+    # Modify vocabularies (for plotting purposes)
+    vocab_in.add(in_keys[1], vocab_in.parse(in_keys[1]).v)
+    vocab_out.add(out_keys[0], vocab_out.parse(out_keys[0]).v)
+
+    plt.subplot(2, 1, 1)
+    plt.plot(t, similarity(sim.data[in_p], vocab_in))
+    plt.ylabel("Input: " + ', '.join(in_keys))
+    plt.legend(vocab_in.keys, loc='best')
+    plt.ylim(top=1.1)
+    plt.subplot(2, 1, 2)
+    plt.plot(t, similarity(sim.data[out_p], vocab_out))
+    plt.plot(t[t_item1], np.ones(t.shape)[t_item1] * 0.9, c='r', lw=2)
+    plt.plot(t[t_item2], np.ones(t.shape)[t_item2] * 0.91, c='g', lw=2)
+    plt.plot(t[t_item2], np.ones(t.shape)[t_item2] * 0.89, c='b', lw=2)
+    plt.ylabel("Output: " + ', '.join(out_keys))
+    plt.legend(vocab_out.keys, loc='best')
+
+    assert np.mean(similarity(sim.data[out_p][t_item1],
+                              vocab_out.parse(out_keys[0]).v,
+                              normalize=True)) > 0.9
+    assert np.mean(similarity(sim.data[out_p][t_item2],
+                              vocab_out.parse(out_keys[1]).v,
+                              normalize=True)) > 0.9
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,9 @@ Release History
 - Added ``y0`` attribute to ``WhiteSignal``, which adjusts the phase of each
   dimension to begin with absolute value closest to ``y0``.
   (`#1064 <https://github.com/nengo/nengo/pull/1064>`_)
+- Allow the `AssociativeMemory` to accept Semantic Pointer expressions as
+  `input_keys` and `output_keys`.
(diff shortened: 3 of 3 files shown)


jgosmann:
Seem like it could be useful. Can you add unit tests for this?

nengo/spa/assoc_mem.py
```diff
@@ -68,17 +68,23 @@ def __init__(self, input_vocab, output_vocab=None,  # noqa: C901
...
+            # parse keys (allows expressions as keys)
+            input_vectors = [input_vocab.parse(key).v for key in
+                             input_keys]
 
         # If output vocabulary is not specified, use input vocabulary
         # (i.e autoassociative memory)
         if output_vocab is None:
             output_vocab = input_vocab
             output_vectors = input_vectors
+            output_keys = input_keys
```
jgosmann: This line seems unecessary?

pblouw:
+1 for this PR - I ran into a similar scenario when I put together my Cogsci submission a while ago.

jgosmann:
Nit-pick: [We prefer commit messages to be in the imperative.](https://docs.google.com/document/d/1YJUEAmrB6y_x0fDUYyRBSIPUNFLLsA3VQNKGRz8T8dI/edit#heading=h.jwjj96dkvmpe)

nengo/spa/assoc_mem.py
xchoo: Don't think so...

jgosmann: It only gets used in the `else` block which initializes it itself if necessary.

jgosmann: And it wasn't there before ..

ikajic: Yeah, looking at it now I think that line shouldn't be there. Unless @xchoo wants to share a bit more of his thoughts :)

 I added it because I thought keys would be used as class attributes/passed to the `network.AssociativeMemory`, but that doesn't seem to be the case. It seems like those keys are only used to create input and output vectors, if I'm not mistaking...

ikajic:
Would this be the right place to add tests: https://github.com/nengo/nengo/blob/master/nengo/spa/tests/test_assoc_mem.py?

jgosmann:
Exactly that place. :dart:

nengo/spa/assoc_mem.py
xchoo: :+1:

ikajic:
I've added a few tests but I would like to discuss how useful they are since the associative memory seems to have a lot of noise when doing this sort of mapping. So I allowed a pretty big error margin -- does this make sense at all? @xchoo @jgosmann

Also, did I just mess up stuff in this branch by merging with the master?  :confused:

jgosmann:
It's the right idea, but the error margin is too large. The proper fix would probably be to improve the associative memory to produce a more precise output instead of overshooting, but not sure how easily that can be achieved.

We could also say that we don't support exact outputs like `0.3 * A` (do you need those?) and thus not test for those. We could still support expressions like `0.5 * A + 0.7 * B` or `A * POS1` and compare the relative strengths of the vectors.

There are a few assertions in that I don't understand, but I mostly skimmed it so far.

> Also, did I just mess up stuff in this branch by merging with the master?

Yes (but it's fixable). Never do merges in the Nengo development model, always to a rebase instead. (It might also make sense to configure `git pull` to do rebase instead of merge, but that would not have helped here.)

xchoo:
Hmmmm. I think we might want to document that the associative memory works by computing the dot product between the input vector and the output vector (i.e. it is expecting the input vector magnitudes to be nominally 1). 

There might be some confusion if this is not stated, because a user might put in, as an input sp phrase, something like '0.5*A', and expect the output of the associative memory to be maximal when the input vector is '0.5*A'. In reality though, the output would be '0.25*A'.

nengo/spa/tests/test_assoc_mem.py
```diff
@@ -35,3 +36,57 @@ def input_func(t):
...
+
+        model.inp = Input(am=lambda t: 'A' if t < .5 else 'B')
+        prob = nengo.Probe(model.am.output, synapse=0.03)
+
+    with Simulator(model) as sim:
+        sim.run(1.0)
+        vec_sim = sim.data[prob]
+
+        final_vecA = vec_sim[100:490].mean(axis=0)
+        simA = np.dot(voc2.vectors, final_vecA)
```
xchoo: I think in this case, we want to use the vector norm of the vector difference (or the cosine angle) as a comparison here. The dot product is not accurate when the input vectors do not have a magnitude of 1 (`-C+D` and `0.3*E` are not guaranteed to have a magnitude of 1).

ikajic:
> [...] exact outputs like 0.3 * A (do you need those?) [..] and compare
the relative strengths of the vectors.

What do you mean by "compare the relative strengths of the vectors"? I
actually had a case where I needed a mapping such as X->0.3*Y, but an
approximation/noisy version was fine for me.

jgosmann:
> What do you mean by "compare the relative strengths of the vectors"?

Normalizing the vectors to length 1 for comparison (i.e. computing the cosine angle; cp. [@xchoo's comment](https://github.com/nengo/nengo/pull/982#discussion_r57260038).

ikajic:
I've updated tests to use cos similarity and simplified them by presenting just one input. What the current tests are doing is:
- testing whether a SP expression can be provided in the input keys
- testing whether a SP expression can be provided in the output keys
- testing for the right output values with the 0.2 error margin

I agree with Xuan that it might not be clear what are expected outputs, so I think documenting how AM works would fit well in the docstring for the AM module (which is currently a one-liner and would contribute to closing #885), but I think it'd make sense to make another PR for documentation.

jgosmann:
> testing whether a SP expression can be provided in the input keys

What would be the use case of providing an expression like `0.9 * A` (vs. just `A`)? It is not clear to me how this would change the operation of the AM and unit tests are also to some degree a usage example. So I might go with an expression like `A + B` or `A * B` where I clearly can see a use case.

You're relying on the order of the pointers in the vocab. It would be nicer (for readability) to access them by name instead. For selecting the probe values `sim.trange() > 0.15` might be more expressive (it doesn't require me to recall what `dt` is).

jgosmann:
> It might also make sense to configure git pull to do rebase instead of merge, but that would not have helped here.

I guess it would have helped because there is now an additional merge commit which looks exactly like it was produced by a git pull doing a merge. To change that setting add
```ini
[pull]
    rebase = true
```
to your `~/.gitconfig`. I'll see if I can get this branch fixed.

jgosmann:
That was pretty easy to fix actually. Just a `git rebase -i master` (and a `git rebase --skip` at one point) did it. Note that I had to force push this branch. To get your local branch in sync don't do a `git pull`, but the following:
1. `git fetch`
2. `git checkout spaassocmem-keys-as-expr` (if not already on that branch)
3. `git reset --hard origin/spaassocmem-keys-as-expr` (WARNING: You lose all local changes and commits on this branch. If you have any local changes/commits you want to keep the process is slightly more complicated. In case you already lost them, because you read this too late, they can still be recovered.)

xchoo:
I've updated the tests. :)

ikajic:
Thanks for the update @xchoo, looks good to me! Btw. Is there a way to get plots when only running these tests? This line: `python -m py.test test_assoc_mem.py --plot` doesn't work for me.

Regarding use-cases: I had an example where I needed keys such as `0.9*A+0.3*B+0.1*C` in the input. I had examples where mappings to the output keys were depending on the scaling for A, B and C.

tbekolay:
To get plots, the command line option is `--plots` (note the plural), and the plots get saved to a directory like `nengo.simulator.plots` in whatever directory you ran `py.test` from.

jgosmann:
> Is there a way to get plots when only running these tests? This line: python -m py.test test_assoc_mem.py --plot doesn't work for me.

It will save the plots in `nengo.simulator.plots`.

> I had an example where I needed keys such as 0.9*A+0.3*B+0.1*C in the input. I had examples where mappings to the output keys were depending on the scaling for A, B and C.

Does this only depend on the relative scaling of A, B, and C? Or is the actual length of the vector important?

ikajic:
> Does this only depend on the relative scaling of A, B, and C? 
Yeah, it was just the scaling.

nengo/spa/tests/test_assoc_mem.py
```diff
@@ -35,3 +37,60 @@ def input_func(t):
...
+
+        in_p = nengo.Probe(model.am.input)
+        out_p = nengo.Probe(model.am.output, synapse=0.03)
+
+    with nengo.Simulator(model) as sim:
+        sim.run(0.2)
+
+    # Specify t ranges
+    t = sim.trange()
+    t_item1 = (t > 0.075) & (t < 0.1)
```
jgosmann: I didn't know that you can use the `&` operator; I always used `np.logical_and`. This is so much more convenient! :)

nengo/spa/tests/test_assoc_mem.py
```diff
@@ -35,3 +37,60 @@ def input_func(t):
...
+    plt.subplot(2, 1, 1)
+    plt.plot(t, similarity(sim.data[in_p], vocab_in))
+    plt.ylabel("Input: " + ', '.join(in_keys))
+    plt.legend(vocab_in.keys, loc='best')
+    plt.ylim(top=1.1)
+    plt.subplot(2, 1, 2)
+    plt.plot(t, similarity(sim.data[out_p], vocab_out))
+    plt.plot(t[t_item1], np.ones(t.shape)[t_item1] * 0.9, c='r', lw=2)
+    plt.plot(t[t_item2], np.ones(t.shape)[t_item2] * 0.91, c='g', lw=2)
+    plt.plot(t[t_item2], np.ones(t.shape)[t_item2] * 0.89, c='b', lw=2)
```
jgosmann: Somewhat of a nit pick: The color might be different, if the user messed with the `matplotlibrc`. To ensure matching colors it might be better to assign the return value of `plt.plot` to a variable (e.g. `lines`) and then use that to get the colors with `lines[i].get_color()`. But it complicates the code a little bit, so I'm not sure.

nengo/spa/assoc_mem.py
```diff
@@ -68,7 +68,9 @@ def __init__(self, input_vocab, output_vocab=None,  # noqa: C901
             input_keys = input_vocab.keys
             input_vectors = input_vocab.vectors
         else:
-            input_vectors = input_vocab.create_subset(input_keys).vectors
+            # parse keys (allows expressions as keys)
```
jgosmann: I don't think this sort of comment is necessary, but whatevs.

jgosmann:
> Yeah, it was just the scaling.

In that case I would argue that it is different from `0.9 * A`.

Made a few comments in the code, but apart from that it LGTM. :cake:

nengo/spa/tests/test_assoc_mem.py
xchoo: Haha. Yeah. This is just with the default color settings. I don't think it's worth the trouble adding the extra code to do the exact color matching (it's not in the plot code in the other tests, and it's really just for debugging the output of the tests).

pblouw:
It would be great to get this into master :)

tcstewar:
> It would be great to get this into master :)

I also ran into a situation today where this would have been handy....  So I'm kicking myself for not reviewing this earlier....

Seanny123:
LGTM.

jgosmann:
* [ ] This should get a changelog entry.

jgosmann:
I'll add the changelog entry and clean up the history.

Seanny123:
Looks even better to me.

jgosmann:
I cleaned the history up, added the changelog entry, and added another fixup commit. If I get an ok on this, I can merge this.

Seanny123:
I am now utterly convinced that this is the best this commit can possibly be.

PR #983, opened 2016-03-16
Reset RC before every test

jgosmann:
[As mentioned in this discussion] (PR #946) the rc params should be reset before every test to prevent test interactions. This PR does that.

In addition it also sets the matplotlib backend to 'AGG' for the tests. Certain other backends are not straight-forward to install in virtualenvs and make problems when runnings tests with tox. This also makes the test backend independent from what the user has set in their config. 'AGG' should be a good default and work on pretty much any system

### DIFF
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -20,8 +20,6 @@ test_seed = 0  # changing this will change seeds for all tests
 
 def pytest_configure(config):
     matplotlib.use('agg')
-    rc.reload_rc([])
-    rc.set('decoder_cache', 'enabled', 'False')
 
 
 @pytest.fixture(scope="session")
@@ -195,6 +193,9 @@ def pytest_generate_tests(metafunc):
 
 
 def pytest_runtest_setup(item):
+    rc.reload_rc([])
+    rc.set('decoder_cache', 'enabled', 'False')
+
     if not hasattr(item, 'obj'):
         return
     for mark, option, message in [
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -3,6 +3,7 @@ import inspect
 import os
 import re
 
+import matplotlib
 import numpy as np
 import pytest
 
@@ -18,6 +19,7 @@ test_seed = 0  # changing this will change seeds for all tests
 
 
 def pytest_configure(config):
+    matplotlib.use('agg')
     rc.reload_rc([])
     rc.set('decoder_cache', 'enabled', 'False')
 
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -195,6 +195,7 @@ def pytest_generate_tests(metafunc):
 def pytest_runtest_setup(item):
     rc.reload_rc([])
     rc.set('decoder_cache', 'enabled', 'False')
+    rc.set('exceptions', 'simplified', 'False')
 
     if not hasattr(item, 'obj'):
         return
--- a/nengo/utils/testing.py
+++ b/nengo/utils/testing.py
@@ -179,7 +179,7 @@ class WarningCatcher(object):
     def __enter__(self):
         self.catcher = warnings.catch_warnings(record=True)
         self.record = self.catcher.__enter__()
-        warnings.simplefilter('always', append=True)
+        warnings.simplefilter('always')
 
     def __exit__(self, type, value, traceback):
         self.catcher.__exit__(type, value, traceback)
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,6 @@ install:
 
 # Run the tests
 script:
-  - "echo 'backend : Agg' > matplotlibrc"
   - if [[ -n $SETUP_CMD ]]; then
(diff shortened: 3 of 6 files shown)


nengo/conftest.py
```diff
@@ -18,8 +19,7 @@
 
 
 def pytest_configure(config):
-    rc.reload_rc([])
-    rc.set('decoder_cache', 'enabled', 'False')
+    matplotlib.use('AGG')
```
tbekolay: In Matplotlib's list of backends (`matplotlib.rcsetup.non_interactive_bk`) it's listed as `u'agg'` (no capitals) so I would favor all lowercase here to be consistent.

tbekolay:
I'm going to add a commit to remove some stuff I did in the `exceptions` notebook to manage `rc` state, but aside from that and the nitpick inline, this looks great!

jgosmann:
Should we also turn off simplified exceptions in the tests? Might make debugging easier ...

hunse:
> Should we also turn off simplified exceptions in the tests? Might make debugging easier ...

That's a good idea.

tbekolay:
Added a commit to remove the RC resetting in `exceptions.ipynb`. I think this looks good to me now; @hunse do you want to look at this? Or @drasmuss, whoever gets to it first!

nengo/utils/testing.py
```diff
@@ -179,7 +179,7 @@ class WarningCatcher(object):
     def __enter__(self):
         self.catcher = warnings.catch_warnings(record=True)
         self.record = self.catcher.__enter__()
-        warnings.simplefilter('always', append=True)
+        warnings.simplefilter('always')
```
hunse: There's now `pytest.warns`, which does the same thing as `pytest.raises` but for warnings. And it does this properly. Maybe we should switch to using that at some point. We would need to restrict ourselves to newer `pytest` though. For now, this is fine.

nengo/conftest.py
```diff
@@ -18,8 +19,7 @@
 
 
 def pytest_configure(config):
-    rc.reload_rc([])
-    rc.set('decoder_cache', 'enabled', 'False')
+    matplotlib.use('agg')
```
hunse: Is this going to affect users running the tests, though? What if users have their matplotlib set up to use a different backend?

hunse:
Other than that question about setting the matplotlib backend, this all looks good.

nengo/conftest.py
jgosmann: Users having set their backend to something else is exactly the reason why I think that this is necessary. For example, on Linux my backend is set to 'Qt4Agg' right now, but `tox` won't install the required modules (e.g., pyside) and fail. On OS X, I have a similar problem where a framework install of Python is needed for the backend which is not supported by the virtualenv (without manual intervention).

Apart from that, the tests should produce reproducible results independent of what some user configuration file says (though, other matplotlib settings probably still have an effect). 'agg' should work on (almost?) all systems for the file output we're producing.

hunse: > but `tox` won't install the required modules

I consider it the user's job to set up their matplotlib properly.

> 'agg' should work on (almost?) all systems for the file output we're producing.

Sometimes, when I'm testing, I use `plt.show` to actually pop up a plot window. But I guess that's considered a hack, and not something we need to support. Also, I can still hack it by just importing matplotlib myself. I would like to be pretty sure that 'agg' will work on almost all systems for file output, but if you have good reason to believe that I'm fine with this. I don't know enough about it myself to say.

jgosmann: > I consider it the user's job to set up their matplotlib properly.

My matplotlib is properly setup on all my systems. Afaik there is no way to tell tox to install additional modules into the venvs it creates given specific matplotlib backends. Also there is currently no way to solve the framework install issue with tox. So the only two options are: Set a defined default backend or require the user to always edit the config to run tests with `tox`.

> Sometimes, when I'm testing, I use plt.show to actually pop up a plot window.

That might no longer work with this because 'agg' is not an interactive renderer (i.e. can only plot to a file).



> I would like to be pretty sure that 'agg' will work on almost all systems for file output,

The matplotlib documentation says "The canonical renderer for user interfaces is Agg which uses the Anti-Grain Geometry C++ library to make a raster (pixel) image of the figure". That's not about file output, but at least better than nothing

I would also be ok with changing the backend to 'pdf' instead of 'agg' because we save PDFs anyways.

hunse: I've never tried the 'pdf' backend, so I'd be wary of changing to it right before a release unless we have a good reason. But if someone wants to do some more research, this could be a good option down the road. I'm not sure what the pros and cons would be.

hunse:
I'm convinced that matplotlib's 'agg' backend should work almost all the time for what we want to do, so I'm fine with this as-is.

nengo/utils/testing.py
tbekolay: We don't require users to install pytest, so upping the required version only affect developers who should have no problems upgrading. So I'm cool upping it, but yes, in another PR :)

Issue #977, opened 2016-03-08
Copying Nengo objects


tbekolay:
From today's meeting, it would be nice to be able to copy Nengo object's cleanly. I think in general we want shallow copies (i.e., don't deepcopy the parameter values); deepcopies aren't necessary since everything we do to make sure we can make multiple simulators also means we don't need to do deepcopies.

Todo:
- [x] Implement `__getstate__` `__setstate__` (see [pickle-config branch](https://github.com/nengo/nengo/tree/pickle-config); ensure we don't duplicate code with `FrozenObject` which does some copy stuff)
- [x] Implement `.copy` method
  - [x] Include an `add_to_network` arg on the copy method that adds or doesn't add to current `Network.context`
- [x] Make sure it plays well with Python's `copy` module

tbekolay:
Also, related to #941.

jgosmann:
I started implementing this and it seems to be possible in a quite general manner. The added `copy` method on `NengoObject`s can take an `add_to_network` argument, but neither the Python `copy.copy` function can nor the unpickling functions. Should we add the object in those cases to the network or shouldn't we? Or maybe only if we are in the context of a network?

Probably we don't want to add the object always to a network (otherwise unpickling outside of a network would not be possible because it would raise an exception).

hunse:
Yeah, I would say don't add it when unpickling. 

Maybe we shouldn't allow `copy.copy` at all? We could just have a NotImplementedError that tells people to use our `copy` method. That would make things more straightforward, and I don't think it really makes anything inconvenient, either.

jgosmann:
I'm using `__getstate__` and `__setstate__`. I have to, to support pickling/unpickling and it gives me copying basically for free. But because of that it is not possible to raise `NotImplementedError` for the Python `copy` function. Furthermore, I'm using exactly that function in our `copy` implementation.

hunse:
What about something like [this](https://github.com/nengo/nengo/commit/ff00ca6d8cc4a43cc1f55dc14090fd12446fff2c)?

jgosmann:
Isn't this using a private method in the `copy` module?

hunse:
Yep! Probably a bit strange, but I basically just took out the parts of their code that do copying if there's no `__copy__` function. Though we should check if this works in 3.x. Also, maybe better to just write all our own code to do it using `__getstate__` and `__setstate__`, which shouldn't be too hard. I just threw it together quick as a proof of concept.

[...]

PR #984, opened 2016-03-16
Copy

jgosmann:
Addresses #977.

Nengo objects should be copyable and pickable now. There might be some weird effects for objects referencing other objects (e.g., connections). I might have to think about those and whether that works as expected (though, technically it works in some way).

Still todo:

* [ ] Networks
* [ ] FrozenObjects: Process, LearningRule, Synapse (also check for potential code duplication)

### DIFF
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -7,8 +7,8 @@ import numpy as np
 from nengo.config import Config
 from nengo.exceptions import ValidationError
 from nengo.params import (
-    Default, is_param, IntParam, Parameter, params, StringParam,
-    Unconfigurable)
+    CopyableObject, Default, is_param, IntParam, Parameter, params,
+    StringParam, Unconfigurable)
 from nengo.rc import rc
 from nengo.utils.compat import is_integer, reraise, with_metaclass
 
@@ -34,7 +34,7 @@ class NetworkMember(type):
         return inst
 
 
-class NengoObject(with_metaclass(NetworkMember)):
+class NengoObject(with_metaclass(NetworkMember, CopyableObject)):
     """A base class for Nengo objects.
 
     This defines some functions that the Network requires
@@ -46,6 +46,7 @@ class NengoObject(with_metaclass(NetworkMember)):
     seed = IntParam('seed', default=None, optional=True)
 
     def __init__(self, label, seed):
+        super(NengoObject, self).__init__()
         self.label = label
         self.seed = seed
 
@@ -83,23 +84,12 @@ class NengoObject(with_metaclass(NetworkMember)):
             super(NengoObject, self).__setattr__(name, val)
 
     def __getstate__(self):
-        state = {}
-
-        for attr in params(self):
-            param = getattr(self.__class__, attr)
-            if self in param:
-                state[attr] = getattr(self, attr)
-
-        for attr in vars(self):
-            if attr == '_initialized':
-                continue
-            state[attr] = getattr(self, attr)
-
+        state = super(NengoObject, self).__getstate__()
+        del state['_initialized']
         return state
 
     def __setstate__(self, state):
-        for k, v in state.items():
-            setattr(self, k, v)
+        super(NengoObject, self).__setstate__(state)
         setattr(self, '_initialized', True)
 
     def copy(self, add_to_container=True):
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -1,3 +1,4 @@
+from copy import copy
 import sys
 import warnings
 
@@ -6,7 +7,8 @@ import numpy as np
 from nengo.config import Config
 from nengo.exceptions import ValidationError
 from nengo.params import (
-    Default, is_param, IntParam, Parameter, StringParam, Unconfigurable)
+    Default, is_param, IntParam, Parameter, params, StringParam,
+    Unconfigurable)
 from nengo.rc import rc
 from nengo.utils.compat import is_integer, reraise, with_metaclass
 
@@ -81,10 +83,31 @@ class NengoObject(with_metaclass(NetworkMember)):
             super(NengoObject, self).__setattr__(name, val)
 
     def __getstate__(self):
-        raise NotImplementedError("Nengo objects do not support pickling")
+        state = {}
+
+        for attr in params(self):
+            param = getattr(self.__class__, attr)
+            if self in param:
+                state[attr] = getattr(self, attr)
+
+        for attr in vars(self):
+            if attr == '_initialized':
+                continue
+            state[attr] = getattr(self, attr)
+
+        return state
 
     def __setstate__(self, state):
-        raise NotImplementedError("Nengo objects do not support pickling")
+        for k, v in state.items():
+            setattr(self, k, v)
+        setattr(self, '_initialized', True)
+
+    def copy(self, add_to_container=True):
+        c = copy(self)
+        if add_to_container:
+            from nengo.network import Network
+            Network.add(c)
+        return c
 
     @classmethod
     def param_list(cls):
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -303,6 +303,10 @@ class Connection(NengoObject):
         since="v2.1.0",
         url="https://github.com/nengo/nengo/issues/632#issuecomment-71663849")
 
+    _param_init_order = [
+        'pre', 'post', 'synapse', 'transform', 'eval_points', 'function_info',
+        'solver', 'learning_rule_type']
+
     def __init__(self, pre, post, synapse=Default, transform=Default,
                  solver=Default, learning_rule_type=Default, function=Default,
                  eval_points=Default, scale_eval_points=Default,
--- a/nengo/ensemble.py
+++ b/nengo/ensemble.py
@@ -98,6 +98,15 @@ class Ensemble(NengoObject):
         self.noise = noise
         self._neurons = Neurons(self)
 
+    def __getstate__(self):
+        state = super(Ensemble, self).__getstate__()
+        del state['_neurons']
+        return state
+
+    def __setstate__(self, state):
+        state['_neurons'] = Neurons(self)
+        super(Ensemble, self).__setstate__(state)
+
     def __getitem__(self, key):
         return ObjView(self, key)
 
--- a/nengo/params.py
+++ b/nengo/params.py
@@ -29,7 +29,8 @@ def is_param(obj):
 
 def params(obj):
     """Return list with the names of all parameters of an object."""
-    return [name for name, v in vars(obj.__class__).items() if is_param(v)]
+    return [name for name, v in vars(obj.__class__).items()
+            if is_param(v) and not isinstance(v, ObsoleteParam)]
 
 
 class Parameter(object):
@@ -367,9 +368,12 @@ FunctionInfo = collections.namedtuple('FunctionInfo', ['function', 'size'])
 
 class FunctionParam(Parameter):
     def __set__(self, instance, function):
-        size = (self.determine_size(instance, function)
-                if callable(function) else None)
-        function_info = FunctionInfo(function=function, size=size)
+        if isinstance(function, FunctionInfo):
+            function_info = function
+        else:
+            size = (self.determine_size(instance, function)
+                    if callable(function) else None)
+            function_info = FunctionInfo(function=function, size=size)
         super(FunctionParam, self).__set__(instance, function_info)
 
     def function_args(self, instance, function):
--- a/nengo/params.py
+++ b/nengo/params.py
@@ -27,6 +27,11 @@ def is_param(obj):
     return isinstance(obj, Parameter)
 
 
+def params(obj):
+    """Return list with the names of all parameters of an object."""
+    return [name for name, v in vars(obj.__class__).items() if is_param(v)]
+
+
 class Parameter(object):
     """Simple descriptor for storing configuration parameters.
 
--- a/nengo/params.py
+++ b/nengo/params.py
@@ -433,3 +433,44 @@ class FrozenObject(object):
         for k in self._paramdict:
             setattr(self, k, state.pop(k))
         self.__dict__.update(state)
+
+
+class CopyableObject(object):
+    """Mixin to allow to create copies of instances of classes with parameters.
+
+    You can add a list of strings as class attribute _param_init_order to
+    declare the order in which parameters have to be initialized. Missing
+    parameters will be initialized last in an undefined order.
+    """
+
+    def __getstate__(self):
+        sp = super(CopyableObject, self)
+        if hasattr(sp, '__getstate__'):
+            state = sp.__getstate__()
+        else:
+            state = dict(self.__dict__)
+
+        for attr in params(self):
+            param = getattr(self.__class__, attr)
+            if self in param:
+                state[attr] = getattr(self, attr)
+
+        return state
+
+    def __setstate__(self, state):
+        if hasattr(self.__class__, '_param_init_order'):
+            for attr in getattr(self.__class__, '_param_init_order'):
+                setattr(self, attr, state[attr])
+                del state[attr]
+
+        for attr in params(self):
+            if attr in state:
+                setattr(self, attr, state[attr])
+                del state[attr]
+
+        sp = super(CopyableObject, self)
+        if hasattr(sp, '__setstate__'):
+            sp.__setstate__(state)
+        else:
+            for k, v in state.items():
+                setattr(self, k, v)
--- a/nengo/tests/test_copy.py
+++ b/nengo/tests/test_copy.py
@@ -1,3 +1,4 @@
+import numpy as np
 import pytest
 
 import nengo
@@ -80,9 +81,12 @@ class TestCopyProbe(CopyTest, PickleTest):
         return p
 
 
-# Probe
-# Ensemble
-# Node
+class TestCopyNode(CopyTest, PickleTest):
+    def create_original(self):
+        with nengo.Network() as _:
+            n = nengo.Node(np.min, size_in=2, size_out=2)
+        return n
+
 # Connection
 
 # Network
--- a/nengo/tests/test_copy.py
+++ b/nengo/tests/test_copy.py
@@ -3,7 +3,17 @@ import pytest
 import nengo
 from nengo.exceptions import NetworkContextError
 from nengo.params import params
+from nengo.utils.compat import pickle
 
+
+def assert_is_copy(copy, original):
+    assert copy is not original  # ensures separate parameters
+    for param in params(copy):
+        if isinstance(getattr(copy, param), nengo.solvers.Solver):
+            continue  # FIXME
+        if isinstance(getattr(copy, param), nengo.base.NengoObject):
+            continue  # FIXME
+        assert getattr(copy, param) == getattr(original, param)
 
 
 class CopyTest(object):
@@ -17,7 +27,7 @@ class CopyTest(object):
             copy = original.copy()
         assert copy in model.all_objects
 
-        self.assert_is_copy(copy, original)
+        assert_is_copy(copy, original)
 
     def test_copy_in_network_without_adding(self):
         original = self.create_original()
@@ -26,7 +36,7 @@ class CopyTest(object):
             copy = original.copy(add_to_container=False)
         assert copy not in model.all_objects
 
-        self.assert_is_copy(copy, original)
+        assert_is_copy(copy, original)
 
     def test_copy_outside_network(self):
         original = self.create_original()
@@ -36,16 +46,20 @@ class CopyTest(object):
     def test_copy_outside_network_without_adding(self):
         original = self.create_original()
         copy = original.copy(add_to_container=False)
-        self.assert_is_copy(copy, original)
-
-    @staticmethod
-    def assert_is_copy(copy, original):
-        assert copy is not original  # ensures separate parameters
-        for param in params(copy):
-            assert getattr(copy, param) == getattr(original, param)
+        assert_is_copy(copy, original)
 
 
-class TestCopyEnsemble(CopyTest):
+class PickleTest(object):
+    def create_original(self):
+        raise NotImplementedError()
+
+    def test_pickle_roundtrip(self):
+        original = self.create_original()
+        copy = pickle.loads(pickle.dumps(original))
+        assert_is_copy(copy, original)
+
+
+class TestCopyEnsemble(CopyTest, PickleTest):
     def create_original(self):
         with nengo.Network() as _:
             e = nengo.Ensemble(10, 1, radius=2.)
@@ -56,6 +70,14 @@ class TestCopyEnsemble(CopyTest):
         copy = original.copy(add_to_container=False)
         assert original.neurons.ensemble is original
         assert copy.neurons.ensemble is copy
+
+
+class TestCopyProbe(CopyTest, PickleTest):
+    def create_original(self):
+        with nengo.Network() as _:
+            e = nengo.Ensemble(10, 1)
+            p = nengo.Probe(e, synapse=0.01)
+        return p
 
 
 # Probe
--- a/nengo/tests/test_copy.py
+++ b/nengo/tests/test_copy.py
@@ -62,7 +62,7 @@ class PickleTest(object):
 
 class TestCopyEnsemble(CopyTest, PickleTest):
     def create_original(self):
-        with nengo.Network() as _:
+        with nengo.Network():
             e = nengo.Ensemble(10, 1, radius=2.)
         return e
 
@@ -75,7 +75,7 @@ class TestCopyEnsemble(CopyTest, PickleTest):
 
 class TestCopyProbe(CopyTest, PickleTest):
     def create_original(self):
-        with nengo.Network() as _:
+        with nengo.Network():
             e = nengo.Ensemble(10, 1)
             p = nengo.Probe(e, synapse=0.01)
         return p
@@ -83,11 +83,17 @@ class TestCopyProbe(CopyTest, PickleTest):
 
 class TestCopyNode(CopyTest, PickleTest):
     def create_original(self):
-        with nengo.Network() as _:
+        with nengo.Network():
             n = nengo.Node(np.min, size_in=2, size_out=2)
         return n
 
-# Connection
+class TestCopyConnection(CopyTest, PickleTest):
+    def create_original(self):
+        with nengo.Network(self):
+            e1 = nengo.Ensemble(10, 1)
+            e2 = nengo.Ensemble(10, 1)
+            c = nengo.Connection(e1, e2, transform=2.)
+        return c
 
 # Network
 
--- a/nengo/tests/test_copy.py
+++ b/nengo/tests/test_copy.py
@@ -1,0 +1,73 @@
+import pytest
+
+import nengo
+from nengo.exceptions import NetworkContextError
+from nengo.params import params
+
+
+
+class CopyTest(object):
+    def create_original(self):
+        raise NotImplementedError()
+
+    def test_copy_in_network(self):
+        original = self.create_original()
+
+        with nengo.Network() as model:
+            copy = original.copy()
+        assert copy in model.all_objects
+
+        self.assert_is_copy(copy, original)
+
+    def test_copy_in_network_without_adding(self):
+        original = self.create_original()
+
+        with nengo.Network() as model:
+            copy = original.copy(add_to_container=False)
+        assert copy not in model.all_objects
+
+        self.assert_is_copy(copy, original)
+
+    def test_copy_outside_network(self):
+        original = self.create_original()
+        with pytest.raises(NetworkContextError):
+            original.copy()
+
+    def test_copy_outside_network_without_adding(self):
+        original = self.create_original()
+        copy = original.copy(add_to_container=False)
+        self.assert_is_copy(copy, original)
+
+    @staticmethod
+    def assert_is_copy(copy, original):
+        assert copy is not original  # ensures separate parameters
+        for param in params(copy):
+            assert getattr(copy, param) == getattr(original, param)
+
+
+class TestCopyEnsemble(CopyTest):
+    def create_original(self):
+        with nengo.Network() as _:
+            e = nengo.Ensemble(10, 1, radius=2.)
+        return e
+
+    def test_neurons_reference_copy(self):
+        original = self.create_original()
+        copy = original.copy(add_to_container=False)
+        assert original.neurons.ensemble is original
+        assert copy.neurons.ensemble is copy
+
+
+# Probe
+# Ensemble
+# Node
+# Connection
+
+# Network
+
+# Process
+# Learning rule
+# Synapse
+
+
+# copy, copy with add to network, pickle and unpickle
--- a/nengo/tests/test_params.py
+++ b/nengo/tests/test_params.py
@@ -283,6 +283,7 @@ def test_params():
     class Test(object):
         p1 = params.IntParam('p1')
         p2 = params.IntParam('p2')
+        obsolete = params.ObsoleteParam('obsolete', 'not included in params')
 
     assert set(params.params(Test())) == {'p1', 'p2'}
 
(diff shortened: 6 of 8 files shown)


hunse:
FrozenObjects already have __getstate__ and __setstate__, so hopefully getting them working is as easy as just calling `copy.copy` in our own `copy` function.

jgosmann:
I'm slightly surprised by the two failing test runs ... as the other test envs pass it is probably something going wrong with the warnings filtering?

jgosmann:
:angry: The debug commit made the tests pass ... what the hell?

jgosmann:
I got a bit closer to the issue: `Network.context` is empty ... despite being in a `with network` block. Just not sure why it is empty.

jgosmann:
@tbekolay Is py.test running tests in parallel on Travis-CI?

tbekolay:
Yes, it runs with -n 2

jgosmann:
I'm wondering whether threads are influencing each othen via `Network.context` ...

jgosmann:
At least I'm finally able to reproduce the problem locally.

jgosmann:
If I add debug code, it makes the tests pass, but I'm pretty convinced that we do not handle `Network.context` and potentially configs correctly in threads. Fixing that is however in the scope of a different PR.

nengo/tests/test_copy.py
```diff
@@ -0,0 +1,124 @@
...
+
+class TestCopyConnection(CopyTest, PickleTest):
+    def create_original(self):
+        with nengo.Network(self):
+            e1 = nengo.Ensemble(10, 1)
+            e2 = nengo.Ensemble(10, 1)
+            c = nengo.Connection(e1, e2, transform=2.)
+        return c
+
+# Network
```
Seanny123: Leftover comments that should be deleted?

nengo/tests/test_copy.py
```diff
@@ -0,0 +1,228 @@
...
+@pytest.fixture(scope='module', autouse=True)
+def error_on_not_added_to_network_warning(request):
+    catcher = warnings.catch_warnings(record=True)
+    catcher.__enter__()
+    warnings.simplefilter(
+        'error', category=NotAddedToNetworkWarning)
+    request.addfinalizer(lambda: catcher.__exit__(None, None, None))
+
+
+def assert_is_copy(cp, original):
```
Seanny123: This might just be me being a noob at `pytest`, but how are these parameters defined? I thought they had to be `pytest` fixtures? I see that `original` is paramterized at the bottom of the file.

nengo/tests/test_copy.py
```diff
@@ -0,0 +1,228 @@
...
+        if isinstance(param_inst, nengo.solvers.Solver) or isinstance(
+                param_inst, nengo.base.NengoObject):
+            assert_is_copy(param_inst, getattr(original, param))
+        elif is_array_like(param_inst):
+            assert np.all(param_inst == getattr(original, param))
+        else:
+            assert param_inst == getattr(original, param)
+
+
+class CopyTest(object):
```
Seanny123: I think I'm asking for a silly comment again, but I'd just like to put this out there. A comment indicating that this is a base class that is inherited for testing copying specific object might be helpful? I just spend a couple of minutes trying to figure out how this works, but are we just assuming that whoever is reading these tests is better at PyTest than I am?

nengo/tests/test_copy.py
Seanny123: Nvm. I just figured out that all utility functions aren't appended with the word `test`.

jgosmann: It's just a utility function because it doesn't start with `test_`

On May 24, 2016 6:36:33 PM GMT+01:00, Sean Aubin <notifications@github.com> wrote:
>> +from nengo.exceptions import NetworkContextError,
>NotAddedToNetworkWarning
>> +from nengo.params import params
>> +from nengo.utils.compat import is_array_like, pickle
>> +from nengo.utils.testing import warns
>> +
>> +
>> +@pytest.fixture(scope='module', autouse=True)
>> +def error_on_not_added_to_network_warning(request):
>> +    catcher = warnings.catch_warnings(record=True)
>> +    catcher.__enter__()
>> +    warnings.simplefilter(
>> +        'error', category=NotAddedToNetworkWarning)
>> +    request.addfinalizer(lambda: catcher.__exit__(None, None, None))
>> +
>> +
>> +def assert_is_copy(cp, original):
>
>This might just be me being a noob at `pytest`, but how are these
>parameters defined? I thought they had to be `pytest` fixtures? I see
>that `original` is paramterized at the bottom of the file. Or are these
>just utility functions? What's the distinction between a utility
>function and a test?
>
>---
>You are receiving this because you were assigned.
>https://github.com/nengo/nengo/pull/984/files/465d6a9e7b01caece896adee1916793911f9e5d2#r64436948

-- 

Seanny123:
LGTM, other than the few test cases that I had a hard time reading due to their complexity.

jgosmann:
Added the requested documentation.

jgosmann:
doing the rebase now

jgosmann:
done

nengo/utils/least_squares_solvers.py
```diff
@@ -69,6 +69,10 @@ def __call__(self, A, y, sigma, rng=None):
         info = {'rmses': rmses(A, x, y)}
         return x, info
 
+    def __eq__(self, other):
```
hunse: This is a FrozenObject, so I don't think this should be necessary. Is this just a holdover from before it was a FrozenObject?

nengo/tests/test_params.py
```diff
@@ -275,3 +277,25 @@ class Test(object):
     # Not OK: not a function
     with pytest.raises(ValidationError):
         inst.fp = 0
+
+
+def test_params():
```
hunse: Maybe call this `test_obsolete_excluded` or something? `test_params` seems to general.

nengo/base.py
```diff
@@ -54,14 +56,35 @@ class NengoObject(with_metaclass(NetworkMember, SupportDefaultsMixin)):
...
-        raise NotImplementedError("Nengo objects do not support pickling")
+        state = super(NengoObject, self).__getstate__()
+        del state['_initialized']
+        return state
 
     def __setstate__(self, state):
-        raise NotImplementedError("Nengo objects do not support pickling")
+        from nengo.network import Network
+        super(NengoObject, self).__setstate__(state)
+        setattr(self, '_initialized', True)
```
hunse: Why not `self._initialized = True`?

nengo/base.py
```diff
@@ -54,14 +56,35 @@ class NengoObject(with_metaclass(NetworkMember, SupportDefaultsMixin)):
...
+        if len(Network.context) > 0:
+            warnings.warn(
+                "{obj} was not added to the network. When copying objects, "
+                "use the copy method on the object instead of Python's copy "
+                "module. When unpickling objects, they have to be added to "
+                "networks manually.".format(obj=self),
+                NotAddedToNetworkWarning)
```
hunse: I don't like the fact that unpickled objects don't get added to networks automatically. It would be nice to be able to pickle a whole complex network structure and unpickle it. That doesn't have to be in this PR, just sayin'.

nengo/params.py
```diff
@@ -416,6 +426,47 @@ def validate(self, instance, function_info):
...
+
+        for attr in params(self):
+            param = getattr(self.__class__, attr)
+            if self in param:
+                state[attr] = getattr(self, attr)
+
+        return state
+
+    def __setstate__(self, state):
+        if hasattr(self.__class__, '_param_init_order'):
```
hunse: Not a big deal, but can we define this in `CopyableObject` so that we don't have to do the `hasattr` here?

nengo/utils/least_squares_solvers.py
jgosmann: Probably.

nengo/base.py
jgosmann: I think the formulation might be ambiguous. If you pickle and unpickle (or copy) a whole network, you should get back the complete network. What is not respected is the current context in which you are unpickling or copying. If you use the `copy` method on the objet or network, it will be respected (and you have the usual `add_to_container` to change this behaviour). The reason that this is not supported by Python's `copy` function:
* There is no way to pass an `add_to_container` argument.
* Not adding the object to the active network allows to use Python's `copy` in the object's `copy` function for implementing both behaviours. Otherwise we would implement all of `copy` ourselves.

nengo/params.py
```diff
@@ -416,6 +426,47 @@ def validate(self, instance, function_info):
         super(FunctionParam, self).validate(instance, function)
 
 
+class CopyableObject(object):
```
hunse: I'm not sure if this should be a mixin class. We're just using it in one place right now. If we just build this stuff right into `NengoObject`, then we don't have to do the `hasattr` checks for `__getstate__` and `__setstate__` below, and I think it's clearer.

This may at some point be reuseable functionality that we want to abstract out like this. But until we find a potential second use, I'd just build this into `NengoObject`.

nengo/params.py
jgosmann: I believe Python doesn't work that way? I would be happy to get rid of it if possible.

nengo/base.py
hunse: Ahh, okay, I see. The thing I was really worried about---pickling and unpickling complex networks---seems to work fine in the current implementation, so this isn't the problem I thought it was.

hunse:
I've put my proposal in [`copy2`](https://github.com/nengo/nengo/tree/copy2). It has the changes I suggested here.

nengo/params.py
```diff
@@ -27,6 +27,13 @@ def is_param(obj):
     return isinstance(obj, Parameter)
 
 
+def params(obj):
```
hunse: I would suggest `get_params` for the name of this function. `params` is too easily confusable with the module.

nengo/params.py
jgosmann: > I'm not sure if this should be a mixin class.

It encapsulates the copy related code from other `NengoObject` instead of mixing it all in one class (and same functions). To me that seems a bit clearer, but I think I'm ok with it either way as long as we don't have any other classes with parameters that should support copies.

tbekolay:
@jgosmann any thoughts on 820ea18e4a32f5c77a340d, the unique commit in the `copy2` branch?

jgosmann:
In addition to my comment above why I prefer the mixin:

> It encapsulates the copy related code from other NengoObject instead of mixing it all in one class (and same functions). To me that seems a bit clearer, but I think I'm ok with it either way as long as we don't have any other classes with parameters that should support copies.

I'll add the following points:

* The documentation for `_param_init_order` is lost (though easily fixable).
* Other object with parameters that should support copying/pickling need to reimplement this. We don't have any such objects right now, so for the time being I'm fine either way, but that could change in the future. So it would be nice, to keep `copy` version around to pull it out when needed.

jgosmann:
Can we get to a decision whether to use `copy` or `copy2` so that I can figure out why this has the `needs changes` label and do those changes to get this merged in anticipation of spaopt-v5?

I stated the reasons why I prefer in the previous comment, but if everyone else prefers `copy2` I'm fine with that for the moment.

hunse:
I looked at this again, since I couldn't remember the difference between the two. I'm in favour of `copy2` because it is simpler and clearer. If you want to know how a `NengoObject` is copied, you can just read the `__getstate__` and `__setstate__` functions, rather than have to try to figure out which one of the three subclasses of `NengoObject` is implementing these functions, and track down the relevant code. If in the future we have other objects that are copyable and we want to implement it then, I'd be fine with that, but right now it's over-engineering. Keep it simple.

The other thing I don't like about `CopyableObject` is that we have objects that are copyable that don't use it, namely `Network`. I find this is a little confusing. It also points to the fact that if we add other copyable objects in the future, there are no guarantees that the `CopyableObject` interface will work for them.

tbekolay:
I'm also +1 on `copy2`. @hunse explained why better than I can :rainbow: :octopus:

jgosmann:
Merged the commits from `copy2` and fixed remaining problems. This needs a rebase, but @tbekolay is assigned, so I cannot rewrite the history.

tbekolay:
My bad, go ahead and rebase and I'll look at this post-rebase.

jgosmann:
rebased

nengo/config.py
```diff
@@ -84,6 +84,10 @@ def __repr__(self):
         return "<%s[%s]{%s}>" % (type(self).__name__,
                                  self._configures.__name__, ", ".join(params))
 
+    def __setstate__(self, state):
+        for k, v in state.items():
+            setattr(self, k, v)
```
hunse: Just wondering why this is necessary. Isn't this basically the default implementation? Also, why do we have to implement this but not `__getstate__`? That seems atypical.

hunse:
Glanced over this one more time, and just had the one minor question. Otherwise LGTM.

nengo/config.py
jgosmann: Without it a bunch of tests fail:
```
________________________________________________ TestCopyNetwork.test_pickle_roundtrip _________________________________________________
[gw3] darwin -- Python 2.7.12 /Volumes/Home/blubb/.virtualenvs/nengodev/Python.app/Contents/MacOS/Python
self = <nengo.tests.test_copy.TestCopyNetwork object at 0x1112b6650>

    def test_pickle_roundtrip(self):
        original = self.create_original()
>       cp = pickle.loads(pickle.dumps(original))

nengo/tests/test_copy.py:121:
nengo/config.py:50: in __getattr__
    return self.get_param(key).get_default(self)
nengo/config.py:100: in get_param
    if key in self._extraparams:
nengo/config.py:50: in __getattr__
    return self.get_param(key).get_default(self)
nengo/config.py:100: in get_param
    if key in self._extraparams:
nengo/config.py:50: in __getattr__
    return self.get_param(key).get_default(self)
!!! Recursion detected (same locals & position)
__________________________________________ TestCopyNetwork.test_unpickling_warning_in_network __________________________________________
[gw3] darwin -- Python 2.7.12 /Volumes/Home/blubb/.virtualenvs/nengodev/Python.app/Contents/MacOS/Python
self = <nengo.tests.test_copy.TestCopyNetwork object at 0x1112b61d0>

    def test_unpickling_warning_in_network(self):
        original = self.create_original()
        pkl = pickle.dumps(original)
        with nengo.Network():
            with warns(NotAddedToNetworkWarning):
>               pickle.loads(pkl)

nengo/tests/test_copy.py:129:

self = <nengo.utils.testing.warns object at 0x112588110>
type = <type 'exceptions.RuntimeError'>
value = RuntimeError('maximum recursion depth exceeded',)
traceback = <traceback object at 0x1127533f8>

nengo/utils/testing.py:205: Failed
_________________________________________ TestCopyEnsemble.test_unpickling_warning_in_network __________________________________________
[gw1] darwin -- Python 2.7.12 /Volumes/Home/blubb/.virtualenvs/nengodev/Python.app/Contents/MacOS/Python
self = <nengo.tests.test_copy.TestCopyEnsemble object at 0x10b73e450>

    def test_unpickling_warning_in_network(self):
        original = self.create_original()
        pkl = pickle.dumps(original)
        with nengo.Network():
            with warns(NotAddedToNetworkWarning):
>               pickle.loads(pkl)

nengo/tests/test_copy.py:129:

self = <nengo.utils.testing.warns object at 0x10e9b6090>, type = None

nengo/utils/testing.py:205: Failed
========================================== 3 failed, 762 passed, 45 skipped in 41.83 seconds ==========================================
```

But I have to dig dipper to tell you the reason.

jgosmann: It might be because `__setattr__` is also overwritten.

jgosmann: So this is was happening: `pickle` checks whether `ClassParams` defines it's own `__setstate__` by calling `__getattribute__`. Now, if we define a `__setstate__` with default behaviour, it will find `__setstate__` and return it. But if we don't, it will delegate the request to `__getattr__` which is overwritten in `ClassParams` and leads to an infinite loop because it calls `get_param` which uses `getattr` (not sure why `self._configures` seems to end up as self).

So an alternate solution to defining `__setattr__` would be to check whether the `key` in `__getattr__` is `__setstate__` or starts with underscores or something similar and just invoke the super `__getattr__` in that case.

It also seems slightly weird to me that `get_param` and `__getattr__` invoke each other ...

jgosmann: Just realized, it's not the `getattr(self._configures, key)` line in `get_param` that leads to the loop, but three lines above. The check `if key in self._extraparams` needs to look up the `_extraparams` attribute and does so via  `__getattr__` which again invokes `get_param`. That's the inifinite recursion!

nengo/base.py
```diff
@@ -50,18 +51,61 @@ class NengoObject(with_metaclass(NetworkMember, SupportDefaultsMixin)):
...
+
+    def copy(self, add_to_container=True):
+        with warnings.catch_warnings():
```
tbekolay: I want to add a comment here explaining why we ignore these warnings in this case... but I wasn't quite sure why we do that. Could you explain briefly? I'll add it in as a comment.

nengo/config.py
tbekolay: So yeah, this happens because `self._extraparams` doesn't exist at this point in the copy/pickle... since it does exist, `if key in self._extraparams` doesn't use `__getattr__` since it can be retrieved in `__getattribute__`.

I managed to fix that particular issue by defining `__slots__` on `ClassParams`, which I think makes sense anyway, and might speed some things up. However, when you define `__slot__` you have to implement a `__getstate__` since the default behavior is to copy `self.__dict__`...

In another branch that I haven't made a PR for, I implemented `__getstate__` and `__setstate__` for `ClassParams` because the default behavior here doesn't actually work -- it may copy `_configures` and `_extraparams`, but it doesn't modify the param instances to point to the right instance of `ClassParams` since the new `ClassParams` is a copy. I'll bring over my implementation from that other branch and add some tests, as unless I'm missing something, the current tests aren't testing all of what `ClassParams` does.

nengo/base.py
jgosmann: The initial copy is not added to a network and produces a warning, but the copy will be added to the network afterwards (if `add_to_container == True`) which means that the warning should not be shown (after all the copy is added to a network).

I assume you cannot do a `Network.add(copy(self))` either because the initial copied instance will still not be added to a network.

nengo/config.py
jgosmann: Indeed, there's a problem:
```python
    model.config[nengo.Ensemble].radius = 2.
    ens = nengo.Ensemble(10, 1)

model2 = model.copy()

with model2:
    ens2 = nengo.Ensemble(10, 1)

print(ens2.radius)  # prints 1.0, but should be 2.0
```

hunse: I also seem to be able to fix it by checking if `key.startswith('_')` in `__getattr__`, and calling the super if it does (like we do with `__delattr__` and `__setattr__`). Is there any reason not to do this?

tbekolay: Hm, I was pretty sure that there is no `__getattr__` on the superclass... so what's happening is it's raising the attribute error, which `pickle` takes to mean that `__setstate__` doesn't exist. So yeah, I would avoid that -- it's another result of the asymmetry between `__setattr__` and `__getattr__` (there's more symmetry with `__setattr__` and `__getattribute__`)

hunse: What about raising an AttributeError rather than calling the super `__getattr__`? This also seems to work for me. Since `__getattr__` is only ever called as a fallback if the attribute isn't found normally, this doesn't seem problematic to me.

tbekolay: Yeah, I had that solution originally before implementing `__slots__`... I could add that back in if you think it's more clear

hunse: I like having `__getattr__`, `__setattr__`, and `__delattr__` all respond to keys starting with underscores in a special manner. It's clearer for me when I look at all three functions and see they have the same general form. So I'd make that change for sure. Would it make sense to do that and implement `__slots__`? I haven't seen the `__slots__` implementation, so I can't really say what I think of it.

tbekolay: I'll push it this afternoon, but yeah, with that check in `__getattr__` it really just amounts to adding a `__slots__` class attribute. We have to implement `__getstate__` and `__setstate__` regardless.

hunse: Can you make an implementation without `__slots__` first, then add `__slots__` in a separate commit if you still think it's a good idea? Then we can decide.

tbekolay: Yep... actually, doing the `__slots__` stuff probably makes more sense in a separate PR anyhow, since there are other classes like `Signal` that would benefit from it.

hunse: Yeah, that sounds like a plan.

nengo/connection.py
```diff
@@ -119,14 +119,16 @@ class ConnectionFunctionParam(FunctionParam):
 
     def __set__(self, conn, function):
         if function is None:
-            function_info = self.Info(function=None, size=None)
+            function_info = FunctionInfo(function=None, size=None)
```
tbekolay: What was the reason for separating out `FunctionInfo`? If I recall, it used to be separate but was put together in a somewhat recent PR... was this accidentally added in the rebase, or added back intentionally?

jgosmann: It needs to be pickle-able and for that it needs to be at the top-level of the module.

tbekolay: Makes sense!

jgosmann:
We have again a test failure because of a missing warning ... :/ (but it is not the open simulator warning)

jgosmann:
Relevant output in case anyone restarts the build:
```
_______ TestCopyEnsemble.test_python_copy_warns_about_adding_to_network ________
[gw1] linux2 -- Python 2.7.12 /home/travis/miniconda/envs/test/bin/python
self = <nengo.tests.test_copy.TestCopyEnsemble object at 0x7f91a0713b10>
    def test_python_copy_warns_about_adding_to_network(self):
        original = self.create_original()
        copy(original)  # Fine because not in a network
        with nengo.Network():
            with warns(NotAddedToNetworkWarning):
>               copy(original)
nengo/tests/test_copy.py:99: 
self = <nengo.utils.testing.warns object at 0x7f91a0713a10>, type = None
nengo/utils/testing.py:205: Failed
============== 1 failed, 749 passed, 60 skipped in 178.03 seconds ==============
```

jgosmann:
This isn't rebased to the current master branch and doesn't include #1141. So this should be fixed once it is rebased and includes those commits.

hunse:
Yeah, on my machine that error was only popping up sometimes, even when running exactly the same block of tests/pytest command. So hopefully rebasing to master will fix it for good.

nengo/base.py
```diff
@@ -3,6 +3,7 @@
 
 import numpy as np
 
+import nengo
```
jgosmann: I'm a bit surprised that this doesn't give a circular import. :astonished: 
But if it works, great!

nengo/config.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, configures):
         assert inspect.isclass(configures)
         self._configures = configures
         self._extraparams = {}
-        self._default_params = tuple(
+        self._defaultparams = tuple(
```
jgosmann: I would change `_extraparams` to `_extra_params` instead, but whatever.

tbekolay:
OK, I made a bunch of commits with recommended changes... enough that it's probably good for @Seanny123 and @hunse to give this another look over.

One note is that I did change the public interface for `NengoObject` in this commit: 1b5290a32a3e5da46d647

However, I don't think anyone is using this right now (it didn't require any changes to unit tests or examples) so I'm inclined to do this without deprecation and requiring a new major version...

nengo/config.py
tbekolay: Yeah, I'm fine either way; it's kind of lame right now actually that there's both `_extraparams` and `extra_params`, probably they shouldn't differ in the underscore.

Anyone have preferences for or against the underscore? I like them without, but it's not a strong preference.

nengo/config.py
```diff
@@ -62,10 +66,32 @@ def __setattr__(self, key, value):
...
+            param = self.get_param(attr)
+            if self in param:
+                state[attr] = param.get_default(self)
+
+        return state
+
+    def __setstate__(self, state):
+        self._configures = state['_configures']
+        self._defaultparams = tuple(state['_defaultparams'])
+        self._extraparams = dict(state['_extraparams'])
```
hunse: Why do we have to cast/copy these?

nengo/config.py
hunse: I agree, since the properties are called `default_params` and `extra_params`

nengo/config.py
tbekolay: We may not have to... I copied this from my old branch that implemented partial copies. I can take them out and see if it matters

nengo/config.py
hunse: I prefer them with, and that's also what the API is like right now. Easier to change the internals to conform with that.

jgosmann:
LGTM

tbekolay:
Oh, also for the record I didn't make these squash/fixup commits because it was too difficult to figure out what commit to squash all these things into... I'll deal with how to organize the history once we're good with the final state.

jgosmann:
Looking up the touched lines in ``git blame HEAD -- <file>`` can often be a good starting point, but if a lot of different lines are touched, it's less useful.

jgosmann:
(works of course only before you make the fixup/squash commit, otherwise you have to figure out the commit hash to use instead of `HEAD`)

tbekolay:
Made the changes requested by @hunse and tests pass, so hopefully the copies were unnecessary. If the current final state looks good to everyone I'll look into splitting up the history.

jgosmann:
LGTM :cake:

nengo/ensemble.py
```diff
@@ -156,6 +156,15 @@ def __init__(self, n_neurons, dimensions, radius=Default, encoders=Default,
         self.noise = noise
         self._neurons = Neurons(self)
 
+    def __getstate__(self):
+        state = super(Ensemble, self).__getstate__()
+        del state['_neurons']
+        return state
+
+    def __setstate__(self, state):
+        state['_neurons'] = Neurons(self)
```
hunse: Just realized: should this be set after the rest of the state is set?

hunse:
I just had trouble with this code:
```python
import nengo
from nengo.utils.compat import pickle

    a = nengo.Ensemble(100, 1)
    b = nengo.Ensemble(100, 1)
    nengo.Connection(a.neurons, b.neurons)

filename = 'test_ensemble_pickle.pkl'
with open(filename, 'wb') as f:
    pickle.dump(model, f)
```
It gives the error ``TypeError: can't pickle weakref objects``, but pickle isn't good about giving stack traces, so I'm not sure exactly what's causing it. Anyone want to help me track it down?

hunse:
Oh, I think it's because even though we remove the `Neurons` object from `Ensemble`, it's still in the connection, thus it gets pickled, and it has a weakref in it. Any ideas how to address this? Do we need a get/setstate for `Neurons`, too?

hunse:
Ok, I think I've got a solution. It also removes the `Ensemble.__setstate__` that I had a question about before, so if my commit looks good to others, this PR is good to go IMO.

hunse:
Also, I'm fine squashing my commit into wherever is most appropriate.

hunse:
Just thinking: We'll probably have to do something similar to what I just did for `LearningRule`. I'll get on that.

jgosmann:
We should add tests for those cases.

hunse:
Ok, I've added some tests. I realized my fix doesn't work, because if you're copying an Ensemble, the Neurons object will end up pointing to the old Ensemble, not the new one.

Anyway, I won't have time to work on this more today, so if someone else wants to look at it over the next few days, feel free. If you can make all the tests pass, you're the winner!

jgosmann:
I added a commit 41a0bba8be7ac633d4254baed930b2433731efa5 that makes all the tests pass, see the commit message for details and #1152 for related discussion.

I set the label back to "needs review" because these changes are more far reaching than just a simple fixup.

hunse:
Those changes look okay to me.

Seanny123:
Found a bug that won't let connections with slices be pickled.

Given the following code:
```
import nengo
import cPickle as pickle

    out_nd = nengo.Node(size_in=1)

    osc = nengo.networks.Oscillator(0.1, 1, 300)


    nengo.Connection(osc.ensemble[0], out_nd)

with open("model", "w") as ff:
    pickle.dump(model, ff)
```

I get the error:
```
  File "pickle_test.py", line 14, in <module>
    pickle.dump(model, ff)
  File "/usr/lib/python2.7/copy_reg.py", line 84, in _reduce_ex
    dict = getstate()
  File "/home/saubin/git/nengo-copy/nengo/base.py", line 175, in __getstate__
    raise NotImplementedError("Nengo objects do not support pickling")
NotImplementedError: Nengo objects do not support pickling
```

jgosmann:
I just looked over the code; it might work if you just remove the `__getstate__` and `__setstate__` methods on `ObjView`.

Seanny123:
Yep, just removing the code solved the problem! Thanks Jan!

jgosmann:
> One note is that I did change the public interface for NengoObject in this commit: 1b5290a
> 
> However, I don't think anyone is using this right now (it didn't require any changes to unit tests or examples) so I'm inclined to do this without deprecation and requiring a new major version...

nengo_gui is using it: nengo/nengo_gui#819

I suppose the options are:
1. Revert those changes.
2. Fix nengo_gui breaking compatibility to older Nengo versions.
3. Fix nengo_gui with keeping compatibility to previous Nengo versions.

The last option might lead to a lot of conditionals. So I would probably prefer option 1 or 2. The new interface is nice, but I'm fine with 1 if people think that the compatibility to older Nengo versions is important.

hunse:
I'd be in favour of partially reverting the changes, and making `params` back into a property.

I could also go for option 2. The situation I worry about is someone upgrading Nengo, and not not realizing they also have to upgrade nengo_gui. How quickly does it fail?

I don't think option 3 really needs that many conditionals. We should just be able to do `params = self.config._clsparams.params; params = params if is_iterable(params) else params()` or something like that.

Seanny123:
> I could also go for option 2. The situation I worry about is someone upgrading Nengo, and not not realizing they also have to upgrade nengo_gui. How quickly does it fail?

It fails immediately by stopping you from running `nengo_gui` in any way.

jgosmann:
> I don't think option 3 really needs that many conditionals.

Yeah, it seems to be only that one place in the GUI.

jgosmann:
I made the necessary changes to allow copies of object views and added some tests for that. Still waiting on @tbekolay to give his ok to revert `params` back into a property.

jgosmann:
I added a commit to turn `params` back into a property. I haven't tested whether this fixes nengo/nengo_gui#819. Right now the property returns an iterator, but to me it seems that it is uncommon for properties to do that. So maybe we want to turn the interator into a tuple or list. This should not affect any existing code.

tbekolay:
After a bit of head scratching, I decided the history of this branch is not too important and squashed this all into one commit with co-authors. I added a fixup (4790cbb) to change the properties that were returning iterators to instead return lists (or tuples where appropriate), which I think is the right call. I'll test this with the snippet in https://github.com/nengo/nengo_gui/issues/819 to see if it works, but assuming it does, I'll merge this tomorrow morning (in case anyone wants to take a final look at this branch today).

nengo/tests/test_copy.py
```diff
@@ -11,15 +10,6 @@
 from nengo.utils.testing import warns
 
 
-@pytest.fixture(scope='module', autouse=True)
-def error_on_not_added_to_network_warning(request):
-    catcher = warnings.catch_warnings(record=True)
-    catcher.__enter__()
-    warnings.simplefilter(
-        'error', category=NotAddedToNetworkWarning)
-    request.addfinalizer(lambda: catcher.__exit__(None, None, None))
```
jgosmann: Why did you remove this test?

tbekolay: It makes all the `with warns(NotAddedToNetworkWarning)` tests fail (since they error when the warning is raised). It seemed more complicated than it was worth to disable this check in the cases in which we wanted to check the warning. Any ideas on how to do it simply?

jgosmann: Ah right, this is a fixture not a test. Removing this LGTM.

Seanny123:
LGTM. Excited to see this merged!

PR #985, opened 2016-03-17
Fixed minor bugs in `test_neurons.test_reset`

hunse:
The WhiteSignal being used had a low frequency and duration,
resulting in a signal that was acutally all zeros.

### DIFF
--- a/nengo/tests/test_neurons.py
+++ b/nengo/tests/test_neurons.py
@@ -230,7 +230,7 @@ def test_izhikevich(Simulator, plt, seed, rng):
     simulated in Nengo (but doesn't test any properties of them).
     """
     with nengo.Network() as m:
-        u = nengo.Node(output=WhiteSignal(0.6, 8), size_out=1)
+        u = nengo.Node(output=WhiteSignal(0.6, high=10), size_out=1)
 
         # Seed the ensembles (not network) so we get the same sort of neurons
         ens_args = {'n_neurons': 4, 'dimensions': 1, 'seed': seed}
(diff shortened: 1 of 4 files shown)


drasmuss:
Do we need to make a similar change here https://github.com/nengo/nengo/pull/985/files#diff-273f7dd17cf9940b99c916464def1fabR284?

Only other comment is typo in commit message (acutally -> actually).

hunse:
No, the fact that it was all zeros was quite acute to me.

Yeah, we should definitely change that other one, too.

tbekolay:
https://github.com/nengo/nengo/pull/985/files#diff-273f7dd17cf9940b99c916464def1fabR284 only runs the simulator for 0.1 seconds, so there's no need to generate more signal. But I guess for short runs the high frequency should be at least 10, I guess?

tbekolay:
Pushed a commit to up the `high` of other `WhiteSignal`s in the test suite. And I can fix the commit message in the merge ;) This look OK to you, @drasmuss and @hunse ?

nengo/tests/test_neurons.py
```diff
@@ -281,7 +281,7 @@ def test_dt_dependence(Simulator, nl_nodirect, plt, seed, rng):
     """Neurons should not wildly change with different dt."""
     with nengo.Network(seed=seed) as m:
         m.config[nengo.Ensemble].neuron_type = nl_nodirect()
-        u = nengo.Node(output=WhiteSignal(0.1, 5), size_out=2)
+        u = nengo.Node(output=WhiteSignal(0.1, high=8), size_out=2)
```
hunse: So this one is still all zero. Basically, we need `high < 1. / period` to get a non-zero signal. Is this check worth putting in WhiteSignal itself?

hunse: Also, when I fix this, this test breaks because the WhiteSignal changes depending on the `dt` used. In general, I think it would be very difficult to have our random processes be agnostic to dt. But do you think users will expect this? In this respect, our old method of generating WhiteSignal was better, where we had sine waves at fixed frequency intervals and we just generated random coefficients for them, since that method is agnostic to `dt`.

hunse:
Okay, I've added a few fixes to make one more signal non-zero and check in WhiteSignal that we don't make zero signals in the future. This should be good to go now.

tbekolay:
Looks great to me. @drasmuss, want to quickly rereview? I'll merge on your +1.

drasmuss:
Looks good, but Travis-CI error?

hunse:
That's a weird error. Not my fault!! :) That notebook doesn't even seem to use WhiteSignal, so I don't see how the error could be related to any of the changes here. Also, the error itself seems like some strange IPython thing.

hunse:
Why is there a flake8 error saying ValidationError is not defined in processes.py? It clearly is, and I don't get this error locally.

tbekolay:
I'll wait on #996 for this since they touch some of the same code, then merge after that's merged. I'll keep the `allclose` in the rebase.

PR #987, opened 2016-03-17
Let connections accept functions without names

### LINKED ISSUE #981: Support partial functions in a Connection
It would be nice to be able to use partial functions for the `function` argument in a `Connection`. Currently an `AttributeError` is thrown when you try to do this.

Simple example:
```
import nengo
from functools import partial

def add(x, val):
    return x + val

model = nengo.Network()

with model:
    a = nengo.Ensemble(10,1)
    b = nengo.Ensemble(10,1)
    nengo.Connection(a, b, function=partial(add,val=2))

sim = nengo.Simulator(model)
sim.run(1)
```

and the error is:
```
  File "partial_func_nengo_issue.py", line 14, in <module>
    sim = nengo.Simulator(model)
  File "/home/bjkomer/nengo/nengo/simulator.py", line 108, in __init__
    self.model.build(network)
  File "/home/bjkomer/nengo/nengo/builder/builder.py", line 34, in build
    return Builder.build(self, obj, *args, **kwargs)
  File "/home/bjkomer/nengo/nengo/builder/builder.py", line 79, in build
  File "/home/bjkomer/nengo/nengo/builder/network.py", line 72, in build_network
    model.build(conn)
  File "/home/bjkomer/nengo/nengo/builder/builder.py", line 34, in build
    return Builder.build(self, obj, *args, **kwargs)
  File "/home/bjkomer/nengo/nengo/builder/builder.py", line 79, in build
  File "/home/bjkomer/nengo/nengo/builder/connection.py", line 196, in build_connection
    model.sig[conn]['weights'] = Signal(weights, name="%s.weights" % conn)
  File "/home/bjkomer/nengo/nengo/connection.py", line 329, in __str__
    return "<Connection %s>" % self._label
  File "/home/bjkomer/nengo/nengo/connection.py", line 326, in _label
    if self.function is not None else "")
AttributeError: 'functools.partial' object has no attribute '__name__'
```

Seanny123:
Resolves #981

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -374,10 +374,9 @@ class Connection(NengoObject):
         if self.label is not None:
             return self.label
 
-        return "from %s to %s%s" % (
-            self.pre, self.post,
-            " computing '%s'" % self.function.__name__
-            if self.function is not None else "")
+        desc = "" if self.function is None else " computing '%s'" % (
+            getattr(self.function, '__name__', str(self.function)))
+        return "from %s to %s%s" % (self.pre, self.post, desc)
 
     def __str__(self):
         return "<Connection %s>" % self._str
--- a/nengo/tests/test_connection.py
+++ b/nengo/tests/test_connection.py
@@ -1,3 +1,5 @@
+from functools import partial
+
 import numpy as np
 import pytest
 
@@ -798,3 +800,19 @@ def test_connectionlearningruletypeparam():
         with pytest.raises(ValueError):  # transform must be correct shape
             nengo.Connection(a, b, transform=np.ones((10, 11)),
                              learning_rule_type=nengo.BCM())
+
+
+def test_function_with_no_name(Simulator):
+
+    def add(x, val):
+        return x + val
+
+    model = nengo.Network()
+
+    with model:
+        a = nengo.Ensemble(10, 1)
+        b = nengo.Ensemble(11, 1)
+        nengo.Connection(a, b, function=partial(add, val=2))
+
+    with Simulator(model) as sim:
+        assert sim


nengo/connection.py
```diff
@@ -374,10 +374,16 @@ def _str(self):
         if self.label is not None:
             return self.label
 
+        desc = None
+        if self.function is not None and hasattr(self.function, "__name__"):
+            desc = self.function.__name__
+        elif not hasattr(self.function, "__name__"):
+            desc = str(self.function)
```
hunse: You can do this in one line with `desc = getattr(self.function, '__name__', str(self.function)) if self.function is not None else ""`, and then you don't have to check `desc is not None` below either.

Seanny123:
@hunse is this better?

hunse:
Well except now you're going to say computing twice :)

Seanny123:
@hunse Gah. Fixed that now and tested it more carefully with this script to visually inspect the strings:

    from functools import partial

    def add(x, val):
        return x + val

    model = nengo.Network()

    with model:
        a = nengo.Ensemble(10, 1)
        b = nengo.Ensemble(11, 1)
        c = nengo.Ensemble(12, 1)
        nengo.Connection(a, b, function=partial(add, val=2))
        nengo.Connection(a, c)
    model.all_connections

tbekolay:
This is pretty small, so seems worth having in 2.1.0. @hunse, look good to you now? If so I'll merge it in.

nengo/connection.py
```diff
@@ -374,10 +374,14 @@ def _str(self):
...
+            desc = " computing '%s'" % (
+                getattr(self.function, '__name__', str(self.function)))
+        else:
+            desc = ""
+
         return "from %s to %s%s" % (
-            self.pre, self.post,
-            " computing '%s'" % self.function.__name__
-            if self.function is not None else "")
+            self.pre, self.post, desc)
```
hunse: This can be one line with the above line now.

nengo/connection.py
```diff
@@ -374,10 +374,14 @@ def _str(self):
         if self.label is not None:
             return self.label
 
+        if self.function is not None:
+            desc = " computing '%s'" % (
+                getattr(self.function, '__name__', str(self.function)))
+        else:
+            desc = ""
```
hunse: Generally I try to avoid `if` blocks that set a variable one way in one part of the block and a different way in another part without doing much else, because it seems like an inline if can do the same usually with more clarity. But that's just personal preference, what you have is okay.

hunse:
Just turn those two lines into one and it's good to go. You could even just do that on merge, it's so small.

tbekolay:
Yep, I'll do it on the merge. Thanks!

PR #988, opened 2016-03-18
Improvements to operator strings

hunse:
- TimeUpdate now has a `tag` so that the `__repr__` doesn't throw
  an exception.
- Operators now have a `_descstr` function that they should overload.
  This function returns a describing string, which the operator uses
  in creating its `__str__`. This means that Operator subclasses
  don't have to overload `__str__`, they can just overload `_descstr`.

Really the only bug here was TimeUpdate not having a tag. If people are unsure about the second part, I can rip that out and we can just put the TimeUpdate tag in 2.1.0.

### DIFF
--- a/examples/usage/strings.ipynb
+++ b/examples/usage/strings.ipynb
@@ -255,6 +255,8 @@
      "collapsed": false,
      "input": [
       "sig = nengo.builder.signal.Signal(np.array([0.]), name=\"sig\")\n",
+      "print(nengo.builder.operator.TimeUpdate(sig, sig))\n",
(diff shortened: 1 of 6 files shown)


drasmuss:
Looks good.  I'm ambivalent about the `_descstr` thing.  It feels a bit excessively hierarchical to have to go read a superclass to understand the `__str__` output of an object.  But it does cut down on some repetitive boilerplate.  Since it's already implemented I'd say go ahead and put it in.

tbekolay:
I'm not a big fan of the names `_descstr` and `_tagstr` but I don't have any better suggestions so this looks good. Merging!

PR #991, opened 2016-03-19
Assoc mem v3

xchoo:
Rewrite of the AssociativeMemory code to support new features:
- Added input utilities output node.
- Utility mapping functions are now specifiable when creating an output mapping. This allows the user to specify utility mappings other than the default filtered_step_func mapping.
- Created several utility mapping helper functions
- Renamed 'threshold_output' to 'cleanup_output' to better illustrate the operation of that function.
- Improved documentation
- Updated test_assoc_mem.py to accommodate changes to assoc_mem.py code.
- Removed extra synapses on connections from nodes to ensembles.

### DIFF
--- a/nengo/networks/assoc_mem.py
+++ b/nengo/networks/assoc_mem.py
@@ -10,10 +10,6 @@ from nengo.utils.compat import is_iterable, range
 from nengo.utils.network import with_self
 
 
-def filtered_step(t, shift=0.5, scale=50, step_val=1):
-    return np.maximum(-1 / np.exp((t - shift) * scale) + 1, 0) * step_val
-
-
 class AssociativeMemory(nengo.Network):
     """Associative memory network.
 
@@ -47,6 +43,10 @@ class AssociativeMemory(nengo.Network):
 
     exp_scale = 0.15  # Scaling factor for exponential distribution
     n_eval_points = 5000
+
+    cleanup_output_prefix = 'cleaned'
+    default_ens_suffix = 'default_ens'
+    utility_output_suffix = 'utilities'
 
     def __init__(self, input_vectors, output_vectors=None,  # noqa: C901
                  n_neurons=50, threshold=0.3, input_scales=1.0,
@@ -82,6 +82,7 @@ class AssociativeMemory(nengo.Network):
             threshold = threshold * np.ones(input_vectors.shape[0])
         else:
             threshold = np.array(threshold)
+        self.thresholds = threshold
 
         # --- Check preconditions
         self.n_items = input_vectors.shape[0]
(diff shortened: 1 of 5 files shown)


nengo/networks/assoc_mem.py
```diff
@@ -137,6 +133,100 @@ def __init__(self, input_vectors, output_vectors=None,  # noqa: C901
...
+        x_shift: float
+            Amount to shift the linear map on the 'x'-axis.
+            E.g. If the value of x_shift = 0.3, then the output of the
+                 mapping function would be 0 at x == 0.3.
+        x_scale: float
+            Scaling factor to be applied to the 'x' input. Affects the slope
+            of the linear mapping function.
+        """
+        return lambda x, x_shift=x_shift, x_scale=x_scale: \
+            (x_scale * x) - x_scale * x_shift
```
hunse: Isn't this more efficient and clearer as `x_scale * (x - x_shift)`?

nengo/spa/assoc_mem.py
```diff
@@ -13,22 +13,23 @@ class AssociativeMemory(Module):
         The vocabulary (or list of vectors) to be produced for each match. If
         not given, the associative memory will act like an auto-associative
         memory (cleanup memory).
+
+    input_keys: list of strings, optional
```
hunse: Could some of these doc changes go into the 2.1.0 doc update?

nengo/spa/assoc_mem.py
```diff
@@ -60,8 +54,9 @@ def __init__(self, input_vocab, output_vocab=None,  # noqa: C901
                  default_output_key=None, threshold=0.3,
                  inhibitable=False, wta_output=False,
                  wta_inhibit_scale=3.0, wta_synapse=0.005,
-                 threshold_output=False, label=None, seed=None,
-                 add_to_container=None):
+                 cleanup_output=False,
+                 replace_output_with_cleaned_output=True,
+                 label=None, seed=None, add_to_container=None):
```
hunse: Unless you're going to have docstrings for `label`, `seed`, and `add_to_container`, I would maybe do them as `**network_kwargs` and just note in the docstring that they go straight to the Network constructor so users can look them up.

nengo/networks/assoc_mem.py
xchoo: Oh yeah. I changed it for the other two functions. Forgot to change it for this one. :laughing:

nengo/spa/assoc_mem.py
```diff
@@ -60,8 +54,9 @@ def __init__(self, input_vocab, output_vocab=None,  # noqa: C901
                  default_output_key=None, threshold=0.3,
                  inhibitable=False, wta_output=False,
                  wta_inhibit_scale=3.0, wta_synapse=0.005,
```
hunse: Can we or would we want to modify this to use the parameter/config system, so that we can set default values for these things?

nengo/networks/assoc_mem.py
```diff
@@ -330,46 +481,109 @@ def add_wta_network(self, inhibit_scale=1.5, inhibit_synapse=0.005):
...
-                for e in self.thresh_ens.ensembles:
+                for e in self.bias_ens2.ensembles:
                     nengo.Connection(
                         self.inhibit, e, transform=-self._inhib_scale,
                         synapse=None)
+
+            # --- Connect default output vector to cleaned outputs
+            #     (if available)
+            default_vector_ens_name = '_'.join([output_name,
+                                                self.default_ens_suffix])
```
hunse: I would do `vector_ens = getattr(self, '_'.join([output_name, self.defaut_ens_suffix]), None)` and then on the next line do `if vector_ens`. That saves you having to store the name and use `hasattr`.

nengo/spa/assoc_mem.py
xchoo: Hmmm. kk!

nengo/spa/assoc_mem.py
xchoo: I'll defer to @tbekolay or @tcstewar for opinions on this.

tcstewar: I think I'm happiest with these as constructor arguments.  They're relatively well-named, there aren't a ridiculous number of them, and they're relatively separate from existing config items.

hunse: Oh, I didn't mean remove them from the constructor. I just meant use parameters for them internally, so that a user could set the e.g. wta_synapse for all associative memories with one line. Or is that not possible with our current config system?

xchoo: So are you thinking of something like this?
```
am_config = nengo.Config[nengo.networks.AssociativeMemory]
am_config.n_neurons = 20
am_config.wta_synapse = 0.01

with am_config:
    am = nengo.networks.AssociativeMemory()
```

xchoo: Oh. And this is the SPA associative memory code. I'd leave this as a separate PR (this PR is just for the networks associative memory code changes). :)

hunse: Yeah, that's what I was thinking. Totally fine as a separate PR, I just thought of it now.

nengo/spa/assoc_mem.py
```diff
@@ -100,18 +95,22 @@ def __init__(self, input_vocab, output_vocab=None,  # noqa: C901
...
+            if cleanup_output and not replace_output_with_cleaned_output:
+                self.cleaned_output = self.am.cleaned_output
+
             if inhibitable:
                 self.inhibit = self.am.inhibit
 
-            self.utilities = self.am.utilities
-            if threshold_output:
-                self.thresholded_utilities = self.am.thresholded_utilities
+            self.utilities = self.am.output_utilities
```
hunse: Why is it called `utilities` in the module but `output_utilities` in the network?

xchoo: Ah. It's a side effect of how the new networks associative memory code works. The networks code now creates a utilities node for each output, the default output that is created in the constructor makes a default `output_utilities` node as well (it's just the `output name` + '_utilities'). I kept it as `utilities` in the spa code so that minimal spa code changes are needed.

nengo/networks/assoc_mem.py
```diff
@@ -40,16 +36,27 @@ class AssociativeMemory(nengo.Network):
...
+        ensembles in the AssociativeMemory network.
+
+    Additional network parameters are passed to the Network constructor through
+    **network_kwargs
     """
 
     exp_scale = 0.15  # Scaling factor for exponential distribution
     n_eval_points = 5000
 
+    cleanup_output_prefix = 'cleaned'
```
Seanny123: A quick comment for what these are used for would be appreciated.

nengo/networks/assoc_mem.py
```diff
@@ -94,49 +102,139 @@ def __init__(self, input_vectors, output_vectors=None,  # noqa: C901
...
 
         # -- Create the core network
         with self, self.am_ens_config:
             self.bias_node = nengo.Node(output=1)
             self.elem_input = nengo.Node(
                 size_in=self.n_items, label="element input")
-            self.elem_output = nengo.Node(
-                size_in=self.n_items, label="element output")
-            self.utilities = self.elem_output
+            self.elem_utilities = nengo.Node(
```
Seanny123: You use the word "element" all over the place in the code, but you never formally define it and how it's different from a vector. Describing this somewhere (I don't know where) would be helpful.

xchoo: The 'elements' aren't really meant to be outward facing. Users really don't need to know what an 'element' in the code corresponds to. All they need are the input and output nodes.

Seanny123: True, but if I have to go to this code later, I'd like to know what
organisation you had in mind.

xchoo: An 'element' is one 'item' in the list of input vectors.

Seanny123: Yeah, so adding that as a code comment before these objects are initialised would be super cool.

xchoo: I have renamed them 'ensembles' if that makes it clearer.

Seanny123: Thanks. That makes it a lot clearer to me, since that maps onto the actual objects these things represent.

Seanny123:
LGTM. Can proceed to second review. I nominate @jgosmann .

jgosmann:
SPA PRs are on hold for the `nengo_spa` extraction. Also, there is a number of issues regarding the associative memory that have to looked over and to check whether we can resolve all of this in `nengo_spa`.

First PR to get merged for `nengo_spa` will be #1006, next I want to merge #1014. For further PRs, I might wait until the deferrables (#995) and spaopt discussion is resolved which I might take a few more weeks (though I think we're slowly getting to a point where we know all the pro and contra arguments and things we will keep disagreeing on).

nengo/networks/assoc_mem.py
```diff
@@ -94,49 +106,139 @@ def __init__(self, input_vectors, output_vectors=None,  # noqa: C901
...
+        x_shift: float, optional
+            Amount to shift the linear map on the 'x'-axis.
+            E.g. If the value of x_shift = 0.3, then the output of the
+                 mapping function would be 0 at x == 0.3.
+        x_scale: float, optional
+            Scaling factor to be applied to the 'x' input. Affects the slope
+            of the linear mapping function.
+        """
+        return lambda x, x_shift=x_shift, x_scale=x_scale: \
+            (x_scale * (x - x_shift))
```
jgosmann: To met it seems that
```python
return lambda x: x_scale * (x - x_shift)
```
would be equall valid code or are the `x_shift=x_shift, x_scale=x_scale` actually required (in this particular case, there are certainly other cases were this idiom is required)?

If both forms are valid, what is the preferred one?

tcstewar: I'm pretty sure you're right that it's not needed here, and I'd say we should only use this idiom when it is required (otherwise it looks odder than it needs to...)

xchoo: Hmmm. I wrote it this way to prevent closure problems. But it might not apply here. haha

tcstewar:
> Hmmm. I wrote it this way to prevent closure problems. But it might not apply here. haha

Yup...  the general rule of thumb is to make sure you do this if you're defining your function inside a ```for``` loop where the variables change on each iteration through the loop.  But in this case, the variable never changes its value, so there's no closure confusion.

jgosmann:
This is obsolete now as the associative memory has been rewritten in nengo_spa.

PR #993, opened 2016-03-23
Fixed deprecated filt and filtfilt functions

tcstewar:
Looks like a ```return``` got left out of https://github.com/nengo/nengo/commit/3fbbba9917e78e4a0ae4d3cdefa5dbeeff41b0d8

This allows the deprecated ```nengo.synapse.filt``` and ```nengo.synapse.filtfilt``` functions to actually return their filtered results, rather than throwing them away and returning None.


tbekolay:
Whoops, good catch! @hunse, care to look and merge?

hunse:
The nice thing about the old way is it did a much better job of discouraging people from using these functions. But I guess this is all right...

PR #994, opened 2016-03-24
Disable IPython history in TravisCI

tbekolay:
We are getting sporadic errors in TravisCI when two threads try to access the IPython history when exporting a notebook to a Python script. This disables the IPython history entirely.

This doesn't actually solve all of the problems; e.g., see [this build](https://travis-ci.org/nengo/nengo/jobs/118276959) which gives an error in an `atexit` function. However, I believe this error won't cause builds to fail, and is just a bug in IPython, which I will file upstream. This PR should fix the error that actually causes builds to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,10 @@ install:
 
 # Run the tests
 script:
+  - mkdir -p "$HOME/.ipython/profile_default"
+  - "echo 'c.HistoryAccessor.enabled = False\n' > $HOME/.ipython/profile_default/ipython_config.py"
   - if [[ -n $SETUP_CMD ]]; then
-      python -c "import numpy; numpy.show_config()"
+      python -c "import numpy; numpy.show_config()";
       python setup.py -q install;
       eval python setup.py "$SETUP_CMD";
     fi


jgosmann:
LGTM :cake:

.travis.yml
```diff
@@ -52,8 +52,10 @@ install:
 
 # Run the tests
 script:
+  - mkdir -p "$HOME/.ipython/profile_default"
+  - "echo 'c.HistoryAccessor.enabled = False\n' > $HOME/.ipython/profile_default/ipython_config.py"
```
hunse: Why is this whole line in quotes?

tbekolay: I just copied the line that used to echo `Agg` to `matplotlibrc` and changed the details; I recall having to use quotes there because of how TravisCI parses YAML but I don't know if it's still necessary, or if it ever was

hunse: Okay, just curious.

tbekolay:
As an update to the intermittent issue that I mentioned above, the [Python exporter](https://github.com/jupyter/nbconvert/issues/273) was the culprit. Fortunately, the issue has [already been fixed](https://github.com/jupyter/nbconvert/pull/229), so when `nbconvert` 4.2 is released, the issue will go away.

PR #996, opened 2016-03-26
Enable CI testing on Windows with AppVeyor

tbekolay:
I figured this would be helpful for ensuring #946 works well on Windows.

This PR ensures that all tests pass on Windows using AppVeyor. Right now, we're using conda (AppVeyor has 2.7, 3.4 and 3.5 versions of Miniconda installed by default). With the first two commits, the builds succeed [for all versions but one](https://ci.appveyor.com/project/nengo/nengo/build/1.0.7). Python 3.5 (32-bit) fails for `test_reset` with `LIFRate` neurons (see [here](https://ci.appveyor.com/project/nengo/nengo/build/1.0.7/job/9l62j27yo3ub85sm#L1002)). Any idea why this might be @hunse?

One thing up for discussion in this PR is how many test environments we actually want to run with AppVeyor.

The way I see it, TravisCI should be where we test as many situations as possible (since it allows us to run tests in parallel, which AppVeyor does not), while AppVeyor should be used just to make sure that things work in Windows. Still, I feel like there are enough Windows-specific differences between Python 2 and Python 3. So, I think I would prefer to test with the latest version of Python 2 and Python 3, but not all versions available, which is what is currently done in this PR.

The alternatives are to only run one test (which I think should be the latest version, so Python 3.5 right now), all tests (the six that are running right now), some other subset of the six tests (e.g., 32-bit and 64-bit Python 3.5), or to mirror exactly in AppVeyor what we do in TravisCI.

As I said above, testing in AppVeyor takes quite a bit longer than TravisCI (since you can use cool docker-based infrastructure in Windows), so I would err on the side of fewer tests.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,21 @@
+environment:
+  matrix:
+    - PYTHON: "C:\\Miniconda"
+    - PYTHON: "C:\\Miniconda3"
+    - PYTHON: "C:\\Miniconda35-x64"
+
+init:
+  - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
+
+install:
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  - conda info -a
+  - conda install numpy matplotlib jupyter pytest flake8
+
+build: false  # Not a C# project
+
+test_script:
+  - python setup.py -q install
+  - python setup.py test -a "nengo -v --durations=20"
+  - flake8 -v nengo
--- a/nengo/tests/test_connection.py
+++ b/nengo/tests/test_connection.py
@@ -299,7 +299,7 @@ def test_dist_transform(Simulator, seed):
         conn = nengo.Connection(a, b)
 
     sim = Simulator(net)
-    assert np.all(w == sim.model.params[conn].weights)
+    assert np.allclose(w, sim.model.params[conn].weights)
 
 
 def test_weights(Simulator, nl, plt, seed):
--- a/nengo/tests/test_ensemble.py
+++ b/nengo/tests/test_ensemble.py
@@ -322,8 +322,8 @@ def test_noise_gen(Simulator, nl_nodirect, seed, plt):
 
     assert np.all(sim.data[pos_p] >= sim.data[normal_p])
     assert np.all(sim.data[normal_p] >= sim.data[neg_p])
-    assert not np.all(sim.data[normal_p] == sim.data[pos_p])
-    assert not np.all(sim.data[normal_p] == sim.data[neg_p])
+    assert not np.allclose(sim.data[normal_p], sim.data[pos_p])
+    assert not np.allclose(sim.data[normal_p], sim.data[neg_p])
 
 
 def test_noise_copies_ok(Simulator, nl_nodirect, seed, plt):
--- a/nengo/tests/test_learning_rules.py
+++ b/nengo/tests/test_learning_rules.py
@@ -208,7 +208,7 @@ def test_unsupervised(Simulator, rule_type, solver, seed, rng, plt):
     plt.xlabel("Time (s)")
     plt.ylabel("Weights")
 
-    assert not np.all(sim.data[weights_p][0] == sim.data[weights_p][-1])
+    assert not np.allclose(sim.data[weights_p][0], sim.data[weights_p][-1])
 
 
 def learning_net(learning_rule, net, rng):
@@ -264,7 +264,7 @@ def test_dt_dependence(Simulator, plt, learning_rule, seed, rng):
     plt.ylabel("Presynaptic activity")
 
     assert np.allclose(trans_data[0], trans_data[1], atol=3e-3)
-    assert not np.all(sim.data[trans_p][0] == sim.data[trans_p][-1])
+    assert not np.allclose(sim.data[trans_p][0], sim.data[trans_p][-1])
 
 
 @pytest.mark.parametrize('learning_rule', [nengo.PES, nengo.BCM, nengo.Oja])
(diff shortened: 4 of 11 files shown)


hunse:
No idea. Can anyone reproduce this failure locally on a Windows machine? It would make it much easier to debug.

drasmuss:
Doesn't happen on my machine (Python 3.5 64bit)

tbekolay:
Hrm, this is puzzling. The past 3 builds have randomly failed on different environments ([1](https://ci.appveyor.com/project/nengo/nengo/build/1.0.7), [2](https://ci.appveyor.com/project/nengo/nengo/build/1.0.8), [3](https://ci.appveyor.com/project/nengo/nengo/build/1.0.9)). The only commonality is that they're all 32-bit versions of Python. I tried to recreate the environment in a virtualbox by installing a 32-bit version version of Anaconda, and then running the tests with different versions of Python (confirming that they're 32-bit versions of Python) but I can't get that test to fail ... The only difference I can think of is that AppVeyor uses Windows Server and my Virtualbox is Windows 10 :P

So... one possibility is to limit the builds to 64-bit Pythons? Any other thoughts?

drasmuss:
On Travis-CI we're only testing on 64-bit Pythons right? Are we sure this is a Windows issue, and not just a 32-bit Python issue?

tbekolay:
Could very likely be a 32-bit Python issue in general -- of course, in non-Windows OSes it's super hard to install 32-bit anythings in a 64-bit OS. TravisCI doesn't have 32-bit environments available, so perhaps that's good reason to test in 32-bit on AppVeyor.

If it is a 32-bit Python issue, then I think the first thing to do would be to replace all `dtype=float` or `dtype=int`s to be explicitly `np.float/int64`s.

That said, I can't reproduce this in a 32-bit Python in Windows 10, so I can't test to see if that would fix anything :P

jgosmann:
I just looked the failing assert:
```python
assert np.all(sim.data[square_p] == first_square_p) 
```
It is always problematic to do an exact comparison of floating point numbers. Even though it would seem reasonable that the floating point computations in this test are completely deterministic and producing the same bit patterns, this might not be true? This could also very well be related to 32-bit vs. 64-bit.

arvoelke:
> It is always problematic to do an exact comparison of floating point numbers. Even though it would seem reasonable that the floating point computations in this test are completely deterministic and producing the same bit patterns, this might not be true? 

:+1: For all intents and purposes we should consider the epsilon to be pseudo-random (since the behaviour is undefined at this level no guarantees of determinism are required by implementation). Top two results on google: [this](http://stackoverflow.com/questions/27583780/deterministic-python-script-behaves-in-non-deterministic-way) and [this](https://mail.scipy.org/pipermail/numpy-discussion/2011-January/054484.html). Seems to come up often with the BLAS on Windows.

hunse:
That's just stupid. Why would the BLAS be non-deterministic? I say we just drop support for Windows...

This failing test is the same one that gets fixed up in #985. How about we merge that first and get this rebased onto that? I'm curious to see if this test still fails with the modifications, though I'm not sure why those modifications would help. If it's still failing after that, then we can switch to checking with `allclose`.

arvoelke:
> That's just stupid. Why would the BLAS be non-deterministic?

The first link offers an explanation:

> Part of the numerical noise comes from data alignment in memory, ie whether it happens to lie directly in a position suitable for cpu's sse etc. instructions or not --- the two cases do operations in different order, so floating point rounding error is different. win32 standard memory allocator afaik tends to return blocks of memory that are on average less aligned than linux allocators, so there is more jitter there. If your results depend significantly on which way FP rounding error goes, your problem formulation is numerically unstable.

This isn't the only platform we are going to run into this issue on either. This is the double-edged sword of undefined behaviour; different implementations are free to do what they want given their own degrees of freedom. And these concepts only become more relevant when considering neuromorphic backends.

>  I say we just drop support for Windows...

I don't think this is any reason to throw the baby out with the bathwater.

hunse:
> I don't think this is any reason to throw the baby out with the bathwater.

image

jgosmann:
But it can be such an annoying baby ... ;)

hunse:
> This isn't the only platform we are going to run into this issue on either.

That's a really good point. I've gone and switched to using `np.allclose` in tests where I could find us doing direct comparison of floats, since we don't want to send the message to backends that these things need to be exact.

hunse:
This might be something that you have to change on the Github side, but right now Appveyor is running on both the branch and the PR. The PR details say "master" as the branch; does that mean it's doing a merge of this branch into master and running on that? I'm not sure if it's necessary to run these both, though I guess there could be a bug that doesn't show up in the branch but does once it's merged into master. I'm partly wondering because Appveyor seems to run everything sequentially, versus TravisCI which runs multiple configurations at the same time.

tbekolay:
I couldn't figure it out through AppVeyor, but I looked at the webhook through Github and disabled the PR build, so it should hopefully just build on branch pushes, not PR updates. Since we rebase all branches to master before merging, we shouldn't get into any issues with the PR being different from the branch itself.

But yeah, the sequential builds are why I think we should do minimal testing on AppVeyor, maybe just 2.7 and 3.5 32-bit?

hunse:
> But yeah, the sequential builds are why I think we should do minimal testing on AppVeyor, maybe just 2.7 and 3.5 32-bit?

I would do one 64-bit test too, whichever of 2.7 and 3.5 we think is more common/are more concerned about.

hunse:
> so it should hopefully just build on branch pushes, not PR updates. Since we rebase all branches to master before merging, we shouldn't get into any issues with the PR being different from the branch itself.

A lot of the time, though, I don't like to wait for the tests once I rebase to master before merging. Could we do it the other way, and just test the PR (merged)? I guess there are chances the merge doesn't quite work the same as a rebase would.

tbekolay:
I assume it will only test if the merge can be done cleanly, which means that a rebase would also be done cleanly and shouldn't affect anything. But yeah, probably it's not worth building branches that aren't PRs anyhow, so seems like it's about the same to build PRs and not pushes, maybe a bit better, so I'll switch to that.

tbekolay:
> I would do one 64-bit test too, whichever of 2.7 and 3.5 we think is more common/are more concerned about.

Hm, I was assuming that TravisCI would take care of that, but if we can hit the most cases in the same number of builds that's cool with me. Are you thinking something like Py2 64-bit + Py35 32-bit, or Py2 64-bit + Py35 32-bit + Py35 64-bit?

hunse:
I was thinking like a Py2 32-bit (necessary), Py35 32-bit (necessary), and Py2 64-bit (probably covered by Travis but I'd like to test some 64-bit thing on Windows just in case). If you'd prefer just two tests, then I'm not sure if it's better to do one 32 and one 64, or both 32. There's pros and cons to both. I don't think it really matters a whole lot, really as long as we're testing both Python 2 and 3, with at least one being 32-bit, on Windows.

hunse:
I'm happy with this. @tbekolay, are you fine with my additional commit? @drasmuss, do you want to take a look and merge if you're happy?

tbekolay:
New commit looks good; lemme add a commit to just do those 3 tests then this is good to go.

tbekolay:
OK, should be ready to review/merge now @drasmuss!

PR #998, opened 2016-03-28
Learning rule error size is connection size out

hunse:
As discussed in #957, our error signal should always be in the
dimensionality of the transform output. This fixes the learning
rule error size to adhere to that.

Even though this removes a ValidationError, we're already checking that the pre object is an Ensemble or Neurons in ConnectionLearningRuleTypeParam above, so I think we're fine.

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -450,14 +450,9 @@ class LearningRule(object):
         elif self.error_type == 'scalar':
             return 1
         elif self.error_type == 'decoded':
-            if isinstance(self.connection.pre_obj, Neurons):
-                return self.connection.pre_obj.ensemble.dimensions
-            elif isinstance(self.connection.pre_obj, Ensemble):
-                return self.connection.size_out
-            else:
-                tname = self.connection.pre_obj.__class__.__name__
-                raise ValidationError("Cannot learn on type %r" % tname,
-                                      attr='size_in', obj=self)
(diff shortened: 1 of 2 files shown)


tbekolay:
@drasmuss and/or @arvoelke can you look at this? I'm pretty sure this is correct.

tbekolay:
Buuut it fails unit tests so something about this isn't right. Hm.

hunse:
Okay, fixed it.

drasmuss:
Looks good.

arvoelke:
With fixes like this I think it's essential to add a new unit test that passes with the change and fails without it. It not only helps document the fix, but ensures that no future change accidentally reverts what you've done here.

tbekolay:
OK, I added 374ff1a838 which modifies `test_pes_multidim_error` so that it fails without this PR, and passes with it. The test is a bit longer with this, but making it into a `parametrized` thing or some other generic thing would be even longer, so i think the repetition is fine here.

If two of @hunse @arvoelke and @drasmuss are cool with the change I'll merge :rainbow: :octopus:

arvoelke:
Thanks, LGTM

hunse:
LGTM too. We could also test the neurons->neurons case in that test, but I think we sort of test that somewhere else. (At some point we should make the learning rule tests more systematic. Possibly as part of fixing #1001.)

Issue #997, opened 2016-03-27
WhiteSignal not suitable for integration with different dt


arvoelke:
```
for dt in (0.0001, 0.002):
    with nengo.Network() as model:
        stim = nengo.Node(output=nengo.processes.WhiteSignal(1.0, rms=0.5))
        p = nengo.Probe(stim, synapse=0.02)

    sim = nengo.Simulator(model, dt=dt, seed=0)
    sim.run(1.0)
    
    pylab.figure()
    pylab.title("dt=%s (RMS=%.3f)" % (dt, nengo.utils.numpy.rms(sim.data[p])))
    pylab.plot(sim.trange(), sim.data[p])
    pylab.xlabel("Time (s)")
    pylab.show()
```

![download 1](https://cloud.githubusercontent.com/assets/5420057/14063778/998e8172-f3ae-11e5-97d7-3d711164467d.png)
![download 2](https://cloud.githubusercontent.com/assets/5420057/14063779/9ab1b45c-f3ae-11e5-9408-1f69047b09ff.png)

Even though the RMS of the unfiltered signal is invariant to `dt` (note: it will be `0.5` for both cases when the synapse is `None`), this becomes no longer true once the signal is filtered/integrated.

As mundane as this seems, this is actually a pretty serious issue, because it means that any network using a `WhiteSignal` process as input to either a LIF neuron or a synapse will almost certainly break if the `dt` is decreased significantly. To see why, consider that a LIF neuron with default `tau_rc = 0.02` will lowpass filter its input in basically the same way as a synapse with the same time constant. And then the above example shows that such a filter will change the dynamic range (and so you must rescale the input, or change the RMS).

It took me way too long to debug this issue on my network as I was varying `dt`, because I thought that things would be okay as long as the unfiltered RMS was staying the same (I suspect others would fall into this same trap). We could possibly make the RMS dependent on `dt`, as we do for the `WhiteNoise` process, and then make this the default behaviour since all input signals to a network inevitably end up being filtered somewhere? However something that is not obvious to me is whether this can be made to play nicely with the `high` parameter.

hunse:
[...]

PR #999, opened 2016-03-28
Require high parameter for WhiteSignal

arvoelke:
Resolves issue #997. TL;DR: `high=None` was a bad default, because changing the `dt` will change which frequencies are present in the input signal. Instead, we should always require a cutoff, so that things stay the same for different `dt`. We now also check during build time that this does not exceed the Nyquist frequency for the given `dt`.

Note: Two of the tests were merged into one by use of `pytest.mark.parametrize`. GitHub's differ makes this hard to see...

### DIFF
--- a/nengo/processes.py
+++ b/nengo/processes.py
@@ -155,17 +155,18 @@ class WhiteSignal(Process):
         Samples will repeat after this duration.
     high : float, optional
         The cut-off frequency of the low-pass filter, in Hz.
-        If not specified, no filtering will be done.
+        Must not exceed the Nyquist frequency for the simulation
+        timestep, that is `high <= 0.5/dt`.
     rms : float, optional
         The root mean square power of the filtered signal. Default: 0.5.
     seed : int, optional
         Random number seed. Ensures noise will be the same each run.
     """
     period = NumberParam('period', low=0, low_open=True)
-    high = NumberParam('high', low=0, low_open=True, optional=True)
+    high = NumberParam('high', low=0, low_open=True)
     rms = NumberParam('rms', low=0, low_open=True)
 
-    def __init__(self, period, high=None, rms=0.5, seed=None):
+    def __init__(self, period, high, rms=0.5, seed=None):
         super(WhiteSignal, self).__init__(seed=seed)
         self.period = period
         self.high = high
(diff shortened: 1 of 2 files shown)


arvoelke:
Also note that I removed the ability to even pass `high=None`. I think this is for the best, since it seems especially unusual to want all of the frequencies if you don't even know how high those frequencies go. The extra hassle of needing to specify the upper limit in this case should then serve as a useful reminder that it does depend on dt.

nengo/tests/test_processes.py
```diff
@@ -99,7 +100,7 @@ def test_gaussian_whitenoise(Simulator, rms, seed, plt):
 def test_whitesignal_rms(Simulator, rms, seed, plt):
     t = 1.
     d = 500
-    process = WhiteSignal(t, rms=rms)
+    process = WhiteSignal(t, high=500, rms=rms)
```
hunse: This isn't a huge deal, but since `high` is no longer a keyword argument I'd like to remove the `high=` everywhere we use WhiteSignal, in both tests and examples.

tbekolay: Why would you want to remove it? If were were using 3.5 I would force it to be passed as a keyword arg even if it is required; bare numbers and booleans are rarely understandable in an argument list without a keyword or a well-named variable.

hunse:
> Also note that I removed the ability to even pass `high=None`.

I think that's a good idea.

Other than my one comment, I think this is good to me.

nengo/tests/test_processes.py
hunse: Yeah, I guess it's just the way I've always done it. I'm used to programming languages where all you have is parameter order. But you're right, it does help with readability, so it's probably good to keep it the way it is. But why don't we do the same for the period, then?

tbekolay: We should!

arvoelke: I also like the added readability. But also curious, @tbekolay where can I find info on that about 3.5? I tried a couple Google searches, but evidently I'm not using the right keywords (pun intended).

nengo/processes.py
```diff
@@ -164,7 +166,7 @@ class WhiteSignal(Process):
     high = NumberParam('high', low=0, low_open=True, optional=True)
```
arvoelke: Whoops, forgot to remove `optional=True`.

nengo/tests/test_processes.py
tbekolay: It's [PEP-3102](https://www.python.org/dev/peps/pep-3102/)... googling around a bit, I think it's in all of Python 3? So I misspoke in saying that it's 3.5 specific!

tbekolay:
Was merging this, but had a thought that a changelog should be added. But then had another thought that that should actually happen in #955 as processes weren't part of 2.0. So continuing to merge!

PR #1003, opened 2016-03-31
Respect input dtype in Synapse.filt

hunse:
This ensures that the output dtype is the same as the input dtype,
and fixes some tests that were failing in Nengo OCL.

### DIFF
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -209,11 +209,12 @@ class Process(FrozenObject):
     default_dt = NumberParam('default_dt', low=0, low_open=True)
     seed = IntParam('seed', low=0, high=maxint, optional=True)
 
-    def __init__(self, default_size_in=0, default_size_out=1, seed=None):
+    def __init__(self, default_size_in=0, default_size_out=1,
+                 default_dt=0.001, seed=None):
         super(Process, self).__init__()
         self.default_size_in = default_size_in
         self.default_size_out = default_size_out
-        self.default_dt = 0.001
+        self.default_dt = default_dt
         self.seed = seed
 
     def get_rng(self, rng):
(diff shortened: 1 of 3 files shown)


hunse:
Also, this is something that we used to do, and got lost when I redid Synapses. So it shouldn't need a changelog entry.

tbekolay:
Cool cool, mostly LGTM. Should we be using `dtype=np.float64`  rather than `float`? Is there a reason to prefer one over the other? `float64` seems more explicit to me, but that's the only argument I have one way or the other.

hunse:
They're the same. I used "float" because it's shorter :) But maybe `np.float64` is clearer. I'm happy either way.

hunse:
Okay, I changed to using ``np.float64`` for clarity.

I also added another commit to use the process ``default_dt`` in ``filt``, so that ``filt`` and ``apply`` work (mostly) the same.

hunse:
Added another commit to make `default_dt` actually changeable and therefore useful.

tbekolay:
I'm not a huge fan of all the `**kwargs` here, but I think the alternative is not much better, so I'm cool with it. Since the synapse changes haven't been released, this doesn't need a changelog, correct?

@drasmuss can you give this a second review?

hunse:
Yes, should be fine without a changelog.

drasmuss:
Looks good.  

The explicit `float64`s reminded me of something I've thought of before, which is that it'd be handy to be able to specify Nengo's float precision through the config.  Definitely a wishlist item for later on, just writing it down here while I think of it.

tbekolay:
> it'd be handy to be able to specify Nengo's float precision through the config

Bro, I totally got you covered: #640

Merging!

PR #1004, opened 2016-03-31
Ensure cache doesn't crash Nengo

tbekolay:
I had an issue in which Nengo crashed opening the cache's index file. I had just run some examples in Python 3, which seems to save with pickle protocol 4; when I ran the same example in Python 2, I got a `ValueError` that the protocol was too high.

This PR patches that by effectively disabling the cache if we encounter a `ValueError`. But this isn't a great solution because it means you'll *never* be able to use the cache as it'll keep hitting the `ValueError`.

A few ways forward that I can see:

1. Detect a wrong pickle version earlier in the process so the cache can be invalidated and started over.
2. Use a format other than pickle (json, csv, npz, custom plaintext) so that the Python version doesn't matter (the NCO format should work the same regardless of Python version right?)

If we can implement any of these quickly, then I'd prefer to go with that than this PR. Especially pinging @jgosmann for thoughts / opinions!

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -168,6 +168,7 @@ class DecoderCache(object):
     _INDEX = 'index'
     _LEGACY = 'legacy.txt'
     _LEGACY_VERSION = 1
+    _PICKLE_PROTOCOL = pickle.HIGHEST_PROTOCOL
 
     def __init__(self, readonly=False, cache_dir=None):
         self.readonly = readonly
@@ -284,16 +285,21 @@ class DecoderCache(object):
         legacy_file = os.path.join(self.cache_dir, self._LEGACY)
         if os.path.exists(legacy_file):
             with open(legacy_file, 'r') as lf:
-                version = int(lf.read().strip())
+                text = lf.read()
+            try:
+                lv, pp = tuple(int(x.strip()) for x in text.split('.'))
+            except ValueError:
+                # Will be raised with old legacy.txt format
+                lv = pp = -1
         else:
-            version = -1
-        return version == self._LEGACY_VERSION
+            lv = pp = -1
+        return lv == self._LEGACY_VERSION and pp == self._PICKLE_PROTOCOL
 
     def _write_legacy_file(self):
         """Writes a legacy file, indicating that legacy files do not exist."""
         legacy_file = os.path.join(self.cache_dir, self._LEGACY)
         with open(legacy_file, 'w') as lf:
-            lf.write("%d\n" % self._LEGACY_VERSION)
+            lf.write("%d.%d\n" % (self._LEGACY_VERSION, self._PICKLE_PROTOCOL))
 
     def _remove_legacy_files(self):
         """Remove files from now invalid locations in the cache.
--- a/nengo/tests/test_processes.py
+++ b/nengo/tests/test_processes.py
@@ -278,7 +278,7 @@ def test_present_input(Simulator, rng):
         sim.run(1.0)
 
     t = sim.trange()
-    i = np.floor((t - sim.dt) / pres_time + 1e-7) % n
+    i = (np.floor((t - sim.dt) / pres_time + 1e-7) % n).astype(int)
     y = sim.data[up].reshape(len(t), c, ni, nj)
     for k, [ii, image] in enumerate(zip(i, y)):
         assert np.allclose(image, images[ii], rtol=1e-4, atol=1e-7), (k, ii)


jgosmann:
> This PR patches that by effectively disabling the cache if we encounter a ValueError

The patch here keeps the cache enabled, but it will forget all previously cached results.

> (the NCO format should work the same regardless of Python version right?)

Nope, NCO is also writing some pickled data with the highest available protocol.

We could use different index files depending on the pickle protocol version. Only downside to this is that this could more easily lead to entries in the index file that are outdated and won't get removed from the index. This wouldn't lead to any crashes, but increases the index size.

tbekolay:
> The patch here keeps the cache enabled, but it will forget all previously cached results.

Effectively disables ;) Which is to say, you won't get any of the benefits of having the cache. The thing is that this error happens in the `CacheIndex` object, so you can't affect the cache itself here. You'd need to catch the error higher up in the stack, I was just too lazy to figure out where that is.

tbekolay:
> Nope, NCO is also writing some pickled data with the highest available protocol.

OK, then I'll go with 1. and invalidate the cache if the index has the wrong pickle version.

jgosmann:
> Effectively disables ;)

It will still cache newly computed decoders and these will be loadable afterwards.

tbekolay:
OK, I tried to do this a bit more intelligently. After a bit of hacking I realized that I was essentially duplicating what the legacy file does.

So, the current commit augments the legacy file to include the pickle protocol. The cache will be wiped and started over if the pickle protocol in `legacy.txt` doesn't match the current `cache._PICKLE_PROTOCOL`, which is set to `pickle.HIGHEST_PROTOCOL`. This is a bit of a pain to unit test, but starting from master, if I check this out in a Python 3 venv and run a model, the cache is wiped and the `legacy.txt` reads `1.4`. If I then checkout a Python 2 venv and run a model, the cache is wiped and the `legacy.txt` reads `1.2`.

This PR also includes a tiny fix for a test error I was getting in Python 2 on a dev version of NumPy; we were indexing into something with a float, which in a dev version raises an exception. I can move this to another PR if desired, but it's pretty innocuous.

nengo/cache.py
```diff
@@ -284,16 +285,21 @@ def _check_legacy_file(self):
...
+            try:
+                lv, pp = tuple(int(x.strip()) for x in text.split('.'))
+            except ValueError:
+                # Will be raised with old legacy.txt format
+                lv = pp = -1
         else:
-            version = -1
-        return version == self._LEGACY_VERSION
+            lv = pp = -1
+        return lv == self._LEGACY_VERSION and pp == self._PICKLE_PROTOCOL
```
jgosmann: We could check for `pp <= self._PICKLE_PROTOCOL`. That would allow to switch to newer protocols without wiping the cache. However, we have to make sure that the legacy file is updated with the higher pickle protocol version.

jgosmann:
Made a comment inline, apart from that LGTM. :cake:

nengo/cache.py
tbekolay: Hmm... is it worth it, you think? I think switching Python / pickle protocols is going to be a pretty rare event, I'm fine wiping it unnecessarily to keep the code simpler.

tbekolay:
@drasmuss Can you give this a review? And an opinion on the inline comment ;)

nengo/cache.py
drasmuss: I'd just wipe the cache whenever things change.  Wiping the cache is not that bad, just (maybe) makes things slightly slower on the next build.  Seems worth it to just keep things clean and not  have to worry about much more disruptive errors.

drasmuss:
Looks good.

tbekolay:
Care to merge @drasmuss ?

nengo/cache.py
```diff
@@ -284,16 +285,21 @@ def _check_legacy_file(self):
         legacy_file = os.path.join(self.cache_dir, self._LEGACY)
         if os.path.exists(legacy_file):
             with open(legacy_file, 'r') as lf:
-                version = int(lf.read().strip())
+                text = lf.read()
+            try:
+                lv, pp = tuple(int(x.strip()) for x in text.split('.'))
```
drasmuss: Is there any chance we would have periods in the legacy version (e.g. `1.1`)?  Could just write the `lv`/`pp` to separate lines if we wanted to avoid any issues.

jgosmann: The legacy version should always be an int, I would be more worried about the pickle version.

tbekolay: I assume the legacy version will always be an integer... I can't imagine we'll ever go so high that ints aren't sufficient. I just thought that `x.y` looked more like typical versions than having two separate things, or having a comma separating them (which is what I did initially).

PR #1005, opened 2016-04-05
Updated parameter definition to new syntax

tcstewar:
I'm not sure why this didn't cause an error before, but this update is needed due to adding the ```name``` into the Parameter system.

### DIFF
--- a/nengo/spa/__init__.py
+++ b/nengo/spa/__init__.py
@@ -11,5 +11,5 @@ from .pointer import SemanticPointer
 from .spa import SPA
 from .state import State
 from .thalamus import Thalamus
-from .utils import enable_spa_params, similarity
+from .utils import similarity
 from .vocab import Vocabulary
--- a/nengo/spa/spa.py
+++ b/nengo/spa/spa.py
@@ -6,7 +6,6 @@ import nengo
 from nengo.exceptions import SpaModuleError
 from nengo.spa.vocab import Vocabulary
 from nengo.spa.module import Module
-from nengo.spa.utils import enable_spa_params
 from nengo.utils.compat import iteritems
 
 
@@ -82,7 +81,6 @@ class SPA(nengo.Network):
                  vocabs=None):
         super(SPA, self).__init__(label, seed, add_to_container)
         vocabs = [] if vocabs is None else vocabs
-        enable_spa_params(self)
         self._modules = {}
         self._default_vocabs = {}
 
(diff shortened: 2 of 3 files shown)


tbekolay:
I guess we don't actually use this in any tests?

The actual `VocabularyParam` needs to be changed as well: https://github.com/nengo/nengo/blob/master/nengo/spa/vocab.py#L486

tbekolay:
And, we should probably add a test for this...

jgosmann:
I'm currently working on allowing hierarchial SPA modules and might touch the exact same lines. But I'm wondering what exactly is that additional `VocabularyParam` used for? Is it just for display purposes in the GUI?

jgosmann:
[I dug around a bit why this code was introduced.](https://github.com/nengo/nengo/pull/604) Apparently for the javaviz visualizer. But are we using it in the nengo_gui? If not, shall we remove it?

tcstewar:
> I dug around a bit why this code was introduced. Apparently for the javaviz visualizer. But are we using it in the nengo_gui? If not, shall we remove it?

I think you're right that it shouldn't be used any more...   That might explain why no tests triggered on it!

jgosmann:
If we don't want to wait for #1006 being merged (though it is marked as reviewed), these lines should just me removed instead of merging this PR because they are obsolete.

hunse:
I think I agree, but in this PR the default for the `VocabularyParam` is set to `None`, whereas in master it's not set at all, making it `Unconfigurable` and thus the parameter readonly. Is this what was intended? Apparently I changed the line in master. I think what's in this PR is actually what we want, though. Is this tested anywhere? Can someone with SPA knowledge test this please?

jgosmann:
The code changed in this PR is never hit (except maybe if someone still uses javaviz). Thus, it is pretty much irrelevant what this code does and it is just confusing to have unused and outdated code around. The vocab param that's getting manipulated isn't used anywhere either I believe.

(In #1006 I actually did delete this code and couldn't find any problems caused by that.)

jgosmann:
I added a commit (to be squashed) that completely removes the function as there seems to be no reason to keep it around (see previous comments). As #1006 won't be merged into master but will become it's own package, it still makes sense to fix this on Nengo master.

hunse:
So it looks like this has changed into something completely different. The change that @tcstewar made in the original commit isn't even used anymore, since all that code is just removed. @jgosmann, why don't you make your commits into a stand-alone one about removing the unnecessary javaviz stuff (might as well make another PR while you're at it, just for clarity). I think at this point @tcstewar's commit is just confusing things.

jgosmann:
:point_right: #1165

Issue #886, opened 2015-10-26
Default vocab assignment in SPA modules is awkward


Seanny123:
As discussed [here](https://github.com/nengo/nengo/pull/849), the way default vocabs are assigned in SPA modules is badly documented.

s72sue:
These are two separate things:
1) Default vocabs get assigned when no vocab is passed to a SPA module (i.e., vocab=None)
2) An error is thrown when a vocab is passed to a SPA module, but its dimensions do not match the dimensions of the module. 

1) needs to be documented, but I don't see a problem with the error thrown in 2). It makes it clear that the dimensions of the vocab should be the same as that of the module.

Seanny123:
Yeah, I made a mistake and thought that there was no error thrown. You're right the error thrown is actually fine and the other thing just needs to be documented.

jgosmann:
Moved to nengo/nengo_spa#17.

Issue #749, opened 2015-06-18
Can't make SPA subnetwork in SPA network


Seanny123:
I was struggling with getting this code to work.

```
import nengo
import nengo.spa as spa
import numpy as np
from nengo.dists import Uniform
from nengo.dists import Choice

# ------------- Network functions --------------

def feature_extract(dimensions, subvocab, feature, label="feature extract"):
    """Extract and clean"""
    with spa.SPA(label=label) as net:
        net.input = nengo.Node(size_in=dimensions)
        net.cleaner = spa.AssociativeMemory(subvocab)
        net.output = nengo.Node(size_in=dimensions)

        nengo.Connection(net.input, net.cleaner.input, transform=((~feature).get_convolution_matrix()), synapse=None)
        nengo.Connection(net.cleaner.output, net.output, synapse=None)

    return net

# ------------- Prameters to use --------------
D = 64

vocab = spa.Vocabulary(D, unitary=True )
vocab.parse('SHAPE')
vocab.parse('COLOR')
vocab.parse('POS')
vocab.parse('LEFT')
vocab.parse('BOTTOM')
vocab.parse('RIGHT')
vocab.parse('SQR')
vocab.parse('TRI')
vocab.parse('CIRC')
vocab.parse('BLUE')
vocab.parse('RED')
vocab.parse('GREEN')

color_vocab = vocab.create_subset(["BLUE", "RED", "GREEN"])
shape_vocab = vocab.create_subset(["SQR", "TRI", "CIRC"])

model = spa.SPA()
with model:
    model.sample = spa.Buffer(D, vocab)
    model.sample_input = spa.Input(sample="SHAPE*SQR+COLOR*RED")
    shape_extract = feature_extract(D, shape_vocab, vocab["SHAPE"], "shape extract")

    nengo.Connection(model.sample.state.output, shape_extract.input)
```

The fix ended up being that I had to change the `spa.SPA()` network to a plain old `nengo.Network()`. Why is this?

s72sue:
Were you getting any errors while running it with spa.SPA() ? I just tried it and it seems to run fine with no errors.

Seanny123:
Unfortunately, I made this issue so long ago that I can't remember if the code didn't work or I was getting errors. At this point, I don't get errors, but I'm also not getting the functionality I expected... I'll take a look at this again later.

Issue #695, opened 2015-04-07
Set up SPA to use config system


drasmuss:
This is just a general wishlist item, but it would be nice if we could use the config system with SPA models (e.g., to set `neurons_per_dimension` for all `Buffers` at once).  I don't think this would require any changes to the existing setup if we did want to do it, it would just be a matter of refactoring `spa` objects to be `NengoObjects` and using `Parameters`.  Not urgent though.

drasmuss:
I should also say, there's a reasonable case for not doing this at all (to preserve the distinction between core Nengo objects and the SPA system).

drasmuss:
A sub-issue of this one: In some places SPA manually sets a parameter that is normally controlled by the config system (as opposed to adding new parameters that aren't in the config system like I was talking about above).  For example, `Buffer` has a separate `direct` parameter that controls the neuron type (by allowing it to be either `Direct` or `LIF`).  If we just didn't set the neuron type within `Buffer`, we could control it like this:
```python
with nengo.Network() as net:
    net.config[nengo.Ensemble].neuron_type = nengo.LIFRate()
    buff = spa.Buffer(...)
```
which would be more consistent with the rest of Nengo, and more flexible.  I suspect this is just because the `spa` code was written before we had the `config` system in place.  I could make this change myself, but I want to check if people like it first (since it would require small modifications to existing models that use the current parameter system).

tcstewar:
I definitely agree this should happen.  Especially for things like the Direct option.  It's the way it currently is just because the config stuff wasn't in place when the spa stuff got written.  This change would make a lot of sense to me....

Seanny123:
+1 for extending config to SPA

I think the distinction between core Nengo and SPA is important, but it's generally pretty obvious once you start making SPA-based models that you're not in Kansas anymore.

jgosmann:
Basic classes to do:
- [x] Scalar
- [x] State
- [ ] Bind
- [ ] Compare
- [ ] Input

More complex classes:
- [ ] AssociativeMemory
- [ ] BasalGanglia
- [ ] Thalamus
- [ ] Cortical

Deprecated classes to delete:
- [ ] Buffer
- [ ] Memory
- [ ] SPA

PR #1006, opened 2016-04-05
Allow hierarchical SPA modules.

### LINKED ISSUE #556: SPA: on_add is not called on submodules
I was trying to write a SPA `Module` containing a `Cortical` object:

``` python
class MyModule(Module):
    def __init__(self):
        super(MyModule, self).__init__()
        # add a bunch of Buffer and Memory objects
        self.cortical = Cortical(Actions(
            # a bunch of actions
        )
```

In this case the `on_add` function of `self.cortical` never gets called and the required connections are missing in the final network. This might be the case for other kinds of modules too, but might not matter because no special things are done in `on_add`.

In my situations the cortical rules are pretty simple and I can add the required connections myself manually. But it took me a while to figure out, why it wasn't working.

jgosmann:
There are a few more things that should be done before reviewing/merging (see below), but it would be nice to get this looked over and see what the general opinion (and especially @tcstewar's opinion) on this is. Also because I plan a number of other changes to the SPA system that might be based on this.

Fixes #556, #749. Might touch on  #695, #886, #1005.

This allows to have SPA modules within other modules (see `test_module.py` for a simple example). As a result the distinction between `spa.SPA` and `spa.Module` is no longer needed. We could mark remove `spa.SPA` or mark it as obsolete. Any opinions on this? (This will of course require a number of tests and examples to be updated.)

Also, it probably makes sense to move the tests from `test_spa.py` to `test_module.py`.

Note that I removed some obsolete code as discussed in #1005.

### Using the config system for SPA modules

You might notice that the the `spa.Module` class now has a `vocabs` parameter. This makes me wonder whether it (or `Network`) should inherit from `NengoObject`) or whether there is some part that could or should be extracted into a separate mix-in class. But so far it this parameter requires only one more line.

But we might add (probably should add) more parameters as discussed in #695. Might that change things? This is also related to #977, #984.

### Addressing sub modules in actions

Currently, SPA Actions can only access modules defined on the same level of the hierarchy. I believe, it should not be to troublesome to extend it so that submodules can be accessed, but we need to agree on a syntax for that. I think the dot notation `module.submodule` is natural because it is the same as in actual Python.

Along those lines, I might want to change the syntax to access non-default inputs and outputs where we currently use an underscore (`module_input`). This might clash with other module names and seems less natural to me. I would like to change this to the dot notation too (`module.input`). This might of course break existing models.

### Adding modules twice

It is an error to add a module twice to to SPA module. Is there a technical reason for that? Or is there some kind of easy to overlook error that could get introduced by adding a module twice? I'm asking because this happens in the `Thalamus` module (it adds the basal ganglia to itself, but usually they will also be added to the top level module). It is not to complicated to circumvent this (see code for details), but maybe it makes more sense to allow adding a module multiple times. Of course this would allow to use different names to access the same module, but would this cause any problems?

(Note with adding I mean assigning it to a SPA module, independent of that a module gets only added once to a network at the moment it is instantiated.)


jgosmann:
Forgot to explain why this touches on #886. I'm not completely sure what exactly that issue is about, but with these changes `spa.Module` has a `vocabs` argument that allows to pass in a set of default vocabularies. These will be inherited by submodules. Or you can pass in a new `VocabularySet` to use a different set of default vocabularies for this submodule.

xchoo:
Could we rename `VocabularySet` to `VocabularyMap`?

jgosmann:
Sure, it has properties of both.

On April 5, 2016 8:58:02 PM EDT, xchoo <notifications@github.com> wrote:
>Could we rename `VocabularySet` to `VocabularyMap`?
>
>
>---
>You are receiving this because you authored the thread.
>https://github.com/nengo/nengo/pull/1006#issuecomment-206057548

-- 

xchoo:
Could we use the VocabularyMaps / sets to partially solve #1007 (it'll set up a way of defining the mapping, but doesn't quite solve the syntax issue)?

xchoo:
i.e. the how to define it in the spa actions system syntax issue

jgosmann:
I don't think that it brings us a lot closer to a solution.

On April 5, 2016 9:01:31 PM EDT, xchoo <notifications@github.com> wrote:
>Could we use the VocabularyMaps / sets to partially solve #1007 (it'll
>set up a way of defining the mapping, but doesn't quite solve the
>syntax issue)? 
>
>---
>You are receiving this because you authored the thread.
>https://github.com/nengo/nengo/pull/1006#issuecomment-206058534

-- 

tcstewar:
Nice.  I definitely like getting rid of the distinction between the spa.SPA and spa.Module.  

I also like the long-term plan of getting rid of the ```_``` notation.  I agree that ```.``` is the way to go, and to use ```.``` for both multiple inputs/outputs and submodules.

As for adding modules twice, yes, getting rid of that restriction seems fine to me.  And  I don't see a problem with having multiple ways to reference the same object.

As for the vocabs thing, I'm still generally unhappy with that weird "default for a number of dimensons" thing, and the whole ```self.inputs``` and ```self.outputs``` approach but I still don't see a nicer design.  Definitely open to other options....

jgosmann:
> As for adding modules twice, yes, getting rid of that restriction seems fine to me. And I don't see a problem with having multiple ways to reference the same object.

I guess, one problem might be if the `on_add` function is not idempotent.

jgosmann:
And there actually seems to be no check about adding things twice, but in the Basal Ganglia/Thalamus case an assigned module gets overwritten which is an error (and should be one).

jgosmann:
> I also like the long-term plan of getting rid of the _ notation. I agree that . is the way to go, and to use . for both multiple inputs/outputs and submodules.

This won't allow to have a `spa.Input` feeding into a submodule or into a non-default input. At least not in an obvious way. The non-obvious way would be `spa.Input(**{'module.submodule': 'A', 'module.input': 'B'})`. Is that good enough? Or should we change spa.Input to take an explicit dictionary instead of implicit keyword args to make it more obvious how to address names with dots in them? Or should we switch to a completely different syntax?

jgosmann:
To dos:
- [ ] Address submodules in actions
- [ ] Address non-default inputs in actions with dot notation
- [ ] Handle submodules in `spa.Input`
- [ ] Handle non-default inptus in `spa.Input` with dot notation
- [ ] Move tests in `test_spa.py` to `test_module.py`
- [ ] Mark `spa.SPA` as deprecated
- [ ] Make everything using `spa.SPA` use `spa.Module` (tests and examples)

jgosmann:
I think this should be ready for initial review. There is probably one remaining discussion point on how to handle submodules and non-default inputs in `spa.Input`. I added a commit to still support the underscore notation for inputs (`module_input` instead of `module.input`) with a `DeprecationWarning`. I would be happy to get rid of it as it complicates the source code (and I think the name resolution code is already complex enough). But not including it will break backwards compatibility (so maybe we want to keep it in for one minor version). Also we have to decide for a `spa.Input` syntax. Here are the options:

1. Keep the keyword arguments: `spa.Input(a='A', b='B')`, in case of submodules/non-default inputs: `spa.Input(**{'module.input': 'A'})`. Pros: backwards compatibility, most common use case is convenient; Cons: Input to submodules uses obscure and non-obvious syntax.
2. Have a single dict argument instead: `spa.Input({'a': 'A', 'b': 'B'})`. Pros: Addressing of submodules is clearer; Cons: breaks backward compatibility.
3. Mix 1 and 2. There are different ways in which this could be done, but for example keyword arguments could be interpreted as before while accepting up to one positional dict argument. Pros: best of both worlds?? Cons: Examples would primarily use one form, thus the other form is non-obvious. Doesn't follow the Zen of Python: "There should be one-- and preferably only one --obvious way to do it."
4. Something completely different?

tcstewar:
Gah.  This ```spa.Input``` thing really seems to be the only place that's left where the dot notation is problematic (in my original design of ```spa``` there were many more of these).

I think I'd lean towards 3 or 1.  I definitely don't like 2, as it makes the most common use cases awkward.  I'm leaning toward 3 just because it's not clear to me yet what the best way is to go....

jgosmann:
Just got another idea:
```python
model_input = spa.Input()
model_input.module.input = 'A'
```
This could also easily be combined with 1.

tcstewar:
Oooo... pretty.  I like that.   And I could see supporting both that and method 1, giving us backwards compatibility.....   And maybe even gradually transition over to always using that new syntax so that there's one standard way to do it.

Seanny123:
Is there anything that still needs to be discussed/developed or can this now be reviewed?

jgosmann:
Potential remaining discussion point:

> I added a commit to still support the underscore notation for inputs (module_input instead of module.input) with a DeprecationWarning. I would be happy to get rid of it as it complicates the source code (and I think the name resolution code is already complex enough). But not including it will break backwards compatibility (so maybe we want to keep it in for one minor version).

Otherwise, ready for review.

jgosmann:
Who will review this? @tcstewar?

tcstewar:
> Who will review this? @tcstewar?

Yup, that's my plan over the next two days.

nengo/spa/module.py
```diff
@@ -25,3 +51,159 @@ def on_add(self, spa):
...
+        except KeyError:
+            raise SpaModuleError("Could not find module input %r" % name)
+
+    def get_module_inputs(self):
+        for name, module in iteritems(self._modules):
+            for inp in module.inputs:
+                if inp == 'default':
+                    yield name
+                else:
+                    yield '%s_%s' % (name, inp)
```
tcstewar: Should this now use the "." syntax rather than "_"?

nengo/spa/spa.py
```diff
@@ -80,153 +76,20 @@ def __init__(self):
 
     def __init__(self, label=None, seed=None, add_to_container=None,
                  vocabs=None):
-        super(SPA, self).__init__(label, seed, add_to_container)
+        warnings.warn(DeprecationWarning(
+            "nengo.spa.SPA is deprecated. Use nengo.spa.Module instead."))
```
tcstewar: Do we need to deprecate SPA?

nengo/spa/spa.py
```diff
@@ -80,153 +76,20 @@ def __init__(self):
...
+
+        if seed is not None:
+            rng = np.random.RandomState(seed)
+        else:
+            rng = None
         vocabs = [] if vocabs is None else vocabs
```
tcstewar: Why does the SPA constructor behave differently than the standard Module constructor?  Does this mean there are situations where we can't just replace a SPA with a Module?

nengo/spa/module.py
```diff
@@ -25,3 +51,159 @@ def on_add(self, spa):
...
+
+        module_list = frozenset(self._modules.values())
+        for net in self.networks:
+            # Since there are no attributes to distinguish what's been added
+            # and what hasn't, we have to ask the network
+            if isinstance(net, Module) and (net not in module_list):
+                raise SpaModuleError("%s must be set as an attribute of "
+                                     "a SPA network" % (net))
+
+    def get_module(self, name, strip_output=False):
```
tcstewar: A quick docstring saying what ```strip_output``` does would be handy.  It looks like it allows ```get_module``` to ignore the input or output identifier (so ```compare.A``` would return the ```compare``` module).  

The default in the old code was to always allow this sort of ambiguity -- was there a reason for changing this default and turning it into a parameter?

nengo/spa/module.py
jgosmann: I believe that function (and `get_module_outputs`) is only there for backwards compatibility (and thus deprecated). I think, it's not even called anymore.

But for keeping backwards compatibility it might be better to not change it?

nengo/spa/spa.py
jgosmann: I think it's cleaner because there is no reason to have two classes around basically doing the same thing.

nengo/spa/spa.py
jgosmann: `SPA` expects `vocabs` to be a list whereas `Module` expects a `VocabularyMap`. Thus, a replacement of `SPA` with `Module` might require the `vocabs` argument to be adjusted (which is trivial and should not change behavior, but has to be done).

Technically, this line would not be required as `VocabularyMap` does the exact same thing if it gets passed a `None`. Creating the `VocabularyMap` here allows to use a common `super` call, otherwise branching would be required to do the call with a varying number of arguments. Passing in the `VocabularyMap` to the constructor means that a potential config of `vocabs` will be overwritten, but prior to this PR this config param did not exist and `SPA` had to be the root module and thus this can't break any existing code (but another reason to deprecate SPA to avoid potential problems from that slightly different behavior). Creating the `VocabularyMap` in this place requires an `rng` which explains the lines above the `vocabs = ` line.

Not sure how clear this answer was ... I feel like it's a lot of minute details why exactly I implemented it in this way.

nengo/spa/module.py
jgosmann: Yes, ambiguity is bad. I want to be explicit. The function name is `get_module` and if I give it an input/output it should raise an error because it's not a module. Except in some circumstances where I actually have an input or output name (but potentially without the `.default` suffix) where I want to get the module corresponding to that input and output.

Where this actually might be important is with the SPA compiler (#1014). At least I would not be surprised if changing this would introduce some weird bugs. But the main reason might actually be the transition from the current BG/Thalamus to the compiler system. The current system needs `strip_output=True` for backwards compatibility, but the compiler uses `strip_output=False` in most places (except one). Not sure if this is strictly necessary, but as I said I don't know what sorts of bugs could otherwise be introduced and in case of future changes debugging might be easier if the system fails early because an input/output was passed instead of a module instead of returning a module anyways.

tcstewar: > Yes, ambiguity is bad. I want to be explicit.

Ah, I finally figured out why I was confused about this ambiguity.  In the past, the purpose of this function was primarily to get the module *for a particular input or output*.  So I was interpretting it as the unambiguous "get me the module for whatever name I give you -- if I give you a module name, then give me the module, and if I give you an input or output inside that module, then still give me the module".  

In any case, that change seems reasonable to me, and I can't find any place that it breaks, so it's all good.  :)

nengo/spa/module.py
tcstewar: It still seems like a useful function to me, just for introspection purposes.  So it may as well use the new syntax.

tcstewar:
Other than updating ```get_module_inputs``` and ```get_module_outputs``` to the new syntax, this all looks good to me.

I also ran it past @xchoo to let him know that this change causes a slight change to Spaun (the "." syntax rather than the "_" syntax for referring to particular inputs and outputs in a module).  He's fine with that change, and I've put together a small diff that seems to make Spaun work with this PR.

nengo/spa/module.py
jgosmann: But for introspection one can directly look into the `inputs` and `outputs`. The function is probably not even doing what one would expect it to do because it lists the inputs of all direct submodules (but not subsubmodules) (which makes sense without hierarchical modules, but not so much with hierarchical modules). So I would either keep it for backwards compatibility (and maybe add a deprecation warning) or delete it. We could discuss implementing a similar function which takes hierarchical modules into account (we would have to define what exactly it should return), but I don't see the need.

jgosmann:
> Other than updating get_module_inputs and get_module_outputs to the new syntax, this all looks good to me.

Commented on that inline.

nengo/spa/module.py
tcstewar: Hmm...  can we get rid of these two functions completely?  Are there any examples or demos or notebooks that use these functions?  As far as I can tell they're only used by the spa.Actions system, which you've rewritten, so it looks to me like we can just remove them.

jgosmann: Yes, I would be happy to remove them. I just did not do it because they are part of the public class interface and removing them would break theoretically break backwards compatibility (though it might be unlikely that it breaks anything in practice). I assume this would push this PR to 2.2? Though, it is not clear to me that it should be included in 2.1.x anyways because it's clearly a new feature and not a bugfix.

jgosmann: Also, with all the SPA refactoring in follow-up PRs it would in general make things a bit easier to not keep backwards compatibility intact up to the smallest bits.

tcstewar: Tricky....  I'm very tempted to just declare that these shouldn't have been part of the public interface in the first place, and so I'm not all that worried about removing them.  I think it's less of a change than the change that's requiring replacing "_" with "." in Spaun.  That actually breaks existing code (but in an easy-to-fix way that results in better code), whereas removing this is only potentially breaking things.  So I don't think what we decide here would be the cause of pushing off this PR, and so let's just get rid of the functions.  You're quite right that they can be inspected in other ways, and the approach used in this function doesn't make sense in hierarchical SPA.

tcstewar: > Also, with all the SPA refactoring in follow-up PRs it would in general make things a bit easier to not keep backwards compatibility intact up to the smallest bits.

Exactly.  That's why I think the level of backwards compatibility we need is basically just "do our existing demos and examples still work".

jgosmann:
Removed those functions. Who can be the second reviewer?

tcstewar:
> Removed those functions.

Hmm, looks like you removed ```get_input_vocab``` and ```get_output_vocab``` too.  Those get used a fair bit, and still seem useful, even if they don't yet support hierarchical SPA.  So I'm pretty sure those should stay.

>  Who can be the second reviewer?

Not sure...  maybe @Seanny123 ?  He at least had to play with some of these things when working on gui components....

jgosmann:
> Hmm, looks like you removed get_input_vocab and get_output_vocab too. Those get used a fair bit, and still seem useful, even if they don't yet support hierarchical SPA. So I'm pretty sure those should stay.

It didn't break any tests ... 😐 but I can bring them back.

tcstewar:
> It didn't break any tests ...

But it does break examples, such as 

https://github.com/nengo/nengo/blob/master/examples/spa/spa_parser.ipynb

Which indicates that our tests have some pretty massive holes in them....

jgosmann:
But aren't the examples run as part of the (slow) tests?

tcstewar:
> But aren't the examples run as part of the (slow) tests?

Yup -- I just tried running them on this branch and they failed with ```AttributeError: 'Module' object has no attribute 'get_output_vocab'```

(the command I used was ```py.test --slow .\test_examples.py```)

jgosmann:
Aren't the slow tests then not run on Travis? (@tbekolay?)

jgosmann:
Added the functions back in. Currently running all tests.

jgosmann:
All tests passed.

tcstewar:
This looks good to me!  Do we want to have a second reviewer as well?  @tbekolay what are your thoughts?

nengo/spa/actions.py
```diff
@@ -59,14 +56,21 @@ def validate_string(self, text):
...
-                    "Semantic pointers must begin with a capital letter.")
-            item = Symbol(key)
-            self.objects[key] = item
-        return item
+        if key == '__tracebackhide__':  # gives better tracebacks in py.test
+            return False
+        if key in self.objects:
+            return self.objects[key]
+        else:
+            try:
```
Seanny123: This logic took me a while to understand. An inline comment would be appreciated. Something like "attempt to create a submodule if it exists, otherwise create a normal object"?

Seanny123:
LGTM, other than a few parts where I got a little confused.

Made a note in #822 that we should remove support for the underscore notation and `spa.SPA()`, which unfortunately means the tutorial Su and I wrote is already going to be broken by Nengo 2.2.

tcstewar:
> Made a note in #822 that we should remove support for the underscore notation and spa.SPA(), which unfortunately means the tutorial Su and I wrote is already going to be broken by Nengo 2.2.

Why?  I thought our deprecation would last longer than that.  And when do you use the underscore notation in the tutorial?

Seanny123:
I don't use the underscore notation, but I use spa.SPA() which is also
deprecated

tcstewar:
> I don't use the underscore notation, but I use spa.SPA() which is also
> deprecated

Then I think we should consider keeping spa.SPA() around for longer.  No sense breaking published things...

Seanny123:
So when do you want to delay it until? 2.3?

tcstewar:
> So when do you want to delay it until? 2.3?

Or just keep it forever.  Or rename spa.Module to spa.SPA.  The only reason that @jgosmann got rid of it is that the hierarchical version doesn't need two separate classes for spa.SPA and spa.Module.  So he got rid of spa.SPA.  But we could also go with the other naming.

In fact, the more I think about it, the more it makes sense to call the generic class "SPA" rather than "Module", because it's not a Module any more.....  hmm....

jgosmann:
If we remove either class full backwards compatibility isn't given anymore. Making the main class `SPA` instead of `Module` will also decrease backwards compatibility slightly because the `vocabs` argument is of a different type.

tcstewar:
> If we remove either class full backwards compatibility isn't given anymore.

It's not about full backwards compatibility.  It's about sufficient backwards compatibility that the existing tutorials and examples still work.  There are no examples where we make a new subclass of Module, but all of the spa examples involve creating an object of type ```spa.SPA```.  I'm also fairly sure that none of the examples in the tutorials involve passing in a ```vocabs``` argument (although I'd want to double-check that).

Seanny123:
>  I'm also fairly sure that none of the examples in the tutorials involve passing in a vocabs argument (although I'd want to double-check that).

The publication that Su and I just submitted does use the `vocabs` argument. Consequently, how long do we want to delay deprecation of `spa.SPA()`? Nengo 2.3, which I'm assuming will happen two years from now?

tcstewar:
The more I think about it, the more I want to keep this syntax:

```python
model = spa.SPA()
with model:
    ...
```

rather than this:

```python
model = spa.Module()
with model:
    ...
```

mostly because the ```model``` is *not* a Module (in the standard cognitive science sense of a module as a self-contained component within a larger system).  It's the full model itself.  So I'd prefer either:

 - keeping spa.SPA and spa.Module separate (with slight different vocabs arguments)
 - renaming spa.Module as spa.SPA (and having a dummy deprecated spa.Module just to be nice to code that defines its own subclass of spa.Module, and maybe handling either version of the vocabs argument)

jgosmann:
> keeping spa.SPA and spa.Module separate (with slight different vocabs arguments)

-1
"There should be one-- and preferably only one --obvious way to do it."



> renaming spa.Module as spa.SPA

Fine with me, but I don't think we can handle backwards compatibility in a clean way.

I'd propose to move SPA into its own package separate from Nengo and have a `nengo_legacy_spa` with the current state and a rewritten `nengo_spa` that does not have to worry about backwards compatibility. (We could even have it that `nengo.spa` tries to import `nengo_legacy_spa`.) Let's discuss this at the dev meeting.

jgosmann:
Potentially, we can keep enough backwards compatibility by making sure that the `VocabularyMap` parameters accepts a list (but some behaviour not relevant to the tutorials might change).

tcstewar:
> I'd propose to move SPA into its own package separate from Nengo and have a nengo_legacy_spa with the current state and a rewritten nengo_spa that does not have to worry about backwards compatibility. (We could even have it that nengo.spa tries to import nengo_legacy_spa.) Let's discuss this at the dev meeting.

I very much like spinning off a new ```nengo_spa``` package that does not have to worry about backwards compatibility.  However, if we do that, I don't see the advantage to pulling the existing ```nengo.spa``` out to a separate ```nengo_legacy_spa``` package, as that just complicates the install process and tracking of dependencies.  And if people just install nengo, all the spa examples will fail to work.  Why not just leave ```nengo.spa``` as it is, and see how things go with the eventual new ```nengo_spa``` once it exists?   

(And yes, this seems like a good point for discussion at the dev meeting)  :)

tcstewar:
> Potentially, we can keep enough backwards compatibility by making sure that the VocabularyMap parameters accepts a list (but some behaviour not relevant to the tutorials might change).

That sounds really good to me.

jgosmann:
Removing @tbekolay as asignee and asigning myself because the current plan is to not merge this into nengo master, but a separate branch that will later be split of into a separate nengo_spa package.

As this has the "reviewed" label, but was lying around for a while: @tcstewar, @Seanny123 this still looks good to you? We can make final decisions on naming details (`spa.Module` vs `spa.SPA`) later before an official release of `nengo_spa`. Right now I would like to start the integration process to get things moving. 🚗

Seanny123:
This still looks good to me. I'm fine with moving it to `nengo_spa` and renaming `spa.Module()` to `spa.SPA()`. I understand that despite having the same name, it won't be backwards compatible, but I fel like people will expect this since the code is living in a separate repository.

jgosmann:
I will go ahead and merge this branch, @tcstewar said that this looks good to him before and had enough time to respond if he changed his mind.

jgosmann:
Merged into `new-spa`.

Issue #323, opened 2014-04-11
Specifying a function with data points, rather than a callable


tcstewar:
Sometimes you just have a set of input and output data points for the function we want to approximate, rather than an explicit function.  It'd be nice to be able to do this:

```python
points = [[1,1], [1, -1], [-1, 1], [-1, -1]]
result = [[1], [-1], [-1], [1]]

with model:
    a = nengo.Ensemble(neurons=100, dimensions=2)
    xor = nengo.Ensemble(neurons=50, dimensions=1)
    nengo.Connection(a, xor, eval_points=points, decoder_result=result)
```

or possibly this, which is closer to the old syntax for doing this:

```python
with model:
    a = nengo.Ensemble(neurons=100, dimensions=2)
    xor = nengo.Ensemble(neurons=50, dimensions=1)
    nengo.Connection(a, xor, eval_points=points, function=result)
```

I'm not sure which approach would be preferred (allowing function to be either a callable or an array versus adding a new parameter).

Either way, this type of Connection would need to raise an Exception if pre is a Node or is an Ensemble running in Direct mode.

hunse:
Reopening this exactly two years later! I'm starting to do some stuff with MNIST in Nengo, and using the dictionary approach, it's not super fast (6 seconds with 10000 training images, 32 seconds with 60000, just for turning the eval points back into an array of target points that I already had to start with). If we scale up to larger data sets (possibly with greater than 1 million data points) with larger dimensionality (so hashing the tuple will take longer), this could become burdensome. So I think it's worth it to build something like this in. It might make Nengo an easier sell to the ML crowd, too, since this is the kinda thing they love to do, and it would great to have it be easy and fast.

drasmuss:
This is in master now.

PR #1010, opened 2016-04-11
Connections can take points for functions

hunse:
When specifying a Connection function, points can now be passed to
implicitly define the function along with the eval points. This
addresses #323.

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -13,7 +13,7 @@ from nengo.params import (Default, Unconfigurable, ObsoleteParam,
                           BoolParam, FunctionParam)
 from nengo.solvers import LstsqL2, SolverParam
 from nengo.synapses import Lowpass, SynapseParam
-from nengo.utils.compat import is_iterable, iteritems
+from nengo.utils.compat import is_array_like, is_iterable, iteritems
 
 logger = logging.getLogger(__name__)
 
@@ -117,25 +117,70 @@ class EvalPointsParam(DistOrArrayParam):
 class ConnectionFunctionParam(FunctionParam):
     """Connection-specific validation for functions."""
 
+    def __set__(self, conn, function):
+        if function is None:
+            function_info = self.Info(function=None, size=None)
+        elif is_array_like(function):
+            array = np.array(function, copy=False, dtype=np.float64)
+            self.validate_array(conn, array)
+            function_info = self.Info(function=array, size=array.shape[1])
+        elif callable(function):
+            function_info = self.Info(function=function,
+                                      size=self.determine_size(conn, function))
+            self.validate_callable(conn, function_info)
+        else:
+            raise ValidationError("Invalid connection function type %r "
+                                  "(must be callable or array-like)"
+                                  % type(function).__name__,
+                                  attr=self.name, obj=conn)
+
+        self.validate(conn, function_info)
+        self.data[conn] = function_info
+
     def function_args(self, conn, function):
         x = (conn.eval_points[0] if is_iterable(conn.eval_points)
              else np.zeros(conn.size_in))
         return (x,)
 
-    def validate(self, conn, function_info):
-        super(ConnectionFunctionParam, self).validate(conn, function_info)
-        fn_ok = (Node, Ensemble)
-        function, size = function_info
-
-        if function is not None and not isinstance(conn.pre_obj, fn_ok):
+    def validate_array(self, conn, ndarray):
+        if not isinstance(conn.eval_points, np.ndarray):
             raise ValidationError(
-                "function can only be set for connections from an Ensemble or "
-                "Node (got type %r)" % type(conn.pre_obj).__name__,
+                "In order to set 'function' to specific points, 'eval_points' "
+                "must be also be set to specific points.",
                 attr=self.name, obj=conn)
 
+        if ndarray.ndim != 2:
+            raise ValidationError("array must be 2D (got %dD)" % ndarray.ndim,
+                                  attr=self.name, obj=conn)
+
+        if ndarray.shape[0] != conn.eval_points.shape[0]:
+            raise ValidationError(
+                "Number of evaluation points must match number "
+                "of function points (%d != %d)"
+                % (ndarray.shape[0], conn.eval_points.shape[0]),
+                attr=self.name, obj=conn)
+
+    def validate_callable(self, conn, function_info):
+        super(ConnectionFunctionParam, self).validate(conn, function_info)
+
+    def validate(self, conn, function_info):
+        function, size = function_info
         type_pre = type(conn.pre_obj).__name__
+
+        if function is not None:
+            if not isinstance(conn.pre_obj, (Node, Ensemble)):
+                raise ValidationError(
+                    "function can only be set for connections from an Ensemble"
+                    " or Node (got type %r)" % type_pre,
+                    attr=self.name, obj=conn)
+
+            if isinstance(conn.pre_obj, Node) and conn.pre_obj.output is None:
+                raise ValidationError(
+                    "Cannot apply functions to passthrough nodes",
+                    attr=self.name, obj=conn)
+
+        size_mid = conn.size_in if size is None else size
         transform = conn.transform
-        size_mid = conn.size_in if function is None else size
 
         if isinstance(transform, np.ndarray):
             if transform.ndim < 2 and size_mid != conn.size_out:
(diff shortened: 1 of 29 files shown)


nengo/connection.py
```diff
@@ -117,25 +117,70 @@ def validate(self, conn, distorarray):
...
+            self.validate_array(conn, array)
+            function_info = self.Info(function=array, size=array.shape[1])
+        elif callable(function):
+            function_info = self.Info(function=function,
+                                      size=self.determine_size(conn, function))
+            self.validate_callable(conn, function_info)
+        else:
+            raise ValidationError("Invalid connection function type %r "
+                                  "(must be callable or array-like)"
+                                  % function.__class__.__name__,
```
arvoelke: If the function failed the above checks then it might not have a class name attribute.

hunse: I was assuming that everything has a `__class__` attribute, but I don't think that's quite true. Better to use `type()` instead? There are probably other places we make this assumption, too (e.g. node outputs).

arvoelke: I think there are only a couple rare cases in Python 2 where this attribute won't exist. I'm not really up to snuff on Python 3 so I'm afraid I don't know what the different cases are, or what the best suggestion would be. `type` does sound more reliable, but again I don't know.

hunse: Ok, well I fixed it everywhere anyway.

hunse:
Okay, fixed this up. The first three commits are fixes to the code that help with the final commit, which actually allows function points to be passed to connections.

The first three commits can be separated out and merged independently if there are questions about the fourth. I'd like to get them in ASAP, since the second one in particular touches large amounts of the codebase and is annoying to rebase.

nengo/connection.py
```diff
@@ -233,9 +272,12 @@ class Connection(NengoObject):
     synapse : Synapse, optional \
               (Default: ``nengo.synapses.Lowpass(tau=0.005)``)
         Synapse model to use for filtering (see `~nengo.synapses.Synapse`).
-    function : callable, optional (Default: None)
+    function : callable or (n_eval_points, size_mid) array_like, \
+               optional (Default: None)
         Function to compute across the connection. Note that ``pre`` must be
         an ensemble to apply a function across the connection.
+        If an array is passed, the function is implicitly defined by the
+        array points and the provided ``eval_points``.
```
Seanny123: Maybe say "the array points referenced" since you explicitly don't copy them?

Seanny123: Also maybe mention that the array has to be 2D?

Seanny123:
I did my best at reviewing. Other than some documentation nitpicks, it LGTM.

nengo/connection.py
hunse: It gives the required shape above. Does that make it clear that it's 2D?

If I read "array points referenced" I'm not sure I would think that means they're passed by reference, i.e. not copied. Also, I think the assumption is that anything passed as a parameter is not copied, so if you pass an array and then change a value it will change that in your model. Of course, when everything is built, we make copies as needed so that the built model would never be changed.

Seanny123: > It gives the required shape above. Does that make it clear that it's 2D?

Aw jeez, I totally missed that notation.

> I think the assumption is that anything passed as a parameter is not copied

That's true. I forgot about that assumption when it comes to Python.

hunse:
Could this be looked at soon? It would be helpful for my tutorial.

nengo/tests/test_node.py
```diff
@@ -139,25 +139,31 @@ def test_circular(Simulator, seed):
     assert np.allclose(sim.data[a_p], sim.data[b_p])
 
 
-def test_function_args_error(Simulator):
+def test_outputparam_errors(Simulator):
     with nengo.Network() as model:
+        # type errors
```
jgosmann: This comment is misleading. The next few lines are perfectly fine and don't produce type errors. (Took me a few moments to realize why there is no check for raised exceptions for this lines.)

hunse: The last line of this section is a "type" error (i.e. the node output is not an acceptable type, not literally a TypeError, though). But you're right, this is mostly checking that the supported types work fine. Would you prefer no comment? Or a different comment?

nengo/connection.py
```diff
@@ -117,25 +117,70 @@ def validate(self, conn, distorarray):
...
+    def validate(self, conn, function_info):
+        function, size = function_info
         type_pre = type(conn.pre_obj).__name__
+
+        if function is not None:
+            if not isinstance(conn.pre_obj, (Node, Ensemble)):
+                raise ValidationError(
+                    "function can only be set for connections from an Ensemble"
+                    " or Node (got type %r)" % type_pre,
+                    attr=self.name, obj=conn)
```
jgosmann: From what object other then Node or Ensemble could a connection come?

nengo/tests/test_node.py
jgosmann: Maybe move the comment down to where actually a type error occurs and change this line to something like `# valid values`?

jgosmann:
Comments aside LGTM 🍰

nengo/connection.py
hunse: An `Ensemble.Neurons` object is one. Not sure if there are many others though.

nengo/tests/test_node.py
hunse: Yeah, okay.

nengo/connection.py
tbekolay: `Connection`, `Connection.learning_rule` and `Probe` can act as `post`, but I think we disallow them as `pre`.

tbekolay:
This looks good to me too! The speed concern I think outweighs the concern of making the `function` kwarg too complex... and we have validation now so that `eval_points` have to be set first, so surprise should be minimized.

I made two small fixup commits to reword some strings and add a deprecation warning to `nengo.utils.connection.target_function` (which we implemented as a helper in lieu of doing this previously). I think in 2.2 we should switch `target_function` to raise a `DeprecationError`, then someday remove it entirely; I noted this in #822.

This also needs a changelog entry. If my changes look good to @hunse I'll merge when a changelog entry is pushed!

tbekolay:
Oh also, this is a relatively big change and we've already done several minor bugfixes; is this a good time to cut a 2.1.1 release? My thinking is yes to start getting in the habit of more frequent, smaller releases.

jgosmann:
+1 for more releases
Though, it is not clear to me what decides whether we do a patch level or minor release. According to semantic versioning: "MINOR version when you add functionality in a backwards-compatible manner". This clearly adds functionality, so shouldn't this be 2.2?

hunse:
> "MINOR version when you add functionality in a backwards-compatible manner". This clearly adds functionality, so shouldn't this be 2.2?

I wouldn't be too strict about that. Especially if we want to do more releases (which I agree with), if we do minor increments for stuff like this we'll be at 2.100.0 before we know it, and that just looks ridiculous.

Do we want to do that mini-sprint we were talking about before the release, @tbekolay? I'd be fine releasing after this, but I think there are also a number of things waiting that we might as well get merged and released.

hunse:
If we're going to go to 2.2, I'd try to get a few more of the waiting PRs in.

jgosmann:
> I wouldn't be too strict about that.

Yes, but at the moment it looks pretty random to me which makes the versioning scheme somewhat useless.

We could also do a 2.1.1 bugfix release with what's been merged (and maybe a few other small bugfix PRs) and then do this and more in a 2.2 release.

tbekolay:
> if we do minor increments for stuff like this we'll be at 2.100.0

Indeed, Linux switched arbitrarily from 2.6.39 to 3.0 because the .39 was just too much. I think we will eventually want to do API-breaking changes, so we shouldn't be that opposed to bumping up to 3 sometime.



> could also do a 2.1.1 bugfix release with what's been merged (and maybe a few other small bugfix PRs)

This crossed my mind also. My thoughts for other small bugfix PRs to merge for 2.1.1 (in no particular order): #1074, #1027, #1005 (if that's still valid)

jgosmann:
Replied in #1005 regarding validity.

tbekolay:
I rebased this and fixed up the fixups. It's been reviewed already, but I'm going to give people the weekend to give this a look-over as it's a relatively big change; I'll merge Monday morning if no issues come up.

tbekolay:
Pushed a fixup to change a `nengo.Simulator` in a test to `Simulator`, and one to add a changelog entry. I'll merge once the tests finish.

Issue #1007, opened 2016-04-05
Getting rid of magic transforms in SPA


jgosmann:
I was trying to find an existing issue for this, but apparently there is none? :cry: That's sad because we had already a lot of discussion on this. Anyways, I'm creating this issue now to get the ball rolling.
### Possible ways to match up vocabularies

I think, first of all we need a list of potential ways one might want to match up two different vocabularies.
- "reinterpret": No transformation at all is performed, vectors will just be compared to a different set semantic pointers.
- "transform": Create a projection matrix projection semantic pointers from one vocab into another vocab. There are different variants possible as keys might only exist in one of the vocabularies:
  - Base transform only on the intersection of the sets of keys.
  - Transform semantic pointers missing in the target set to some special vector (might be the null vector).
  - Add new semantic pointers to the target vocabulary for missing keys.
  - Keys only in the target vocabulary, should not pose a problem?
### Syntax

I would propose something similar to function calls (e.g. "reinterpret(output)"). I have to leave right now so I'll flesh that out later.

tcstewar:
Thanks for starting this issue.  :)  The picture in my head has been something like:

```module1 = translate(module2, module1)```

with arguments allowed in there, giving things like:

```module1 = translate(module2, module1, keys=['DOG', 'CAT', 'RAT'])```

I could also see doing

```module1 = module2.translate_to(module1)```

tcstewar:
(note: I used ```translate``` instead of ```transform``` just because we use the word ```transform``` in ```Connection```s, and I wanted to avoid two uses of the term)

jgosmann:
With #1137 merged (with the branch incidentally called `no-magic-transform`), this has been resolved.

PR #1014, opened 2016-04-18
SPA compiler

jgosmann:
Implementation of a SPA compiler to address #1007.

For now I use this PR just to track the remaining to dos. I will provide more detailed information what's done here once it's finished. ~~Discussion points in bold.~~

* [x] Naming `actions.py` and `actions2.py`. Move `Parser` to same file as `Actions`.
* [x] Refactor and document `Parser` and `Actions`.
* [x] Refactor convolution and product construction functions.
* [x] Implement dot product construction.
* [x] Documentation of grammar and how AST + type inference works.
* [x] Make sure `infer_types` does all required type checks.
* [x] Have all exceptions provide useful information.
* [x] Implement explicit transforms
* [x] TODOs and FIXMEs
* [x] test spaces around `.` module separator
* [x] test hierarchical modules in rules
* [x] fix pep8
* [x] fix Python 3 compatibility
* [x] clean up history
* [x] Documenation and examples for `reinterpret` and `translate`?
* [x] Allow instantiation of `Cortical`, `BasalGanglia`, and `Thalamus` without `Actions` instance and provide these classes to the `Actions` instance for compilation and make this the preferred instantiation.
* [x] Placement and naming of gates and channels.
* [x] check unit test coverage


jgosmann:
This PR is now feature complete in the sense that everything that was working before should still be working with this PR. There are a few more things to do before a proper review of the code base can be done, but these should 100% backwards compatible to the current state of the PR.

Because this PR is rather huge and I'm not sure that the unit tests cover everything, **it would be awesome if some of you could test this PR with your existing SPA models** (if those include `Cortical`, `BasalGanglia`, and/or `Thalamus`).

# What does this PR do?

This PR is an almost complete rewrite of the SPA action system, implementing a compiler for it. To find out more about how the compiler works check the docstring at top of `spa_ast.py`. Almost on the side, but in fact the reason for this rewrite, is that no implicit casting¹ (i.e. transform) from one vocabulary to another are done anymore. Each cast has to be explicit and there are currently to possible cast:

* `reinterpret`: Reinterprets a vector as if it is a semantic pointer in another vocabulary without changing it.
* `translate`: Translates vectors with a linear transform matrix to another vocabulary using [this function](https://github.com/nengo/nengo/blob/master/nengo/spa/vocab.py#L336)

Each cast can be invoked in three different ways. I will give examples for `translate`, but you can use `reinterpret` in the same way:

1. `spa.Actions('state = translate(source)')`: This infers the target vocabulary. In this case it is inferred from the target `state`, but it could also be inferred from another operand as in `dot(translate(source), other_source)` or `translate(source) * other_source. The inference might not work in all cases (not because my code doesn't cover those cases, but because there can be cases where it is just impossible). In that case a `SpaTypeError` will be raised. 
2. `spa.Actions('state = translate(source, state)')` or `spa.Actions('state = translate(source, vocab=state)')`: Obtains the target vocabulary from the `state` module.
3. `spa.Actions('state = translate(source, v)', vocabs={'v': v})` or `spa.Actions('state = translate(source, vocab=v)', vocabs={'v': v})`: Uses the vocabulary `v`. Because the SPA parser needs to now the available vocabularies and their names they have to be passed into the `Actions` object with the `vocabs` argument.

¹: I'm using the term 'casting' here, because that's what is actually done here. Each vocabulary represents a type and the compiler will only allow matching types. Besides vocabulary types there is also a type for scalars and some more which are not as important as they only represents things above the expression level.

# Stuff for future PRs

There are a few things I will do as separate PRs because this one is already pretty big. One thing is changing default syntax in examples and test to new preferred versions (this includes the new syntax for inputs from PR #1006 and instantiation of actions and required modules).

# Backwards compatibility

All existing examples are still passing. All SPA models doing not too crazy stuff (like messing with synapses in thalamus or BG) should still work. One example used the `Action.effect` attribute; this is still supported with a `DeprecationWarning` (because of that one example), but the preferred attribute is not `Action.effects` (plural) and I changed the example accordingly.

If someone has a model with an action named `vocabs`, it won't work anymore because this became it's own keyword argument to the `Actions` object. There might be a few other of such obscure things that break backwards compatibility.

# Some thoughts

In general I like how the new structure makes it easy to implement additions because it is possible to focus on one node in the syntax tree at a time. This can still give four different things to implement for one binary node (each of left-hand and right-hand side can be a semantic pointer or scalar, that is 2x2 combinations), but that is manageable.

However, when I started this I did not realize that we had some 'optimizations' like not feeding a constant effect through a channel, but just doing it within a transform from the thalamus output. The current implementation does the same, but leads to slightly more complicated code and some interdependency between the `Effect` and `Action` node. That is not as pristine as I hoped, but probably not too bad. Also, there is some sort of interdependency between the compiler and the `Cortical`/`BasalGanglia`/`Thalamus` modules as these modules trust the compiler to build what they need and the compiler will call out to them to construct certain things. But to me this seems to be the correct thing to do.

# Not all magic is gone ... :sparkles: :(

Semantic pointers still get added automatically to vocabularies. I noticed that this can have unexpected effects on the translation matrices from one vocabulary to another. Those matrices depend on what elements are in the vocabulary and if more elements are added to a vocabulary later these won't be included. This could potentially be hard to track down. Also it is possible to add further actions after the instantiation of an `Actions` object (this was already possible before this PR, though it would give a warning related to another problem not existent anymore) and if at that second time pointers with new names will be used in the `Actions` the transform matrices generated at both points in time will differ. Same for independent `Actions` objects which is a more common use case as you might have `Cortical` and `BasalGanglia` actions.

I really would like to get rid of this automagic adding of pointers to vocabularies, but at the same time I would like to keep the convenience of not having to specify all pointers in advance.

arvoelke:
So cool. Lots of very nice ideas going on here that I've been wanting to see, and more. It's too bad I don't have any way to test this PR. But I do have a couple quick questions:
 - Are there clear ways for the user to probe the intermediate ensembles built by the compiler? Thinking in terms of debugging and finding bottlenecks in the approximation error.
 - If someone wanted to extend the language without changing the Nengo codebase how hard would that be? For example if someone from the summer school wanted to add an interesting function or support a new operation in the language.

jgosmann:
> Are there clear ways for the user to probe the intermediate ensembles built by the compiler (while being able to understand their semantics)? Thinking in terms of debugging and finding bottlenecks in the approximation error.

Nope, unfortunately not yet (except for the GUI). Would be certainly be useful to have, but requires some thought how to address those intermediate ensembles. The easiest thing might be to store the generated ensembles on the AST and make the AST accessible?

> If someone wanted to extend the language without changing the Nengo codebase how hard would that be?

Should be fairly easy. Just implement a `Node` class with the respective functions for type inference and network constructions and add it to the `builtins` dictionary of the `Actions` class.

arvoelke:
Great. As long as that first part is kept in mind and not ruled out by anything here in particular, I'm happy.

As for unit testing, have you looked at our [coveralls](https://coveralls.io/github/nengo/nengo?branch=no-magic-transform)? It will show you any _dead code_, which would give you an idea of what extra unit tests are needed (noting that obviously the converse does not guarantee correctness). But navigating this seems like a big pain. I've found codecov to be way-way better (and it even integrates with GitHub's [source](https://www.youtube.com/watch?v=d6wJKODB8_g) and [pull requests](https://github.com/codecov/example-python/pull/14)).

jgosmann:
> As for unit testing, have you looked at our coveralls? It will show you any dead code, which would give you an idea of what extra unit tests are needed (noting that obviously the converse does not guarantee correctness).

That's what I had in mind. I might look at codecov, but I only have to check the coverage of one file.

tbekolay:
codecov does look way better than coveralls, thanks for the pointer @arvoelke! I'll switch Nengo over soon.

nengo/spa/actions.py
```diff
@@ -1,151 +1,162 @@
...
+            if strict:
+                raise SpaParseError("Not an action, '-->' missing.")
+            else:
+                return self.parse_effects(action, channeled=False)
+        else:
+            return Action(
+                self.parse_expr(condition),
+                self.parse_effects(effects, channeled=True), index)
+
+    def parse_effects(self, effects, channeled=False):
```
jgosmann: flake8 is comlaining about this function:
```
nengo/spa/actions.py:60:1: C901 'Parser.parse_effects' is too complex (14)
```

But it is not obvious how to reduce the complexity. I'm afraid that splitting it up into smaller function won't make it any easier to understand. I suppose it could be helpful to add another comment on how this functions parsing stuff using a push down automaton, but such a comment would not silence flake8.

jgosmann:
I looked at the coverage report and most stuff is covered. What's mostly missing are test for the exceptions raised and a few minor things where it seems to be too much overhead to test every possibility. The only major thing:
* [ ] add a test for dot product with two dynamic sources

However, the coverage basically only tells me that all the code doesn't horribly break with exceptions. It cannot tell me that all generated networks work exactly as they should. There are some tests testing actual networks, but the more people could test it with their own models the better.

nengo/spa/actions.py
arvoelke: Using a more generic parser/lexer that uses a lookup table in a PDA/DFA respectively would prevent all this branching. But I'm not sure if this would be useful for the main repo, and we seem morally opposed to importing additional dependencies. For now you can just add `noqa: C901`?

jgosmann: Generic parser/lexer would be nice, but there seems to be nothing in the Python standard lib and this function doesn't seem that complex that we should add a new dependency.

> For now you can just add noqa: C901?

If people are fine with that and think that the code has a reasonable level of complexity/readability.

jgosmann:
New syntax for the Action construction:
```python
actions = spa.Actions(
    # all your rules/actions/effects here
)
actions.build(spa)
```
This will automatically add `Cortical`, `BasalGanglia`, and `Thalamus` to `spa` (named `spa.cortical`, `spa.bg`, and `spa.thalamus`) as needed. If you want to construct those yourself you can pass them in to the `build` function.

Because I'm keeping backwards compatibility the code of `Actions`, `BasalGanglia`, and `Thalamus` is slightly convoluted. My proposal would be to keep the backwards compatibility for one minor version and then get rid of it.

jgosmann:
Regarding the new syntax for the Action construction:
```python
with spa.Module() as spa:
    # your model
    actions = spa.Actions(
        # all your rules/actions/effects here
    )
    actions.build(spa)
```
At the moment you have to pass in the root module to the `build` function, but I suppose it could be inferred from the network context provided by the `with` statement. Any preferences? I'm not sure if one is better than the other.

jgosmann:
I decided to put all the channels and gate related to one action into one subnetwork. This allows to give the network a useful name (the expression of that particular action) and gives a somewhat logical grouping. Note that the names might get rather long, but there could be an argument made that this should be handled on the GUI side (cutting long names of and show the full name on hover?). A slight issue on the Nengo side is that the labels might contain more parenthesis as necessary, but figuring out when parenthesis are not required is somewhat complicated.

tcstewar:
> I decided to put all the channels and gate related to one action into one subnetwork. This allows to give the network a useful name (the expression of that particular action) and gives a somewhat logical grouping. Note that the names might get rather long, but there could be an argument made that this should be handled on the GUI side (cutting long names of and show the full name on hover?). A slight issue on the Nengo side is that the labels might contain more parenthesis as necessary, but figuring out when parenthesis are not required is somewhat complicated.

Nice.  Good idea grouping based on functionality like that.

As for the names, in the current system it's also possible to specify Actions by ```**kwargs```, so you can do:

```spa.Actions(
        greet='dot(vision,PERSON) --> speech=HELLO',
        run_away='dot(vision,THREAT) --> motor=RUN, heart=FASTER',
)```

The intent behind that was to allow the kwargs to be used in graphs and displays, but it'd also make sense for a label on your routing Network.

**Comment by jgosmann** [comment]
I'm using that kwarg name right now, but only in addition, not as replacement. But this is definitely open to discussion.

**Comment by tcstewar** [comment]
> I'm using that kwarg name right now, but only in addition, not as replacement. But this is definitely open to discussion.

Interesting...  I think it makes sense as a replacement -- one of those things people will only do when they care about what the label is (or when the text version gets really long....)

**Comment by jgosmann** [comment]
> Interesting... I think it makes sense as a replacement -- one of those things people will only do when they care about what the label is (or when the text version gets really long....)

ok, I changed it

**Comment by jgosmann** [comment]
I realized that the changes in this PR allow us to get rid of the `on_add` stuff (see last commit). I think that is great and makes `Module`s behave more like normal networks. The benefit of this will be that I can actually create `Bind` modules etc. (instead of `CircularConvolution` networks) in the compiler without the need to assign them to another module (which might always conflict with user names) and then these generated modules support the SPA plots in the GUI. :D

**Comment by jgosmann** [comment]
Also changed the syntax for inputs. Instead of
```python
with spa.Module() as model:
    model.state = spa.State(16)
    model.stimuli = spa.Input()
    model.stimuli.state = 'A'
```
You can just do:
```python
with spa.Module() as model:
    model.state = spa.State(16)
    model.stimuli.state = 'A'
```
which will automatically create the input module (it's tied to the `stimuli` attribute, other names are suggestions are welcome).

tcstewar:
> I realized that the changes in this PR allow us to get rid of the on_add stuff (see last commit).

Awesome!!!!!  That makes me very very happy!  :)

> Also changed the syntax for inputs.

Hmm...  I think I'd prefer explicitly having to do the ```model.stimuli = spa.Input()``` line, just to reduce magic.

jgosmann:
> Hmm... I think I'd prefer explicitly having to do the model.stimuli = spa.Input() line, just to reduce magic.

Though in that case it should be `model.stimuli = spa.Input(model)`. (`spa.Input()` still works, but I would consider it deprecated after removing `on_add` because the way it looks for the `model` if it is not given explicitly)

tcstewar:
> Though in that case it should be model.stimuli = spa.Input(model). (spa.Input() still works, but I would consider it deprecated after removing on_add because the way it looks for the model if it is not given explicitly)

Ah, I see...  interesting...  hmm...  now I'm wavering back to your original approach...  Probably worth just leaving it how you have it and seeing what people think.  :)

nengo/spa/vocab.py
```diff
@@ -480,15 +485,63 @@ def create_subset(self, keys):
         return subset
 
 
-class VocabularyParam(nengo.params.Parameter):
-    """Can be a Vocabulary."""
+class VocabularyMap(Mapping):
```
Seanny123: It would be nice to have a quick one-line docstring about the general intended purpose of this object.

nengo/spa/tests/test_module.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+
+import nengo
+from nengo import spa
+from nengo.spa.utils import similarity
+
+
+class SpaCommunicationChannel(spa.Module):
```
Seanny123: I'm tempted to ask for documentation for this class, so that a person could understand it's typical use without looking through all the tests it's used in, but I can't decide if that's excessive or not. Thoughts?

xchoo: It's just a class / module used in the test functions... I don't think documentation is necessary.

nengo/spa/tests/test_spa_ast.py
```diff
@@ -0,0 +1,271 @@
...
+    d = 16
+    with spa.Module() as model:
+        model.state = spa.State(d)
+
+    ast = Parser().parse_effect('state = 0')
+    ast.infer_types(model, None)
+    assert ast.source.type.vocab == model.state.vocabs[d]
+
+
+# def test_actions():
```
Seanny123: Why is all of this commented-out code left in?

nengo/spa/spa_ast.py
```diff
@@ -0,0 +1,994 @@
...
+1. If something with unknown vocabulary is assigned to a module, that module's
+   vocabulary provides the type.
+2. If a binary operation has an operand with unknown vocabulary it is
+   determined from the other operand.
+
+Once all the types have been determined, the AST can be used to construct
+Nengo objects to perform the operations represented with the AST. In this
+process each node in the syntax tree can create :class:`Artifact`s. These
+give generated Nengo objects to be connected to the appropriate places
+including the transform that should be used for that. This is necessary
```
Seanny123: "including the transform that should be used for that." It's not clear what "that" is referring to in this sentence.

nengo/spa/spa_ast.py
```diff
@@ -0,0 +1,994 @@
...
+
+    def __str__(self):
+        return '{}<{}>'.format(self.name, self.vocab)
+
+    def __eq__(self, other):
+        return (super(TVocabulary, self).__eq__(other) and
+                self.vocab is other.vocab)
+
+
+class Artifact(object):
```
Seanny123: Why choose to name this "Artifact"?

nengo/spa/vocab.py
jgosmann: Yes, but that change is part of PR #1006 (on which this PR is based).

nengo/spa/spa_ast.py
arvoelke: I remember being asked the same question when I did the same thing. :)
https://en.wikipedia.org/wiki/Artifact_(software_development)

arvoelke: The wiki article isn't very clear on this, but the term has connections to compilers and in particular the byproducts of a build process.

jgosmann: Because it is something produced by the build process (somewhat analogous to "Build tools often refer to source code compiled for testing as an artifact" and or a byproduct).

It basically is the compilation result of part of the syntax tree, but at that stage we cannot translate it to actual Nengo object because we only have the source object and associated transform and we need to keep that information around until we know the destination object and create the connection. At the same time the source object and associated transform are closely tied to each other and I think that warrants to put them into one object (which also simplifies passing them around). Artifact was the best name I found for this kind of thing.

nengo/spa/tests/test_module.py
jgosmann: I think it's also pretty self explanatory: It is a SPA communication channel module. *duh*

nengo/spa/vocab.py
jgosmann: Added a line in #1006.

jgosmann:
Will need rebase after #1006 has been merged (which should happen first).

jgosmann:
Will reopen a PR on the actual branch this gets merged into to make the review process easier.

PR #1016, opened 2016-04-19
Add `expand_dims` with multiple axis support

hunse:
This allows `norm` and `rms` to take multiple axes.

### DIFF
--- a/nengo/utils/numpy.py
+++ b/nengo/utils/numpy.py
@@ -137,6 +137,29 @@ def expm(A, n_factors=None, normalize=False):
     return np.linalg.matrix_power(Y, a) if normalize else Y
 
 
+def _expand_dims(a, axis):
+    """Like numpy._expand_dims, but with support for multiple axes.
+
+    Parameters
+    ----------
+    a : array_like
+        Array to expand dimensions of.
+    axis : int or sequence, optional
+        Indices where to insert axes. Each index is in relation to all an array
+        where all prior axes have been inserted. For example given an array
+        with shape `(n,)` ``axis=(0, 1)`` would expand the array to shape
+        `(1, 1, n)`.
+    """
+    a = np.asarray(a)
+    shape = list(a.shape)
+    axes = axis if is_iterable(axis) else (axis,)
+    n = len(shape) + len(axes)
+    axes = sorted([i + n if i < 0 else i for i in axes])
+    for i in axes:
+        shape.insert(i, 1)
+    return a.reshape(shape)
+
+
 def norm(x, axis=None, keepdims=False):
     """Euclidean norm
 
@@ -150,8 +173,10 @@ def norm(x, axis=None, keepdims=False):
         If True, the reduced axes are left in the result. See `np.sum` in
         newer versions of Numpy (>= 1.7).
     """
+    if axis is None:
+        axis = tuple(range(x.ndim))
     y = np.sqrt(np.sum(x**2, axis=axis))
-    return np.expand_dims(y, axis=axis) if keepdims else y
+    return _expand_dims(y, axis=axis) if keepdims else y
 
 
 def meshgrid_nd(*args):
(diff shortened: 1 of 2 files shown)


drasmuss:
Looks good

jgosmann:
Is there a specific reason we need this feature?

jgosmann:
I added a test for the `expand_dims` function and it is failing because it is not clearly specified what the desired semantics are when giving multiple axes to expand.

I would assume that all axes indices for newly inserted axes are in relation to the original array. For example I would assume the following to pass:
```python
a = np.array([0, 1])
assert expand_dims(a, (0, 0, 1, -1)).shape == (1, 1, 2, 1, 1)
```
as it (in my opinion) should insert two dimensions in the front and two at the end. The actual implementation however inserts one dimension after another and thus the index 1 in the third place is not in relation to the original array, but an itermediary array. Those semantics seem more complicated to me because one has to think about how indices change as dimensions are inserted, but if there is a good reason to prefer these semantics I'm fine with that too as long as it is adequately documented. Note that with having all indices in relation to the original array it is not possible to specify what the order of the newly inserted dimensions is (e.g. I cannot say whether to insert the dimension for the second 0 index as first or second dimension), but this is irrelevant as all new dimensions have a size of 1 and are equivalent.

I can probably do the required changes myself (different semantics or better documentation), but we should agree on the semantics we want here.

hunse:
So the motivation behind this PR is to allow `norm` and `rms` to take multiple axes, like `np.sum` and `np.mean`, and maintain the `keepdims` functionality when multiple axes are passed.

I wrote `expand_dims` to help with this. So at a minimum, the semantics need to support this functionality. But it could be that there are multiple sets of semantics that allow for what I want. @jgosmann, the semantics you're suggesting do make sense for a stand-alone `expand_dims` function, I'd just want to make sure that they'll still work properly where they're being used in `rms` and `norm`. This PR should definitely add tests that check that `rms` and `norm` are doing what we want when multiple axes are passed and `keepdims == True`. A test for `expand_dims` would also be good, but less critical IMO since I mostly see it as a helper for `rms` and `norm` (we could even make it a hidden function if we want).

@jgosmann, if you have time to look at this that would be great. I'm going to be busy here at Telluride for the next month, so it might be a while before I could get back to this.

jgosmann:
I've added tests for the `norm` and `rms` function (62c2a4d0c838e6e58e83eb2e7b7690fd48970868) and changed the semantics of `expand_dims` as I proposed (4b47c911c915a9ad8253a633b8a9d30a594dd65d). Even though it makes the interface of the function more logical, it adds complexity to `norm` and `rms`. So I am not sure if we want to keep this.

Passing this on to a second reviewer now ...

hunse:
Can we keep the old `expand_dims` as a dedicated helper for `rms` and `norm`? I think it gives us all the functionality we need there, so I'm very hesitant to add all this complexity. Currently we have no use-case for the new `expand_dims`.

jgosmann:
Ok, but in that case I would make it private and want to have it documented.

hunse:
Absolutely, I'm fine with both those things.

jgosmann:
I made the final changes, if those look good to @hunse this should be ready to merge?

tbekolay:
According to TravisCI this fails for NumPy 1.6. Should we bump up the requirement or fix that?

jgosmann:
I vote bumping the requirement. Otherwise the proper fix IMO would be either to disallow multiple axes which makes this PR moot or to implement all of that functionality by ourselves.

tbekolay:
> I vote bumping the requirement.

+1 here as well, 1.7 was released Feb 2013 and 1.8 is present in both Debian stable and Ubuntu 14.04, which is a pretty old LTS release (the only supported Ubuntu that we would be impacting is 12.04, which I'm not that worried about).

hunse:
So if we bump to 1.7, I'm not sure if we need this at all. Then we can just pass the `keepdims` argument straight to `sum`.

tbekolay:
> if we bump to 1.7, I'm not sure if we need this at all

Adding a note to next week's dev meeting agenda to discuss dropping NumPy 1.6 support.

tbekolay:
With 1.6 support dropped in #1147, this PR is no longer needed. Appreciated nonetheless!

PR #1021, opened 2016-04-21
Check that parameter names are strings

### LINKED ISSUE #1020: Nengo parameters let you set silly names
As @drasmuss and I found out, the following code using the old parameter syntax throws no error:
```
net.config[nengo.Connection].set_param(
    "insert_weights", nengo.params.BoolParam(False))
```

Under the new parameter syntax, the correct code would be:
```
net.config[nengo.Connection].set_param(
    "insert_weights", nengo.params.BoolParam("test", default=False))
```

Can we check the `type` of `name` [on initialisation](https://github.com/nengo/nengo/blob/master/nengo/params.py#L48) so that it throws an error for anything that isn't a string?

hunse:
Fixes #1020 

Sorry @Seanny123, just saw that you had self-assigned the issue after I finished this and was about to submit.

### DIFF
--- a/nengo/params.py
+++ b/nengo/params.py
@@ -45,13 +45,19 @@ class Parameter(object):
 
     def __init__(self, name,
                  default=Unconfigurable, optional=False, readonly=None):
+        # freeze Unconfigurables by default
+        readonly = default is Unconfigurable if readonly is None else readonly
+
+        if not is_string(name):
+            raise ValueError("'name' must be a string (got %r)" % name)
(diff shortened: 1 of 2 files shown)


Seanny123:
No worries, that was just a reminder. I'm glad you fixed it!

tbekolay:
This is tiny and kinda sorta a bug, so I'm inclined to include in 2.1.0. @drasmuss can you take a quick look? I'll merge on your +1.

drasmuss:
Looks good

PR #1025, opened 2016-04-21
Processes use same time as Simulator

### LINKED ISSUE #1022: Discrepancy between Process.run_steps and sim.run
When you run a process via `Process.run_steps`, it gets called starting at `t=0`:

https://github.com/nengo/nengo/blob/master/nengo/base.py#L242
```
for i in range(n_steps):
    output[i] = step(i * dt)
```

but when you run a process as part of a Node during a simulation (e.g., via `sim.run_steps`), that starts at `t=dt`.

They should probably both work the same way.  I would suggest modifying `Process.run_steps`, just because that would have less overhead.

hunse:
Previously, process time was one time-step behind simulator time.
This changes them to be the same. Fixes #1022.

### DIFF
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -240,7 +240,7 @@ class Process(FrozenObject):
         step = self.make_step(shape_in, shape_out, dt, rng, **kwargs)
         output = np.zeros((n_steps,) + shape_out)
         for i in range(n_steps):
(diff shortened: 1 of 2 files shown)


drasmuss:
Sweet, I was going to do this tomorrow but you're on the ball.  Looks good.

PR #1027, opened 2016-04-22
Fix NNLS weight solvers

### LINKED ISSUE #1019: Nnls solvers fail when weights=True
The non-negative solvers don't seem to work on full weight matrices.  Here's a quick example:

```python
model = nengo.Network()
with model:
    a = nengo.Ensemble(100, 1)
    b = nengo.Ensemble(100, 1)
    nengo.Connection(a, b, solver=nengo.solvers.Nnls(weights=True))
sim = nengo.Simulator(model)
```

The error is:
```
BuildError                                Traceback (most recent call last)
<ipython-input-22-b1fdc30f942d> in <module>()
----> 1 sim = nengo.Simulator(model)
      2 sim.run(2)

/home/tcstewar/github/nengo/nengo/simulator.pyc in __init__(self, network, dt, seed, model)
    115         if network is not None:
    116             # Build the network into the model
--> 117             self.model.build(network)
    118 
    119         self.model.decoder_cache.shrink()

/home/tcstewar/github/nengo/nengo/builder/builder.pyc in build(self, obj, *args, **kwargs)
     83 
     84     def build(self, obj, *args, **kwargs):
---> 85         return Builder.build(self, obj, *args, **kwargs)
     86 
     87     def add_op(self, op):

/home/tcstewar/github/nengo/nengo/builder/builder.pyc in build(cls, model, obj, *args, **kwargs)
    128                 "Cannot build object of type %r" % obj.__class__.__name__)
    129 
--> 130         return cls.builders[obj_cls](model, obj, *args, **kwargs)

/home/tcstewar/github/nengo/nengo/builder/network.pyc in build_network(model, network)
     68         # TODO: Except perhaps if the connection being learned
     69         # is in a subnetwork?
---> 70         model.build(conn)
     71 
     72     logger.debug("Network step 4: Building probes")

/home/tcstewar/github/nengo/nengo/builder/builder.pyc in build(self, obj, *args, **kwargs)
     83 
     84     def build(self, obj, *args, **kwargs):
---> 85         return Builder.build(self, obj, *args, **kwargs)
     86 
     87     def add_op(self, op):

/home/tcstewar/github/nengo/nengo/builder/builder.pyc in build(cls, model, obj, *args, **kwargs)
    128                 "Cannot build object of type %r" % obj.__class__.__name__)
    129 
--> 130         return cls.builders[obj_cls](model, obj, *args, **kwargs)

/home/tcstewar/github/nengo/nengo/builder/connection.pyc in build_connection(model, conn)
    206                     in_signal,
    207                     signal,
--> 208                     tag="%s.weights_elementwiseinc" % conn))
    209 
    210     # Add operator for filtering

/home/tcstewar/github/nengo/nengo/builder/builder.pyc in add_op(self, op)
     90         signals = SignalDict()
     91         op.init_signals(signals)
---> 92         op.make_step(signals, self.dt, np.random)
     93 
     94     def has_built(self, obj):

/home/tcstewar/github/nengo/nengo/builder/operator.pyc in make_step(self, signals, dt, rng)
    386         A = signals[self.A]
    387         Y = signals[self.Y]
--> 388         reshape = reshape_dot(A, X, Y, self.tag)
    389 
    390         def step_dotinc():

/home/tcstewar/github/nengo/nengo/builder/operator.pyc in reshape_dot(A, X, Y, tag)
    350     if (badshape or incshape != Y.shape) and incshape != ():
    351         raise BuildError("shape mismatch in %s: %s x %s -> %s"
--> 352                          % (tag, A.shape, X.shape, Y.shape))
    353 
    354     # Reshape to handle case when np.dot(A, X) and Y are both scalars

BuildError: shape mismatch in <Connection from <Ensemble (unlabeled) at 0x7fda44dc74d0> to <Ensemble (unlabeled) at 0x7fda44edd610>>.weights_elementwiseinc: (1, 100) x (100,) -> (100,)
```

If I try it with the standard solvers, it runs fine.

hunse:
Fixes #1019, as well as some other small issues.

### DIFF
--- a/nengo/solvers.py
+++ b/nengo/solvers.py
@@ -456,17 +456,28 @@ class NnlsL2(Nnls):
         super(NnlsL2, self).__init__(weights=weights)
         self.reg = reg
 
-    def _solve(self, A, Y, rng, E, sigma):
+    def _solve(self, A, Y, rng, E, sigma=0.):
+        import scipy.optimize
+
         tstart = time.time()
+        Y, m, n, _, matrix_in = format_system(A, Y)
+        Y = self.mul_encoders(Y, E, copy=True)
+        d = Y.shape[1]
+
         # form Gram matrix so we can add regularization
         GA = np.dot(A.T, A)
-        GY = np.dot(A.T, Y)
         np.fill_diagonal(GA, GA.diagonal() + A.shape[0] * sigma**2)
-        X, info = super(NnlsL2, self).__call__(GA, GY, rng=rng, E=E)
+        GY = np.dot(A.T, Y.clip(0, None))
+        # ^ TODO: why is it better if we clip Y to be positive here?
+
+        X = np.zeros((n, d))
+        residuals = np.zeros(d)
+        for i in range(d):
+            X[:, i], residuals[i] = scipy.optimize.nnls(GA, GY[:, i])
+
         t = time.time() - tstart
-        # recompute the RMSE in terms of the original matrices
-        info = {'rmses': rmses(A, X, Y), 'gram_info': info, 'time': t}
-        return X, info
+        info = {'rmses': rmses(A, X, Y), 'residuals': residuals, 'time': t}
+        return X if matrix_in or X.shape[1] > 1 else X.ravel(), info
 
     def __call__(self, A, Y, rng=None, E=None):
         return self._solve(A, Y, rng, E, sigma=self.reg * A.max())
--- a/nengo/solvers.py
+++ b/nengo/solvers.py
@@ -379,6 +379,10 @@ class Nnls(Solver):
     """Non-negative least-squares solver without regularization.
 
     Similar to `.Lstsq`, except the output values are non-negative.
+
+    If solving for non-negative **weights**, it is important that the
+    intercepts of the post-population are also non-negative, since neurons with
+    negative intercepts will never be silent, affecting output accuracy.
     """
 
     def __init__(self, weights=False):
@@ -405,7 +409,7 @@ class Nnls(Solver):
 
         tstart = time.time()
         Y, m, n, _, matrix_in = format_system(A, Y)
-        Y = self.mul_encoders(Y, E)
+        Y = self.mul_encoders(Y, E, copy=True)
         d = Y.shape[1]
 
         X = np.zeros((n, d))
@@ -422,6 +426,10 @@ class NnlsL2(Nnls):
     """Non-negative least-squares solver with L2 regularization.
 
     Similar to `.LstsqL2`, except the output values are non-negative.
+
+    If solving for non-negative **weights**, it is important that the
+    intercepts of the post-population are also non-negative, since neurons with
+    negative intercepts will never be silent, affecting output accuracy.
     """
 
     reg = NumberParam('reg', low=0)
@@ -468,6 +476,10 @@ class NnlsL2nz(NnlsL2):
     """Non-negative least-squares with L2 regularization on nonzero components.
 
     Similar to `.LstsqL2nz`, except the output values are non-negative.
+
+    If solving for non-negative **weights**, it is important that the
+    intercepts of the post-population are also non-negative, since neurons with
+    negative intercepts will never be silent, affecting output accuracy.
     """
 
     def __call__(self, A, Y, rng=None, E=None):
--- a/nengo/solvers.py
+++ b/nengo/solvers.py
@@ -350,7 +350,7 @@ class LstsqDrop(Solver):
 
     def __call__(self, A, Y, rng=None, E=None):
         tstart = time.time()
-        Y, m, n, d, matrix_in = format_system(A, Y)
+        Y, m, n, _, matrix_in = format_system(A, Y)
 
         # solve for coefficients using standard solver
         X, info0 = self.solver1(A, Y, rng=rng)
@@ -372,7 +372,7 @@ class LstsqDrop(Solver):
         t = time.time() - tstart
         info = {'rmses': rmses(A, X, Y), 'info0': info0, 'info1': info1,
                 'time': t}
-        return X if matrix_in else X.flatten(), info
+        return X if matrix_in or X.shape[1] > 1 else X.ravel(), info
 
 
 class Nnls(Solver):
@@ -404,8 +404,9 @@ class Nnls(Solver):
         import scipy.optimize
 
         tstart = time.time()
-        Y, m, n, d, matrix_in = format_system(A, Y)
+        Y, m, n, _, matrix_in = format_system(A, Y)
         Y = self.mul_encoders(Y, E)
+        d = Y.shape[1]
 
         X = np.zeros((n, d))
         residuals = np.zeros(d)
@@ -414,7 +415,7 @@ class Nnls(Solver):
 
         t = time.time() - tstart
         info = {'rmses': rmses(A, X, Y), 'residuals': residuals, 'time': t}
-        return X if matrix_in else X.flatten(), info
+        return X if matrix_in or X.shape[1] > 1 else X.ravel(), info
 
 
 class NnlsL2(Nnls):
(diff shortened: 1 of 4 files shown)


hunse:
@tcstewar, can you review, since you filed the original #1019 bug that this addresses?

tcstewar:
This looks good to me, and fixes the original problem.  I also confirmed that if you use the nnls with weights=True, you get the same result as if you used weights=False (if the encoders are all positive).

However, oddly, if we use normal encoders and weights=True, the results are worse than I'd expect.  For example:

```python
model = nengo.Network()
with model:
    stim = nengo.Node(lambda t: np.sin(t*np.pi*2))
    a = nengo.Ensemble(50, 1)
    b = nengo.Ensemble(50, 1)#, encoders=nengo.dists.Choice([[1]]))
    
    nengo.Connection(stim, a)
    nengo.Connection(a, b, solver=nengo.solvers.Nnls(weights=True))
    
    p = nengo.Probe(b, synapse=0.03)
    p_stim = nengo.Probe(stim, synapse=0.03)
    
sim = nengo.Simulator(model)
sim.run(2)    
    
pylab.plot(sim.trange(), sim.data[p_stim])
pylab.plot(sim.trange(), sim.data[p])
pylab.show()    
```

produces this:

image


It looks to me like it could have gotten a better fit by just multiplying all the weights by ~1.4.  Why didn't it find that solution?

hunse:
That is a good question. Here's my theory:
- We pose our normal decoder solving problem as `dot(A, X) = Y` where A is our activation, X is the decoders, and Y is the targets (in the case of representation, equal to our eval points).
- To solve for the weights, I take the targets and multiply them by the encoders E, so we get `dot(A, W) = dot(Y, E)`, where W is the weights we want to solve for.
- If we're doing NNLS, all the weights must be non-negative, and so are the activities, so their product is non-negative. `dot(Y, E)` can be quite negative, though.
- Solving the NNLS problem minimizes the RMS amplitude of `dot(A, W) - dot(Y, E)`. Since there are negative terms that we can never get close to, the best we can hope to do is zero for those terms, so those terms end up contributing a lot of the error.
- All this is being done in the *input* space of the post-synaptic neurons, and doesn't account for the bias and nonlinearity of the neurons. So in the example above, doubling the weights helps, but only because there are a lot of neurons that are already silent who are unaffected by the weight change. If all the output neurons were linear, doubling the weights doesn't help. So in the eyes of the solver, it's found the best solution to the linear problem it's given. (I don't fully understand this yet, though.)
- Choosing all non-zero intercepts allows the solver to perform almost perfectly. i.e. `intercepts=nengo.dists.Uniform(0, 1)` in the output population. I don't fully understand this yet either, but basically my thinking is that this makes the active (i.e. firing, non-silent) range of the neuron the same as the controllable range of the neuron. If you have a neuron with a negative intercept, this means that you need a negative current to make the neuron silent, and we can't get a negative current with non-negative weights. So essentially we always have some off and some on neurons fighting against each other, making the amplitude of the result too small. Here's the result with non-negative intercepts:
figure_1

tcstewar:
Ah, that makes total sense.  For about half the neurons, it's finding a pretty perfect set of weights.  For the other half, it's supposed to find weights that give a negative number for the input current, but it can't, so it just gets that as small as possible.  But that screws up the *decode*, giving a smaller value than it should (since those neurons that are getting too much input are screwing up the decode, and will always be pushing the decoded value towards zero).  When I did the "just multiply by 1.4" thing, I was actually making the input current wrong for all the neurons, but in a way that things kinda cancelled as far as the decode was concerned.  But in this scenario the weights that are best at getting the right input current to each neuron are not the weights that are best at getting the right decoded value out of the population.

It'd be nice to toss this observation into the NNLS documentation, along with your suggestion for the intercepts.  But that's a separate PR from this one.  Now that I (think I) understand what's happening here, this PR looks great to me.

hunse:
There's still a few things I'm working on adding, so I can throw that bit of documentation in as well.

hunse:
Ok @tcstewar, two more little commits you can look at. There was some strange bug in the `NnlsL2` solver and I couldn't figure out why it wasn't working when calling out to the `Nnls` solver for the core of the solving, so I just copied over the code and now it works.

Also, one thing worth mentioning is that `Nnls` solves the least squares problem directly, whereas `NnlsL2` solves the Gram system (i.e. the normal equations). These give slightly different answers, where `NnlsL2(weights=True, reg=0.0)` does slightly worse than `Nnls(weights=True)` (0.215 RMSE vs 0.186). However, `NnlsL2` is considerably faster, at least for the system I was trying it on, because the Gram system has much smaller matrices to deal with. We should probably document this somewhere, though I'm not sure if it's best to put it in the docstring or a notebook. I'm already not totally happy with the redundancy between the docstrings for the different types of Nnls solvers (i.e. I put that comment about non-negative intercepts in all of them).

nengo/solvers.py
```diff
@@ -404,8 +408,10 @@ def __call__(self, A, Y, rng=None, E=None):
         import scipy.optimize
 
         tstart = time.time()
-        Y, m, n, d, matrix_in = format_system(A, Y)
-        Y = self.mul_encoders(Y, E)
+        Y, m, n, _, matrix_in = format_system(A, Y)
+        Y = self.mul_encoders(Y, E, copy=True)
+        d = Y.shape[1]
+        Y[Y < 0] = 0  # makes the final RMS error more reasonable
```
tcstewar: I'm not sure we should be doing this.  The RMS error is what it is -- this is returning the RMS error where any negative target value is treated as if its target value was 0, so it's not really the RMS error any more.  Having this happen silently seems a bit odd to me...

tcstewar: Err, never mind -- I misinterpreted what was happening here (I thought it was just about giving a more reasonable RMS report out of the solver, rather than something that affects the actual decoders)

tcstewar:
Looks good to me.  I think doing the extra documentation in a notebook makes the most sense, as there are a lot of possibilities of different ways of making use of this solver, and it's not something that people have done much of yet.

nengo/solvers.py
hunse: Oh no, you're right, this is bad. I liked it because it helped to compare different solvers when I was just playing around, but you're right that it misrepresents things. It could also be problematic if we ever have neurons that can have negative activities.

hunse: Oh that's so strange. It does make a difference in the Gram system (if you do it before multiplying Y by the transposed activities), but not in the regular system. I have no idea why that is.

hunse:
Ok, I redid those last two commits so they don't affect the RMSE. We just clip Y to be non-negative in the Gram system, because this helps for some reason.

tcstewar:
Cool.  That looks great to me.  Thanks!  (and thanks for adding the TODO about figuring out what's happening there in the future... )   :)

hunse:
Yeah, I mean it kind of makes sense. Making Y non-negative doesn't change the original system, because the solution can never achieve those negative values, so it just does what it can to minimize the error on the positive ones and keeps the negatives zero. The value that minimizes `f(x) + large_number` is the same that minimizes `f(x)`. Though that said, clipping things to zero in the original system might change the balance of things, for example if the target value is -10, the squared error difference between returing 0 and 1 is larger than if the target value is 0.

In the Gram system, if we don't clip Y, then when we multiply by A transpose, some of the positives and negatives might cancel out, and we end up with a different system than if we clip Y first. So it makes sense that things change, and the whole Gram system/normal equation idea is based on things being linear (i.e. linear least squares), and when we have the non-negative constraint that's no longer the case.

hunse:
Added changelog entries. I think this is ready for merge.

tbekolay:
I'll look at this shortly as it's the last thing for 2.1.1!

tbekolay:
We found a minor issue which Eric fixed; history is clean now so I'll merge this tomorrow morning unless there are objections!

Issue #1026, opened 2016-04-22
Allow distributions for neuron parameters


Seanny123:
As discussed during Aaron's presentation, parameters for the neurons, such as `tau_rc`, and `tau_ref` in the LIF case, as well as `tau_n` and `inc_n` in the AdaptiveLIF case, should be able to be set as either numbers or distributions, but not arrays.

hunse:
The reason we can't do arrays is that we want the `NeuronType` to be able to work for different ensembles, which could have different numbers of neurons so we don't know the size the array should be. To get around this, a user can use a `Tile` distribution (currently housed in nengo-extras).

tbekolay:
I believe you can also use `Samples` (https://github.com/nengo/nengo/blob/master/nengo/dists.py#L305).

PR #1028, opened 2016-04-22
Distributions for neuron parameters

hunse:
Working on allowing distributions for neuron parameters like `AdaptiveLIF.tau_n`, as requested in #1026.

This will require a number of changes, including making these into distribution parameters that get sampled when the neuron is built, and get passed to `gain_bias` and `step_math` as needed. One option is to package these together with `gain` and `bias` and just have a `params` object that gets passed around together (or don't package in `gain` and `bias`, but still have such an object.

We currently do checking to make sure e.g. `tau_rc > 0`, and it would be nice to have this for distributions. I think this is difficult/impossible to do exhaustively for distributions, but we could have either validation for some distributions when they're passed (e.g. Uniform, Choice), or validation once the samples are generated in the builder, or both.

### DIFF
--- a/nengo/dists.py
+++ b/nengo/dists.py
@@ -528,6 +528,17 @@ class DistributionParam(Parameter):
 
     equatable = True
 
+    def __init__(self, name, default=Unconfigurable,
+                 optional=False, readonly=None, cast_number=False):
+        super().__init__(name, default=default, optional=optional,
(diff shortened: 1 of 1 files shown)


hunse:
This PR never got off the ground. Closing it for now (the issue still exists).

PR #1031, opened 2016-04-22
Implement SupportDefaultsMixin.

jgosmann:
This makes it easy for other classes than NengoObject to support the
`Default` parameter value.

I'm finding that I want to have this in SPA modules too and with this change I can prevent code duplication.

### DIFF
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -1,15 +1,13 @@
-import sys
 import warnings
 
 import numpy as np
 
-from nengo.config import Config
+from nengo.config import SupportDefaultsMixin
 from nengo.exceptions import ValidationError
 from nengo.params import (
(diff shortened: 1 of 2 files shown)


Seanny123:
LGTM.

hunse:
Yeah, I'm fine with this too. I still have some open questions about the config system in general, but this is a simple improvement so those can wait.

tbekolay:
Hm... I'm not a fan of these tiny mixin classes personally, though this one seems OK. There are some typos in the `SupportsDefaultsMixin` docstring. Sorry, this PR wasn't on my radar as it's not in a milestone...

jgosmann:
> I'm not a fan of these tiny mixin classes personally, though this one seems OK

To me it seems better than duplicating the code ...

Issue #937, opened 2016-01-26
Print time spent on building model


jgosmann:
The progress bar shows how long it took a simulation to finish. It would be nice to have an output how long it took to build the model, too. (A progress bar for that would be even more awesome, but more complicated to implement.)

hunse:
I do this in Nengo OCL with the logging system. I log info level messages timing various parts of the build process. I guess in Nengo we don't split up the build process in the same way, so it's harder to get info about how much time is spent building connections, ensembles, etc. 

An overall timing that we display to the user no matter what, as @jgosmann is suggesting here, should be quite easy and makes sense.

tcstewar:
It'd also be nice if such a system could also have hooks in it that would allow nengo_gui to pass the information along as well....  :)

xchoo:
I currently do it manually (i.e. using timestamps from `time.time`) for spaun builds.

jgosmann:
> I currently do it manually (i.e. using timestamps from time.time) for spaun builds.

I know that I could do that, but sometimes I build a model and wonder afterwards how long it took.

drasmuss:
I'm kind of the opposite, I'm constantly disabling the progress bar for every model I run because I'd rather not have information cluttering the console output unless I ask for it.  So not saying that we shouldn't put this in, just that it'd be nice if there was a way to turn it off.

hunse:
Do we have a config option for the progress bar? If not, we should. And one for this, too.

jgosmann:
Yes, we have a config option for that. Just put
```ini
[progress]
progress_bar: False
```
in your `~/.config/nengo/nengorc`

drasmuss:
Ah sweet, I didn't realize that was used for more than just decoder caching settings :tada:

PR #1035, opened 2016-04-25
Operator graph optimizer

jgosmann:
I updated this description with the new template. The old description can be found below.

**Description:**
This adds an optimizer to the simulator that goes through the operator graph and merges operators and their signals. This reduces the items to iterate over in slow Python (vs. fast NumPy) and gives larger blocks of sequentially aligned memory for improved CPU caching and prefetching.

**Motivation and context:**

Not everyone has a powerful GPU and can use nengo_ocl (especially at the summer school it is common to have a laptop without decent GPU). For those people this PR can significantly improve the run time.

This is also useful for running on high-performance clusters (in a serial-farming setup) like Sharcnet which mostly consist of CPU clusters and only few GPU clusters (and usually much more CPU cores than GPUs are provided).

Furthermore, when using large ensemble arrays (as in SPA, especially with spaopt #941, dot products, circular convolutions) this reduces the overhead that is introduced by many small ensembles.

No real interactions afaik, but #941 will benefit.

**How has this been tested?**
The optimizer is enabled by default and all the tests still pass. Furthermore, I am using the optimizer on my TCM model and on @ikajic's spa-rat model. Especially, the latter one has a large arrays of small ensembles (5018 elements in a cleanup memory) and uncovered a number of memory and runtime issues that have been fixed. @xchoo tested a slightly older version of the optimizer on Spaun. Read through the comments for quotes on run time improvements and additions to the build time (quoted build time for Spaun might be lower now with recent improvements).

I did not add a lot of unit tests because for the most part it is not clear to me how to test the optimizer. It is tested through the existing tests (all those models still work) which verifies its correctness in a way, but doesn't test that it actually speed up things or isn't a no-op. Good ideas welcome.

`optimizer.py` is the main algorithm and should be structured going from the high level to the low level of the algorithm. I tried to document it well, but it might not be sufficient to understand everything. Let me know where more comments would be helpful.

Maybe it is best to go through the code with me as it is somewhat complex.

**Types of changes:**

**Checklist:**

- [x] I have updated the documentation accordingly. (More documentation should be added, separate issue #1119.)

**Still to do:**
#1119, but I think it should be a separate PR.
# Old description for reference

This adds an optimizer to the simulator that goes through the operator graph and merges operators and their signals. This reduces the items to iterate over in slow Python (vs. fast NumPy) and gives larger blocks of sequentially aligned memory for improved CPU caching and prefetching.

At the moment the only operator that's being merged is `SimNeurons`, but I will look and see if this can be done for other operators too.

There is a trade-off, though. Running the simulation gets faster, but the build process will be slower. The algorithmic complexity of the optimizer is O(n^2 + e) where n is the number of operators and e the number of edges in the dependency graph.

A few "benchmarks" I run:
- Simulating an ensemble array with 100 one-dimensional ensembles with 50 neurons: Pure simulation time decreases from 18 to 4 seconds (4.5x speedup! :dash: )
- `py.test -n 6` from 60s down to 51s
- `py.test -n 6 --slow` from 415s down to 274s 
- `py.test` from 83s down to 71s

I haven't tested in on any more complicated models yet, but that should definitely be done. It would be nice to have #937 implemented to easily see the effect on build time.

It probably would be a good idea to have an option to turn the optimization on and off. Probably with an rc setting, but I feel like this could be something people want to change per instantiation of the `Simulator`. Also wondering if there should be some more fine grained control which operators to merge (or if we add other optimizations which set of these should be done).

~~I think there was an issue about adding optimizations to the simulator, but I can't find it right now.~~ Found it, this addresses #778.
# TODO
- [x] Merging of views has been implemented for `Copy` in the `op-opt-view` branch with a lot of code duplication. Refactor this.
- [x] Look at view merging for other operators
- [x] Look at general merging for other operators
- [x] Add more documentation about how the optimization is performed and some of the finecky details (views get broken in the first pass, then fixed later, then view merging is done because first pass introduces more views)
- [x] Could we need more than two passes? Or in other words could view merging introduce more views because of associated non-view signals? (--> only relevant if view merging is supported for other things than `Copy`)
- [x] Add test for signal merging (this should be simple enough to unit test and is part of the public `Signal` API). Not sure if it makes sense to test the optimizer in any other way than checking that it does not break any of the existing tests. Complex interactions are hard to cover in any other way anyways. Not sure if it is worth to construct and check simple cases. Maybe that certain optimizations are done at all?
- [x] Config options
- [x] Under some circumstances the optimizer seems to require a lot of memory. I should investigate this.
- [x] Always performs a minimum of two passes even if the first pass didn't lead to a reduction in the number of operators.
- [x] This currently use the `np.stack` function in one place which requires NumPy 1.10. Replace it with an equivalent available in earlier versions. Also, add a test because that part of the code is apparently never hit in the tests (no failure on Travis-CI).


jgosmann:
General note:
* The optimization algorithm is greedy in figuring out what to merge. Finding the optimal solution is equivalent to finding cliques in a graph which is NP-complete.

Important note to self to fix:
* [ ] If two operators access the same signal and both get merged with other operators, things will (probably) break.

jgosmann:
Another todo:
* [ ] add some logging for debugging purposes

Seanny123:
Whenever I hear NP-complete, I think "throw Simulated Annealing at it!". Would this be feasible for this optimisation and would it offer any significant advantage over the greedy-algo?

jgosmann:
It's probably more effort that it's worth (either in developer time or computation time). Given the constraints under which a merge is performed it might even be that the greedy algorithm is already optimal (at least I can't easily construct a counter example). If you'd loosen those constraints, such situation could occur more easily, but that would make correct merging of the operators also much more complicated (and probably wouldn't allow for a single NumPy operation anymore which is half the reason for the merge).

jgosmann:
Continuing my todo list:
* [ ] Merging of views has been implemented for `Copy` in the `op-opt-view` branch with a lot of code duplication. Refactor this.
* [ ] Look at view merging for other operators
* [ ] Look at general merging for other operators
* [ ] Add more documentation about how the optimization is performed and some of the finecky details (views get broken in the first pass, then fixed later, then view merging is done because first pass introduces more views)
* [ ] Could we need more than two passes? Or in other words could view merging introduce more views because of associated non-view signals?

jgosmann:
I figured out how to merge `DotInc` operations given SciPy is installed (with `bsr_matrix`) and realized how to merge even more stuff by correctly propagating memory layouts from on operation to another (though this requires somewhat more optimization passes and makes the build a bit slower). On my circular convolution test case this reduces the simulation time for 75 to 5 seconds (15x speedup! :dash: )

Still have to clean up my code quite a bit before I can commit this, though.

hunse:
How fast is your convolution benchmark without combining the DotIncs?

jgosmann:
18s (5227 operators instead of just 79).

hunse:
That's pretty good. Looking at it, it actually just [calls gemv](https://github.com/scipy/scipy/blob/master/scipy/sparse/sparsetools/bsr.h#L750) for all the individual matrices. But the fact that it's in a tight C loop removes all the Python overhead and obviously really helps. I was kind of worried it would be using some specialized sparse matrix code that won't be as fast as BLAS for large matrices, but that's obviously not the case.

The only drawback is that all the matrices have to be the same size, but that is often true because typically we have quite a few EnsembleArrays in large models. So overall, this should really help. Nice!

jgosmann:
According to @xchoo the build times for Spaun are reasonable with the latest improvements to the algorithm (though, I have to get back to him for exact numbers). The simulation time is cut down by 90% apparently. :dash: However, that is still slower than nengo_ocl. But then again, there might be situations where people cannot or don't want to install nengo_ocl or don't have a supported GPU. Also, maybe some of this can be used to speed up nengo_ocl as well?

jgosmann:
If I remember correctly the operator count was decreased to a few hundred thousand from a couple of million for Spaun.

xchoo:
Specifically, 1,334,851 operators to 245,760 operators. The build time was 2588s, and the runtime for 0.1s was 147.3s.

On the master nengo branch, the build time is typically ~1200s, and the runtime for 0.1s is ~700s

jgosmann:
How many optimization passes?

xchoo:
15 passes.

jgosmann:
* [ ] Under some circumstances the optimizer seems to require a lot of memory. I should investigate this.

jgosmann:
I could use a memory allocation guru ... under certain circumstances the optimizer will lead to a very high memory usage (tripled usage during the first pass). My guess right now is that this is due to memory fragmentation because a lot of small memory segments are being merged into one big memory segment which does not fit in any of the freed areas of the small memory segments.

My best idea to solve this is to limit the size of newly created memory segments. Unfortunately, I have no idea what a good limit would be.

jgosmann:
Apparently, the NumPy arrays weren't the problem, but some very long strings I was constructing. Still getting a somewhat high peak memory usage under certain circumstances, but at least the memory gets freeed afterwards.

jgosmann:
The high peak memory usage is caused by the transitive closure ...

hunse:
I'm trying to figure out if there's overlap between this and the planners that I have in nengo_ocl. They're both trying to do a similar thing, in terms of figuring out which operators can be executed in parallel and can be combined.

It's not clear to me what part of this code is figuring out what is parallel and what is not, though. Maybe I'll need an explanation in person sometime.

jgosmann:
> It's not clear to me what part of this code is figuring out what is parallel and what is not, though.

To answer that specific question: It is [this line](https://github.com/nengo/nengo/blob/f4aa0e438d4fcfb0e096d91ab4429d153023764b/nengo/builder/optimizer.py#L217) where `tc` is the transititve closure computed [here](https://github.com/nengo/nengo/blob/f4aa0e438d4fcfb0e096d91ab4429d153023764b/nengo/builder/optimizer.py#L149) (if one operator is in the transitive closure of the other, one depends on the other and they cannot be executed in parallel)

At least for now. Unfortunately, the data structure for the transitive closure can take up a lot of memory, but it seems that everything using less memory will increase the runtime.

> Maybe I'll need an explanation in person sometime.

That might be necessary for the review anyways. Though, it would be good if someone could get an understanding just by reading the code ...

jgosmann:
Some more stats I collected on @ikajic's spa-rat model:
* Build time: ~9min
* Optimizer run time (with latest improvements): ~4min (without those improvements > 15min)
* Simulation time without optimizer: ~4h
* Simulation time with optimizer: ~30min
* Peak memory usage without optimizer: <6GB
* Peak memory usage with optimizer: <9GB

jgosmann:
For comparison with nengo_ocl (GeForce GTX 980):
* Build time: ~7min (excluding semantic pointer generation, included in stats above)
* Planning etc ~70s
* Simulation time: ~1min
* Peak memory usage: <6GB

jgosmann:
Just noticed that the optimizer makes LIF neurons run faster than direct mode (at least in some cases). :open_mouth:

jgosmann:
This should be ready for review now. I updated the PR description with our template.

jgosmann:
This uses `np.stack` which is not available on Sharcnet (Numpy 1.9). Should be an easy fix though. Not sure why our tests are not hitting that.

jgosmann:
Moved all the code specific to the optimizer to `optimizer.py`. This seems much cleaner to me. We might want to discuss the order of function and their naming ...

tbekolay:
> This seems much cleaner to me.

Agreed! I took a quick look and this is much cleaner than the last time I skimmed the diff. Some other general things I saw were that a lot of uses of classes could probably be done with functions instead, and lots of docstrings don't match [PEP-257 conventions](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings), mostly in not having a one-line summary to start. I'll look at this in more detail after I wrap up some 2.2.0 stuff -- this PR's my first priority post 2.2.0.

jgosmann:
> I saw were that a lot of uses of classes could probably be done with functions instead

I think the classes add useful structure here, but I'm happy to be proven wrong.

> mostly in not having a one-line summary to start

Some complicated things are really hard to express in a single line.

tbekolay:
> Some complicated things are really hard to express in a single line.

The complicated bits aren't necessary to get across in the summary line... function/class names manage to fit in one line.

jgosmann:
> function/class names manage to fit in one line.

But a one line summary just repeating the function/class name is not helpful in any way.

jgosmann:
If anyone is looking into reviewing this, I have an early draft of a paper on this I'm working on, that I can send you. Maybe it's helpful to get a high level overview.

arvoelke:
While I'm thinking of this, do you optimize away the dangling leaves? More precisely, the signals that do not eventually reach some probe? Looking at the builder, it looks as though signals are created for every probable attribute for every neuron, even if they are not being probed. Also, many sub-networks and/or models will create superfluous output nodes. I can see this using a reasonable chunk of RAM / time.

jgosmann:
I'm not pruning dangling leaves. I think that is out of scope of this specific optimizer because that information is not contained in the dependency graph itself.

I'm not sure if there are actually that many dangling leaves (but I could be wrong on this). I believe most signals will be used (e.g. the voltage of neurons is usually not probed, but still required to update the neuron state).

For superfluous output nodes, it might be easier to do the optimization on the model instead of the dependency graph? (Prune everything without outgoing connections.)

To me it seems the largest bottle neck (after doing the operator graph optimization) is the single threaded nature of the reference simulator.

hunse:
> I believe most signals will be used (e.g. the voltage of neurons is usually not probed, but still required to update the neuron state).

I agree. I don't think we're making any extra signals to allow something to be probable. When someone makes a `Probe` object, then we make a signal to hold the filtered value of the probed attribute, if necessary.

In #654, I make operators for probes, which would make it much easier to do some of these optimizations, since then it will be obvious what the output signals actually are.

jgosmann:
I ran some memory benchmarks measuring the [unique set size](https://en.wikipedia.org
/wiki/Unique_set_size). With my n-back model both the reference and optimized reference simulatar take slightly below 170MB. Nengo OCL requires up to 320MB. For a circular convolution with 500 neurons per dimension and 500 dimensions the optimizer requires slightly more memory than the reference backend: 345 vs 312MB. Nengo OCL exceeds this again with up to 473MB.

I will run further benchmarks on the Spaun model, but this will take a while

jgosmann:
I made a commit to install scipy on Travis-CI to increase the amount of code hit during the tests.

arvoelke:
Hmm yeah forget my comment about the probeable neuron attributes. I did not think that through.

Also note to determine if something is superfluous, we also need to determine if it eventually connects to a `Node` that has specified an `output` function, since it may produce side-effects.

Seanny123:
What's the state of this pull request? I know it has some conflicts, but the "I've written tests to cover my changes" is also unchecked. Does that make this still a work in progress?

jgosmann:
The optimizer is covered by existing tests. It might be possible to write unit tests to test parts of it (as opposed to the complete system), but I am not sure if that is worthwhile. There a actually a few new tests that I added.

So it's not a work in progress at the moment, but it would be if people think more tests are required.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+    This reduces the number of iterators to iterate over in slow Python code
+    (as opposed to fast C code) and due to the merging operations will be
+    performed on larger chunks of sequential memory making better use of CPU
+    caching and prefetching.
+
+    The optimization algorithm has a worst case complexity of O(n^2 + e) where
+    n is the number of operators and e the number of edges in the dependency
+    graph. In practice the run time will be much better because not all n^2
+    pairwise combinations of operators will be evaluated. A grouping depending
+    on the operator type and view bases will be done with dictionaries. This
```
Seanny123: I think you mean "is done with dictionaries". Using future tense makes it sound like it hasn't been implemented yet.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+    in many cases and gives a runtime much closer to linear in most cases.
+
+    Note that both `model` and `dg` will be changed by the `optimize` method.
+
+    Parameters
+    ----------
+    model : :class:`nengo.builder.builder.Model`
+        Builder output to optimize.
+    dg : dict
+        Dict of the form {a: {b, c}} where b and c depend on a specifying the
```
Seanny123: Could we change this to "`{a: {b, c}}` where `b` and `c` depend on `a` specifying the" to make it easier to parse?

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+    max_passes = IntParam('max_passes', default=None, optional=True)
+
+    def __init__(self, model, dg):
+        self.enabled = RcDefault
+        self.max_passes = RcDefault
+
+        self.model = model
+        self.dg = dg
+
+        # Internal dict mapping Signals to Operators accessing those signals.
```
Seanny123: Phrasing nitpick. I think I prefer the phrasing "Internal dict mapping from a Signal to the Operators accessing the Signal." Or you can fix the plural if it's not a one-to-many mapping.

Seanny123: Also, I think you can remove the word "Internal" from both the `self._sig2op` and the `self._merged` description. I think it's obvious they're internal since they're private variables?

nengo/utils/graphs.py
```diff
@@ -98,6 +98,43 @@ def toposort(edges):
     return ordered
 
 
+def transitive_closure(edges, topo_sorted=None):
+    """Constructs the transitive_closure of a DAG.
```
Seanny123: According to the [first link I found by searching](http://www.geeksforgeeks.org/transitive-closure-of-a-graph/), I can consider a "transitive closure" as a synonym for a "reach-ability matrix". Is that accurate? Or is something else going on here? If yes, could you elaborate? If no, would it be a good idea to insert that layman's terms translation?

nengo/utils/graphs.py
```diff
@@ -98,6 +98,43 @@ def toposort(edges):
...
+    edges : dict
+        Dict of the form {a: {b, c}} where b and c depend on a. Must not
+        contain cycles.
+    topo_sorted : sequence, optional
+        The topological sorting of the vertices. If not passed in the algorithm
+        will do a topological sort.
+
+    Returns
+    -------
+    The transitive closure data structure as `edges`.
```
Seanny123: Could you explain a bit more about what this is returning? From what I can tell, it's returning a dictionary where every key is an a node and the key is all the other nodes it can reach? Is that correct?

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+
+import numpy as np
+
+from nengo.builder.neurons import SimNeurons
+from nengo.builder import operator
+from nengo.builder.operator import DotInc, ElementwiseInc, SlicedCopy
+from nengo.builder.signal import Signal
+from nengo.config import SupportRcDefaultsMixin
+from nengo.params import BoolParam, IntParam, RcDefault
+from nengo.utils.compat import zip_longest
```
Seanny123: Cool. I didn't even know that type of iterator existed!

nengo/builder/tests/test_optimizer.py
```diff
@@ -0,0 +1,119 @@
...
+    assert not compatible(
+        [Signal(np.empty((2, 1))), Signal(np.empty((2, 2)))], axis=0)
+
+    # shape mismatch
+    assert not compatible(
+        [Signal(np.empty((2,))), Signal(np.empty((2, 2)))])
+
+    # mixed dtype
+    assert not compatible(
+        [Signal(np.empty(2, dtype=int)), Signal(np.empty(2, dtype=float))])
```
Seanny123: I have a silly question. Is the cost of up-converting an `int` into a `float` greater than leaving two signals incompatible? Or would that up-conversion cause unintended consequences in some way?

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+
+    In comparison to the `check_*_mergeable` functions, this function does
+    not throw exceptions (either behavior might be desired dependent on the
+    context) and allows for either signals or signal views.
+    """
+    if len(set(signals)) != len(signals):
+        # Signal appears twice in list.
+        return False
+    for s in signals:
+        if s.ndim != signals[0].ndim:
```
Seanny123: In my opinion and [in the opinion of StackOverflow](http://stackoverflow.com/q/22782785/1079075), you should only have a stack of if-statements if the cases aren't mutually exclusive. From what I understand, these conditions aren't mutually exclusive and their ordering is arbitrary, so I think I'm okay with the stacked if-statements.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+    def _update_model_sigs(self, sig_replacements):
+        """Updates the signal dictionary in the builder model with
+        `sig_replacements`.
+        """
+        for k in self.model.sig:
+            for name, v in self.model.sig[k].items():
+                self.model.sig[k][name] = sig_replacements.get(v, v)
+
+
+class AbstractMerger(object):
```
Seanny123: Why not move this into a separate file?

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+                if isinstance(sig, Signal):
+                    setattr(op, key, fn(sig))
+
+            op.sets = [fn(s) for s in op.sets]
+            op.incs = [fn(s) for s in op.incs]
+            op.reads = [fn(s) for s in op.reads]
+            op.updates = [fn(s) for s in op.updates]
+
+    @classmethod
+    def _get_sig_view_replacements(cls, ops, sig_replacements):
```
Seanny123: I'm confused because I think a lot of stuff is being passed by reference and then the references are being modified. Is this function supposed to modify the `ops`?

Seanny123: If yes, could you update this docstring to make that clear?

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+            type(op1) == type(op2) and
+            type(op1) in cls.op_mergers and
+            cls.op_mergers[type(op1)].can_merge(op1, op2))
+
+    @classmethod
+    def merge_ops(cls, ops):
+        return cls.op_mergers[type(ops[0])].merge(ops)
+
+    @classmethod
+    def register_merger(cls, op_type):
```
Seanny123: What is this decorator supposed to stop or enable? I'm getting a bit sleepy so it might be really obvious to me tomorrow morning.

nengo/utils/graphs.py
jgosmann: I think “transitive closure” is a well-defined term and can be easily looked up (e.g. on Wikipedia). Technically, it is not a reach-ability *matrix* here because the graph is not represented with a adjancency matrix, but something more similar to adjancency lists.

nengo/utils/graphs.py
jgosmann: I think the sentence is missing a word. It should say something like it is using the same “data structure as `edges`”. Would that clarify it? Or is it better to repeat the description of the datat structure?

nengo/builder/tests/test_optimizer.py
jgosmann: It definitely has the potential to change the exact model results. Obviously, we can't convert the `float` to `int`. Converting `int` to `float` opens the door to rounding errors and all kind of weird behaviours. If we are using `int` signals somewhere, we probably would not want to convert them to `float`.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+        op_replacements : dict
+            Dictionary that maps from original operators to the newly merged
+            operators. To be used to update the dependency graph.
+        sig_replacements : dict
+            Dictionary that maps from original signals to new signals resulting
+            from the merges. To be used to update the signals referenced by
+            the operators.
+        """
+        step_order = toposort(self.dg)
+        tc = transitive_closure(self.dg, step_order)
```
jgosmann: `transitive_closure` only sorts if the second argument is not given.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+        sig_replacements = {}
+
+        # Note that the following loops will stop as soon as any operators
+        # have been merged, so that merges are performed on at most one type
+        # of operators. This function will be called again once the dependency
+        # graph and signals have been updated with those merges to continue
+        # merging other operator types.
+
+        # Heuristic order to reduce runtime and number of passes
+        for tp in [ElementwiseInc, SlicedCopy, DotInc, SimNeurons]:
```
jgosmann: The comment in the line before is supposed to be that explanation. How can I improve it?

jgosmann: There are other types of operators that can be merged, but these are done in the next loop in random order.

nengo/builder/optimizer.py
jgosmann: No, it is not supposed to modify `ops`.

nengo/builder/optimizer.py
jgosmann: Because it is inherently coupled with the rest of the optimizer?

Seanny123: Right. So maybe move the optimizer to it's own folder so it's components can be split up?

nengo/builder/optimizer.py
Seanny123: Registers it for what? What does "register" mean in this context?

nengo/utils/graphs.py
Seanny123: Cool. Then I think it would be a good idea to include the synonym "list of adjacencies" in the description.

nengo/utils/graphs.py
Seanny123: Repeating the description of the data structure would be useful, I think.

nengo/builder/optimizer.py
Seanny123: If it's not supposed to modify `ops`, then why does `_map_onto_op_signals` iterates through `ops` and assigns new values to their attributes? I'm talking about this block:

            op.sets = [fn(s) for s in op.sets]
            op.incs = [fn(s) for s in op.incs]
            op.reads = [fn(s) for s in op.reads]
            op.updates = [fn(s) for s in op.updates]

nengo/builder/operator.py
```diff
@@ -643,6 +643,70 @@ def step_dotinc():
         return step_dotinc
 
 
+class BsrDotInc(Operator):
```
Seanny123: Maybe note the usage of this operator, like you do in your report? Also, how likely is it that a model won't require this operation if it's being optimized? Because, if a person tries to turn on the optimizer, fails the import, it would be nice to send them a warning informing them that if they want an actual performance boost they should use SciPy. Similar to what we do with the Basal Ganglia.

nengo/builder/optimizer.py
jgosmann: Registers the class that provides the logic required to merge a specific type of operator. The merge function call and related functions get dispatched to that registered class.

nengo/builder/optimizer.py
jgosmann: Good question. I have to look more into this. Maybe that function call isn't supposed to be there anymore.

nengo/builder/optimizer.py
jgosmann: I'm not entirely convinced. What would be the advantage of that?

nengo/builder/optimizer.py
jgosmann: The main argument for `elif` seems to be in case of mutually exclusive conditions, it is less likely to introduce an error when someone changes the code. I feel in this case it less likely without the `elif` because the conditions are not mutually exclusive. If one of the `return` statements is removed, it should still return `False` if any of the other `if`-conditions evaluates to false.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+from nengo.builder.signal import Signal
+from nengo.config import SupportRcDefaultsMixin
+from nengo.params import BoolParam, IntParam, RcDefault
+from nengo.utils.compat import zip_longest
+from nengo.utils.graphs import toposort, transitive_closure
+
+logger = logging.getLogger(__name__)
+
+
+class OpMergeOptimizer(SupportRcDefaultsMixin):
```
jgosmann: I could maybe see separating out some of the code checking for mergeability and dispatching that stuff to the `*Merger` classes. However:
* Parts of that code are inherently tied to the rest of the implementation. For example in `_perform_merges_for_view_subset` it is essential that the comparison in the call to `self._check_sequential(merge[-1], op2)` uses `<` and not `<=`. By separating this in separate classes it might appear that there is less coupling than there actually is.
* I'm not sure what sort of construct would be right. All of the methods I would separate out wouldn't require a class that gets instantiated. Should I define all the methods as `@classmethod`? Or should I make them module methods?

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,969 @@
...
+        # order for the memory alignment whereas operators without views could
+        # be merged in a random order. Merging the views of operators will
+        # propagate requirements in the memory ordering via the other
+        # associated signals of the operator to other operators.
+        #
+        # Once no more operators with views can be merged, we try to merge
+        # operators without views and then try again merging views (because
+        # each operator merge might generate new views).
+
+        before, after = None, None
```
jgosmann: I see your argument for not initializing with `None`. My argument for doing so: At that point we don't know the proper values (at least not for `after`), thus I would like to avoid to give the impression that we do. To me the initialization with `None` also seems beneficial in case a bug is introduced that prevents the proper initialization later in the loop. With the `None` initialization the code is more likely to fail early because something requires an `int` instead of silently continuing with a wrong value. Also, when printing the value for debugging purposes it would be obvious that the value is uninitialized.

If you or others still feel strongly about this after considering this, I'm willing to change to it.

nengo/builder/optimizer.py
jgosmann: So the function actually does not modify `ops`. The `get_sig_view_replacement` function that gets mapped onto `ops` with `_map_onto_op_signals` just returns the unmodified signal `s` and stores the replacement in `view_replacements`.

I'll insert a comment that hopefully clarifies this.

nengo/utils/graphs.py
jgosmann: Not sure, it wouldn't be correct to say the transititve closure is a “list of adjancencies” (that is only the data structure used to represent graphs and even that is only partially correct because we're using sets). To define it correctly and precisely it would probably require quite a bit of text. I would prefer to keep it concise because it is a well-defined term that can easily be looked up.

jgosmann: (There is also no explanation of what a topological sort is in the same file.)

nengo/builder/optimizer.py
Seanny123: I have no idea, due to my inexperience in structuring large Python projects and will defer this to the next reviewer.

Seanny123:
@jgosmann has done an awesome job of addressing my concerns and engaging me on various issues. The things we're still discussing could probably use some clarity from a second reviewer, so I'm marking this as "needs second review".

The next reviewer, whoever that brave soul may be, should make sure to read the most recent version of Jan's paper before reviewing this code, as it will greatly help with understanding.

nengo/builder/operator.py
jgosmann: I think my stance on this depends on whether the optimizer is activated by default. If this is the case, the user would see that warning potentially for every model.

Seanny123: Oh god, you're totally right. In that case, since making an offline note of whether a warning has been given recently would be way too painful, I'm totally fine with just a silent warning.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,974 @@
...
+            start = np.searchsorted(
+                offsets, offsets[i] + self._first_view_size(op1), side='left')
+            for op2 in sorted_subset[start:]:
+                can_merge = (
+                    op2 not in self._merged and
+                    self._mergeable(merge[-1], op2, view_indices, tc) and
+                    self._is_memory_access_sequential([merge[-1], op2]))
+                if can_merge:
+                    merge.append(op2)
+                elif self._check_sequential(merge[-1], op2):
```
tbekolay: I'm a bit confused as to why we break out of the loop if this is true, rather than false... we can merge if the memory access is sequential. This check is a simple check for it being sequential. So if that were true, then shouldn't it be possible for subsequent ops to also be sequential?

jgosmann:
I think it is not super costly because ni data should be copied, but if we can avoid doing it on every time step that might still be a good thing.

On February 10, 2017 4:51:30 PM EST, Trevor Bekolay <notifications@github.com> wrote:
>tbekolay commented on this pull request.
>
>
>
>> +        self.sets = []
>+        self.incs = [Y]
>+        self.reads = [A, X]
>+        self.updates = []
>+
>+    def _descstr(self):
>+        return '%s, %s -> %s' % (self.A, self.X, self.Y)
>+
>+    def make_step(self, signals, dt, rng):
>+        X = signals[self.X]
>+        A = signals[self.A]
>+        Y = signals[self.Y]
>+
>+        def step_dotinc():
>+            from scipy.sparse import bsr_matrix
>+            mat_A = bsr_matrix((A, self.indices, self.indptr))
>
>This constructs a new `bsr_matrix` on each timestep. Is it possible to
>avoid this? Is it worth trying to avoid this? I don't have any sense of
>how costly this is, so if you're not sure I can do some profiling.
>
>> +            # Find operators that can be merged.
>+            merge = [op1]
>+            # For a merge to be possible the view of the next operator
>has to
>+            # start where the view of op1 ends. Because we have sorted
>the
>+            # operators by the start of their views we can do a binary
>search
>+            # and potentially skip a number of operators at the
>beginning.
>+            start = np.searchsorted(
>+                offsets, offsets[i] + self._first_view_size(op1),
>side='left')
>+            for op2 in sorted_subset[start:]:
>+                can_merge = (
>+                    op2 not in self._merged and
>+                    self._mergeable(merge[-1], op2, view_indices, tc)
>and
>+                    self._is_memory_access_sequential([merge[-1],
>op2]))
>+                if can_merge:
>+                    merge.append(op2)
>+                elif self._check_sequential(merge[-1], op2):
>
>I'm a bit confused as to why we break out of the loop if this is true,
>rather than false... we can merge if the memory access is sequential.
>This check is a simple check for it being sequential. So if that were
>true, then shouldn't it be possible for subsequent ops to also be
>sequential?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/nengo/nengo/pull/1035#pullrequestreview-20556777

-- 

nengo/builder/operator.py
```diff
@@ -643,6 +643,70 @@ def step_dotinc():
...
+        return '%s, %s -> %s' % (self.A, self.X, self.Y)
+
+    def make_step(self, signals, dt, rng):
+        X = signals[self.X]
+        A = signals[self.A]
+        Y = signals[self.Y]
+
+        def step_dotinc():
+            from scipy.sparse import bsr_matrix
+            mat_A = bsr_matrix((A, self.indices, self.indptr))
```
jgosmann: For some reason my email reply does not show up here ... anyways:

I don't think that this operation is very costly because no data should be copied. But if it is possible to do the construction only once that would of course be great. I have the feeling that I tried that and for some reason it was not possible, but I cannot recall the exact details anymore. I'll think a bit more about it.

jgosmann: oh, the email reply was attached to the general thread ...

nengo/builder/optimizer.py
jgosmann: `_check_sequential` is a misnomer and should be renamed (but not sure what a better name would be, maybe `_check_for_gap`?). It only returns `True` if there is a gap between `merge[-1]` and `op2`.

tbekolay: Ah okay. So is it possible that this check will be True, but the previous check will be True also? In other words, could this be the if instead of the elif?

jgosmann: I cannot come up with a case where both of these could be `True`, but I might be missing something.

nengo/builder/optimizer.py
```diff
@@ -211,7 +211,8 @@ def _perform_merges(self, only_merge_ops_with_view):
             op_replacements.update(opr)
             sig_replacements.update(sigr)
             del by_type[tp]  # no need to process tp again in the second loop
-            if not only_merge_ops_with_view and len(opr) > 0:
+            n_changes = sum(new is not old for old, new in iteritems(opr))
```
jgosmann: We don't need the exact number, do we? An `any` instead of `sum` might be more efficient.

tbekolay:
OK, I've been studying this code for the past couple of weeks... I have some minor refactoring in progress to hopefully address the too-many-methods issue that @Seanny123 brought up in his review, but before I finalize any of that I've been struggling with an issue that I can't figure out.

I noticed, after refactoring, that I was merging more operators than were previously being merged. Looking at [this line](https://github.com/nengo/nengo/pull/1035/commits/4611b952da6b5951db6e0d580c7fb8f3a85d38a9#diff-f23ca6a62a61764f4d9a7c944035d7beL214), I interpreted the `len(opr)` check as attempting to detect if any operators had been merged in the previous subset. If so, then we stop the current optimization pass and finalize the merge. Because of the way in which the `dg` is updated, however, `len(opr)` was always greater than 0 because all ops are added to `opr` at the start of the subset merge, whether any substitutions are made or not. So, I made 4611b952da6b5951db6e0d580c to continue as long as no actual substitutions had been made.

However, when I make this change, 14 unit tests start failing for a few reasons, indicating that the optimization is changing the results of some simulations. It looks like most of the failures are related to learning rules, but they pop up in a few other places too (see [travis CI build results](https://travis-ci.org/nengo/nengo/builds/203927608). I know it's been quite a while since you made the PR @jgosmann (apologies for that) but do you have any thoughts as to why these tests might be failing?

jgosmann:
This is what I found out so far:
* There seems to be a problem merging DotInc operators to BsrDotInc operators. This might be limited to cases where a reshape is necessary, but that might also be a random correlation.
* There seems to be another problems somewhere.

jgosmann:
For some reason the tests are not deterministic if `PYTHONHASHSEED` is not set to a fixed seed.

jgosmann:
Even with PYTHONHASHSEED set the tests are still non-deterministic. 😖

jgosmann:
Ok, of course it's non-deterministic because optimizations can depend on where objects are in memory. Still annoying to debug. 😞

jgosmann:
So, there was a number of different bugs, but I hope I fixed them all now (all tests seems to be consistently passing locally). I have to take a second look at the changes tomorrow as the current fixes are the minimal changes to get it working, but there might be better ways to implement these fixes. Also note that 19a2b942fb868c678626385e2a9e27ace06ea875 has the side effect of disabling the merging of certain DotInc operations at the moment.

jgosmann:
I think, I also have to update the (already submitted) paper and rerun the benchmarks.

jgosmann:
Cleaned up the fixes from yesterday and handing this back to you, @tbekolay.

tbekolay:
OK, I've finished a first draft of a refactoring, which I hope will address the concerns @Seanny123 had in his review. I haven't yet gone through and ensured that all docstrings are present and up to date, but I wanted to get @jgosmann's feedback on the reorganization before putting in that time. An explanation of the main changes is included in the commit message of the main refactoring commit, cde067b10e831175fda5e3d, so see that for the main details.

In addition to polishing up the documentation, one thing I'd like to change (if consensus agrees) is to remove most of the underscores from the beginning of the classes/functions in most of the file... In the refactoring, I attempted to maintain similar visibility levels, in terms of what code is accessible through underscored classes/function and what code is accessible through non-underscored things. However, I generally find underscores to make things more confusing when reading through code, as it indicates that shouldn't actually add that to my mental model of what's going on since I shouldn't be calling it anyhow, it's just an implementation detail. But in many cases, I do end up needing to know those things, and now I additionally have to remember that it's an underscored thing. Perhaps that a personal aesthetic taste thing though, so it would be good to get other thoughts on pros/cons of underscoring (without [distracting from the rest of the changes](https://en.wikipedia.org/wiki/Law_of_triviality)).

nengo/builder/optimizer.py
```diff
@@ -11,386 +10,707 @@
...
+            tomerge.optype in cls.op_checkers and
+            type(op) is tomerge.optype and
+            op not in tomerge.merged and
+            op not in tomerge.all_tc and
+            len(tomerge.tc[op].intersection(tomerge.ops)) == 0 and
+            all(tomerge.check_signals(o, op) for o in tomerge.ops) and
+            all(op not in tomerge.tc[o] and o not in tomerge.tc[op]
+                for o in tomerge.merged) and
+            cls.sig_checkers[tomerge.optype](op, tomerge) and
+            cls.op_checkers[tomerge.optype](tomerge.last_op, op))
```
jgosmann: I find this hard to parse because it is a long list of conditions without their intend. Prior to the merge I assigned groups of these conditions to variables as a way of documenting, for example `indepent_of_prior_merges` -> “these are the conditions for this operator to be independent of prior merges”. This gives then also a much more human readable statement in the end when combining those variables.

nengo/builder/optimizer.py
```diff
@@ -11,386 +10,707 @@
...
 
     @classmethod
-    def get_merger(cls, op):
-        return cls.op_mergers[type(op)]
+    def is_type_mergeable(cls, optype):
+        return optype in cls.op_checkers and optype in cls.op_mergers
```
jgosmann: Why isn't this checking `optype in cls.sig_checkers`?

nengo/builder/optimizer.py
```diff
@@ -11,386 +10,707 @@
...
+
+@_OpMerger.register_op_checker(SimNeurons)
+def simneurons_checker(op1, op2):
+    return (op1.neurons == op2.neurons and
+            all(_SigMerger.check(s) for s in
+                zip(op1.all_signals, op2.all_signals)))
+
+
+def _gather(ops, key):
+    return [getattr(o, key) for o in ops]
```
jgosmann: I was slightly surpresit to find a non-checker/non-merger function here, it seems a bit out of place here even though it is only used in `simneurons_merger`. Maybe move `_gather` into `simneurons_merger`?

nengo/builder/optimizer.py
```diff
@@ -11,386 +10,707 @@
...
+        self.info.clear()
+
+
+_opinfo = _OpInfo()
+
+
+class _OpsToMerge(object):
+    """Analyze and store extra information about a list of ops to be merged."""
+
+    def __init__(self, initial_op, merged, tc):
```
jgosmann: `tc` might be somewhat cryptic without a doc string. Maybe it is also worth renaming it to `transitive_closure`?

nengo/builder/optimizer.py
```diff
@@ -11,386 +10,707 @@
...
+        self.merged = merged
+        self.tc = tc
+        self.ops = [initial_op]
+        self.optype = type(initial_op)
+
+        self.all_signals = set(initial_op.all_signals)
+        self.all_tc = set(self.tc[initial_op])
```
jgosmann: `all_tc` is a slightly weird name because there is only one transitive closure. Maybe use something llike `all_dependants`?

nengo/builder/optimizer.py
```diff
@@ -11,386 +10,707 @@
...
+
+    def add(self, op):
+        self.ops.append(op)
+        self.all_signals.update(op.all_signals)
+        self.all_tc.update(self.tc[op])
+
+    @staticmethod
+    def check_signals(op1, op2):
+        for s1, s2 in zip(op1.all_signals, op2.all_signals):
+            # If one signal's op is a view, the other must be as well
```
jgosmann: I suppose this should say “If one op's signal is a view”?

nengo/builder/optimizer.py
jgosmann: Also `all_tc` makes it seem that `tc` is the transitive closure of a single operator, but it is the complete transitive closure graph (whereas `all_tc` is not a graph).

nengo/builder/optimizer.py
```diff
@@ -11,386 +10,707 @@
...
-            self._first_view_offset(op1) + self._first_view_size(op1) <
-            self._first_view_offset(op2))
+        # At this point, we may have marked some ops for merge.
+        # Some signals may therefore have been replaced.
+        # Those signals might be used in some other op, so we go through
+        # all other ops and do the same signal replacement if necessary.
 
-    def _perform_merges_for_subset(
-            self, subset, tc, only_merge_ops_with_view=True):
+        # TODO: do we have to do all ops or can we get away with a subset?
```
jgosmann: Potentially with `sig2ops`?

To me the following piece is a different level of abstraction and should go in its own function.

nengo/builder/optimizer.py
```diff
@@ -11,386 +10,707 @@
...
+        return repr(self.info)
+
+    def __str__(self):
+        return str(self.info)
+
+    def clear(self):
+        self.info.clear()
+
+
+_opinfo = _OpInfo()
```
jgosmann: I'm worried that this global variable can cause problems in multithreaded programs.

nengo/builder/optimizer.py
```diff
@@ -399,67 +719,37 @@ def _perform_merges_for_view_subset(
...
+                # has been updated
                 continue
-            if any(op1 in tc[op] or op in tc[op1] for op in self._merged):
+
+            if any(op1 in self.tc[op] or op in self.tc[op1]
+                   for op in self.merged):
                 continue
 
-            view_indices = self._get_view_indices(op1)
+            # Find operators to be merged.
```
jgosmann: This comment is misleading to me because the next line does not find operators to merge, but just constructs some supporting data structures.

nengo/builder/optimizer.py
```diff
@@ -469,561 +759,137 @@ def _perform_merges_for_view_subset(
...
+
+    def finalize(self):
+        """Finalizes merges done during the pass."""
+
+        for old, new in iteritems(self.op_replacements):
+            assert old is not new
+            if new not in self.dg:
+                self.dg[new] = set()
+            for dep in self.dg[old]:
+                self.dg[new].add(self.op_replacements.get(dep, dep))
```
jgosmann: I think these two lines can be replaced by
`self.dg[new].update(self.dg[old])` because the replacement is done in the next loop anyways.

jgosmann:
I made some inline comments, but I can make commits to address those things myself if desired.

I like the general direction of the refactoring with one exception: The `*_checker` and `*_merger` functions (and to some degree the `*_sig_checker` functions) don't reflect the tight coupling of these functions for one operator type. The correctness of a `*_checker` will depend on what the `*_merger` is doing, but in the refactoring this functions are independently defined and registered. In a way it also suggest that you could write a checker without a merger (or a merger without a checker), but it only makes sense to write both.

One stylistic thing: I would prefer a top-down order in the file, i.e. start with the `optimize` function and then as you scroll down have increasingly more specific and detailed function. That way you get the gist of whats going on in the beginning instead of tiny pieces where you don't see the big picture how they fit together.

As for the underscores: I'm probably fine with removing them, maybe add an `__all__ = ['optimize']` line to give some indication what the important function is (also, putting it at near the top would help with that imo). I probably would avoid including the “private” operator classes in the documentation to not give the impression that those are the part of any public API.

jgosmann:
One more thing: I'm ok with postponing the discussion about `SupportsRcDefaultMixin`, but there are a few things we should consider given the ad-hoc solution of adding an argument to the simulator class:
* This makes it slightly harder to write code for multiple backends if the `optimize` argument is changed, for example `backend(model, optimize=False)` will not work across backends.
* It also makes it slightly harder to write code compatible across Nengo versions. Older version do not support `optimize=False`, but I suppose the my solution would have the same problem that one cannot set the config option without the class existing. But if we decide to change this in the future and remove the `optimize` argument, it is also a problem. Though also true for any solution, if we decide to change it in the future and there are other things that have been deprecated etc.
* It is not possible to deactivate the optimizer when running models in the GUI or in pytry. Which might be ok depending on our confidence that it will always work as expected?

tcstewar:
> It is not possible to deactivate the optimizer when running models in the GUI or in pytry. Which might be ok depending on our confidence that it will always work as expected?

I can't say much on the other points, but to this one, I'm pretty sure that both pytry and the GUI need to have ways of specifying arguments to be passed to the Simulator.  I've wanted that multiple times in the last two weeks.  So I wouldn't worry about this aspect for now.

nengo/builder/optimizer.py
tbekolay: I'm OK to change it to that, but there is a speed hit... with all of the conditions in one statement, Python will short circuit and stop early if possible. With separate boolean variables, at least one statement in all of those variables will have to be evaluated.

Another way to help readability would be to group the conditions and add comments noting what each group of conditions is doing. That could also be a minor speed hit (if some conditions take longer than others and are in the wrong order) but it's less likely.

jgosmann: I'm an advocate for self-documenting code. If I can express something in code, I prefer that over adding comments. Speed might also be a concern, but I would like to verify that it makes a noticeable difference first.

tbekolay:
Pushed changes for the issues you pointed out @jgosmann. Didn't address the new Simulator argument, but it would be good to talk about that before merging.

One option would be to bake this into the builder rather than the simulator, some kind of post-build function that gets called, which I believe other PRs / projects have wanted. We've already been relatively clear that the builder API isn't well-defined yet, and is next in line for a refactoring, so this would (hopefully) reduce the number of things we would have to deprecate.

Regarding our confidence in the optimization, that reminds me that one other thing I'd like to do before merging is add some tests to verify that the optimization doesn't affect the model at all. Since it's just merging signals / ops, as long as the seed is set, the resulting model should be exactly the same before and after optimization. A few tests of that on relatively complicated models would provide some peace of mind!

drasmuss:
`nengo_deeplearning` also makes use of a lot of custom `Simulator` arguments.  Not that that couldn't be switched to some other method, just weighing in that it seems to be a thing that a lot of projects are gravitating towards.

jgosmann:
Further changes look good to me.

> Regarding our confidence in the optimization, that reminds me that one other thing I'd like to do before merging is add some tests to verify that the optimization doesn't affect the model at all. Since it's just merging signals / ops, as long as the seed is set, the resulting model should be exactly the same before and after optimization. A few tests of that on relatively complicated models would provide some peace of mind!

That would definitely be nice to have! Any ideas what model to take for that?

Given @tcstewar's and @drasmuss's comment it's probably fine to add the `optimize` argument to the simulator.

jgosmann:
I think I have to do some more debugging. For Spaun the optimizer does not finish in an acceptable time anymore.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,911 @@
...
+        sorted_subset = [subset[i] for i in sort_indices]
+
+        for i, op1 in enumerate(sorted_subset):
+            if op1 in self.merged:
+                # Cannot merge merged operator again until dependency graph
+                # has been updated
+                continue
+
+            if any(op1 in self.dependents[op] or op in self.dependents[op1]
+                   for op in self.merged):
```
jgosmann: I think this loop is the (or one of the) reasons for the performance problem with Spaun. I will try if it improves things to keep around a `set` with the union of all `self.dependents[op]` and switch the second part to `op in self.merged for op in self.dependents[op1]`.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,911 @@
...
+    def is_mergeable(cls, op, tomerge):
+        merger = cls.mergers[tomerge.optype]
+
+        independent_of_ops_tomerge = (
+            op not in tomerge.all_dependents and
+            len(tomerge.dependents[op].intersection(tomerge.ops)) == 0)
+        independent_of_prior_merges = (
+            op not in tomerge.merged and
+            all(op not in tomerge.dependents[o] and
+                o not in tomerge.dependents[op] for o in tomerge.merged))
```
jgosmann: This loop is the other culprit, it's pretty much the same as above, so the same solution should apply.

Fixing these two spots, run times seem to be reasonable again, but tests are still in progress.

tbekolay: Might be worth seeing if making them one big boolean statement again yields speedups (unlikely, but speed is speed).

jgosmann: I think that is only a constant factor, while the main problem at the moment is the algorithmic complexity. But I'll probably try it at some point.

jgosmann:
I've got individual passes running in a reasonable time now and will make a commit with those fixes either tonight or tomorrow. However, the optimizer is taking a large number of passes for Spaun. Originally it was 15, but now I aborted after 150 passes with no end in sight. Each pass merges only a comparitively small number of operators. So this will be the next point of investigation.

jgosmann:
Things to try to reduce duration of individual passes:
* [ ] Instead of switching between merging only views and everything, switch between merging only views and only non-views. (Check whether the “everything” step actually merges any views, because the previous step merged views, this should be rare?)
* [ ] Remove operators unmergeable due to their signals from processing until signals have been changed.
* [ ] Directly update `dg`, `dependents`, `step_order`, etc. The update of these on each pass takes quite some time, which used to be ok because there weren't many passes and the loops within the passes where the main factor. But this has changed and it might be possible to be more intelligent with the updating and save some work (e.g. do not recompute the toposort on each pass?).

jgosmann:
If all of that does not suffice, the final measure will me to stop the optimization once we get diminishing returns in the reduction of operators.

jgosmann:
I added several commits with various speed optimizations. I will run more benchmarks over the next week, but I think the additional commits can already be reviewed.

I also noticed that one of the Thalamus network tests is most likely to fail with bugs in the optimizer implementation. Thus, this might be a good candidate for the test that the results with and without optimizer agree.

jgosmann:
Added that test.

tbekolay:
New changes are looking good! I was a bit worried about stopping the optimization early, as optimizing some ops would presumably have a much bigger effect than optimizing other ops. However, in playing around with Spaun, it does seem like a lot of later optimization passes aren't necessary... in fact, for some reason stopping early results in slightly faster sim times for me than letting it complete all possible optimizations (though this is likely to be specific to my machine -- the test suite isn't any faster with optimizations for me, for example).

Added a few commits with some small reorganizations and changes to the test (which uncovered a minor bug -- easily fixed though).

I had one thought about something that might result in some speedups... what do you think about dealing with each op type separately? I.e., instead of the outer loop going through all possible op types on each optimization pass, we instead have an outer loop for each op type, which should reduce the number of ops that have to be considered on each step. The op types are sorted, so it's likely that in those later optimization steps we're spending a long time re-checking ops that have no chance of being merged because that whole op type has been dealt with. It also means that we could deal with the diminishing returns per op type instead of being more aggressive on op types early in the list, possibly at the cost of not optimizing op types later in the list. Do you think this might help @jgosmann? If so I can try this out tomorrow.

nengo/builder/optimizer.py
```diff
@@ -84,7 +84,7 @@ def optimize(model, dg, max_passes=None):
         before = len(dg)
 
         with Timer() as t:
-            single_pass(only_merge_ops_with_view)
+            sig_replacements.update(single_pass(only_merge_ops_with_view))
```
jgosmann: I though about returning this and I decided against it because this make the function call return a value *and* have side effects. I prefer to only do one or the other because
* We're having side effects already (modifying dg for example), why not be consistent and do everything as a side effect?
* Returning a value might give the impression that there are no side effects and I saw bugs (in other projects) introduced because of that.

jgosmann:
> I had one thought about something that might result in some speedups... what do you think about dealing with each op type separately? I.e., instead of the outer loop going through all possible op types on each optimization pass, we instead have an outer loop for each op type, which should reduce the number of ops that have to be considered on each step. The op types are sorted, so it's likely that in those later optimization steps we're spending a long time re-checking ops that have no chance of being merged because that whole op type has been dealt with.

I think I tried something like that without much success, but we can retry with the recent changes. A few things to keep in mind why this approach might not work well:
* After one pass of one op type the only ops of that type that can still be merged are ops that are in the transitive closure of the merged ops (I believe). That means we need to update the transitive closure to reflect the merges which is one of the slowest parts to be done in each pass outside of the main loop. We might still be able to merge other op types before the transitive closure is updated?
* After all passes for each op type have been done, we need to start over because merges of one op type might allow for merges of another op type.
* With the recent changes we are already throwing out all ops that can never be merged. It is not clear to me how a different loop nesting would improve this.

To make this a bit more explicit:
* Our current loop scheme: for n-passes: compute transitive closure, loop over all types, loop over operators of that type.
* Potential variant 1: for n-passes: compute transitive closure, loop over all types, for m-passes: loop over operators of that type
* Potential variant 2: for n-passes: loop over all types, compute transitive closure, for m-passes: loop over operators of that type
* Potential variant 3: for n-passes: loop over all types, for m-passes: compute transitive closure, loop over operators of that type

Writing it out that way, we see that the essential order of what is looped over stays the same. There are only two parameters:
1. Where/how often to we recompute the transitive closure (less often is better, but we should do it when otherwise further merges are unlikely)
2. Do we try to merge the same op type multiple times before moving on to the next op type?

nengo/builder/optimizer.py
tbekolay: Good points; what do you think of e04b6329?

jgosmann: LGTM

tbekolay:
Thanks for the explanation @jgosmann! I played around with doing optypes separately for a little bit, but it requires more changes than I expected, so it's likely not worth the effort if there's no guaranteed benefit anyhow. Can always be revisited later, like everything, so better to try to get this finalized instead.

I pushed two new commits to try to eke out a bit more speed... the majority of the time spent in `replace_op_signals` (7 out of 9 seconds in the profiling I was doing) is iterating through all the attributes of the ops. The new commits replace the signal attributes with properties so that each signal is only stored in one place (one of the `sets`, `reads` etc lists). This makes the signal update easier and reduces the time spent there to to 1.5 seconds in my profiling. Not a huge improvement, but it's something!

jgosmann:
LGTM

jgosmann:
There is still a bug somewhere:
```
  File "gosmann_frontiers2017/benchmarks/circ_conv.py", line 40, in <module>
    sim.run(1.)
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.4/site-packages/gosmann_frontiers2017/optimized/simulator.py", line 267, in run
    self.run_steps(steps, progress_bar=progress_bar)
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.4/site-packages/gosmann_frontiers2017/optimized/simulator.py", line 287, in run_steps
    self.step()
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.4/site-packages/gosmann_frontiers2017/optimized/simulator.py", line 298, in step
    step_fn()
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.4/site-packages/nengo/builder/neurons.py", line 68, in step_simneurons
    self.neurons.step_math(dt, J, output, *states)
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.4/site-packages/nengo/neurons.py", line 316, in step_math
    spiked_mask = voltage > 1
FloatingPointError: invalid value encountered in greater
```

Not sure if I introduced it by copy & paste of the optimizer code into my benchmarks project ...

nengo/params.py
```diff
@@ -19,6 +19,7 @@ def __repr__(self):
         return self.name
 
 Default = DefaultType("Default")
+RcDefault = DefaultType("RcDefault")
```
jgosmann: Note: this line should be removed.

jgosmann:
The bug might not be in the optimizer itself, but introduced into core Nengo between 2.2.0 and 2.3.0. @xchoo mentioned that Spaun's motor system also stopped working with 2.3.x. Did anyone make progress on debugging that issue?

I'll try to do a git bisect to find the commit causing the issue with the optimizer.

jgosmann:
The “problem” is 7220100500315cd325d62daa32c98e58c139a039 which is not actually a bug in Nengo, but requires the optimizer to be updated to handle the tracked offset. @xchoo, it might be worth checking that commit and the parent commit with Spaun's motor system.

Ironically, one of the optimizer benchmarks made me find the problem fixed in 7220100500315cd325d62daa32c98e58c139a039, but I'm using a workaround now. I'm probably gonna run the benchmarks with Nengo 2.2.0 and then rebase this branch and make the neccessary changes.

jgosmann:
I squashed all the `!squash` and `!fixme` commits as a first step of rebasing this. The next step will be to get each commit to pass the tests. Then I will rebase onto master (probably using a different branch to keep the current state around for the running benchmarks).

tbekolay:
> The next step will be to get each commit to pass the tests.

I'm not sure how many of these commits we'll actually want to keep in the final merge... the history of the optimizer from first draft to finished product isn't super pertinent to `master`, and it could cause confusion since the early commits change other files like `builder/signal.py` and then un-change them later. Commits that touch files other than `optimizer.py` like d99ef13 make sense to keep separate, but the rest I think I would sqaush.

jgosmann:
Ok, I finally managed to rebase this properly and every commit should pass the unit tests. To facilitate the review of the changes that made necessary I did not squash 1100d6e4ad3b880574ff7e756121fa1fedf4fe4c yet. I also added 8912887d138dea2154e1fcb535da576696918564. The track offset related changes are, however, already squashed into the main optimizer commit. The relevant changed lines are:

* https://github.com/nengo/nengo/pull/1035/commits/ea776d264c59c9ddb73bc3a8d9cf668189773e58#diff-f23ca6a62a61764f4d9a7c944035d7beR340
* https://github.com/nengo/nengo/pull/1035/commits/ea776d264c59c9ddb73bc3a8d9cf668189773e58#diff-f23ca6a62a61764f4d9a7c944035d7beR653
* https://github.com/nengo/nengo/pull/1035/commits/ea776d264c59c9ddb73bc3a8d9cf668189773e58#diff-f23ca6a62a61764f4d9a7c944035d7beR896

I added an assert that seems useful and helped me in debugging, but it might have a performance impact: https://github.com/nengo/nengo/pull/1035/commits/ea776d264c59c9ddb73bc3a8d9cf668189773e58#diff-752cf85ff95528b7d03a4db5beb03ddbR272


Most of the benchmarks have completed now. The bugfixes have let to an increase in build times. Especially for high-dimensional convolution in direct mode. But run times seem to have slightly decreased. I'll update plots and number in the paper either later today or tomorrow and can send you an updated version.

jgosmann:
Added another commit to improve performance.

jgosmann:
I added another commit that gives an even better performance improvement. Note that with this commit (fd6c711) the previous commit bd5bb5a only make a minor difference. So we could drop that commit depending on how we want to trade off the minor speed improvement with the minor added complexity in code.

jgosmann:
I think it's now as fast as I will get it for now. The `transitive_closure`, `is_mergeable`, and `merge` functions remain hot spots, but there is no obvious way to make them faster (maybe connecting more of the statements in `is_mergeable` with `and`, but at least for the circular convolution that doesn't make a difference).

jgosmann:
*sigh* for some reason the optimizer is consuming way more memory now than before. Not completely sure why, because when I tested these changes before it was fine.

jgosmann:
This doesn't seem to be caused by the last commits. I updated from Python 3.5 to 3.6 today, potentially 3.6 needs more memory for something? Let's see if I get the problem with 3.5 now too.

jgosmann:
It seems that Python 3.6 is indeed to blame. I wonder whether we're seeing here the [“memory usage of the new dict() is between 20% and 25% smaller”](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation) in action? ;)

But in all seriousness, I'm not sure how do go about this. Some of my thoughts:

* It seems that it is not an algorithmic problem or general problem with the data structures as it is fine Python <3.6.
* Though, the memory usage seems to only explode in the second optimizer pass. So maybe there is something in particular that the optimizer does which isn't working very well.
* It might effect only very large models like Spaun, so maybe it's not too problematic for now? (But I should check with the smaller models.)
* We could show a warning or deactivate the optimizer for Python >=3.6 until we find a solution.
* I wonder whether it is worth to post on some Python mailing list or file an issue? But not sure if I have enough information on what's happening to get any helpful response.
* We could try defining `__slots__` for operators and signals to reduce memory consumption.

jgosmann:
When I use `sys.getsizeof` and `gc.get_objects` to determine how much memory is used, I get about 16 to 17GiB. That makes me believe that this is not a pure core-Python issue (`getsizeof` is supposed to be accurate for all built-in objects), but related to NumPy arrays (which I believe are not handled by the standard garbage collector and as such do not appear in `gc.get_objects`). In fact, it might be this issue: numpy/numpy#8503. I will try to verify this, but I will need to build debug versions of Python to do so first.

jgosmann:
I can reproduce numpy/numpy#8503 with both Python 3.6 and 3.5 which rules this out as the cause. It also applies only to object arrays (our arrays should all be float or int).

jgosmann:
The [NumPy allocation tracker](https://github.com/numpy/numpy/tree/master/tools/allocation_tracking) reports pretty much the same numbers for Python 3.5 and 3.6. So I have no idea what is consuming all that memory.

For the circular convolution and n-back benchmarks Python 3.6 consumes less memory.

jgosmann:
I did some further investigation of the Python 3.6 memory problem: It seems to only occur with the Spaun model and when the Python requires more memory than provided by physical RAM (i.e. data has to be written to swap). If decrease the semantic pointer dimensionality in Spaun by 16 dimensions, the problem disappears.

Python 3.5 seems to be slightly more memory efficient for the Spaun model, so that the swap isn't used for the default 512 dimensions. So I increased the dimensionality by 128 dimensions to force things into swap, but even in that case Python 3.5 doesn't consume as much memory as Python 3.6. 

I think this is somewhat good news as it indicates that this will be only a problem for really large models. But it is also somewhat bad news because I still have no idea what is causing this, especially as I thought that swap/virtual memory allocation is transparent to the process. Or maybe the it is just coincidence that the memory requirements grow so much more at a point where my the limit of my physical RAM gets hit ...

jgosmann:
The issue with Python 3.6 has been identified and there is [an issue in the Python bug tracker](https://bugs.python.org/issue29949). It seems that the change introducing this issue might get reverted. There might also be workarounds on our side possible by construction sets/frozensets in a different way, but I would rather not increase the complexity of our code (and their might be also a cost in runtime).

Seanny123:
So is this ready for a final review? *rolls up sleeves and puts on
code-reading goggles*

tbekolay:
Sorry for the slow reply, this is indeed ready for a final review! @Seanny123 do you think you can get to this in the next few days?

Seanny123:
Yep, I'll do a review in the next few days!

Seanny123:
Btw, the fix to the Python 3.6 error got merged and should be in the next release of Python.

nengo/builder/tests/test_optimizer.py
```diff
@@ -0,0 +1,142 @@
...
+
+def test_sigmerger_check_views():
+    s1 = Signal(np.empty((5, 5)))
+    s2 = Signal(np.empty((5, 5)))
+
+    # compatible along first axis
+    SigMerger.check_views([s1[:1], s1[1:]])
+
+    # compatible along second axis
+    SigMerger.check_views([s1[0:1, :1], s1[0:1, 1:]], axis=1)
```
Seanny123: Nitpick: why write `0:1` instead of `:1`. If there's no reason, then I'll make a fixup commit to address it.

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,934 @@
...
+
+        # These variables will be initialized and used on each pass
+        self.dependents = None
+        self.only_merge_ops_with_view = None
+
+        self.merged = set()
+        self.merged_dependents = set()
+        self.opinfo = OpInfo()
+
+    def __call__(self, only_merge_ops_with_view):
```
Seanny123: Silly inexperience coder question: why make this a `__call__` method instead of a regular method, named something like `run_pass`?

nengo/builder/optimizer.py
```diff
@@ -0,0 +1,934 @@
...
+            for tp, ops in iteritems(groupby(dg, type)):
+                logger.debug("%s: %d", tp, len(ops))
+
+        only_merge_ops_with_view = before is None or before != after
+        before = len(dg)
+
+        with Timer() as t:
+            single_pass(only_merge_ops_with_view)
+
+        after = len(dg)
```
Seanny123: Just to make sure I'm understanding this correctly, `dg` is passed into `OpMergePass` by reference, like all things in Python. Once `OpMergePass` is called, the `dg` is modified.

This was a bit hard for me to follow, because if something is modified, I kind of expect:
a) it to be returned, or
b) to be passed in as an argument to a function call

Is this a weird expectation to have? Hypothetically, if the `call` would return and over-write the `dg` instance, would it affect performance?

Seanny123: To be clear, the reason I have this expectation is that I have a mental model of Classes as containers for data that are only interacted with by accessing via attributes or calling methods. I wouldn't expect the data inside of the Class to connected to the outside by a reference, nor would I expect calling the Class to update an outside method.

jgosmann: > Is this a weird expectation to have?

I don't think those are pretty reasonable expectations, but unfortunately things are a bit more complicated here.

> if the call would return and over-write the dg instance, would it affect performance?

Yes, it would increase memory usage; potentially by a lot for large models. The dependency graph can get quite large itself already, but then it also references the initial set of operators and those reference the initial set of signals that require even more memory with their initial values. By keeping the old dependency graph around all those objects would be prevented from being freed.

Maybe it would be clearer to use `single_pass.dg.forward` instead of `dg` to make it clear that it is an attribute of the `single_pass` instance that might get modified by the object?

> I have a mental model of Classes as containers for data

I think that is an appropriate mental model in many cases, but not all. Especially here `OpMergePass` is more like a [function object (sometimes called functor)](https://en.wikipedia.org/wiki/Function_object). It is used pretty much like a function, but by using an object it allows us to split up the algorithm into smaller functions, but still group them. It also allows to easily share data between those functions and preserve some state across invocations.

nengo/builder/optimizer.py
jgosmann: @tbekolay might say more on why he decided for this approach, what comes to my mind is:

* It might highlight the main entry point and sets it apart from the other methods that are not supposed to be called from outside of the class.
* The class is more used like a function (on steroids  because it is actually an object and has some state) so it might make sense to make it a function object?

tbekolay: 👉 👉 nailed it, those were my reasons. Really, the whole class could be seen as one big long function, but doing it that way would be really hard to follow and be nested 5 or 6 indentation levels deep, hence why there are methods to organize things. But, the "user" (`optimize` in this case) should never call those functions, they could treat the class as though it is just one big long function, even though it does some _stuff_ inside the class.

nengo/builder/tests/test_optimizer.py
tbekolay: Can't think of a reason off the top of my head!

nengo/builder/optimizer.py
tbekolay: I initially had a similar thought that it would be nice to return an optimized version of the `dg` and `model` rather than mutating them, but as Jan says, it affects performance significantly. There's a difficult tradeoff to make in this PR, as optimized code tends to be less readable than slow but obvious code. I think this PR manages to strike a good balance between readable and efficient, though it's of course good to get more opinions, and is why we want multiple reviewers for each PR 😄

Seanny123:
I added my tiny fixup commits. If they look good to @jgosmann then this is ready to merge!

jgosmann:
LGTM. I think @tbekolay  wanted to drop bd5bb5af6e0f71a2b80a73a0b3b29497f8852f68 (and https://github.com/nengo/nengo/pull/1035/commits/e525dba953c278fbefb0a80fcbeb9d615a2263f9 accordingly)?

jgosmann:
I'm not sure what caused the authorship information to get lost ... I was expecting rebases to keep it? So I didn't even consider checking that.

tbekolay:
If I had to guess, it would be in situations where you split a commit into multiple commits... in the past when I've done that, only one of the commits retains the authorship information. At least in the way that I was splitting commits at the time ¯\\\_(ツ)\_/¯ In any case no bigs!

jgosmann:
I'll try to keep it in mind for the next complicated rebase.

Seanny123:
Yep. Everything still looks good for me. Merge away!

PR #1037, opened 2016-04-26
2.1.0 release

tbekolay:
Minor fixes for a few things that I noticed when running the full test suite (including the slow tests), looking at the generated plots, and looking at the built documentation, as per our release instructions.

This is based on the `promote-process` branch from #955, which I believe is ready to merge. Once this PR is reviewed and merged, I'll do the 2.1.0 release!


nengo/version.py
```diff
@@ -1,14 +1,14 @@
...
 """
 
 name = "nengo"
 version_info = (2, 1, 0)  # (major, minor, patch)
-dev = True
+dev = 0
 
 version = "{v}{dev}".format(v='.'.join(str(v) for v in version_info),
-                            dev='-dev' if dev else '')
+                            dev='.dev%d' % dev if dev is not None else '')
```
hunse: I'd put parentheses around the inline if, just because I have no idea what the order of operations is for random Python stuff and it's nice to be explicit IMO.

examples/dynamics/controlled_integrator.ipynb
```diff
@@ -232,7 +233,8 @@
       "plt.plot(t, ref, 'k--', label='Exact')\n",
       "plt.plot(t, sim.data[A_probe][:,0], label='A (value)')\n",
       "plt.plot(t, sim.data[A_probe][:,1], label='A (control)')\n",
-      "plt.ylim([-1.1, 1.1])\n",
+      "plt.xlim(right=t[-1])\n",
```
hunse: According to the [docs](http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.xlim), this argument should be called `xmax` not `right`.

hunse:
Just one more minor comment, otherwise LGTM.

examples/dynamics/controlled_integrator.ipynb
tbekolay: Hm, according to [this line in the master branch](https://github.com/matplotlib/matplotlib/blob/6b22bfc4d831c3f615cae87838127ec0e7aebd2d/lib/matplotlib/axes/_base.py#L2776) it's the other way around... since it's okay for `right < left`.

tbekolay:
Made some fixup commits which should address the comments! Hopefully the version order of ops is clear (other parentheses locations give a syntax error). Also fixed the CI tests and the flake8 error. @drasmuss can you give a second review and merge if all looks good?

examples/dynamics/controlled_integrator.ipynb
hunse: Looking in more detail, it looks like that changed over two years ago, at least. They just never changed the `xlim` docs. So I think `right` is actually the right choice...

tbekolay: Looks like a low hanging fruit MPL contribution if anyone wants to do that ;)

PR #1045, opened 2016-04-28
Added ShapeParam to store shapes

hunse:
A variation of TupleParam that is specialized to storing shapes. Allows checking for the length of the shape, that all elements are ints (or castable to ints), and ensuring all values are greater than a given value (often 0 or 1).

### DIFF
--- a/nengo/params.py
+++ b/nengo/params.py
@@ -292,6 +292,30 @@ class TupleParam(Parameter):
         super(TupleParam, self).validate(instance, value)
 
 
+class ShapeParam(TupleParam):
+    """A parameter where the value is a tuple of integers."""
+
+    equatable = True
+
+    def __init__(self, name, default=Unconfigurable, length=None, low=0,
+                 optional=False, readonly=None):
+        super(ShapeParam, self).__init__(name, default=default, length=length,
+                                         optional=optional, readonly=readonly)
+        self.low = low
+
+    def validate(self, instance, value):
+        super(ShapeParam, self).validate(instance, value)
+        for i, v in enumerate(value):
+            if not is_integer(v):
+                raise ValidationError("Element %d must be an int (got type %r)"
+                                      % (i, type(v).__name__),
+                                      attr=self.name, obj=instance)
+            if self.low is not None and v < self.low:
+                raise ValidationError(
+                    "Element %d must be >= %d (got %d)" % (i, self.low, v),
+                    attr=self.name, obj=instance)
+
+
 class DictParam(Parameter):
     """A parameter where the value is a dictionary."""
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,8 @@ Release History
   discoverable for other Python packages. In the future all backends should
   declare an entry point accordingly.
   (`#1127 <https://github.com/nengo/nengo/pull/1127>`_)
+- Added ``ShapeParam`` to store array shapes.
+  (`#1045 <https://github.com/nengo/nengo/pull/1045>`_)
 
 **Bug fixes**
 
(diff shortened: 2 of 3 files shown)


Seanny123:
LGTM

hunse:
To see this in action, see https://github.com/nengo/nengo_extras/blob/master/nengo_extras/convnet.py. It's been working well there.

nengo/params.py
```diff
@@ -292,6 +292,34 @@ def validate(self, instance, value):
...
+
+    def __init__(self, name, default=Unconfigurable, length=None, low=0,
+                 optional=False, readonly=None):
+        super(ShapeParam, self).__init__(name, default=default, length=length,
+                                         optional=optional, readonly=readonly)
+        self.low = low
+
+    def __set__(self, instance, value):
+        try:
+            value = tuple(int(v) for v in value)
```
jgosmann: My NumPy says about non-integer number in shapes "DeprecationWarning: using a non-integer number instead of an integer will result in an error in the future". I would say, in anticipation of this, we should not allow non-integers at all.

jgosmann:
Add commits regarding my comment and to add a changelog entry.

nengo/params.py
```diff
@@ -304,9 +304,7 @@ def __init__(self, name, default=Unconfigurable, length=None, low=0,
         self.low = low
 
     def __set__(self, instance, value):
-        try:
-            value = tuple(int(v) for v in value)
-        except TypeError:
+        if not all(is_integer(v) for v in value):
```
hunse: So the reason I didn't do this, is that I often do something like `np.floor(a / b)` to get some of my shape indices, and it's just convenient to not have to explicitly cast to an int. That said, maybe it's better to have to be explicit. But would it be worth checking if a floating point number is actually exactly an integer, and in that case just doing the cast ourselves?

jgosmann: You wouldn't have that convenience in the future anyways: https://github.com/nengo/nengo/pull/1045#discussion_r71457562

I would be fine with doing the cast ourselves, but I am not sure how you can reliably detect whether a floating point number is an integer (considering rounding errors).

hunse: > You wouldn't have that convenience in the future anyways

I wouldn't necessarily have that convenience in Numpy (though it might be quite a while before they actually remove that), but I would still have it here.

So I think `floor` and `ceil` remove any rounding errors for reasonably sized numbers, but to really make a robust system we'd need to check tolerances or something, and that's all getting too complex. So I'm fine having to be explicit. That way, there's no magic going on, making it easier for users to debug.

jgosmann: > I wouldn't necessarily have that convenience in Numpy

But wouldn't the shape parameter usually be used to create a Numpy array?

hunse: Perhaps, but not necessarily. Anyway, I'm fine with the explicit approach.

tbekolay:
Was merging this and noticed that with @jgosmann's changes, there's no longer a need for a `__set__` and it should all be done in `validate`. Made 4b22c87d346a0e1937d29b5c48611 to change that; will fix that up and merge tomorrow morning if there are no objections.

hunse:
So one thing that got lost in @jgosmann's changes is that the input is no longer cast to a tuple. I had set this up so e.g. an iterable could be passed in and it would still work fine. So I think we need to override the `__set__` for that reason. I'll fix that up.

tbekolay:
> input is no longer cast to a tuple

I think we do this in `TupleParam` so it should do this anyhow? https://github.com/nengo/nengo/blob/master/nengo/params.py#L279

hunse:
Oh, right you are Ken. Ok, should be good to go!

tbekolay:
Cool, I'll go ahead now then! :rainbow:

PR #1046, opened 2016-04-28
Make Solvers be FrozenObjects

### LINKED ISSUE #1044: Cache does not consider solver regularization
```
import nengo

nengo.cache.get_default_decoder_cache().invalidate()

with nengo.Network() as model_foo:
    x = nengo.networks.EnsembleArray(100, 1, seed=0)
    x.add_output('foo', function=None, solver=nengo.solvers.LstsqL2(reg=0.001))
    p_foo = nengo.Probe(x.foo, synapse=0.01)
    
with nengo.Simulator(model_foo) as sim_foo:
    sim_foo.run(1.0)

# nengo.cache.get_default_decoder_cache().invalidate()
    
with nengo.Network() as model_bar:
    x = nengo.networks.EnsembleArray(100, 1, seed=0)
    x.add_output('bar', function=None, solver=nengo.solvers.LstsqL2(reg=1000))
    p_bar = nengo.Probe(x.bar, synapse=0.01)
    
with nengo.Simulator(model_bar) as sim_bar:
    sim_bar.run(1.0)

print nengo.utils.numpy.rms(sim_foo.data[p_foo])
print nengo.utils.numpy.rms(sim_bar.data[p_bar])
```

Running this gives the output:
```
0.0329802255974
0.0329802255974
```

But if I uncomment the line in between to invalidate the cache:
```
0.0329802255974
3.57156978133e-07
```

The latter result is the correct one (since `bar` is supposed to be heavily regularized). This means that the cache is not key'ing on the solver's regularization.

hunse:
So they can take advantage of the built-in pickling and hashing.

This fixes #1044.

### DIFF
--- a/nengo/solvers.py
+++ b/nengo/solvers.py
@@ -6,7 +6,6 @@ depending on whether the post-population encoders `E` are provided (see below).
 Solvers that are only intended to solve for either decoders or weights can
 remove the `E` parameter or make it manditory as they see fit.
 """
-import collections
 import logging
 import time
 
@@ -14,8 +13,8 @@ import numpy as np
 
 import nengo.utils.least_squares_solvers as lstsq
 from nengo.exceptions import ValidationError
-from nengo.params import BoolParam, NumberParam, Parameter
-from nengo.utils.compat import range, with_metaclass, iteritems
+from nengo.params import BoolParam, FrozenObject, NumberParam, Parameter
+from nengo.utils.compat import range, with_metaclass
 from nengo.utils.least_squares_solvers import (
     format_system, rmses, LeastSquaresSolverParam)
 from nengo.utils.magic import DocstringInheritor
@@ -23,7 +22,7 @@ from nengo.utils.magic import DocstringInheritor
 logger = logging.getLogger(__name__)
 
 
-class Solver(with_metaclass(DocstringInheritor)):
+class Solver(with_metaclass(DocstringInheritor, FrozenObject)):
     """Decoder or weight solver."""
 
     def __call__(self, A, Y, rng=None, E=None):
@@ -55,33 +54,6 @@ class Solver(with_metaclass(DocstringInheritor)):
             Other keys are unique to particular solvers.
         """
         raise NotImplementedError("Solvers must implement '__call__'")
-
-    def __hash__(self):
-        items = list(self.__dict__.items())
-        items.sort(key=lambda item: item[0])
-
-        hashes = []
-        for k, v in items:
-            if isinstance(v, np.ndarray):
-                if v.size < 1e5:
-                    a = v[:]
-                    a.setflags(write=False)
-                    hashes.append(hash(a))
-                else:
-                    raise ValidationError("array is too large to hash", attr=k)
-            elif isinstance(v, collections.Iterable):
-                hashes.append(hash(tuple(v)))
-            elif isinstance(v, collections.Callable):
-                hashes.append(hash(v.__code__))
-            else:
-                hashes.append(hash(v))
-
-        return hash(tuple(hashes))
-
-    def __str__(self):
-        return "%s(%s)" % (
-            self.__class__.__name__,
-            ', '.join("%s=%s" % (k, v) for k, v in iteritems(self.__dict__)))
 
     def mul_encoders(self, Y, E, copy=False):
         """Helper function that projects signal ``Y`` onto encoders ``E``.
@@ -133,9 +105,13 @@ class Lstsq(Solver):
         If False, solve for decoders. If True, solve for weights.
     """
 
+    weights = BoolParam('weights')
+    rcond = NumberParam('noise', low=0)
+
     def __init__(self, weights=False, rcond=0.01):
-        self.rcond = rcond
+        super(Lstsq, self).__init__()
         self.weights = weights
+        self.rcond = rcond
 
     def __call__(self, A, Y, rng=None, E=None):
         tstart = time.time()
@@ -176,6 +152,7 @@ class _LstsqNoiseSolver(Solver):
         weights : bool
             If False, solve for decoders. If True, solve for weights.
         """
+        super(_LstsqNoiseSolver, self).__init__()
         self.weights = weights
         self.noise = noise
         self.solver = solver
@@ -233,6 +210,7 @@ class _LstsqL2Solver(Solver):
         weights : bool
             If False, solve for decoders. If True, solve for weights.
         """
+        super(_LstsqL2Solver, self).__init__()
         self.weights = weights
         self.reg = reg
         self.solver = solver
@@ -302,6 +280,7 @@ class LstsqL1(Solver):
         """
         import sklearn.linear_model  # noqa F401, import to check existence
         assert sklearn.linear_model
+        super(LstsqL1, self).__init__()
         self.weights = weights
         self.l1 = l1
         self.l2 = l2
@@ -369,6 +348,7 @@ class LstsqDrop(Solver):
         weights : bool
             If False, solve for decoders. If True, solve for weights.
         """
+        super(LstsqDrop, self).__init__()
         self.weights = weights
         self.drop = drop
         self.solver1 = solver1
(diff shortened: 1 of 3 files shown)


Seanny123:
FYI, this passed all tests except for PEP8.

nengo/params.py
```diff
@@ -416,7 +417,7 @@ def validate(self, instance, function_info):
         super(FunctionParam, self).validate(instance, function)
 
 
-class FrozenObject(object):
+class FrozenObject(with_metaclass(DocstringInheritor)):
```
tbekolay: Hm... are we sure we actually want this on the `FrozenObject` base class? I think it could be confusing if a user subclasses dist, for example, and doesn't write a docstring... calling `help` will give something wrong instead of blank.

The alternative is to do multiple inheritance in the solver; i.e., `class Solver(FrozenObject, with_metaclass(DocstringInheritor)):`.

nengo/utils/least_squares_solvers.py
```diff
@@ -121,7 +131,11 @@ def __call__(self, A, Y, sigma, rng=None):
 class Conjgrad(LeastSquaresSolver):
     """Solve a least-squares system using conjugate gradient."""
 
+    tol = NumberParam('tol', low=0)
+    maxiters = IntParam('maxiters', low=1, optional=True)
+
```
tbekolay: Why is `X0` not included here?

nengo/utils/least_squares_solvers.py
```diff
@@ -181,7 +195,10 @@ def _conjgrad_iters(calcAx, b, x, maxiters=None, rtol=1e-6):
 class BlockConjgrad(LeastSquaresSolver):
     """Solve a multiple-RHS least-squares system using block conj. gradient."""
 
+    tol = NumberParam('tol', low=0)
+
```
tbekolay: Ditto here with `X0`.

nengo/utils/least_squares_solvers.py
```diff
@@ -253,7 +270,11 @@ class RandomizedSVD(LeastSquaresSolver):
     ``sklearn.utils.extmath.randomized_svd`` for details about the parameters.
     """
 
+    n_components = IntParam('n_components', low=1)
+    n_oversamples = IntParam('n_oversamples', low=0)
+
```
tbekolay: Should add a `DictParam` for the kwargs too? In which case, it should maybe be passed in as an explicit dict, rather than as `**kwargs`?

tbekolay:
Made some inline comments; other than those this is a definite improvement!

nengo/params.py
hunse: Multiple inheritance was my original plan, but I was having some trouble getting it working (can't remember what). If you can do it, I'm all for it.

tbekolay: Will give it a shot!

nengo/utils/least_squares_solvers.py
hunse: I had a reason... maybe because it could be big and I didn't want to worry about hashing/pickling it? Also, it's kind of a strange parameter in general because it has to be the right size for the system being solved, meaning that when it's set the solver will only work for one size of connection. But maybe it should be included...

nengo/utils/least_squares_solvers.py
hunse: Does `DictParam` have a proper hashing function right now? Maybe we should just add all the possible parameters manually. There aren't too many of them.

nengo/params.py
jgosmann: > The alternative is to do multiple inheritance in the solver; i.e., class Solver(FrozenObject, with_metaclass(DocstringInheritor)):.

Shouldn't it be `Solver(with_metaclass(DocstringInheritor, FrozenObject)):`? At least I believe that is what I used in some of my PRs.

hunse: From what I vaguely know about metaclasses, I think that would make sense. Maybe that's why it wasn't working for me before. I definitely had it like @tbekolay did above.

tbekolay: All 3 possibilities (`with_metaclass(DocstringInheritor), FrozenObject`; `FrozenObject, with_metaclass(DocstringInheritor)`; `with_metaclass(DocstringInheritor, FrozenObject)`) passed unit tests for me, but @jgosmann's is undoubtedly the better solution (results in a shorter `mro`, I think), so I pushed a fixup commit with that.

hunse: Did you check that the docstrings are being properly inherited, though? There's no test for that, and I think that's what wasn't working for me.

hunse: I just checked, and it seems to be working this way, so it's all good!

hunse: It also seems to work the other ways, so I'm not sure what I was doing wrong before...

tbekolay: Sometimes we just have to have faith in the gods of programming. Surely they have a glorious plan for us!

nengo/solvers.py
```diff
@@ -67,16 +72,14 @@ def mul_encoders(self, Y, E, copy=False):
...
-                    "Encoders must be 'None' for decoder solver", attr='E')
-            return Y.copy() if copy else Y
+        if self.weights and E is None:
+            raise ValidationError(
+                "Encoders must be provided for weight solver", attr='E')
+        if not self.weights and E is not None:
+            raise ValidationError(
+                "Encoders must be 'None' for decoder solver", attr='E')
+
+        return np.dot(Y, E) if E is not None else Y.copy() if copy else Y
```
tbekolay: The nested conditionals here makes it a bit hard to parse... could you put `Y.copy() if copy else Y` in parentheses?

hunse: That's interesting, I don't really consider this nested, and feel like it reads exactly like the logic goes. That's one thing I like about Python: the statement `x if a else y if b else z if c else q` does exactly what you think it would do if someone said it to you (at least to me). If those parentheses would be helpful, though, I'm fine to add them.

tbekolay: Nah, not a critical thing, I see your point! I think for me, I parse these lines with an eye to the value, rather than the condition... so when I see `x if a else y if b else z if c else q` I'm thinking about what is the situation in which you get `b`, rather than reading it left to right. So I'm looking at `q` for example, then I scan left and see that `q` happens if `a` `b` and `c` are False, which takes a few seconds to think through. For some reason the parens help with the mental organization of that, but like I said, not a critical thing so no need to change.

jgosmann:
This makes the tests fail for some reason?

tbekolay:
Hmm... concerning that that didn't fail [pylint](https://travis-ci.org/nengo/nengo/jobs/130609753)

hunse:
Ok, hopefully those two commits fix everything.

jgosmann:
> Hmm... concerning that that didn't fail pylint

I assume pylint isn't able to tell the the super call is needed, because of the `with_metaclass` inheritance.

hunse:
I think this is ready to go.

tbekolay:
OK, I'll look at this now

nengo/utils/least_squares_solvers.py
```diff
@@ -142,14 +143,15 @@ def __init__(self, tol=1e-2, maxiters=None, X0=None):
 
     def __call__(self, A, Y, sigma, rng=None):
         Y, m, n, d, matrix_in = format_system(A, Y)
+        X = np.zeros((n, d)) if self.X0 is None else np.array(self.X0)
+        if X.shape != (n, d):
+            raise ValueError('X0 must be shape %s, got %s' % ((n, d), X.shape))
```
tbekolay: Should be a ValidationError rather than ValueError

nengo/utils/least_squares_solvers.py
```diff
@@ -207,14 +210,16 @@ def __call__(self, A, Y, sigma, rng=None):
         sigma = np.asarray(sigma, dtype='float')
         sigma = sigma.reshape(sigma.size, 1)
 
+        X = np.zeros((n, d)) if self.X0 is None else np.array(self.X0)
+        if X.shape != (n, d):
+            raise ValueError('X0 must be shape %s, got %s' % ((n, d), X.shape))
```
tbekolay: ValidationError here too

tbekolay:
I'll do the `ValueError` -> `ValidationError` change in the merge!

PR #1047, opened 2016-04-28
Test solver fingerprinting.

jgosmann:
Addresses #1044.

Note that this test will fail without PR #1046.

### DIFF
--- a/nengo/tests/test_cache.py
+++ b/nengo/tests/test_cache.py
@@ -10,6 +10,7 @@ import pytest
 import nengo
 from nengo.cache import DecoderCache, Fingerprint, get_fragment_size
 from nengo.exceptions import FingerprintError
+from nengo.solvers import LstsqL2
 from nengo.utils.compat import int_types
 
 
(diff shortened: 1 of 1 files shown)


hunse:
LGTM

hunse:
The test is failing now, though. I'm going to rebase and see if that helps.

hunse:
Which it should, because #1046 is now merged. duh

tbekolay:
;) This LGTM too btw so if CI testing passes feel free to merge.

hunse:
One question though: do we want to test this for other FrozenObjects too? I mean really, we'd need a more automated way of testing every parameter on every FrozenObject if we want to be robust. Might be something to consider down the road.

tbekolay:
> do we want to test this for other FrozenObjects too?

Hmm... this could be doable with the registry in #1068, which would be an argument for that approach over #1066.

jgosmann:
> do we want to test this for other FrozenObjects too?

I don't think, this issue is specific to frozen objects, but could occur with any object having parameters (or even some sort of descriptor). For parameters this is solved by either making the class a FrozenObject or copyable. For the general case the class has to correclty implement pickling. So we can't test this for every potential case anyways.

Issue #1048, opened 2016-04-29
Cache index lock sometimes not released


jgosmann:
Instead of relying on `__del__`, context managers should be used. This should also resolve #1041.

jgosmann:
Should be fixed with #1053.

Issue #1041, opened 2016-04-28
Cache throws 'open' not defined error in Python 3


xchoo:
After running a model in python3, the cache system throws the following error when closing down the simulation (after the model has successfully ran):

```
Simulation finished in 0:01:32.                                                 
Exception ignored in: <bound method CacheIndex.__del__ of <nengo.cache.CacheIndex object at 0x7f57521f8588>>
Traceback (most recent call last):
  File "/home/abruser/git/nengo/nengo/cache.py", line 110, in __del__
  File "/home/abruser/git/nengo/nengo/cache.py", line 126, in sync
  File "/home/abruser/git/nengo/nengo/cache.py", line 115, in _load_index
NameError: name 'open' is not defined
```

tbekolay:
Found [this relevant StackOverflow question](http://stackoverflow.com/questions/23422188/why-am-i-getting-nameerror-global-name-open-is-not-defined-in-del)...

jgosmann:
@xchoo Can you test if PR #1053 solves the problem?

jgosmann:
@xchoo Did PR #1053 (now merged into master) solve the problem and can this issue be closed?

jgosmann:
@xchoo It's been 8 days. Can I close the issue?

xchoo:
Uh yup. I'll open it again if another issue arises with Python3 (I don't use it very often). :smile:

PR #1053, opened 2016-05-02
Use cache as context manager.

jgosmann:
Do not rely on `__del__` which can cause problems.

Addresses #1041, #1048.

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -85,8 +85,7 @@ class CacheIndex(object):
     def __init__(self, filename):
         self.filename = filename
         self._lock = FileLock(self.filename + '.lock')
-        with self._lock:
-            self._index = self._load_index()
+        self._index = None
         self._updates = {}
         self._deletes = set()
         self._removed_files = set()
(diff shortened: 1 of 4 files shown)


tbekolay:
Pushed a commit to fix two PEP8 issues; this also needs a bugfix changelog entry. Aside from that looks good to me!

jgosmann:
Added the changelog entry.

tbekolay:
Cool; will merge on a second reviewer's +1.

jgosmann:
Any second reviewer available?

nengo/cache.py
```diff
@@ -177,14 +181,7 @@ def __init__(self, readonly=False, cache_dir=None):
...
         self._fragment_size = get_fragment_size(self.cache_dir)
-        try:
-            self._remove_legacy_files()
-            self._index = CacheIndex(os.path.join(self.cache_dir, self._INDEX))
-        except TimeoutError:
-            warnings.warn(
-                "Decoder cache could not acquire lock and was deactivated.")
-            self._index = {}
-            self.readonly = True
+        self._index = {}
```
hunse: Why does this default to an empty dict, rather than say None? If it were None, then you could just check for that on `__exit__`, rather than having to use `hasattr`.

hunse:
Other than that one little question, it's all fine by me.

nengo/cache.py
jgosmann: You're right. I fixed this. I guess, I was worried about having always to check whether it is `None` in that case, but all the code where it would be necessary doesn't get executed in that case for other reasons.

Seanny123:
@hunse just a friendly reminder that this issue appears to be waiting on your OK given Jan's last modification of switching from an empty `dict` to `None` (:

hunse:
Oh yeah, that's fine.

Issue #1043, opened 2016-04-28
One-liner for defining Configs


tcstewar:
Right now if you want to use a ```nengo.Config``` object to set parameters in something, you do it this way:

```python
cfg = nengo.Config(nengo.Connection)
cfg[nengo.Connection].solver=nengo.solvers.NnlsL2()
with cfg:
    a = nengo.networks.EnsembleArray(n_neurons=100, n_ensembles=3)
```

It would be nice to collapse that into a one-liner.  @hunse suggested this:

```python
with nengo.Config(nengo.Connection).set('solver', nengo.solvers.NnlsL2()):
    a = nengo.networks.EnsembleArray(n_neurons=100, n_ensembles=3)
```

Another option is this:

```python
with nengo.Config(nengo.Connection).set(solver=nengo.solvers.NnLsL2()):
    a = nengo.networks.EnsembleArray(n_neurons=100, n_ensembles=3)
```

or this:

```python
with nengo.Config(nengo.Connection, solver=nengo.solvers.NnLsL2()):
    a = nengo.networks.EnsembleArray(n_neurons=100, n_ensembles=3)
```

I like the third option (for simplicity, and giving a natural way to set multiple parameters), but the second option looks good too.  

However, all of them do have the problem of being a bit ambiguous if someone does

```python
with nengo.Config(nengo.Connection, nengo.Ensemble).set('solver', nengo.solvers.NnLsL2()):
    a = nengo.networks.EnsembleArray(n_neurons=100, n_ensembles=3)
with nengo.Config(nengo.Connection, nengo.Ensemble).set(solver=nengo.solvers.NnLsL2()):
    a = nengo.networks.EnsembleArray(n_neurons=100, n_ensembles=3)
with nengo.Config(nengo.Connection, nengo.Ensemble, solver=nengo.solvers.NnLsL2()):
    a = nengo.networks.EnsembleArray(n_neurons=100, n_ensembles=3)
```

jgosmann:
```python
with nengo.Config({nengo.Connection: {'solver': nengo.solvers.NnLsL2()}, nengo.Ensemble: {}}):

Seanny123:
I don't think the Connection and Ensemble case is ambiguous. It seems obvious to me that it would be set on both, if the attribute exists. But maybe that's just my mental model? I can see the down-side of someone seeing that line of code and having to check if either Nengo object has that attribute.

[...]

PR #1055, opened 2016-05-03
WIP: simplified config

hunse:
The config system has seemed needlessly complicated to me for a while. Here's an idea of an easier way to do it.

I was prompted by #1043, because this method no longer requires a list of classes to configure when making a `Config` object.

There are still some things that need touching up, but I feel like this covers the vast majority of use cases for the config system. What are the drawbacks to this way of doing things?

### DIFF
--- a/nengo/config.py
+++ b/nengo/config.py
@@ -21,227 +21,111 @@ from nengo.rc import rc
 from nengo.utils.threading import ThreadLocalStack
 
 
-class ClassParams:
-    """A class to store extra parameters and defaults on configured classes.
+class KeyAccessor:
+    def __init__(self, config, key):
+        self._config = config
+        self._key = key
 
-    This is used by `.Config` to associate defaults and new `.Parameter`
-    instances with existing objects. It should not be instantiated outside of
-    `.Config.configures`.
+        self._values = {}
+        self._params = {}
 
-    Parameters
-    ----------
-    configures : class
-        The class with which to associate new defaults and parameters.
-    """
+        self._is_class = inspect.isclass(key)
+        self._key_class = key if inspect.isclass(key) else type(key)
 
-    def __init__(self, configures):
-        assert inspect.isclass(configures)
-        self._configures = configures
-        self._extra_params = {}
-        self._default_params = tuple(
-            attr for attr in dir(self._configures)
-            if is_param(getattr(self._configures, attr)))
-
-    def __contains__(self, key):
-        return self in self.get_param(key)
-
-    def __delattr__(self, key):
-        if key.startswith("_"):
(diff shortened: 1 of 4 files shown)


jgosmann:
What things aren't possible anymore wtih this that were before?

hunse:
So right now, setting parameters (previously done by `ClassParams.set_param`) is not possible, but I'm not really sure if anyone was using that. It might be something that we could add.

Printing all defaults is also currently not implemented. That should also be possible, though it's now a bit less clear what "all" means. We were printing the defaults for all the classes being configured by that Config, so now we would need some way to specify what classes we want to print.

hunse:
But in terms of things that aren't actually possible, I'm not sure. As I said, the old config system confused me, and I was hoping somebody who knows it better could say if there is something that it could do that this can't. @tbekolay or @drasmuss?

One thing might be the validation of defaults. Because there's no more `Parameter.set_default` method, it becomes a little less clear where to do this validation. One could always just call `Parameter.validate` in `Config`, but some Parameters override the set method to cast objects, etc., on set, and this would circumvent that. But looking at the code, it seems to me the old system was circumventing that too, since typically when someone overrode the `__set__` method they did not also override `set_default`. One solution to this might be to do casting in `validate`, which should help both in this framework and the old one.

tbekolay:
I'll all for simplifying the config stuff! There are lots of weird cases that need to be properly managed, which is why the code got so weird and complicated. But if this stuff passes all of the unit tests and looks right when executing the `config.ipynb` notebook, then we can work out the nits I'm sure. I'll take a closer look at Friday or early next week so I understand what's going on!

> I'm not really sure if anyone was using `set_param`

[nengo_gui](https://github.com/nengo/nengo_gui/blob/master/nengo_gui/config.py) and [nengo_spinnaker](https://github.com/project-rig/nengo_spinnaker/blob/master/nengo_spinnaker/config.py) use it, so we definitely need that functionality.

jgosmann:
I'm making use of the config system in PR #941 in some slightly weird ways:

* There is a class configuring itself.
* And it's configuring other classes which have the a parameter with the same name.

Not sure if this PR affects this in any way, but maybe something to consider.

hunse:
I got this passing the `test_config` tests, but it's still failing a lot of other ones. I don't think it's worth pouring more time into, since it's become considerably less simple since I started it (to handle the `set_param` stuff, as well as a number of other features).

I think the config system can still be simplified (e.g. by not storing config values on parameters themselves, so that there's less interaction between `config.py` and `params.py`), but I think it will be much less a drastic change than I was hoping. I'm going to close this for now.

PR #1057, opened 2016-05-05
Fix missing super call.

jgosmann:
The missing super call causes problems with other branches I'm working on; but independent of those it won't hurt to fix this.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,9 @@
+[MESSAGES CONTROL]
+
+disable = all
+enable = W0231
+
+
+[REPORTS]
+
+reports = no
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,11 @@ matrix:
         PIP_DEPS="$PIP_DEPS pytest-cov coveralls"
     - env: >
         PYTHON="2.7"
+        NUMPY=""
         SETUP_CMD=""
-        EXTRA_CMD="flake8 -v nengo"
+        EXTRA_CMD="flake8 -v nengo && pylint nengo"
         CONDA_DEPS="flake8"
-        PIP_DEPS=""
+        PIP_DEPS="pylint"
     - env: PYTHON="2.7" NUMPY="1.6"
     - env: PYTHON="2.7" NUMPY="1.7"
     - env: PYTHON="2.7" NUMPY="1.8"
@@ -59,4 +60,4 @@ script:
       python setup.py -q install;
       eval python setup.py "$SETUP_CMD";
     fi
-  - if [[ -n $EXTRA_CMD ]]; then $EXTRA_CMD; fi
+  - if [[ -n $EXTRA_CMD ]]; then eval "$EXTRA_CMD"; fi
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -64,12 +64,12 @@ class SimBCM(Operator):
 
     def __init__(self, pre_filtered, post_filtered, theta, delta,
                  learning_rate, tag=None):
+        super(SimBCM, self).__init__(tag=tag)
         self.pre_filtered = pre_filtered
         self.post_filtered = post_filtered
         self.theta = theta
         self.delta = delta
         self.learning_rate = learning_rate
-        self.tag = tag
 
         self.sets = []
         self.incs = []
@@ -152,13 +152,13 @@ class SimOja(Operator):
 
     def __init__(self, pre_filtered, post_filtered, weights, delta,
                  learning_rate, beta, tag=None):
+        super(SimOja, self).__init__(tag=tag)
         self.pre_filtered = pre_filtered
         self.post_filtered = post_filtered
         self.weights = weights
         self.delta = delta
         self.learning_rate = learning_rate
         self.beta = beta
-        self.tag = tag
 
         self.sets = []
         self.incs = []
@@ -243,6 +243,7 @@ class SimVoja(Operator):
 
     def __init__(self, pre_decoded, post_filtered, scaled_encoders, delta,
                  scale, learning_signal, learning_rate, tag=None):
+        super(SimVoja, self).__init__(tag=tag)
         self.pre_decoded = pre_decoded
         self.post_filtered = post_filtered
         self.scaled_encoders = scaled_encoders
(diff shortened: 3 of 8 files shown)


arvoelke:
LGTM

hunse:
Good catch.

This type of thing is something I've run into before, too. It's easy to forget to call the FrozenObject constructor, but this is necessary for FrozenObjects to work properly. Is there some way we could check this? Or some fancier way of doing things that makes sure the necessary attributes get added (e.g. with metaclassing), or is that too much magic/complexity? This might also depend on if we expect users to create FrozenObjects, which could happen if they're making their own Synapses and NeuronTypes.

jgosmann:
I think the best way to catch this sort of errors is to use pylint, but we would have to decide on the appropriate set of warnings and go through the code base to fix a bunch of other stuff.

On May 6, 2016 12:11:54 PM EDT, Eric Hunsberger <notifications@github.com> wrote:
>Good catch.
>
>This type of thing is something I've run into before, too. It's easy to
>forget to call the FrozenObject constructor, but this is necessary for
>FrozenObjects to work properly. Is there some way we could check this?
>Or some fancier way of doing things that makes sure the necessary
>attributes get added (e.g. with metaclassing), or is that too much
>magic/complexity? This might also depend on if we expect users to
>create FrozenObjects, which could happen if they're making their own
>Synapses and NeuronTypes.
>
>---
>You are receiving this because you authored the thread.
>https://github.com/nengo/nengo/pull/1057#issuecomment-217486965

-- 

hunse:
And pylint isn't going to help users. Maybe the best thing to help them is just a good set of examples, since there are many things to consider when creating one's own FrozenObject (like how to use Parameters).

jgosmann:
Having to call the super constructor is normal Python and should always be done. I don't think we should add special code to ensure this in this one special case. I can't speak for the "many other things to consider".

tbekolay:
We used to do the kind of magic that @hunse is referring to in networks because we were encouraging users to subclass network. Ultimately, the consensus was to remove all that even though we expect that people will still make their own network subclasses, so I feel like any magic that we add will suffer the same fate. Better / more examples would definitely be good!

tbekolay:
As suggested by @arvoelke previously, I added a test that fails with this PR. The test is ... a bit complicated (maybe too generic?) but ensures that new distributions added to Nengo will fail this test if the superclass constructor isn't called. Also added a changelog entry. These probably need to be reviewed by at least one person.

nengo/tests/test_dists.py
```diff
@@ -1,10 +1,17 @@
...
+
 import numpy as np
 import pytest
 
 import nengo.dists as dists
 import nengo.utils.numpy as npext
 from nengo.exceptions import ValidationError
 
+all_dists = [getattr(dists, attr) for attr in dir(dists)
+             if (inspect.isclass(getattr(dists, attr))
```
jgosmann: Can something be a subclass of `Distribution` without being a class?

tbekolay: No, but `issubclass` raises an error is the first argument isn't a class.

nengo/tests/test_dists.py
```diff
@@ -295,3 +302,57 @@ def test_frozen():
...
+            elif isinstance(args[to_change], int):
+                args[to_change] = 0.5
+            elif isinstance(args[to_change], float):
+                args[to_change] = np.ones(1)
+            else:
+                print(to_change, arg_str, args)
+                raise
+            fixes += 1
+
+    assert hasattr(inst, '_paramdict')  # Created by FrozenObject.__init__
```
jgosmann: I don't like that the test has to check for an implementation detail (as opposed to the public interface of class).

tbekolay: Hmm... how do you propose to test the public interface? `FrozenObject` only defines private variables and special functions. We could test the `hash` stuff like `test_frozen` does, but a dist subclass could presumably implement that stuff too, if it wanted.

jgosmann:
Apart from my inline comments it looks good to me, but I still feel that the proper way to check for constructor calls would be pylint. But maybe such a test is better for know until we manage to sort out which pylint warnings to fix and which to ignore?

tbekolay:
Yeah, I would be fine with using pylint instead. We could just whitelist that test, and add in more incrementally? I'll try that out!

nengo/tests/test_dists.py
jgosmann: Not sure ... I just try to avoid testing for implementation details.
Maybe one could monkey patch the base classes constructor to check that it gets called, but that would things more complicated.

tbekolay: The only downside that I can see with testing for implementation details (since those implementation details are contained in this package) is that the test might have to change if the `FrozenDict` implementation changes. That doesn't seem like a big deal to me... anyhow, if the pylint stuff works then this will be moot anyhow, so I'll look into that first!

tbekolay:
OK! I removed that hacky test commit and added pylint checking for the superclass call specifically. This identified that we don't do this in `Operator` subclasses, which is fixed here also. It wasn't strictly necessary for operators, but we have in the past forgotten to include the `tag`, which checking for the superclass constructor call would have identified.

jgosmann:
You're pylint invocation isn't working:
```
checking &&

&&:1:1: E902 IOError: [Errno 2] No such file or directory: '&&'

checking pylint

pylint:1:1: E902 IOError: [Errno 2] No such file or directory: 'pylint'
```

tbekolay:
Yeah, travis CI is weird. Amended the commit, fingers crossed.

jgosmann:
Nope, `/home/travis/build.sh: line 45: exec: flake8 -v nengo && pylint nengo: not found`

tbekolay:
Works now.

jgosmann:
LGTM

tbekolay:
@hunse or @drasmuss can you merge?

hunse:
So my only issue is that the pylint output colorization looks bad with my terminal colours. So I'd prefer to have it off by default, and if people want to turn it on themselves it's an easy command line option.

If and when someone okays my fixup that does that, I can merge.

tbekolay:
Fair enough! Merge away!

PR #1060, opened 2016-05-10
Switch to codecov

tbekolay:
This PR switches to codecov from coveralls.io. Coverage from this branch can be seen at https://codecov.io/gh/nengo/nengo/branch/codecov

The second commit also simplifies `.travis.yml` a fair bit, making it possible / easy to test coverage on both Python 2 and 3. Codecov automatically merges the two reports together, so we don't have to ignore lines/files that are specific to one of Python 2 / 3.

Once we do a release with this PR in it, I'll remove Nengo from coveralls.io.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,58 +6,67 @@ notifications:
 
 env:
   global:
-    - NUMPY="1.10"
-    - SETUP_CMD="test --addopts 'nengo -n 2 -v --durations 20'"
-    - EXTRA_CMD=""
-    - CONDA_DEPS="matplotlib jupyter pygments pytest"
-    - PIP_DEPS="pytest-xdist"
+    - NUMPY="1.11"
+    - COVERAGE="false"
+    - STATIC="false"
+    - CONDA_DEPS="matplotlib jupyter"
+    - PIP_DEPS="pytest pytest-xdist"
 
 matrix:
   include:
-    - env: >
-        PYTHON="2.7"
-        SETUP_CMD="test --addopts '--cov-config .coveragerc --cov nengo nengo -n 2 -v --durations 20'"
-        EXTRA_CMD="bash <(curl -s https://codecov.io/bash)"
-        CONDA_DEPS="$CONDA_DEPS coverage"
-        PIP_DEPS="$PIP_DEPS pytest-cov"
-    - env: >
-        PYTHON="2.7"
-        NUMPY=""
-        SETUP_CMD=""
-        EXTRA_CMD="flake8 -v nengo && pylint nengo"
-        CONDA_DEPS="flake8"
-        PIP_DEPS="pylint"
+    - env: PYTHON="2.7" COVERAGE="true"
+    - env: PYTHON="2.7" STATIC="true"
     - env: PYTHON="2.7" NUMPY="1.6"
     - env: PYTHON="2.7" NUMPY="1.7"
     - env: PYTHON="2.7" NUMPY="1.8"
     - env: PYTHON="2.7" NUMPY="1.9"
+    - env: PYTHON="2.7" NUMPY="1.10"
     - env: PYTHON="3.4"
-    - env: PYTHON="3.5"
+    - env: PYTHON="3.5" COVERAGE="true"
 
 # Setup Miniconda
 before_install:
   - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
-  - bash miniconda.sh -b -p $HOME/miniconda
+  - bash miniconda.sh -b -p "$HOME/miniconda"
   - export PATH="$HOME/miniconda/bin:$PATH"
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   - conda info -a
-  - conda create -q -n test python=$PYTHON pip
+  - conda create -q -n test python="$PYTHON" pip
   - source activate test
 
 # Install packages with conda, then pip
 install:
-  - if [[ -n $NUMPY ]]; then export CONDA_DEPS="$CONDA_DEPS numpy=$NUMPY"; fi
-  - if [[ -n $CONDA_DEPS ]]; then conda install $CONDA_DEPS; fi
-  - if [[ -n $PIP_DEPS ]]; then eval pip install "$PIP_DEPS"; fi
+  - if [[ "$COVERAGE" == "true" ]]; then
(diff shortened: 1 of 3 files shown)


jgosmann:
LGTM, I forsee merge conflicts with #1057.

There will still be adjustments to the coverage config necessary. For example, imports, class and function definitions are counted as missed lines. (That is also currently the case, so I would be fine merging this PR and fixing the rest with another PR.)

tbekolay:
Yep, no worries about the merge conflicts, I'll rebase this once #1057 is merged. Adding pylint here will be a bit cleaner, actually.

The actual coverage amount has nothing to do with codecov or coveralls really, both services parse Python's `coverage` module's output. So the fixes for coverage will be with that (perhaps some `.coveragerc` options?)

tbekolay:
This is rebased and ready to go now, btw. Needs another review.

jgosmann:
LGTM

jgosmann:
except: The codecov "details" link isn't working? But that seems to be codecov's fault?

tbekolay:
Pretty sure that's because there's been no coverage collected for master yet, so it can't do a diff between this branch and master.

hunse:
I added a small fixup to make the travis.yml logic a bit clearer IMO. If someone can check that the changes look good (especially that they'll do the same thing), then I can merge.

.travis.yml
```diff
@@ -56,14 +56,13 @@ script:
   - export TEST_EXTRA=""
   - if [[ "$COVERAGE" == "true" ]]; then
       export TEST_EXTRA="--cov-config .coveragerc --cov nengo";
-    fi
-  - if [[ "$STATIC" == "false" ]]; then
+    elif [[ "$STATIC" == "true" ]]; then
```
tbekolay: Making this an elif means that the tests won't get run if `"$COVERAGE" == "true"`; pretty sure it needs to stay an if/else.

hunse: Oh, I thought that was the point. I thought that STATIC and COVERAGE would never both be true. Also, in the install section above, it's an elif.

hunse: Oh, though I see, the last tests have to run... yeah, okay.

tbekolay: It is true that STATIC and COVERAGE won't both be true, but that's not what's happening in this section...

In `install`, we install certain things by default, then different things in the COVERAGE and STATIC cases.

In `script`, we run tests by default, but not in the STATIC case (we run the static analyzers). In the COVERAGE case, we add a few arguments to the tests run by default. So, either the `py.test` invocation has to be added to the COVERAGE case (which feels like unnecessary duplication to me), or keep the logic separate.

hunse: Check my new commit.

tbekolay: Also, the "default" case is `"$STATIC" == "false"`, hence why that's the first condition in the if; it'll be hit more often so it should be first. The static tests are the exceptional case.

hunse: Is that how you do things?? I usually go by length, putting shorter cases near the top of an if block, and longer cases near the end. That's just a rule of thumb, though, I do make exceptions.

.travis.yml
```diff
@@ -54,12 +54,14 @@ script:
...
-      export TEST_EXTRA="--cov-config .coveragerc --cov nengo";
-    elif [[ "$STATIC" == "true" ]]; then
+  - if [[ "$STATIC" == "true" ]]; then
       flake8 -v nengo;
       pylint nengo;
     else
+      if [[ "$COVERAGE" == "true" ]]; then
+        export TEST_EXTRA="--cov-config .coveragerc --cov nengo";
+      fi
+
```
tbekolay: Huh, I'm surprised TravisCI will let you have blank lines like this. good to know!

.travis.yml
```diff
@@ -37,11 +37,11 @@ before_install:
 
 # Install packages with conda, then pip
 install:
-  - if [[ "$COVERAGE" == "true" ]]; then
-      export PIP_DEPS="$PIP_DEPS pytest-cov";
-    elif [[ "$STATIC" == "true" ]]; then
+  - if [[ "$STATIC" == "true" ]]; then
```
tbekolay: I think I'd prefer the static case to be the else as it was before, both here and below... we run 9 tests total, and STATIC is only true in 1 of those tests, so it's always the exception.

hunse: It wasn't the else here originally; you had it first. I just changed it back. I'm fine having it be the else (i.e. test STATIC == false), I just want it to be the same in both install and script.

.travis.yml
tbekolay: Yeah, my heuristic is what will be executed most often ;) I think it's partly because the 'if' is assumed to be true for branch prediction way way under the hood, which is probably not relevant here, but I definitely read the if before the else, so it works in that way too.

.travis.yml
tbekolay: Before as in your first commit changing it then ;)

.travis.yml
hunse: > I think it's partly because the 'if' is assumed to be true for branch prediction way way under the hood, which is probably not relevant here

I think modern branch predictors are likely so advanced that this makes no difference. They might assume true the very first time, but after one false that's all out the window.

I just hate scrolling through a huge if condition to get to the smaller conditions. Again, it doesn't matter here since they're all quite small, but if you had HUGE->SMALL->SMALL you'd never get the starts of all of them on the screen at once, whereas if you have SMALL->SMALL->HUGE you can see the start of each condition on one screen, giving you a sense of what the condition is and what it's doing. I read it in a programming book one time, it's a thing :)

tbekolay: Haha, fair enough, I agree with the S->S->H ordering!

.travis.yml
```diff
@@ -54,17 +54,16 @@ script:
...
   - "echo 'c.HistoryAccessor.enabled = False\n' > $HOME/.ipython/profile_default/ipython_config.py"
   - export TEST_EXTRA=""
-  - if [[ "$STATIC" == "true" ]]; then
-      flake8 -v nengo;
-      pylint nengo;
-    else
+  - if [[ "$STATIC" == "false" ]]; then
       if [[ "$COVERAGE" == "true" ]]; then
         export TEST_EXTRA="--cov-config .coveragerc --cov nengo";
       fi
```
tbekolay: Thanks :D I think the TravisCI error is here, need a semicolon after `fi`

tbekolay:
LGTM! Sorry for the many comments, apparently I care about shell script style more than I thought

PR #995, opened 2016-03-24
Deferrables

jgosmann:
PR separated out from PR #941.

~~Based on PR #977.~~ Copying is not required for the functionality in this PR, so I rebased it to master to remove that part.

PR #1062, opened 2016-05-11
Deferrable neuron number

jgosmann:
This is a proof of concept implementation of determining the neuron number to meet a certain accuracy with the deferrables system (#995).

Before this can be merged with master a least the following things have to be done:

* [ ] Refactor
* [ ] Document
* [ ] Very slow, are any improvements possible? It should be possible to cache the `opt_slope`, `opt_inter` and `first_p` values for a given neuron type.
* [ ] More testing.


jgosmann:
[Given the decision regarding deferrables](https://github.com/nengo/nengo/pull/995#issuecomment-241622867) in #995 this needs to be reimplemented as a helper function.

Please keep this branch around for reference as some of the code could be reused.

PR #1063, opened 2016-05-11
Fix coverage

jgosmann:
Changes the coverage command to fix lines detected as missed which are not missed ([compare](http://stackoverflow.com/a/16524426)).
Also excludes lines raising NotImplementedErrors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_install:
 # Install packages with conda, then pip
 install:
   - if [[ "$COVERAGE" == "true" ]]; then
-      export PIP_DEPS="$PIP_DEPS pytest-cov";
+      export PIP_DEPS="pytest coverage";
     elif [[ "$STATIC" == "true" ]]; then
       export CONDA_DEPS="";
       export PIP_DEPS="flake8 pylint";
@@ -53,14 +53,15 @@ install:
 script:
   - mkdir -p "$HOME/.ipython/profile_default"
   - "echo 'c.HistoryAccessor.enabled = False\n' > $HOME/.ipython/profile_default/ipython_config.py"
-  - export TEST_EXTRA=""
   - if [[ "$STATIC" == "false" ]]; then
-      if [[ "$COVERAGE" == "true" ]]; then
-        export TEST_EXTRA="--cov-config .coveragerc --cov nengo";
-      fi;
       python -c "import numpy; numpy.show_config()";
       python setup.py -q develop;
-      eval py.test nengo -n 2 -v --duration 20 "$TEST_EXTRA";
+      if [[ "$COVERAGE" == "false" ]]; then
+        py.test nengo -n 2 -v --duration 20;
+      else
+        coverage run --rcfile .coveragerc --source nengo -m py.test nengo -v --duration 20;
+        coverage report;
+      fi;
     else
       flake8 -v nengo;
       pylint nengo;
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,6 @@
 [run]
 source = nengo
+[report]
 omit = */tests/test*
+exclude_lines =
+    raise NotImplementedError
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ wintest.sh
 Vagrantfile
 *.class
 *.eggs/
+.coverage
+htmlcov


jgosmann:
Looks like the fix make exactly the opposite set of lines "missed" :angry:

jgosmann:
Seems like the fix does not play well with xdist, disabled it for now when coverage information is collected. Will check if I can get both to work.

jgosmann:
I'm also wondering if it would make sense to exclude `nengo/utils/testing.py` from the coverage?

Also, can we merge coverage from tests run on Windows too?

tbekolay:
I think that's the main selling point of `pytest-cov`, that it works with xdist... not sure exactly how much xdist speeds up the builds, so if it means we wait another couple minutes, that's better than incorrect coverage!

I'm not sure how exactly codecov merges reports, but from what I can tell, it should accept them from AppVeyor as long as the commit hash is the same, which it should be. Does different code get executed in Windows?

jgosmann:
> Does different code get executed in Windows?

[Just came randomly across this.](https://github.com/nengo/nengo/blob/master/nengo/utils/paths.py#L5). Maybe there are a few other rare places.

tbekolay:
Also it seems as though codecov is having some issues right now, in case you're looking for the results on codecov.io: https://github.com/codecov/support/issues/194 https://github.com/codecov/support/issues/195

> Just came randomly across this.. Maybe there are a few other rare places.

Ah yeah. Makes sense, those things should get tested as they're likely to be things we won't catch ourselves.

jgosmann:
I created a bug report upstream (pytest-dev/pytest-cov#117).
For now it is probably best to run the coverage analysis without xdist. It doesn't seem to be any slower on Travis-CI anyways.

tbekolay:
Running without xdist sounds good to me!

I also suspect the way codecov merges reports might be problematic for our workflow, so I made an issue upstream too: https://github.com/codecov/support/issues/196

hunse:
> I'm also wondering if it would make sense to exclude `nengo/utils/testing.py` from the coverage?

I would not exclude any files. There may be files that we don't expect 100% coverage for, or maybe any coverage for. But it's still nice to have the information about what parts of those files are run during the tests.

jgosmann:
Not sure why I had this still marked as WIP, this should be ready for review.

.travis.yml
```diff
@@ -53,14 +53,15 @@ install:
...
   - "echo 'c.HistoryAccessor.enabled = False\n' > $HOME/.ipython/profile_default/ipython_config.py"
-  - export TEST_EXTRA=""
   - if [[ "$STATIC" == "false" ]]; then
-      if [[ "$COVERAGE" == "true" ]]; then
-        export TEST_EXTRA="--cov-config .coveragerc --cov nengo";
-      fi;
       python -c "import numpy; numpy.show_config()";
       python setup.py -q develop;
-      eval py.test nengo -n 2 -v --duration 20 "$TEST_EXTRA";
+      if [[ "$COVERAGE" == "true" ]]; then
```
tbekolay: Very minor nitpick, but since I did it to @hunse: could this be the else instead? I.e., Check it coverage is false and do the normal tests, otherwise collect coverage.

.travis.yml
```diff
@@ -53,14 +53,15 @@ install:
...
+        coverage run --rcfile .coveragerc --source nengo -m py.test nengo -v --duration 20;
+        coverage report;
+      else
+        py.test nengo -n 2 -v --duration 20;
```
tbekolay: We are duplicating the `-v` and `--duration 20` options here so they could get out of sync... but I think leave it as is for now; if we add more `py.test` options we should make an environment variable for them, but it's easier to read as it is here.

tbekolay:
Just one minor nitpick; otherwise LGTM!

.travis.yml
jgosmann: Sure, whatever you prefer. Feel free to change it. I'm leaving (tomorrow) until end of next week, not sure how much I'll get done myself.

tbekolay: Will do!

tbekolay:
Fixed up the history and made my nitpick change. Since this doesn't touch any code I'll merge once the CI tests pass (unless anyone has an objection while they're running).

PR #1064, opened 2016-05-16
Added initial y0 to WhiteSignal process

arvoelke:
I ran into a subtle issue with using `WhiteSignal` without a lowpass; you sometimes get a large infinite frequency at the start of the simulation. This can be problematic in some situations with principle 3 (can give details if needed). Setting `y0=0` helps prevent this problem.

### DIFF
--- a/nengo/processes.py
+++ b/nengo/processes.py
@@ -159,6 +159,9 @@ class WhiteSignal(Process):
         timestep, which is ``0.5 / dt``.
     rms : float, optional (Default: 0.5)
         The root mean square power of the filtered signal
+    y0 : float, optional (Default: None)
+        Align the phase of each output dimension to begin at the value
+        that is closest (in absolute value) to y0.
     seed : int, optional (Default: None)
         Random number seed. Ensures noise will be the same each run.
     """
@@ -166,12 +169,14 @@ class WhiteSignal(Process):
     period = NumberParam('period', low=0, low_open=True)
     high = NumberParam('high', low=0, low_open=True)
     rms = NumberParam('rms', low=0, low_open=True)
+    y0 = NumberParam('y0', optional=True)
 
-    def __init__(self, period, high, rms=0.5, **kwargs):
+    def __init__(self, period, high, rms=0.5, y0=None, **kwargs):
         super(WhiteSignal, self).__init__(default_size_in=0, **kwargs)
         self.period = period
         self.high = high
         self.rms = rms
+        self.y0 = y0
 
         if self.high is not None and self.high < 1. / self.period:
             raise ValidationError(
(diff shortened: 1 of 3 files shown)


nengo/processes.py
```diff
@@ -208,6 +213,13 @@ def make_step(self, shape_in, shape_out, dt, rng):
         coefficients *= np.sqrt(2 * n_coefficients)
         signal = np.fft.irfft(coefficients, axis=0)
 
+        if self.y0 is not None:
+            # Starts each dimension off where it is closest to y0
+            def shift(x):
+                offset = np.argmin(abs(self.y0 - x))
+                return np.roll(x, -offset+1)  # +1 since t starts at dt
+            signal = np.apply_along_axis(shift, 0, signal)
```
arvoelke: Note: `roll` here takes an axis argument, but it doesn't help prevent the `apply_along_axis` because `offset` could be different per dimension in `shape[1:]`.

tbekolay:
This LGTM, just needs a second review and a changelog entry.

arvoelke:
Added changelog entry and shortened test simulation time.

jgosmann:
LGTM :cake:

hunse:
> you sometimes get a large infinite frequency at the start of the simulation

What do you mean by this?

nengo/tests/test_processes.py
```diff
@@ -170,8 +170,9 @@ def test_whitesignal_y0(Simulator, seed, y0, d):
     with Simulator(model, seed=seed) as sim:
         sim.run(t)
     values = sim.data[up]
+    error = np.min(abs(y0 - values), axis=0)
 
-    assert np.allclose(values[0, :], y0, atol=1e-3)
+    assert ((y0 - error <= values[0, :]) & (values[0, :] <= y0 + error)).all()
```
tbekolay: How is this more careful / reliable than `allclose`? Would adjust to tolerance be sufficient?

arvoelke:
The initial state of all the signals are 0, and then suddenly at `t = dt` they receive a large delta. From the perspective of the system, it's as though it's walked into the middle of a simulation. From the perspective of the state-space, it is filtering a delta impulse because its initial state is `x_0 = 0`. From a theory standpoint this change doesn't fix the problem, because there are still non-zero phases in the input, but in practice this has made my problems go away.

nengo/tests/test_processes.py
arvoelke: That part of the message was referring to the `error` calculation. For short signals with high frequencies there is no strict guarantee that we'll get anything close to `y0`. This checks that we at least got as close as we possibly could, by making it sure it matches the available minimum (note each dimension will match either the lower or upper bound).

hunse:
Ahh, I see what you're saying. Your system is underdamped, so that essentially high-frequency change creates problems. Personally, I would use `FilteredNoise` instead, but this PR seems like not a bad thing to have, too.

Issue #1054, opened 2016-05-03
Custom solver without default args will break cache


arvoelke:
```
class CustomSolver(nengo.solvers.LstsqL2):
    def __call__(self, A, Y, *args, **kwargs):
        return super(CustomSolver, self).__call__(A, Y, *args, **kwargs)
    
with nengo.Network(seed=0) as model:
    a = nengo.Ensemble(10, 1)
    nengo.Connection(a, a, solver=CustomSolver())

nengo.Simulator(model)
```

This gives the following error:
```
    354             except TypeError:
    355                 args, _, _, defaults = inspect.getargspec(solver.__call__)
--> 356             args = args[-len(defaults):]
    357             if rng is None and 'rng' in args:
    358                 rng = defaults[args.index('rng')]

TypeError: object of type 'NoneType' has no len()
```

Changing the signature to `(self, A, Y, rng=None, E=None)` resolves this issue.

arvoelke:
Also... there is a slight inherent problem with developing a custom solver when the cache is enabled: changing the method's implementation won't change the fingerprint (unless the attributes are changed as well). This means that someone trying to write their own solver might be confused when they keep getting the same results, despite changing their code.

This could also be hazardous with Nengo solvers if, say, a bugfix was made to one of the solvers. The cache would not pick up on the fix and continue using the previous incorrect results. 

Perhaps cache misses should be forced when the solver is _not_ within the Nengo directory (to alleviate the first problem), and the Nengo `__version__` should be a part of the cache key (to alleviate the second problem)?

Obviously this is hard problem in general, since it's always possible to construct weird counter-examples, but I think we should aim to be slightly less aggressive with how we detect hits, since false-hits are way more "costly" in terms of user time than false-misses. Also a way to force cache misses on specific connections (#1044) would be nice when all else fails.

hunse:
> changing the method's implementation won't change the fingerprint

This is a general problem with caches and development, and unfortunately I don't think there's much we could do about it. I guess we could have the cache disabled by default in the development version. Either way, I think it's good to make developers aware that the cache is there, and perhaps recommend developing with the cache off.

> This could also be hazardous with Nengo solvers if, say, a bugfix was made to one of the solvers.

This is more worrying, but I think something like you suggested (incorporating the version in some way), should address it. Do we not do that right now? My inclination is that we should always be clearing cache files that don't match the current version, but I'm not sure how difficult that is. Actually, we should probably have the cache files organized with separate folders for each version.

[...]

PR #1066, opened 2016-05-16
Only cache things with explicit mark.

jgosmann:
Addresses #1054.

Code can change and we can't detect it, so we allow only things that
are marked as supporting fingerprints to be cached and only mark
objects that should be stable as supporting this. Note that an object
marked as supporting fingerprints will retain this mark even if it
references objects that are not marked as supporting fingerprints
(because the fingerprint is created by pickling and we cannot hook into
that process).

This also removes the solver_fn from the fingerprint because it is not
needed in there. It should always be the same function which is part of
the builder (the solver itself is still part of the fingerprint of
course). This in turn allows to simplify the SolverMock in the test
code a little bit. (It actually is slightly misnamed as it mocks the
solver_fn and not a Solver.)

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -16,7 +16,7 @@ from nengo.exceptions import FingerprintError, TimeoutError
 from nengo.rc import rc
 from nengo.utils import nco
 from nengo.utils.cache import byte_align, bytes2human, human2bytes
-from nengo.utils.compat import is_string, pickle, PY2
+from nengo.utils.compat import int_types, is_string, pickle, PY2
 from nengo.utils.lock import FileLock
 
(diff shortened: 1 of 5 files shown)


nengo/cache.py
```diff
@@ -70,7 +70,20 @@ class Fingerprint(object):
 
     __slots__ = ['fingerprint']
 
+    NON_REFERENCE_TYPES = (
+        bool, float, complex, bytes, unicode, str) + int_types
```
tbekolay: `unicode` doesn't exist in Python 3; should use `string_types` from compat.

drasmuss:
Merged in #1068

PR #1067, opened 2016-05-17
Ensure flake8 errors fail TravisCI tests

tbekolay:
I noticed that in some other PRs, flake8 errors weren't resulting in a TravisCI failure.

TravisCI can only inspect the return code of the last command run in the `script` list, so in our big if/else chain, we need to ensure that the last command returns what we consider to be success or failure.

In the static case, we were not looking at the return value of flake8. This PR uses `&&` to use both return values.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,12 +59,10 @@ script:
       if [[ "$COVERAGE" == "false" ]]; then
         py.test nengo -n 2 -v --duration 20;
       else
-        coverage run --rcfile .coveragerc --source nengo -m py.test nengo -v --duration 20;
-        coverage report;
-      fi;
+        coverage run --rcfile .coveragerc --source nengo -m py.test nengo -v --duration 20 && coverage report;
+      fi
     else
-      flake8 -v nengo;
-      pylint nengo;
+      flake8 -v nengo && pylint nengo;
     fi
 
 after_success:


hunse:
Okay!

jgosmann:
Is it possible to execute both commands always and just 'and' the return values?
If not it might be better to do pylint first as it can check many things that flake8 checks. Just to get as many errors displayed in one check as possible.

tbekolay:
> Is it possible to execute both commands always and just 'and' the return values?

It's possible, but it's not pretty... see [this SO question](http://stackoverflow.com/questions/16357624/anding-exit-codes-in-bash). We can do something like this, but it's pretty hard to read, IMO.

One approach would be to use `tox` for the static testing; should be as simple as conda or pip installing `tox` and then `tox -e static`.... I'll try that out and make sure it's not too slow.

As for which to do first, I'd switch to pylint first if we enable more pylint checks, but right now we just check for the superclass init call... can deal with it in future PRs that add pylint checks!

tbekolay:
`tox` doesn't play well when using conda, as it turns out... so I did the ugly solution in a26edca53688fc8b. What do you think @jgosmann, worth it?

jgosmann:
It's not extremely horrible ... I'm fine either way.

hunse:
For short commands like these, I prefer the original `&&`. Maybe for longer commands we can bust out this messy stuff. Also, if we do do it the "ugly" way now or in the future, is there a way to add a comment?

tbekolay:
I think I prefer the original `&&` too so I'll go back to that. If more things get added to `.travis.yml` we can consider hiding the ugliness away in a separate shell script with some functions to make it less ugly.

hunse:
This is good to go, n'est-ce pas?

tbekolay:
Yep, as long as @jgosmann is ok with the `&&`

hunse:
He's away, and said previously he was fine either way.

jgosmann:
Sure

On May 24, 2016 9:56:08 PM GMT+01:00, Trevor Bekolay <notifications@github.com> wrote:
>Yep, as long as @jgosmann is ok with the `&&`
>
>---
>You are receiving this because you were mentioned.
>https://github.com/nengo/nengo/pull/1067#issuecomment-221388800

-- 

PR #1068, opened 2016-05-17
Only cache things in a registry

tbekolay:
This is an alternative implementation of #1066 that uses a class registry in the `Fingerprint` rather than modifying frontend objects.

It occurred to me that the cache system (at least for now) is specific to the reference backend, so it would be best to keep any cache-specific code local to the cache, rather than modifying frontend objects. So that's what this branch does.

I think the amount of code is, in the end, about the same. `Fingerprint.__init__` is more clear this way (IMO) but we don't get the benefit of subclassing inheriting things, so each class that should be fingerprinted needs a separate call to `Fingerprint.whitelist`. But perhaps that's desirable anyhow?

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -16,7 +16,7 @@ from nengo.exceptions import FingerprintError, TimeoutError
 from nengo.neurons import (AdaptiveLIF, AdaptiveLIFRate, Direct, Izhikevich,
                            LIF, LIFRate, RectifiedLinear, Sigmoid)
 from nengo.rc import rc
-from nengo.solvers import (Solver, Lstsq, LstsqL1, Nnls, NnlsL2,
+from nengo.solvers import (Lstsq, LstsqL1, Nnls, NnlsL2,
                            NnlsL2nz, LstsqNoise, LstsqMultNoise, LstsqL2,
                            LstsqL2nz, LstsqDrop)
 from nengo.utils import nco
@@ -24,7 +24,7 @@ from nengo.utils.cache import byte_align, bytes2human, human2bytes
 from nengo.utils.compat import (
     int_types, is_string, pickle, replace, PY2, string_types)
 from nengo.utils.least_squares_solvers import (
-    LeastSquaresSolver, Cholesky, ConjgradScipy, LSMRScipy, Conjgrad,
+    Cholesky, ConjgradScipy, LSMRScipy, Conjgrad,
     BlockConjgrad, SVD, RandomizedSVD)
 from nengo.utils.lock import FileLock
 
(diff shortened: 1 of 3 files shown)


jgosmann:
@tcstewar mentioned he is using the cache to cache something else. Not sure if that use is affected by this?

The downside I see with this approach is that there `check_*` functions required that are not tied to the actual class and might come out of sync with the actual implementation.

> it would be best to keep any cache-specific code local to the cache, rather than modifying frontend objects.

Except that there is still the `whitelist` call in those files.

I think both approaches have their advantages and disadvantages and overall I don't have a strong preference either way.

tbekolay:
> Except that there is still the whitelist call in those files.

Yes, I considered having these in `nengo/cache.py`, which we could still do. Does that make more sense?

> I think both approaches have their advantages and disadvantages and overall I don't have a strong preference either way.

I agree, pros and cons to both! I have an aversion to modifying user-facing objects whenever possible, and would like to reduce the surface area of the API (i.e., not introduce new classes like `Cacheable*`) which is why I wanted to try out this implementation. But when I realized I needed the `check_*` functions and that those functions would have to access `Fingerprint` internals I was less than 100% on this approach. So I'm open to arguments both ways.

jgosmann:
> Yes, I considered having these in nengo/cache.py, which we could still do. Does that make more sense?

I don't know ...

btw: Not sure if I mentioned this, but my approach was inspired by how the for example pickle allows objects to define the own pickling function.

mundya:
> It occurred to me that the cache system (at least for now) is specific to the reference backend

nengo_spinnaker uses the cache!

tcstewar:
> nengo_spinnaker uses the cache!

Yup!  I think this is another situation like the Parameter system where we need to be careful about changes because other backends are using this too...  nengo_brainstorm also can use the cache, and in a weird way....

drasmuss:
I pushed a commit that moves all the fingerprinting related code to `cache.py`, so we can see what that looks like.  Personally I like it, it keeps the caching apparatus hidden from the rest of the code, as I think it should be.

jgosmann:
I looked at this again and noticed that 9074d8c removed the dependence on solver internals. I added another commit to take this a step further and not just check attributes of the solver that are instances of `Solver` and `LeastSquaresSolver`, but all public attributes and recurse on those. It also add support for tuples and lists.

jgosmann:
I think I'm ok with and without that additional commit. I think the additional commit makes this more conservative because it checks more.

tbekolay:
Interesting! I'm :+1: on 64cf8f329ef67278b3bc9835. Pushed a fixup to remove some unused imports and space things out a bit more (is more readable for me, at least, but maybe I've been doing too much JavaScript). There are some flake8 errors in this branch, but they will be resolved once it's rebased to master.

jgosmann:
> space things out a bit more

Definitely nice when adding/removing things because you don't need to reformat everything to fit the line length. Only downside I see is somewhat more scrolling.

nengo/cache.py
```diff
@@ -109,20 +109,50 @@ class Fingerprint(object):
...
-    CHECKS = dict(
-        [(np.ndarray, check_dtype), (tuple, check_seq), (list, check_seq)] +
-        [(x, check_attrs) for x in SOLVERS + LSTSQ_METHODS + NEURON_TYPES])
+        SOLVERS + LSTSQ_METHODS + NEURON_TYPES
+    )
+    CHECKS = dict([
+        (np.ndarray, check_dtype),
+        (tuple, check_seq),
+        (list, check_seq)
+    ] + [
```
drasmuss: There's no :barf: emoji but if there was I would use it on this line just gonna settle for :scream:

drasmuss:
I'm shrugs on the reformatting; not how I would write it myself, but I think it's fine either way.  And the rest all looks :+1: to me.

PR #1069, opened 2016-05-23
Add a pull request template for better reviewing

tbekolay:
This PR adds a file, `.github/pull_request_template.md`, which will prepopulate the text field in all new pull requests. The intention is to make it easier for PR creators to include all the relevant details to make reviewing easier.

The template is based on a few sources, most notably https://github.com/TalAter/open-source-templates
with changes based on our development practices.

It seemed prudent to also add a CONTRIBUTING.rst file, which is kind of like README.rst in that it's a shorter version of what's in the documentation, hopefully not repeating too much. I rearranged the dev documentation a bit to tighten it up as well.


.github/pull_request_template.md
```diff
@@ -0,0 +1,34 @@
...
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that causes existing functionality to change)
+
+**Checklist:**
+<!--- Go over all the following points. Put an `x` in all the boxes that apply. -->
+<!--- If you're unsure about any of these, don't hesitate to ask. -->
+<!--- We're here to help! -->
+
+- [ ] I have read the **CONTRIBUTING.rst** document.
```
Seanny123: Is every contributor going to have to fill out these check boxes every time they make a pull request? I would find this a bit annoying.

If yes, I would reduce them to:
- [ ] My change requires a change to the documentation and I have updated the documentation accordingly.
- [ ] My change requires a changelog entry and I have included a changelog entry.

tbekolay: Yes, you have to fill out the checkboxes for each PR. The reason to keep them separate is if you're making a WIP pull request and you know it needs documentation, but you haven't written it yet. But you're right, it's a bit cumbersome. I've added a TODO section instead.

tbekolay:
Made some commits updating the PR template; please review!

hunse:
I'm not really looking forward to filling this out for each PR, but I guess it will be helpful. So LATM (that is, looks acceptable to me). :)

Seanny123:
Looks acceptable to me as well. (:

.github/pull_request_template.md
```diff
@@ -0,0 +1,42 @@
...
+- [ ] My change requires a change to the documentation
+      and I have updated the documentation accordingly.
+- [ ] My change requires a changelog entry and I have included a changelog entry.
```
jgosmann: I wonder if we can express these things in a simpler way because it is a bit cumbersome to read "my change requires X and I exactly did X" when the second part doesn't give me any new relevant information as what I have to check. Already the first part tells me that I have to think about whether to do X. Maybe change it to "I did X" and add a note that the creator of the PR should comment on why checkboxes are not checked (e.g. "I believe X is not necessary for this PR, because ...").

.github/pull_request_template.md
```diff
@@ -0,0 +1,42 @@
...
+<!--- If this change depends on another PR, please link to that PR here. -->
+
+**How has this been tested?**
+<!--- Please describe in detail how you tested your changes. -->
+<!--- Reviewers will test your PR in at least this way. -->
+
+**How long should this take to review?**
+<!--- Please estimate if this is a small, average, or large PR. -->
+<!--- Take into account both the size and complexity of the changes. -->
+<!--- Also note if this is a timely PR that should be reviewed by a certain date. -->
```
jgosmann: Do we expect people to give an estimate in some amount of time or just say something like "small" or "large"? I think time estimates are really difficult to give (especially for other people). In the latter case it might be useful to have checkboxes where you just click the right one. It would also be helpful to have some indication of what counts as small, average, and large. E.g. small (less than 10 lines changed and 3 files touched), average (neither small or large), large (more that 150 lines changed)

jgosmann:
Maybe it would be good to have a specific section for dependencies? So they don't get lost in some other text? Also it might be helpful to give some more information for these depenedencies: Have these been merged into this branch? Or is this branch based of another branch? What is the first commit that purely belongs to this PR?

hunse:
Other opinions on a separate section for dependencies? I would be fine with it. It's nice to have that listed somewhere.

tbekolay:
Made the changes requested in 8e8156e8a6302c2! I like the "interactions with other PRs" section, as it's also applicable to situations like #1066 and #1068 which conflict. I moved the length one so that all the checklist-y ones are at the end. The shorter checklist entries are definitely better!

hunse:
LGTM. If @jgosmann thinks that looks good too, I'll merge.

Seanny123:
Just one nitpick that shouldn't block this being merged. Because these are all checkboxes, they are shown as progress markers when browsing pull requests, which we'll now have to ignore, as we do with #1077 . Just wanted to point that out.

jgosmann:
LGTM 🍰 
✔️

tbekolay:
That's a good point which I didn't really think about... There are some
alternative markups we could think about, but yeah, I think for the time
being we should rely on labels for progress markers rather than the task
list thing!
On Jun 8, 2016 23:01, "Sean Aubin" <notifications@github.com> wrote:

Just one nitpick that shouldn't block this being merged. Because these are
all checkboxes, they are shown as progress markers when browsing pull
requests, which we'll now have to ignore, as we do with #1077
<https://github.com/nengo/nengo/pull/1077> . Just wanted to point that out.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/nengo/nengo/pull/1069#issuecomment-224788940>, or mute
the thread
<https://github.com/notifications/unsubscribe/AAMbvYEhmiL-W6GfMKGmHGjtUamPz3kDks5qJ4H5gaJpZM4Ik0Nz>
.

hunse:
@Seanny123, @tbekolay: What about something like 478ddf1? Then we reserve the checkboxes for actual things to be completed, so the progress should reflect actual progress on the PR.

Seanny123:
I would prefer this aesthetically. Is it going to be annoying to edit for
noobs?

hunse:
> Is it going to be annoying to edit for noobs?

If they know enough to be making a PR, hopefully they should be able to figure it out.

Seanny123:
I'm convinced. LGTM.

jgosmann:
How instead of uncommenting lines, removing lines that do not apply? It seems easier to me to delete whole lines than to delete exactly the right number of character at the start and end of line.

tbekolay:
I'm fine with uncommented or removing lines! I agree that removing lines is probably easier; might result in some people leaving all three in there, but that's relatively easy to correct I think.

Seanny123:
I agree with Jan

hunse:
Does that look good?

Seanny123:
That looks awesome. Thanks Eric!

tbekolay:
LGTM!

jgosmann:
Awesome :) :cake:

Issue #1071, opened 2016-05-25
More docstrings in conftest.py


Seanny123:
While documenting Best Practices for Testing I did some trawling through the Nengo test code. Here is some stuff that I found that I think would be useful to document:

- [x] `parametrize_function_name`: I think this is just converting the function name into something that can be passed as an argument, but it seems to be doing other stuff as well.
- [x] `recorder_dirname`: I'm very confused what that long if-statement is doing
- [x] since `load_class` is only used in `pytest_configure`, should it be defined above `pytest_configure` instead of below it?
- [x] why is the `function_seed` function so complicated? I think I'm missing some information on how it evolved.

hunse:
- `parametrize_function_name` is creating a unique name for a test function that accounts for values passed with the `pytest.mark.parametrize` fixture. That fixture basically allows you to run a test multiple times with different arguments, and `parametrize_function_name` adds the argument values onto the test function name so that we have a unique name for the test. This is useful for example when naming plots.
- `recorder_dirname` figures out the directory to put "recorder" things (e.g. plots) in. The directory is based on the simulator first, and then a subdirectory for each neuron type if that's used in the function. So the `if` statement is just figuring out first the value of the `Simulator` argument (if it is an argument, i.e. `'Simulator' in request.funcargnames`), and then the value of the `nl` or `nl_nodirect` argument, if either of them is used.
- I often prefer putting helper functions after the function they're used in. That way, you see the important function first, and aren't distracted by the helper.
- What do you find complicated about `function_seed`?

Seanny123:
Thanks for explaining this to me Eric. I have a few more (potentially silly) questions:

- What dose `nl_nodirect` stand for? I've figured out it has something to do with neuron types, but other than that, I'm not sure how it's supposed to be used.
- What I'm confused about in terms of `function_seed` is more in regards to how seeding works in our tests in general. I know we want a the same seed each time we want a test (otherwise a lot of our tests would have unpredictable outcomes), but do we also want many different seeds so that we aren't just testing the same system all the time? Is the diversity of seeds what `function_seed` is satisfying?

hunse:
- nonlinearity, without `nengo.Direct`. It's the same as `nl` but without Direct. For when we want to test with all neuron types but not direct mode.
[...]

PR #1074, opened 2016-05-28
Add docs to conftest.py

Seanny123:
Addresses some of confusion I had previous in #1071

### DIFF
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -56,7 +56,7 @@ def load_class(fully_qualified_name):
 
 @pytest.fixture(scope="session")
 def Simulator(request):
-    """the Simulator class being tested.
+    """The Simulator class being tested.
 
     Please use this, and not ``nengo.Simulator`` directly. If the test is
     reference simulator specific, then use ``RefSimulator`` below.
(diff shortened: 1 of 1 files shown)


jgosmann:
LGTM :cake:

nengo/conftest.py
```diff
@@ -56,7 +56,7 @@ def load_class(fully_qualified_name):
 
 @pytest.fixture(scope="session")
 def Simulator(request):
-    """the Simulator class being tested.
+    """The Simulator class being tested.
```
tbekolay: FYI this used to be the convention used by py.test for fixture docstrings (which are printed when you do `py.test --fixtures`) but it seems as though they've switched to sentence case :+1:

tbekolay:
Pushed a fixup commit to be closer to [PEP-257 style](https://www.python.org/dev/peps/pep-0257/). I found [pydocstyle](https://github.com/PyCQA/pydocstyle) which I'll get set up so that this is checked in CI. Take a look @Seanny123 and if it looks good to you I'll merge.

Seanny123:
LGTM. I didn't know there was a style for docs! I'm glad it will be checked in the CI from now on.

Issue #1058, opened 2016-05-06
Place for common configs


jgosmann:
I repeatedly run across situations where I want to create thresholding ensembles and it would be nice to sort of have a network for that. But it is not really a network, but only a set of config options set on the ensemble. So I was wondering if we should have a place for functions that create such configs?

Not sure if there are other common sets of configuration options that get used?

In the thresholding case it would look somewhat like this:
```python
def ThresholdingConfig(threshold):
    config = nengo.Config(nengo.Ensemble)
    config[nengo.Ensemble].dimensions = 1
    config[nengo.Ensemble].intercepts = nengo.dists.Exponential(0.15, threshold, 1.)
    config[nengo.Ensemble].encoders = nengo.dists.Choice([[1]])
    config[nengo.Ensemble].eval_points = nengo.dists.Uniform(threshold, 1.)
    return config
```
and be used like this
```python
with ThresholdingConfig(0.3):
    threshold = nengo.Ensemble(50)  # might be better to give dimensions explicitely?
```

tbekolay:
This is a great idea! Never occurred to me, but I can already think of a bunch of places that this would be super useful. It's a nice level of abstraction above the regular objects, but below networks... Could make our network construction code more concise / consistent. Exciting!

tbekolay:
From the meeting: let's steal Handbrake's terminology and call these "presets".

jgosmann:
Any other presets than thresholding that people have and want to see included?

tbekolay:
Presets for different receptor types could be useful; `networks.BasalGanglia` makes configs for AMPA and GABA receptors.

Seanny123:
@xchoo has a wide variety of configs that he uses for Spaun, including one for a ramp network if I remember correctly, but I think those more esoteric configs can wait until later.

Is the next step for this issue to make a pull request? And by pull request, I mean one containing:
- [ ] Threshold config
- [ ] Documentation indicating that these configs are available

jgosmann:
Yes, I will add my thresholding config (I think the example above needed some minor changes, I have a working version lying around).

jgosmann:
Added a PR for this.

tbekolay:
#1077 was merged, so this has been done!

Issue #1011, opened 2016-04-12
Warning on open simulators test may fail


jgosmann:
Relevant Travis-CI output [from this build](https://travis-ci.org/nengo/nengo/builds/122316947)

```
___________________________ test_warn_on_opensim_gc ____________________________

[gw0] linux2 -- Python 2.7.11 /home/travis/miniconda/envs/test/bin/python

Simulator = <class 'nengo.simulator.Simulator'>

    def test_warn_on_opensim_gc(Simulator):

        with nengo.Network() as net:

            nengo.Ensemble(10, 1)

    

        sim = Simulator(net)

        assert sim

    

        with warns(ResourceWarning):

            sim = None

>           gc.collect()

nengo/tests/test_simulator.py:117: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <nengo.utils.testing.warns object at 0x7fb0aaf8c4d0>, type = None

value = None, traceback = None

    def __exit__(self, type, value, traceback):

        if not any(r.category is self.warning_type for r in self.record):

>           self._pytest.fail("DID NOT WARN")

E           Failed: DID NOT WARN

nengo/utils/testing.py:198: Failed

============== 1 failed, 674 passed, 55 skipped in 155.97 seconds ==============
```

jgosmann:
I can't reproduce this locally, but it happen every so often on Travis-CI.

jgosmann:
Might it be better to expicitely invoke `__del__` in this test instead of implicitely relying on the reference counter/garbage collection?

tbekolay:
Huh, I never thought about calling `__del__` explicitly, but it does seem to work! I assume it just gets called again when it's actually garbage collected. I'll make a PR for it after the release (don't think it's important enough to hold it).

jgosmann:
> I assume it just gets called again when it's actually garbage collected.

Yeah, so if we would do some actual cleanup in there it could break, but it doesn't seem to be that we ever want to do that because the proper way is to do that in `close`/`__exit__`.

tbekolay:
> the proper way is to do that in close/__exit__

Good point, which reminds me that I'll call `close` after calling `__del__` just in case!

tbekolay:
Should be fixed in #1091.

jgosmann:
[...]

PR #1077, opened 2016-05-31
Presets

jgosmann:
**Description:**
Adds a thresholding preset as discussed in #1058. Other people are welcome to add their presets.

**Motivation and context:**
See #1058.

**How has this been tested?**
The thresholding preset is being used in the spa-rat model @ikajic and me are working on. I also implemented a unit test.

It's a small PR, shouldn't take long to review.

`nengo/presets.py`, look into `nengo/tests/test_presets.py` for a usage example.

**Types of changes:**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that causes existing functionality to change)

**Checklist:**

- [x] My change requires a change to the documentation
      and I have updated the documentation accordingly.
- [x] My change requires a changelog entry and I have included a changelog entry.

**Still to do:**

- [x] Changelog entry.
- [ ] Link/reference in the documenattion?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,9 @@ Release History
   (`#1127 <https://github.com/nengo/nengo/pull/1127>`_)
 - Added ``ShapeParam`` to store array shapes.
   (`#1045 <https://github.com/nengo/nengo/pull/1045>`_)
+- Added ``ThresholdingPreset`` to configure ensembles for thresholding.
+  (`#1058 <https://github.com/nengo/nengo/issues/1058>`_,
+  `#1077 <https://github.com/nengo/nengo/pull/1077>`_)
 
 **Bug fixes**
 
--- a/nengo/__init__.py
+++ b/nengo/__init__.py
@@ -31,7 +31,7 @@ from .rc import rc, RC_DEFAULTS
 from .simulator import Simulator
 from .synapses import Alpha, LinearFilter, Lowpass, Triangle
 from .utils.logging import log
-from . import dists, exceptions, networks, processes, spa, utils
+from . import dists, exceptions, networks, presets, processes, spa, utils
 
 logger = logging.getLogger(__name__)
 try:
--- a/nengo/presets.py
+++ b/nengo/presets.py
@@ -1,0 +1,40 @@
+"""Configuration presets for common use cases."""
+
+import nengo
+
+
+def ThresholdingEnsembles(threshold):
+    """Configuration preset for a thresholding ensemble.
+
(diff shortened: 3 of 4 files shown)


jgosmann:
I added the WIP label because other people might want to add more presets?
But apart from that my thresholding preset is ready for review.

Seanny123:
I would like to add a preset to make an integrator a ramp. Would that be an appropriate use case for a `preset`?

Seanny123:
What do people think about a `DirectPreset` for setting `neuron_type=nengo.Direct()` and `n_neurons=1`? I use this when debugging my network and when I need to add `spa.State()` populations for visualisation.

jgosmann:
I think `n_neurons` is ignored/irrelevant when setting `neuron_type=nengo.Direct()`. I only realized this myself recently when reading through some parts of the builder. Maybe this should be better documented? Because it is not really obvious. Maybe also something to mention as part of #1119.

jgosmann:
Added a changelog entry which should make this ready for review. It was decided at one of the dev meetings to go ahead with this PR. New PRs can be opened for additional presets.

Only remaining thing might be to add some general documentation? But not sure where.

jgosmann:
Note that the Travis-CI test failure is unrelated to this PR, but I don't want to restart the job to keep the documentation for the corresponding issue #1011.

tbekolay:
> Only remaining thing might be to add some general documentation? But not sure where.

Hm. I think probably we need a new section in the [user guide](http://pythonhosted.org/nengo/user_guide.html) between the modelling API and the networks for user-facing config system documentation and presets. I'd be OK with leaving this for a separate PR!

tbekolay:
This PR LGTM! I added a commit fixing some reST formatting and grammar. In our API docs, we use a default Python role of "object", so the `:class:` part isn't needed (Sphinx figures out that it's a class through introspection).

I also renamed the preset to `ThresholdingEnsembles` ... that can be reverted, but IMO thinking about model scripts, I feel like

```python
with ThresholdingEnsembles(0.5):
    ens = nengo.Ensemble(50, 1)
```

gives more of a clue of what happens than

```python
with ThresholdingPreset(0.5):
    ens = nengo.Ensemble(50, 1)
```

partly because the `Preset` part is not super important, but also because it's not clear from the name that it only affect ensembles and not connections etc. But again, that's my opinion, the second reviewer can arbitrate if needed!

Seanny123:
This still LGTM.

PR #1078, opened 2016-06-03
WIP: Use pydocstyle to check docstrings

tbekolay:
Enable checking docstrings with [pydocstyle](http://pydocstyle.readthedocs.io/en/latest/index.html).

Still a WIP because this PR should also update the docstrings (or the `add-ignore` list) such that this branch passes the pydocstyle test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ install:
       export PIP_DEPS="pytest coverage";
     elif [[ "$STATIC" == "true" ]]; then
       export CONDA_DEPS="";
-      export PIP_DEPS="flake8 pylint";
+      export PIP_DEPS="flake8 pylint pydocstyle";
     fi
   - if [[ -n "$NUMPY" && "$STATIC" == "false" ]]; then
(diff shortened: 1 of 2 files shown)


hunse:
I did an investigation into `pydocstyle`, and it does not seem to be a worthwhile tool to apply regularly as part of the static checks. Most of the errors it catches are spacing errors that have no effect on the output docs, and are not intuitive to someone writing the docstring (e.g. it requires an extra blank line at the end of multiline docstrings that is not part of PEP 257). None of the errors appear worthwhile to check regularly.

I also investigated `pylint`'s docstring checking, which is limited. The main useful feature is checking that the actual function arguments and documented arguments match. Unfortunately, this does not work when the documented arguments have more complex types (e.g. any type with a space in it, like array_like (*, 2)). We use these kinds of types a lot. So we will not use this tool, either.

hunse:
I've opened #1557 to make some of the suggested `pydocstyle` fixups, to make our docstrings a little more uniform.

Issue #1034, opened 2016-04-24
Remove Copy operator


tbekolay:
As it turns out, we only use the `Copy` operator in one place, whereas we use `SlicedCopy` twice. Since `SlicedCopy` can do normal copies as well, perhaps we should remove `Copy` and rename `SlicedCopy` -> `Copy`?

jgosmann:
With respect to #1035 I'm against it because the multiple `Copy` operators can be merged into one, but with `SlicedCopy` this is not as easy because of the slicing.

jgosmann:
I took another look at the `SlicedCopy` code and figured that merging of `SlicedCopy`s is not as complicated as I thought. There are basically three cases:

1. Real slices: These are handled by slicing the signal. This case is already handled in the optimizer code.
2. No slicing/Ellipsis: This is basically a `Copy`, so no problem here too.
3. Slicing with list of indices: Can be handled by joining the lists together (and adjusting the indices accordingly).

So I suppose I'm also in favor of this now ... (though it might be good to test an actual implementation of merging `SlicedCopy`s first to make sure I didn't overlook anything).

jgosmann:
I changed my mind again. The optimizer will be (potentially a lot) faster with two separate classes. If there are *n* `SlicedCopy` and *m* `Copy` it has a complexity of O(*n*² + *m*²) versus a complexity of O((*n* + *m*)²) = O(*n*² + *m*² + 2*nm*). The optimizer will also probably never be able to merge a sliced copy with a non-sliced copy.

hunse:
I don't see how it makes any difference. It seems like just names to me. If we merge things, you can always go through the operators and treat any copy with `src_slice` or `dst_slice` set differently, can't you? All the information you need is still there.

jgosmann:
The code checks (in the worst case¹) for every combination of *a* and *b* where *a* and *b* are from a set of operators whether they can be merged. If `SlicedCopy` and `Copy` get unified, both are in one set. Of course I could go through that set and separate it again in linear time, but it
* complicates the optimizer code which is already quite complex
* adds special case code for just this operator type to the optimizer

¹: I figured that I can sort the operators by the order of the memory access first which allows me to cut the loop short in many cases. So the average runtime is probably much better, but that is something to be tested on some real models.

jgosmann:
The most recent version of the optimizer is doing grouping by the bases of views and thus my concerns might not apply anymore, but this should be tested.

PR #1083, opened 2016-06-09
Combine Copy and SlicedCopy operators

hunse:
**Description:**

Combines the `Copy` and `SlicedCopy` operators. 

**Motivation and context:**

Fixes issue #1034. The operators were very similar, so this makes the code cleaner. Also, some `SlicedCopy` operators had no slices, making them just like `Copy`, and this PR allows those to be optimized better (including in NengoOCL). 

**How has this been tested?**

Ran all the unit tests, and inspected `strings.ipynb` to make sure the new strings look okay.

Look at the new operator. It's quite similar to the old `SlicedCopy` operator, so it should be straightforward. Passed with the default keyword arguments, it should work the same as old `Copy`.

**Types of changes:**

- Breaking change for backends that use our Builder (e.g. NengoOCL)

**Checklist:**

**Still to do:**

- [x] In #1034, @jgosmann had expressed concerns about how this would affect his optimizer. I think overall, this will be beneficial, since previously there were `SlicedCopy` operators that were acting just like `Copy` since they had no slices, but would not get optimized. This will help with that, but it may make his optimizer code slightly more complex.
- [ ] Does this need a changelog entry? We don't have a developer changelog yet, which is where I would put this.


jgosmann:
I volunteer to review this, also to evaluate interactions with #1034.

jgosmann:
Implementation-wise this does not change the optimizer. Build times (including time spent on the optimization) seem to stay about the same. *However*, the estimated run time (on my TCM model) increases from ~50 minutes to over 7 hours (about the same without optimizer)!

In theory, all the same optimizations should still be possible. A not necessarily exhaustive list of possible reasons that these won't work anymore:

* Bug in the optimizer implementation (maybe related to memory sorting).
* The optimizer uses a heuristic to perform optimizations on certain operators early (before other operators). `SlicedCopy` is in this list of operators to optimize early, while `Copy` is not. When there is only one `Copy` operator, either all of those have optimized early or all have to be optimized late (neither way gives a better run time). That might result in operators being optimized and merged in a different order that might prevent further optimizations on other operators using the same signals.
* `SlicedCopy` and `Copy` do a grouping of operators and prevent the merging of a `SlicedCopy` and `Copy` operator. With the change in this PR, this is not given and formerly `SlicedCopy` and `Copy` might get merged in a way preventing further optimization.

jgosmann:
> Also, some SlicedCopy operators had no slices, making them just like Copy

I can't see any of those in the changes of this PR.

hunse:
> > Also, some SlicedCopy operators had no slices, making them just like Copy

> I can't see any of those in the changes of this PR.

I'm not sure what you mean.

To better explain what I meant, currently in `master`, many of the SlicedCopy operators created will have `Ellipsis` for both `src_slice` and `dst_slice`, meaning that the same operation could be done by Copy. By combining them both into one operator, it's easier for e.g. nengo_ocl to check whether the operator is actually doing a copy with indices (which needs a SlicedCopy operator) or just a regular copy, and plan more efficiently because of that.

jgosmann:
I see

jgosmann:
The increased runtime was due to an error in merging this branch with the optimizer branch. I fixed it and the runtimes seem to be the same with and without this PR. The build times with this PR (and .with the optimizer) might be a bit longer, but it's hard to say from a few test due to random variability. It probably won't matter in practice.

Thus, I'm in favour of the general idea of this PR. I will do a review of the actual implementation probably later today (after the review workshop 😄 ).

examples/usage/strings.ipynb
```diff
@@ -263,8 +263,8 @@
       "print(nengo.builder.operator.Reset(sig, tag=\"tag\"))\n",
       "print(nengo.builder.operator.Copy(sig, sig))\n",
       "print(nengo.builder.operator.Copy(sig, sig, tag=\"tag\"))\n",
-      "print(nengo.builder.operator.SlicedCopy(sig, sig, Ellipsis, slice(0, 1)))\n",
-      "print(nengo.builder.operator.SlicedCopy(sig, sig, Ellipsis, slice(0, 1), tag=\"tag\"))\n",
+      "print(nengo.builder.operator.Copy(sig, sig, [0], slice(0, 1)))\n",
+      "print(nengo.builder.operator.Copy(sig, sig, [0], slice(0, 1), tag=\"tag\"))\n",
```
jgosmann: Why did you switch to `[0]` instead of `None`? Because there is no print with a `Copy` operator using integer list indexing?

hunse: Yeah, exactly. I just wanted to see what it looked like, especially since I changed the print function a bit in this PR.

jgosmann:
This LGTM. :cake:
Nominating @tbekolay for 2nd review and merging.

PR #1090, opened 2016-06-11
Fix memory leak in parameter system

### LINKED ISSUE #1089: Memory leak in parameter system.
I think there is a memory leak in the parameter system (I haven't verified it yet). We haven't noticed it yet because parameters tend to be small and at least I usually run only one simulation per process. @ikajic however is running more than one simulation and is experiencing severe memory leaks. Also because in those simulations large SPA vocabularies are stored as part of a parameter (on the bleeding-edge branch using the rewritten SPA system).

First look at [`params.py:76`](https://github.com/nengo/nengo/blob/bleeding-edge/nengo/params.py#L76), the assigned parameters are stored in a `WeakKeyIDDictionary`. Note that the `Params` class is not instantiated per object instance, but for each parameter a such an instance is created when the parameter is defined in a class. Thus, the `Params` instance including the `WeakKeyIDDictionary` lives usually for the whole runtime of the script.

Now take a look at the implementation of [`WeakKeyIDDictionary`](https://github.com/nengo/nengo/blob/bleeding-edge/nengo/utils/stdlib.py#L19). It uses two dictionary-like objects `_keyrefs` and `_keyvalues` to map object ids to the "key" and value respectively. `_keyrefs` is a `WeakValueDictionary`. As such the objects used as key (that is the objects on which parameters are be set) will be garbage collected when no other references exist. `_keyvalues` however is a normal dictionary. That means the value assigned to the parameter will never be garbage collected (and that can be large vocabularies as stated at the beginning).

jgosmann:
**Description:**
Fixes #1089.

We create a weak reference to the key with a callback here that will free the value when no strong references to the key are left. To be able to do this we need the id of the key which will be retrieved
with the _ref2id dictionary. Note that weak reference use the hash function of the object they are referencing. Thus, we have to use the id of the weak reference to index that dictionary. Finally, to support
the del operation we need to be able to map the object id to the weak reference which requires another dictionary _id2ref (instantiating another weak ref would give us an object with a different id). This dictionary has another important function: It stores strong references the weak reference to ensure that the callback gets called (it would not if the weak reference gets garbage collected beforehand and we
need to store ids as keys in the _ref2id dictionary).

**Motivation and context:**
See #1089.

**How has this been tested?**
Added a unit test checking that values in the `WeakKeyIDDictionary` are cleaned up.

Look at #1089 and the test to understand the problem, then look at the changes to `WeakKeyIDDictionary` and information above to see how this is fixed.

**Types of changes:**

**Checklist:**

- [x] I have updated the documentation accordingly. -> no documentation changes required

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -51,6 +51,9 @@ Release History
   the Gram system.
   (`#1027 <https://github.com/nengo/nengo/pull/1027>`_,
   `#1019 <https://github.com/nengo/nengo/issues/1019>`_)
+- Eliminate memory leak in the parameter system.
+  (`#1089 <https://github.com/nengo/nengo/issues/1089>`_,
+  `#1090 <https://github.com/nengo/nengo/pull/1090>`_)
 
 2.1.0 (April 27, 2016)
 ======================
--- a/nengo/utils/stdlib.py
+++ b/nengo/utils/stdlib.py
@@ -26,6 +26,8 @@ class WeakKeyIDDictionary(collections.MutableMapping):
     def __init__(self, *args, **kwargs):
         self._keyrefs = weakref.WeakValueDictionary()
         self._keyvalues = {}
+        self._ref2id = {}
+        self._id2ref = {}
         if len(args) > 0 or len(kwargs) > 0:
             self.update(*args, **kwargs)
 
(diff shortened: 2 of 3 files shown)


tbekolay:
A callback! Well played, Python `weakref` API. This LGTM; nominating @hunse as the other reviewer as he wrote this class in the first place.

hunse:
This code looks fine, but I don't quite understand why this whole thing is necessary. Why were things not getting garbage collected before, and how does this help?

hunse:
Oh, I see, the key was getting freed, but not the value? So we were just storing a bunch of extra values. Maybe put a comment to that effect somewhere, either when you make the callback or at the beginning of the callback.

jgosmann:
> Oh, I see, the key was getting freed, but not the value?

Yes.

> Maybe put a comment to that effect somewhere

Isn't that the reviewer's task now given the PR review guide lines?

hunse:
> Isn't that the reviewer's task now given the PR review guide lines?

Well, technically it's whoever has the time. I had to run earlier so I put it back to you. But I can do it tomorrow.

hunse:
@jgosmann: Does this look good?

jgosmann:
LGTM 🍰

tbekolay:
I'm happy to merge @hunse or you can go ahead; I'd squash it all down to one commit, I think.

hunse:
I'll merge.

Just one comment about commit messages: fd0bb32 has a very long commit message that goes into detail about why two dictionaries are used. This kind of documentation is useful, but I'm not sure if a commit message is the best place for it. If I'm working on that section of code, it's not likely that I'll see that message. I could do a git blame, but often one would need a second- or third-generation blame to get back to a message like that. For example, I added a commit here that fixes the function order, but that's going to make the blame for those important changes one generation older.

hunse:
Also, I did keep the full commit message in the merge, so that comment was more of a general comment than something that we actually need to address here.

tbekolay:
I'd much rather have the documentation in the commit message than in the PR text... there's no downside to long commit messages because most tools will only show the summary line anyway. It's probably true that placing that documentation elsewhere would be good too, but I definitely don't want to discourage anyone from writing long commit messages, as that's definitely something that I read when looking at PRs commit-by-commit.

tbekolay:
But yeah, putting that information in the code itself is often preferable, I will agree on that :) :rainbow: :octopus:

PR #1091, opened 2016-06-11
More robustly generate opensim warning in tests

tbekolay:
**Description:**
This makes the test of warning on open simulators pass 100% of the time instead of 99% of the time.

**Motivation and context:**
Previously, the test would try to force Python to garbage collect, but this doesn't seem to always work. Now we explicitly call `__del__`, which always works. At least it better! Reported in #1011.

**How has this been tested?**
Ran the test a bunch of times... hasn't failed yet!

**Types of changes:**

**Checklist:**

These changes are not user facing, and don't touch any Nengo code, just tests, so they do not require docs or a changelog entry.

### DIFF
--- a/nengo/tests/test_simulator.py
+++ b/nengo/tests/test_simulator.py
@@ -1,5 +1,3 @@
-import gc
-
 import numpy as np
 import pytest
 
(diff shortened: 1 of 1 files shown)


jgosmann:
LGTM :cake:

If `__del__` would be doing more than just producing a warning this might be a problem because it will be called twice (and `sim.close` will be called after `__del__`). But this is just a test which should fail in that case and than we can take measures to fix the test accordingly.

tbekolay:
True true; hopefully we won't have to add anything to `__del__`, just `close`, but I didn't think about the fact that it would be called twice.

drasmuss:
Looks good :+1:, I'll merge

Issue #1080, opened 2016-06-07
2.1.1, 2.2.0 releases


tbekolay:
Making a separate issue to move the discussion out of #1010.

Proposed PRs to include in each release (feel free to suggest others in comments):

**2.1.1**

- [x] #1074
- [x] #1027
- [x] #1052
- [x] #1069 
- [x] #1088 
- [x] #1090 
- [x] #1091

**2.2.0**

- [ ] #1010
- [ ] #984
- [ ] #1045
- [ ] #1016
- [ ] #1107

---

One thing I'm not sure about are the PRs that make relatively large changes to SPA (e.g., #1006, #1014). I think @hunse and I aren't super well equipped to review them, and it doesn't seem quite right to up the minor version of Nengo whenever the SPA API changes (which seems to be happening relatively frequently). We sort of talked about it before, but I don't remember if there was a strong consensus, so I'll raise the issue again: is 2.2 the right time to move SPA to a different repo and Python package?

hunse:
I'd also like #1069 (PR template), #1045 (ShapeParam), #1016 (multiple axis support for some utils), probably all fine in 2.2.0.

hunse:
Also, I'm really in favour of moving SPA to a different repo, and I think 2.2 is a good time to do that since none of the PRs are touching SPA, so it shouldn't be too confusing.

jgosmann:
I think I was in favour of having SPA in core Nengo so far? But I might change my mind on that. Having an external package would us allow to have a legacy SPA package that ensures full backwards compatibility, while we can create a new SPA API without worrying about backwards compatibility. However, there will be a few things where it is not clear whether to put then in core Nengo or the SPA package. That is mainly the spaopt optimization algorithm and the basic ensemble array for unit vectors (that does not need to be tied to any SPA stuff or vocabulary and might be useful for other things to).

jgosmann:
List of PRs looks good. I think, #1069 does not need to be tied to version as it doesn't really affect the code base, but it helpful for all future PRs once it is in master. I prefer to have it merged sooner than later, but not before the remaining open points have been decided.

Another word on the potential Nengo/SPA split: As future SPA (mainly with spaopt) requires some Nengo features (deferrables or the equivalent), it would be nice to have them merged timely to simplify development. It would probably be an easier to develop SPA with just the Nengo master than having both packages on some development branch (that might both change).

Seanny123:
[...]

PR #1095, opened 2016-06-15
Support any synapse for learning rules

arvoelke:
**Description:**
- Replaces `pre_tau`, `post_tau`, and `theta_tau` with `pre_synapse`, `post_synapse`, and `theta_synapse` respectively on all learning rules. This supports the use of arbitrary synapse models on the activities used for learning.
- Makes the `learning_rate` parameter always first in the list for consistency.
- Fixes `theta_tau` to only do one level of filtering (used to be filtered twice).

**Motivation and context:**
We might not always want to filter the activities with a `Lowpass`. This allows arbitrary filters for learning (e.g. `Alpha`). I found this useful for RL in a dynamic context (the filter defines a kernel, which determines the state to update with respect to the reward). This also incidentally fixes a bug (missing issue) where we couldn't pass `pre_tau=None` or `pre_tau=0` as a parameter.

**How has this been tested?**
A test was added for PES using `pre_synapse=Alpha(...)`, which checks that the probed activities from the learning rule are identical to the alpha-filtered neural activities.  The other learning rules use the same code paths (i.e. `build_or_passthrough`), that are also tested with `Lowpass` synapses.

Start in `learning_rules.py`, in particular the `PES` class. Then move on to `builder/learning_rules.py`.

**Types of changes:**

**Checklist:**

**Still to do:**
- [x] Add this URL to `ObsoleteParam`.
- [x] Add tests.
- [x] Add documenting example.
- [x] Apply changes to PES.
- [x] Apply changes to BCM.
- [x] Apply changes to Oja.
- [x] Apply changes to Voja.

### DIFF
--- a/examples/usage/strings.ipynb
+++ b/examples/usage/strings.ipynb
@@ -148,14 +148,14 @@
      "collapsed": false,
      "input": [
       "print(nengo.PES())\n",
-      "print(nengo.PES(pre_tau=0.01, learning_rate=1e-6))\n",
+      "print(nengo.PES(learning_rate=1e-6, pre_synapse=0.01))\n",
       "print(nengo.BCM())\n",
       "print(nengo.BCM(\n",
-      "    pre_tau=0.01, post_tau=0.005, theta_tau=10.0, learning_rate=1e-8))\n",
+      "    learning_rate=1e-8, pre_synapse=0.01, post_synapse=0.005, theta_synapse=10.0))\n",
       "print(nengo.Oja())\n",
-      "print(nengo.Oja(pre_tau=0.01, post_tau=0.005, beta=0.5, learning_rate=1e-5))\n",
+      "print(nengo.Oja(learning_rate=1e-5, pre_synapse=0.01, post_synapse=0.005, beta=0.5))\n",
       "print(nengo.Voja())\n",
-      "print(nengo.Voja(post_tau=None, learning_rate=1e-5))"
+      "print(nengo.Voja(learning_rate=1e-5, post_synapse=None))"
      ],
      "language": "python",
      "metadata": {},
--- a/nengo/learning_rules.py
+++ b/nengo/learning_rules.py
@@ -1,7 +1,10 @@
 import warnings
 
+from nengo.config import SupportDefaultsMixin
 from nengo.exceptions import ValidationError
-from nengo.params import IntParam, FrozenObject, NumberParam, Parameter
+from nengo.params import (Default, IntParam, FrozenObject, NumberParam,
+                          Parameter, Unconfigurable)
+from nengo.synapses import Lowpass, SynapseParam
 from nengo.utils.compat import is_iterable, is_string, itervalues
 
 
@@ -20,7 +23,7 @@ class LearningRuleTypeSizeInParam(IntParam):
                 instance, size_in)  # IntParam validation
 
 
-class LearningRuleType(FrozenObject):
+class LearningRuleType(FrozenObject, SupportDefaultsMixin):
     """Base class for all learning rule objects.
 
     To use a learning rule, pass it as a ``learning_rule_type`` keyword
@@ -70,21 +73,47 @@ class LearningRuleType(FrozenObject):
     modifies = None
     probeable = ()
 
-    learning_rate = NumberParam('learning_rate', low=0)
+    learning_rate = NumberParam(
+        'learning_rate', low=0, readonly=True, default=1e-6)
     size_in = LearningRuleTypeSizeInParam('size_in', low=0)
 
-    def __init__(self, learning_rate=1e-6, size_in=0):
+    def __init__(self, learning_rate=Default, size_in=0):
         super(LearningRuleType, self).__init__()
         self.learning_rate = learning_rate
         self.size_in = size_in
 
     def __repr__(self):
-        return '%s(%s)' % (type(self).__name__, ", ".join(self._argreprs))
+        r = []
+        for name, default in self._argdefaults:
+            value = getattr(self, name)
+            if value != default:
+                r.append("%s=%r" % (name, value))
+        return '%s(%s)' % (type(self).__name__, ", ".join(r))
 
     @property
-    def _argreprs(self):
-        return (["learning_rate=%g" % self.learning_rate]
-                if self.learning_rate != 1e-6 else [])
+    def _argdefaults(self):
+        return ('learning_rate', LearningRuleType.learning_rate.default),
+
+
+def _deprecated_tau(old_attr, new_attr):
+    def get_tau(self):
+        return (None if getattr(self, new_attr) is None else
+                getattr(self, new_attr).tau)
+
+    def set_tau(self, val):
+        if val is Unconfigurable:
+            return
+
+        since = "v2.8.0"
+        url = "https://github.com/nengo/nengo/pull/1095"
+        msg = ("%s has been deprecated, use %s instead (since %s).\n"
+               "For more information, please visit %s" % (
+                   old_attr, new_attr, since, url))
+        warnings.warn(msg, DeprecationWarning)
+
+        setattr(self, new_attr, None if val is None else Lowpass(val))
+
+    return property(get_tau, set_tau)
 
 
 class PES(LearningRuleType):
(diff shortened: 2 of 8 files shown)


tbekolay:
Awesome, this is definitely something I've wanted to do for a while!

nengo/params.py
```diff
@@ -163,6 +163,12 @@ def validate(self, instance, value):
     def raise_error(self):
         raise ObsoleteError(self.short_msg, since=self.since, url=self.url)
 
+    def hashvalue(self, instance):
+        return Unconfigurable  # otherwise __get__ throws an error
+
+    def equal(self, instance_a, instance_b):
+        return True  # otherwise __get__ throws an error
+
```
hunse: Just wondering out loud whether there's a better way to do `__get__` that avoids having to do this. Also, I would bust these out into a separate commit since they're not directly related to the main change.

hunse:
Yeah, this is a great thing to do.

arvoelke:
Updated the BCM/Oja/Voja rules, changelog, made docstrings consistent, made use of the `SupportDefaultsMixin` for all default params, and made the parameter order consistent. Still to do: tests, and an example using the alpha synapse.

@tbekolay Why was this removed from the 2.2.0 release milestone?

nengo/params.py
arvoelke: Not sure... but moved these changes to a separate commit.

arvoelke:
Also if we are happy with the way this uses the param system, then the same changes should be made to `synapses.py` and `processes.py`. In particular, the `Process` should inherit `SupportDefaultsMixin` to fully harness the param system and its defaults. Does this sound good?

nengo/learning_rules.py
```diff
@@ -83,37 +87,50 @@ class PES(LearningRuleType):
...
+
         self.pre_tau = pre_tau
-        super(PES, self).__init__(learning_rate)
 
     @property
     def _argreprs(self):
         args = []
         if self.learning_rate != 1e-4:
             args.append("learning_rate=%g" % self.learning_rate)
-        if self.pre_tau != 0.005:
```
arvoelke: How can I access a param's default?

nengo/learning_rules.py
```diff
@@ -83,37 +87,50 @@ class PES(LearningRuleType):
...
-            args.append("pre_tau=%f" % self.pre_tau)
+        if self.pre_synapse != Lowpass(tau=0.005):
```
arvoelke: How can I access a param's default?

tbekolay: `PES.pre_synapse.default`

tbekolay:
> @tbekolay Why was this removed from the 2.2.0 release milestone?

See the discussion in #1080, which led to #1114.

arvoelke:
Okay this is now ready for review.

To check that my improvements to `__repr__` did not mess anything up, I verified that this snippet from `examples/usage/strings.ipynb`...
```
print(nengo.PES())
print(nengo.PES(learning_rate=1e-6, pre_synapse=0.01))
print(nengo.BCM())
print(nengo.BCM(learning_rate=1e-8, pre_synapse=0.01, post_synapse=0.005, theta_synapse=10.0))
print(nengo.Oja())
print(nengo.Oja(learning_rate=1e-5, pre_synapse=0.01, post_synapse=0.005, beta=0.5))
```

... displays the right thing:
```
PES()
PES(learning_rate=1e-06, pre_synapse=Lowpass(0.01))
BCM()
BCM(learning_rate=1e-08, pre_synapse=Lowpass(0.01), post_synapse=Lowpass(0.005), theta_synapse=Lowpass(10.0))
Oja()
Oja(learning_rate=1e-05, pre_synapse=Lowpass(0.01), post_synapse=Lowpass(0.005), beta=0.5)
```

Perhaps that whole notebook should be made into a series of unit tests?

Also an important note:
 - The `theta` signal for `BCM` used to be filtered twice. I don't think we want this. It's now filtered once. If you want it filtered twice, you can use a double-exponential synapse. This seemed too subtle to include in the changelog?

tbekolay:
I haven't done a detailed review of this yet, but one discussion point from a quick look at the diff is that the API changes in this PR are backwards-incompatible, meaning that we'd need to do a 3.0.0 release when this is merged. Since we allow numbers as a shortcut for `Lowpass(tau=number)`, what do you think about making `pre_tau` etc deprecated parameters that just set `pre_synapse` etc for now, but print a warning and will be removed in 3.0.0?

arvoelke:
Maybe we can save this for 3.0.0 and avoid the duplication of parameters altogether? For me it's more pain than worth to maintain these two code paths.

tbekolay:
The deprecated parameters can be like this:

```python
@property
def pre_tau(self):
    warnings.warn("Deprecated since 2.2.0; use pre_synapse instead")
    return self.pre_synapse

@pre_tau
def pre_tau(self, val):
    warnings.warn("Deprecated since 2.2.0; use pre_synapse instead")
    self.pre_synapse = val
```

But either way, up to you!

arvoelke:
Updated into a form ready for 3.0.0. Also looping in @neworderofjamie since this will affect SpiNNaker if it gets in.

examples/usage/strings.ipynb
```diff
@@ -147,11 +147,11 @@
...
      "input": [
       "print(nengo.PES())\n",
-      "print(nengo.PES(pre_tau=0.01, learning_rate=1e-6))\n",
+      "print(nengo.PES(learning_rate=1e-6, pre_synapse=0.01))\n",
       "print(nengo.BCM())\n",
-      "print(nengo.BCM(pre_tau=0.01, post_tau=0.005, theta_tau=10.0, learning_rate=1e-8))\n",
+      "print(nengo.BCM(learning_rate=1e-8, pre_synapse=0.01, post_synapse=0.005, theta_synapse=10.0))\n",
       "print(nengo.Oja())\n",
-      "print(nengo.Oja(pre_tau=0.01, post_tau=0.005, beta=0.5, learning_rate=1e-5))"
+      "print(nengo.Oja(learning_rate=1e-5, pre_synapse=0.01, post_synapse=0.005, beta=0.5))"
```
arvoelke: TODO: Add one for `Voja`? Also consider making these into unit tests?

neworderofjamie:
Thanks @arvoelke - I don't think we're actually supporting any kind of filtering in our PES or  Voja implementations at the moment - so perhaps we should implement that in a more generic way using @mundya 's code for the synapses

tcstewar:
So, @arvoelke just let me know this branch exists, and it does exactly what @bjkomer and I were wanting with #1256 .  We'll try it out!

Also, if it is possible for this to be done in a backwards-compatible manner with a few deprecation warnings (and no large code duplication), then it would be great to do so, even if we do put off merging it in till 3.0.0.   I'll take a look and see if @tbekolay 's suggestion of how do to it could end up being reasonable....

arvoelke:
If you go to any of the previous commits (other than the last, e.g., e41f7e3ef27ac220b179e73812e5dfb2e56444c4) then it will be  backwards-compatible. @tbeokolay's suggestion would just make it a bit less muddy, but still the same in principle.

drasmuss:
Rebased this to master and fixed things up.  I also made it so that this change is backwards compatible (as suggested by @tbekolay), so no need to wait for a major release.  Should be ready for review.

nengo/learning_rules.py
```diff
@@ -149,54 +187,80 @@ class BCM(LearningRuleType):
...
-        if self.post_tau != self.pre_tau:
-            args.append("post_tau=%g" % self.post_tau)
-        if self.theta_tau != 1.0:
-            args.append("theta_tau=%g" % self.theta_tau)
-        if self.learning_rate != 1e-9:
-            args.append("learning_rate=%g" % self.learning_rate)
-        return args
+        return (('learning_rate', BCM.learning_rate.default),
+                ('pre_synapse', BCM.pre_synapse.default),
+                ('post_synapse', self.pre_synapse),
```
Seanny123: I think this is a typo and should be ` ('post_synapse', BCM.post_synapse.default)` at least for the sake of consistency?

drasmuss: The `post_synapse` defaults to the value of `pre_synapse`, so that's the value we want to check against (to see if the user set it to something other than the default).

nengo/learning_rules.py
```diff
@@ -20,7 +23,7 @@ def coerce(self, instance, size_in):
                 instance, size_in)  # IntParam validation
 
 
-class LearningRuleType(FrozenObject):
```
hunse: So the idea behind FrozenObjects is that they're not supposed to support Defaults. I remember the idea behind this was so that FrozenObjects could be used in multiple places (e.g. the same learning rule type on multiple connections) or copied without issue. But I can't remember the exact details as to why having Defaults and being Frozen could not go together.

Part of the reason might be because FrozenObjects themselves are often used as defaults for other things. For example, the default for ``Ensemble.neuron_type`` is ``LIF()``. If ``LIF`` had its own defaults for ``tau_rc`` and ``tau_ref``, then changing ``model.config[nengo.LIF].tau_rc`` could change the default tau_rc for an ensemble, but also change that default other places, too. Things could get complicated or have unintended consequences.

For learning rules, that's less of an issue, since they're currently not used as defaults anywhere that I know of. My main point is just that I had designed FrozenObject to be separate from the Default system, so if we're bringing them together sometimes, I just want to make sure that that makes sense and doesn't cause any unforeseen problems.

drasmuss: That's the reason all the parameters are marked as `readonly`, so that they can be used with `FrozenObject` (this is explicitly checked for in [FrozenObject](https://github.com/nengo/nengo/blob/master/nengo/params.py#L498), so I think it is part of the intended/expected behaviour).

arvoelke:
Could you please include a link to http://compneuro.uwaterloo.ca/publications/voelker2017c.html alongside the other tech report that is currently linked in the PES notebook? This gives useful information for understanding how different synapses on the error connection will affect the dynamics, and a heuristic for setting the learning rate. Thanks!

tbekolay:
> Could you please include a link to http://compneuro.uwaterloo.ca/publications/voelker2017c.html alongside the other tech report that is currently linked in the PES notebook?

Absolutely, will do :+1:

Issue #1036, opened 2016-04-26
Make piecewise a Process


tbekolay:
As discussed in #955:

@drasmuss 

> Unrelated, but this example just made me think that `nengo.utils.functions.piecewise` and `PresentInput` are doing very similar things. If we made `PresentInput` a little bit more general so that it could have different presentation times, we could merge them into one.

@hunse 

> It would be great to have a process for `utils.functions.piecewise`. I'd keep it separate from `PresentInput`, though, or at least have some version of it with fixed times (maybe if you pass a constant it could have fixed times, or a list of times it could have varying?). It's just easier to do in OCL if the times are fixed.
> 
> Also with piecewise it would be great to have linear interpolation, too.

PR #1100, opened 2016-06-17
Add Piecewise process

hunse:
**Description:**
Turns the `piecewise` function into a Process. This adds all the normal amenities, such as running the process for a length of time outside a Nengo simulator.

**Motivation and context:**

Addresses issue #1036. It should make this easier to use.

I'm also adding interpolation, so that we can do e.g. linear interpolation rather than just the current zero-order hold behaviour.

If we really want to go the whole nine yards, we could actually allow this to take other Processes in addition to callables and constants, so that we can e.g. run one process for 5 seconds, then another for the next 3 seconds, etc.

**How has this been tested?**

Still TODO. I'm planning on pulling over all the tests from the old `piecewise` function, so we maintain all that behaviour.

Look at the new process first, and read the documentation (still TODO). 

**still unknown**

**Types of changes:**

**Checklist:**

**Still to do:**

- [ ] Support callables as outputs
- [ ] Support Processes as outputs if we want to do that (would it be useful?)
- [x] Update old tests so they work with the new Process
- [x] Test interpolation
- [ ] Update examples that use old `piecewise`
- [x] Remove old `piecewise` function and tests, or perhaps just add a deprecation warning.


AllenHW:
After discussing with @hunse, we decided to support callable in another PR. To do it we would require another parameter type that lets user pass an array of callable AND floats/list of floats

nengo/processes.py
```diff
@@ -260,3 +260,121 @@ def step_presentinput(t):
...
+        if self.tp.shape[0] != self.yp.shape[0]:
+            raise ValidationError(
+                "`tp.shape[0]` (%d) must equal `yp.shape[0]` (%d)"
+                % (self.tp.shape[0], self.yp.shape[0]),
+                attr='yp', obj=self)
+
+        if self.interpolation != 'zero':
+            if len(self.yp[0].shape) > 0 and self.yp[0].shape[0] != 1:
+                raise ValidationError(
+                    "Interpolation is only supported for 1d data",
```
tbekolay: Why impose this restriction? I would assume that for multidimensional data, we are always doing interpolation over time, so each dimension would be interpolated separately (which still uses `interp1d`)

nengo/processes.py
```diff
@@ -260,3 +260,121 @@ def step_presentinput(t):
...
+                    attr='interpolation', obj=self)
+            try:
+                import scipy.interpolate
+                self.sp_interpolate = scipy.interpolate
+            except ImportError:
+                    raise ValidationError(
```
tbekolay: This block is overindented by 4 spaces; it should line up with the `try` block above.

tbekolay: Also, I think this should raise an `ImportError` rather than a `ValidationError`.

I am also wondering if, rather than raising an error, we should instead raise a warning and set `interpolate` to `'zero'` instead of crashing. What do you think?

nengo/tests/test_processes.py
```diff
@@ -334,3 +334,108 @@ def test_present_input(Simulator, rng):
     y = sim.data[up].reshape(len(t), c, ni, nj)
     for k, [ii, image] in enumerate(zip(i, y)):
         assert np.allclose(image, images[ii], rtol=1e-4, atol=1e-7), (k, ii)
+
+
+class TestPiecewise(object):
+
+    dt = 0.001
```
tbekolay: Unless there's some reason to change the `dt` in a specific test, it's probably best not to specify it, as specifying it indicates to the reader that we're going to be using a non-standard `dt` in some tests (which I don't believe we are here).

nengo/tests/test_processes.py
```diff
@@ -334,3 +334,108 @@ def test_present_input(Simulator, rng):
...
+            process = nengo.processes.Piecewise(data, 'not-interpolation')
+            assert process
+
+    def test_invalid_interpolation_dimention(self):
+        data = {0.5: [1, 0], 1.0: [0, 1]}
+        with pytest.raises(ValidationError):
+            process = nengo.processes.Piecewise(data, 'cubic')
+            assert process
+
+    def test_interpolation(self, Simulator):
```
tbekolay: At the moment, this test requires scipy, which is not a required dependency of Nengo. In order to not fail unit tests when scipy is not installed, please use `pytest.importorskip` (see [test_dists.py](https://github.com/nengo/nengo/blob/master/nengo/tests/test_dists.py#L176) for an example).

nengo/utils/functions.py
```diff
@@ -3,12 +3,17 @@
...
+import warnings
 
 from nengo.exceptions import ValidationError
 from nengo.utils.compat import is_number, iteritems
 
 
+
 def piecewise(data):
+    warnings.warn("piecewise() function is deprecated. Use Piecewise process instead.",
+                      DeprecationWarning)
```
tbekolay: This line is overindented, according to PEP8. It should line up with the `p` in `piecewise`.

nengo/utils/tests/test_functions_piecewise.py
```diff
@@ -4,6 +4,10 @@
 from nengo.exceptions import ValidationError
 from nengo.utils.functions import piecewise
 
+"""
+note:: piecewise() function is deprecated. Use Piecewise process instead.
+"""
```
tbekolay: This note isn't necessary, as it's very unlikely anyone will be touching this test file. We raise the deprecation warning in `utils/functions.py` already, which is something that will be seen when running these tests anyhow.

hunse:
I looked it over, and it all looks great to me! Thanks @AllenHW!

I added a little fixup to use ``interpolation`` as a keyword argument only, just because I'm nit-picky like that sometimes.

nengo/processes.py
```diff
@@ -260,3 +262,146 @@ def step_presentinput(t):
...
+    def __init__(self, data, interpolation='zero', **kwargs):
+        tp, yp = zip(*data.items())
+        self.tp = tp
+        self.yp = yp
+
+        if self.tp.shape[0] != self.yp.shape[0]:
+            raise ValidationError(
+                "`tp.shape[0]` (%d) must equal `yp.shape[0]` (%d)"
+                % (self.tp.shape[0], self.yp.shape[0]),
+                attr='yp', obj=self)
```
drasmuss: There is no way this ValidationError could ever trigger (since a dictionary always has the same number of keys and values).

tbekolay:
Those changes look good to me @drasmuss! Can you approve this so I can merge it in?

Issue #1097, opened 2016-06-16
Graceful fail when cache index cannot be loaded


jgosmann:
Nengo raises an exception when the cache index cannot be read, but should handle this case graceful.

Here is a backtrace from a particular case where an incomplete index was written due to interrupting the process:
```
---------------------------------------------------------------------------
EOFError                                  Traceback (most recent call last)
<ipython-input-3-0f9dc08c14fc> in <module>()
----> 1 with nengo.Simulator(model) as sim:
      2     pass

/home/jgosmann/Documents/projects/nengo/nengo/simulator.pyc in __init__(self, network, dt, seed, model, shrink_cache)
    130             cache = model.decoder_cache
    131 
--> 132         with cache:
    133             if model is None:
    134                 dt = float(dt)  # make sure it's a float (for division

/home/jgosmann/Documents/projects/nengo/nengo/cache.pyc in __enter__(self)
    199             self._remove_legacy_files()
    200             self._index = CacheIndex(os.path.join(self.cache_dir, self._INDEX))
--> 201             self._index.__enter__()
    202         except TimeoutError:
    203             warnings.warn(

/home/jgosmann/Documents/projects/nengo/nengo/cache.pyc in __enter__(self)
    108     def __enter__(self):
    109         with self._lock:
--> 110             self._index = self._load_index()
    111         return self
    112 

/home/jgosmann/Documents/projects/nengo/nengo/cache.pyc in _load_index(self)
    118         try:
    119             with open(self.filename, 'rb') as f:
--> 120                 return pickle.load(f)
    121         except IOError as err:
    122             if err.errno == errno.ENOENT:

EOFError:
```

Besides handling an error on read graceful, it might also be good to do the writing in a more atomic way (e.g. write to a different file first and than rename the file; keep in mind that the files have to be on the same file system for this to be atomic and work without copying data).

tbekolay:
This is currently in the 2.1.1 milestone. How close are you to a fix @jgosmann? Do you want me to attempt to fix this or should we punt to 2.2.0?

jgosmann:
I haven't looked at this at all yet. It would probably be nice to at least fail gracefully and recreate the index if it can't be loaded in 2.1.1 (or have a 2.1.2 release). Writing the index in a safer way could probably put of to 2.2.0, but might actually be not that complicated to implement.

jgosmann:
If you want to take a stab at it, go ahead.

tbekolay:
[...]

PR #1107, opened 2016-06-23
Make writing of cache index atomic.

jgosmann:
**Description:**
 
Make writing of cache index atomic.

If the process is killed while writing the cache index pickle, we can't load it anymore and thus lose all index information. Thus, we write it to another file first, so we still have the old index in case the process gets killed. Once we're done writing we rename the file (which should be atomic on most file systems and operating systems, but will at least be a much faster operation reducing the likelihood of ending up with an invalid index).

**Motivation and context:**

Adresses #1097.

Besides this we should make sure that the cache fails gracefully if the index cannot be read.

**How has this been tested?**

Tests still pass. The atomicity of the operation is hard to test as it is an implementation detail.

Should be obvious, it's just a two line change ...

**Types of changes:**

**Checklist:**

- [x] I have updated the documentation accordingly. (No changes necessary.)

**Still to do:**

If the process is killed while writing the pickle, we can't load it
anymore and thus lose all index information. Thus, we write it to
another file first, so we still have the old index in case the process
gets killed. Once we're done writing we rename the file (which should
be atomic on most file systems and operating systems, but will at least
be a much faster operation reducing the likelihood of ending up with
an invalid index).

Addresses #1097.

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -16,7 +16,7 @@ from nengo.exceptions import FingerprintError, TimeoutError
 from nengo.rc import rc
 from nengo.utils import nco
 from nengo.utils.cache import byte_align, bytes2human, human2bytes
-from nengo.utils.compat import is_string, pickle, PY2
+from nengo.utils.compat import is_string, pickle, replace, PY2
 from nengo.utils.lock import FileLock
 
 logger = logging.getLogger(__name__)
@@ -133,11 +133,12 @@ class CacheIndex(object):
                 for key in self._deletes:
                     del self._index[key]
 
-                with open(self.filename, 'wb') as f:
+                with open(self.filename + '.part', 'wb') as f:
                     pickle.dump(
                         {k: v for k, v in self._index.items()
                          if v[0] not in self._removed_files},
                         f, pickle.HIGHEST_PROTOCOL)
+                replace(self.filename + '.part', self.filename)
         except TimeoutError:
             warnings.warn(
                 "Decoder cache index could not acquire lock. "
--- a/nengo/utils/compat.py
+++ b/nengo/utils/compat.py
@@ -1,6 +1,8 @@
 from __future__ import absolute_import
 
 import collections
+import os
+import subprocess
 import sys
 
 import numpy as np
@@ -30,6 +32,28 @@ if PY2:
         assert isinstance(s, bytes)
         return s
 
+    if sys.platform.startswith('win'):
+
+        def replace(src, dst):
+            # The Windows implementation of replace calls out to the shell
+            # to do 'move /Y src dst' due to an odd bug in 32-bit versions
+            # of Python 2.7. See https://github.com/nengo/nengo/pull/1107
+            # for the bizarre details.
+            with open(os.devnull, 'w') as devnull:
+                subprocess.check_call(["move", "/Y", src, dst],
+                                      shell=True,
+                                      stdout=devnull,
+                                      stderr=devnull)
+
+    else:
+
+        def replace(src, dst):
+            try:
+                os.rename(src, dst)
+            except OSError:
+                os.remove(dst)
+                os.rename(src, dst)
+
     class TextIO(StringIO):
         def write(self, data):
             if not isinstance(data, unicode):
@@ -53,6 +77,7 @@ else:
     import pickle
     import configparser
     from io import StringIO
+    from os import replace
     TextIO = StringIO
     string_types = (str,)
     int_types = (int,)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,10 @@ Release History
 
 **Bug fixes**
 
+- Fixed some situations where the cache index becomes corrupt by
+  writing the updated cache index atomically (in most cases).
+  (`#1097 <https://github.com/nengo/nengo/issues/1097>`_,
+  `#1107 <https://github.com/nengo/nengo/pull/1107>`_)
 - The synapse methods ``filt`` and ``filtfilt`` now support lists as input.
   (`#1123 <https://github.com/nengo/nengo/pull/1123>`_)
 
--- a/nengo/utils/lock.py
+++ b/nengo/utils/lock.py
@@ -33,7 +33,7 @@ class FileLock(object):
     def release(self):
         if self._fd is not None:
             os.close(self._fd)
-            os.unlink(self.filename)
+            os.remove(self.filename)
             self._fd = None
 
     @property


nengo/cache.py
```diff
@@ -133,11 +133,12 @@ def sync(self):
                 for key in self._deletes:
                     del self._index[key]
 
-                with open(self.filename, 'wb') as f:
+                with open(self.filename + '.part', 'wb') as f:
                     pickle.dump(
                         {k: v for k, v in self._index.items()
                          if v[0] not in self._removed_files},
                         f, pickle.HIGHEST_PROTOCOL)
+                os.rename(self.filename + '.part', self.filename)
```
hunse: Is `self.filename` guaranteed not to exist? That's the only flaw I can see in this implementation. In Unix, the rename will overwrite, but on Windows it won't ([reference](http://stackoverflow.com/questions/8107352/force-overwrite-in-os-rename)).

jgosmann: Can we finally drop support for Windows? It's so annoying. 😠

jgosmann: I made a fixup commit to handle this correctly on Windows.

nengo/utils/compat.py
```diff
@@ -30,6 +32,13 @@ def ensure_bytes(s):
         assert isinstance(s, bytes)
         return s
 
+    def replace(src, dst):
```
tbekolay: Why not use `shutil.move`? It's part of the standard library and does something similar but perhaps more robust...

jgosmann: Because `shutil.move` [depends on `os.rename` semantics](https://docs.python.org/2/library/shutil.html#shutil.move) and thus does not solve any of the problems. `os.replace` provides the correct semantics, but only exists in Python 3.3 and newer.

nengo/utils/compat.py
```diff
@@ -30,6 +32,13 @@ def ensure_bytes(s):
...
+        try:
+            os.rename(src, dst)
+        except OSError:  # Will be raised on Windows if dst exists
+            os.unlink(dst)
```
hunse: Since `os.rename` is identical, and a much clearer name IMO, can we use that?

jgosmann: It is not identical depending on the OS. [`os.replace`](https://docs.python.org/3/library/os.html#os.replace) is identical and the name implies that an existing file might be overwritten.

nengo/utils/compat.py
hunse: Yeah, looking at the `shutil.move` source, it actually does a copy in the case of OSError on rename, and then deletes the source. The copy function they use will actually overwrite I think, but it seems like some extra work to copy the file. So I'd probably just go with this, unless there's a clear way in which `shutil.move` is more robust.

nengo/utils/compat.py
tbekolay: I think he meant `os.remove`.

hunse: The [docs say it's identical](https://docs.python.org/2/library/os.html#os.remove).

jgosmann: I don't care either way. I'm used to the Unix/Posix API which is why I tend to use `unlink`.

hunse: Oh, sorry, I misspoke in my original comment. I meant we should use `os.remove` instead of `os.unlink`.

tbekolay:
For whatever reason, I tracked down the commit in CPython that added `os.replace`: https://github.com/python/cpython/commit/c64442f7ec661e08c6c8d8351ce38eb8be893ab6

It's in C so I'm satisfied with reimplementing this in Python 2 and using it in Python 3.

jgosmann:
Does this need a changelog entry?

hunse:
I would add one as a bugfix, yes.

hunse:
Actually, if we're not adding a changelog entry for #1110, then we probably don't need one here either. So I think it's fine as-is, except for that one test that's failing.

Also, before we merge this, I think we should rebase and re-run all the tests (though don't we always...) since there might be interactions with #1110.

jgosmann:
Is there a way to restart the AppVeyor build?

tbekolay:
Restarted; if you log in with your Github account you should be able to link to the nengo dev user which can restart builds... I think?

Seanny123:
After extensive searching I have determined that it is not possible to restart an appveyor branch. At least not from the type of account Jan and I have. Consequently, should this be rebased to see if the tests pass? And then we can continue with the review?

jgosmann:
Rebase has to be done sooner or later anyways. @tbekolay you're assigned to this, do you want to do the rebase? Or do you want to reassign this to me and let me handle it?

tbekolay:
I'm looking at this now, so I'll do the rebase if you don't mind.

tbekolay:
Rebased, but kept the fixup and squash commits for the time being. I reworded the changelog message in 03d91d7c9b4a161f212942 (was easier to do in the rebase, sorry). The appveyor build still [fails in Python 2](https://ci.appveyor.com/project/nengo/nengo/build/1.0.574/job/3yp6j584deow5rhf) (but succeeds in [Python 3](https://ci.appveyor.com/project/nengo/nengo/build/1.0.574/job/t2k32k4tk6fh7qp7)).

jgosmann:
Can we drop Windows support? :rage:

tbekolay:
> Can we drop Windows support?

[Nope](https://www.netmarketshare.com/operating-system-market-share.aspx?qprid=10&qpcustomd=0&qpf=1)!

tbekolay:
@jgosmann I can fire up Windows to try to debug this if you want?

jgosmann:
Sure.

tbekolay:
Alright... that was legit one of the most annoying bugs I've attempted to fix in some time. But, I think I came up with a solution (ad4ae7d14e9038e79e66). Works locally, and it seems to work in [appveyor too](https://ci.appveyor.com/project/nengo/nengo/build/1.0.582/job/wqkgruvndkrx3y7j)!

Explanation of the bug and fix:

In 32-bit Python 2.7 (but not 2.7 64-bit, and not 32-bit 3.5) there was an issue when doing `replace(src='index.part', dst='index')` (but the absolute path versions of those). When trying to the `replace` in 32-bit 2.7, this is what I was getting while debugging:

```python
>>> root = os.path.dirname(dst)
>>> os.listdir(root)
['d6', 'index', 'index.part', 'legacy.txt']
>>> os.path.isfile(src)
False
>>> os.path.isfile(dst)
False
>>> src == os.path.join(root, 'index.part')
True
>>> dst == os.path.join(root, 'index')
True
```
So yeah, `os.listdir` says that those files are there. `os.path.isfile` says that those files are _not_ there. Other functions like `os.stat` and `os.remove` use the same machinery as `os.path.isfile`, so they were failing to remove the files that did indeed exist (as verified in the terminal; you could `cat` the contents and whatnot, so they were legitimately there).

I have a guess as to the underlying issue, but I haven't verified it... at first, I thought it might be the Windows 260 character path limit, but quickly dismissed that as the paths were only around 100 characters. A bit of searching led me to the fact that Windows will sometimes give you a different path than you actually requested. The most common example is that if you try to access `C:\Windows\SysWOW64` in a 32-bit process, it will give you `C:\Windows\System32` (and vice versa in 64-bit) whether you like it or not. I think that a similar thing is happening with getting temporary directory, as it might be the case that in 64-bit Windows you get a different path for `%TEMP%` as in 32-bits, and Windows will automatically (and without telling you) switch the paths if you try to get them.

That's just a guess though, I don't know if that's actually the case. What I do know is that it works in the terminal, so while debugging I tried calling out to the shell to do the dirty work for me.

```python
>>> import subprocess
>>> subprocess.call(["dir", root])
 Volume in drive C has no label.                                
 Volume Serial Number is 45F3-4A7A                              
                                                                
 Directory of C:\tmp\pytest-of-tbekolay\pytest-7\test_dec_cache 
                                                                
07/27/2016  11:13 PM    <DIR>          .                        
07/27/2016  11:13 PM    <DIR>          ..                       
07/27/2016  11:13 PM    <DIR>          24                       
07/27/2016  11:13 PM    <DIR>          44                       
07/27/2016  11:13 PM               151 index                    
07/27/2016  11:13 PM               151 index.part              
07/27/2016  11:12 PM                 5 legacy.txt               
               3 File(s)            307 bytes                   
               4 Dir(s)  19,363,864,576 bytes free              
>>> subprocess.call(["move", "/Y", src, dst])
        1 file(s) moved.
```

So I cleaned that up a bit, and made `replace` call out to the shell in Windows in Python 2. Which seems to work!

jgosmann:
😩 wtf

Seanny123:
Would this be worth rporting as a bug to someone?

tbekolay:
> Would this be worth rporting as a bug to someone?

Since it works in 32-bit Python 3.5 it is likely that a similar bug has been reported and fixed. Also figuring out the root cause will probably take a while... though yeah, I guess Python 2.7 is still supported for the time being, so if we could find the root cause it would be worth reporting.

nengo/utils/compat.py
```diff
@@ -1,6 +1,9 @@
 from __future__ import absolute_import
 
 import collections
+import errno
```
hunse: I don't think we're using this anymore, so can probably get rid of it.

nengo/utils/compat.py
```diff
@@ -30,6 +33,24 @@ def ensure_bytes(s):
         assert isinstance(s, bytes)
         return s
 
+    if sys.platform.startswith('win'):
+
+        def replace(src, dst):
```
hunse: Maybe add a comment here briefly describing why this is necessary. Might be worth referencing this PR, too, so it's easy to find the full explanation.

tbekolay: Good call, I'll add it in.

tbekolay:
Added @hunse's requested changes; this PR LGTM with all that. Up for second review now. @hunse does this look good to you now?

hunse:
Yep, looks good.

tbekolay:
Cool, I'll merge :+1: :octopus:

PR #1110, opened 2016-06-24
Graceful handing of corrupted cache index

jgosmann:
**Description:**
Gracefully handles a corrupted cache index by clearing the cache an creating a new index.

**Motivation and context:**
In some instances the cache index can get corrupted (i.e. killing the process while the cache index is being written) and that would give an exception whenever one tries to use the cache. See #1097.

Related to #1107 which makes it less likely to end up with a corrupted index.

**How has this been tested?**
Added a regression test, made sure it fails and fixed the problem which makes the test pass.

Either the test or the fix depending on personal preference.

**Types of changes:**

**Checklist:**

- [x] I have updated the documentation accordingly. (no changes necessary imo)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,8 +32,8 @@ Release History
 - Added ``y0`` attribute to ``WhiteSignal``, which adjusts the phase of each
   dimension to begin with absolute value closest to ``y0``.
   (`#1064 <https://github.com/nengo/nengo/pull/1064>`_)
-- Allow the `AssociativeMemory` to accept Semantic Pointer expressions as
-  `input_keys` and `output_keys`.
+- Allow the ``AssociativeMemory`` to accept Semantic Pointer expressions as
+  ``input_keys`` and ``output_keys``.
   (`#982 <https://github.com/nengo/nengo/pull/982>`_)
 
 **Bug fixes**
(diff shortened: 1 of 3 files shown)


tbekolay:
This looks good me, thanks for prepping this quickly! I added a commit to fix the flake8 error, and add some assert statements. The asserts aren't strictly necessary but feels like something should be asserted (I could be persuaded to leave them out).

This probably can use a changelog entry, which I could write in the merge. Can someone else look at this quickly, as it's the last thing for 2.1.1?

hunse:
Yeah, LGTM.

jgosmann:
additional commit LGTM, do you want to also include #1107 in 2.1.1?

hunse:
I would not include #1107, since it still needs a bit more testing I think. Hopefully 2.2.0 will be out in a week or two, so it shouldn't be too long of a wait.

jgosmann:
> This probably can use a changelog entry, which I could write in the merge.

ok, leaving this to you. Didn't do it before because I first need the PR number and after I opened the PR I went for lunch before I was able to write the changelog.

tbekolay:
Fair ;) Will do the merge now, then the release.

tbekolay:
... after my lunch, that is.

PR #1121, opened 2016-07-13
marker size change

celiasmith:
**Description:**
Changes the size of markers on the spike rasters when plotting with rasterplot

**Motivation and context:**
Rasters are usually plotted without overlap of spikes from different neurons, this makes that the case for Nengo.

**How has this been tested?**
Ran raster plots for papers across a few models.  Only on Mac OSX.

It's one line.  I'd start there. 🍶 

10 seconds

**Types of changes:**
?

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,8 @@ Release History
 - Added ``ThresholdingPreset`` to configure ensembles for thresholding.
   (`#1058 <https://github.com/nengo/nengo/issues/1058>`_,
   `#1077 <https://github.com/nengo/nengo/pull/1077>`_)
+- Tweaked ``rasterplot`` so that spikes from different neurons don't overlap.
+  (`#1121 <https://github.com/nengo/nengo/pull/1121>`_)
 
 **Bug fixes**
 
--- a/nengo/utils/matplotlib.py
+++ b/nengo/utils/matplotlib.py
@@ -99,34 +99,34 @@ def implot(plt, x, y, Z, ax=None, colorbar=True, **kwargs):
 
 
 def rasterplot(time, spikes, ax=None, use_eventplot=False, **kwargs):  # noqa: C901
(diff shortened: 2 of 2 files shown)


tbekolay:
I remember spending a long time picking that constant... the issue that I found was that any value that worked well for a certain number of neurons worked poorly for a different number of neurons. I sort of got around it by scaling by axis height, but that was a bit of a hack; in actuality, it should probably scale by some combination of axis height and the number of neurons... I believe the marker size is in units of points squared, so there's probably a square root in there too? I'll look into this more.

celiasmith:
I checked this value for numbers of neurons from 2-100 and it seemed ok... i think that's a good 'reasonable' range.  The previous value didn't look ok at all for small numbers of neurons.

drasmuss:
Updated constant looks slightly better to me, here's an image for comparison (updated constant is bottom row):
comparison
You'll still get overlap for different sized plots, but at least this looks ok for the default plot size, and I'm not super motivated to do some fancy scaling to get it to work in all cases.

I also added a small fixup commit to update the documentation in that rasterplot function, which was wildly out of date.  I squashed it in already (assuming it would be uncontroversial), as I was already fiddling with the history to update @celiasmith commit message to conform to our [guidelines](https://pythonhosted.org/nengo/contributing.html#commit-messages) :wink:.  And rebased to master while I was at it.

celiasmith:
man, that @celiasmith guy didn't even follow the guildines! what a douche.  he was probably overwhelmed with the complexity of his added code.

tbekolay:
New plots LGTM; I pushed a fixup to add a changelog entry and fix up the example formatting a bit (interpreter sessions have a different starting marker for prompts versus continuations of a previous prompt). If it looks good to you @drasmuss fix it up and merge :+1: :rainbow:

PR #1127, opened 2016-07-22
Declare entry point for nengo reference simulator.

jgosmann:
**Description:**
This adds an entry point for the reference simulator so that other Python packages (like nengo_gui) can discover the backend in a general way. Such an entry point should become best practice for all backends.

**Motivation and context:**
See description.

**How has this been tested?**
1. Checkout the branch and do a `python setup.py develop`.
2. Run the following Python code:

```python
from pkg_resources import iter_entry_points
print([ep.load() for ep in iter_entry_points(group='nengo.backends')])
```
It will print "[<class 'nengo.simulator.Simulator'>]".

Click "Files changed".

**Types of changes:**

**Checklist:**

**Still to do:**

- Add some documentation, but we need some general backend developer documentation first. I'll create an issue for that.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,13 @@ Release History
 2.2.0 (unreleased)
 ==================
 
+**Improvements**
+
+- Added a ``nengo.backends`` entry point to make the reference simulator
+  discoverable for other Python packages. In the future all backends should
+  declare an entry point accordingly.
+  (`#1127 <https://github.com/nengo/nengo/pull/1127>`_)
+
 **Bug fixes**
 
 - The synapse methods ``filt`` and ``filtfilt`` now support lists as input.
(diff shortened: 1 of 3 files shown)


tbekolay:
> It will print "[]".

I get `[<class 'nengo.simulator.Simulator'>]` which I'm pretty sure is actually what we want ;)

Aside from that, this looks good to me! I pushed a commit adding a test.

jgosmann:
That's the markdown messing with me ... click the edit icon and you will see I put in in there. I'm innocent. I promise. 😇

jgosmann:
It would also be possible to load a specific entry point without iterating over all. Might that be more appropriate for the test?

tbekolay:
The advantage of loading them all (which shouldn't take long) is that we don't have to write a different test for each backend; also, failing this test will let backends know that they should also declare this entry point.

jgosmann:
Good point!

Seanny123:
Given that the only test that is not passing was fixed by #1134, this branch looks ready to merge.

PR #1129, opened 2016-07-23
Add tutorial about functions and tuning curves.

jgosmann:
**Description:**
Adds a tutorial/example notebook.

**Motivation and context:**
Seemed useful to me to have.

none

**How has this been tested?**
Ran all cells of the notebook.

Read the notebook. 📖 

**Types of changes:**

**Checklist:**

### DIFF
--- a/examples/advanced/functions_and_tuning_curves.ipynb
+++ b/examples/advanced/functions_and_tuning_curves.ipynb
@@ -1,0 +1,520 @@
+{
+ "metadata": {
+  "kernelspec": {
+   "display_name": "nengodev",
+   "language": "python",
+   "name": "nengodev"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.5.3"
+  },
+  "name": ""
+ },
+ "nbformat": 3,
+ "nbformat_minor": 0,
+ "worksheets": [
+  {
+   "cells": [
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "import matplotlib.pyplot as plt\n",
+      "import numpy as np\n",
+      "\n",
+      "import nengo\n",
+      "\n",
+      "%load_ext nengo.ipynb\n",
+      "%matplotlib inline"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "# Improving function approximation by adjusting tuning curves\n",
+      "\n",
+      "This tutorial shows how adjusting the tuning curves of neurons\n",
+      "can help implement specific functions with Nengo.\n",
+      "As an example we will try to to compute\n",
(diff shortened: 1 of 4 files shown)


Seanny123:
Would it be a good idea to introduce the idea of the thresholding preset at the conclusion of this notebook?

jgosmann:
Yes! I wrote most of this quite a while ago (when there was no preset) and managed just now to look it over and finish it up.

jgosmann:
Added the `ThresholdingEnsembles` preset.

examples/advanced/functions_and_tuning_curves.ipynb
```diff
@@ -0,0 +1,531 @@
...
+   "source": [
+    "with nengo.Simulator(threshold_net) as sim:\n",
+    "    sim.run(1.)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let as look at the tuning curves first."
```
Seanny123: Typo. Should be "Let us"

Seanny123:
Other than the typo, this LGTM.

tbekolay:
I will review this shortly, but one thing I noticed in the diff is that this notebook uses notebook version 4 while all the other notebooks we've kept as version 3 so far. The notebook version was changed in IPython 3.0, and can be read by IPython 2.4.1. so we would effectively be ending support for IPython 2.4.0 and below. 2.4.1 was released February 2015, and our tests already depend on `jupyter` rather than IPython. Additionally, if we wanted, we could spit out version 3 notebooks during our doc building process even if the stored notebooks are version 4. So, there are a few options... vote using the emoji responses!

- :laughing: Upgrade all our notebooks to v4 but generate v3 notebooks to be linked in the docs.
- :tada: Upgrade all our notebooks to v4; if people have older versions of IPython they can download the Python script equivalent instead of the notebook.
- :confused: Keep our notebooks as v3 (and convert the notebook in this PR to v3).

Feel free to propose other resolutions as well.

jgosmann:
How easy is it to generate v3 from v4?
Debian stable is still at IPython 2.3. :/

tbekolay:
> How easy is it to generate v3 from v4?

Pretty trivial: `jupyter nbconvert notebook.ipynb --to notebook --nbformat 3 --output notebook-3.ipynb`

And there would be some programmatic way to do the same.

jgosmann:
If we can add that to the clear outputs script, I'm leaning towards keeping them in v3 for now. (Gotta call that script anyways to clear the outputs.)

tbekolay:
Let's keep them v3 for now (path of least resistance) as this would be nice to have for #1150. I'll modify the clear outputs script as well.

tbekolay:
Started looking at this in detail... one issue is that it uses `spa` bits, which we will eventually be moving elsewhere. I think it's pretty clear that this example is a general one and doesn't rely on the spa bits, so is it OK if I change this to a simpler example that doesn't require SPA stuff?

In either case, since this will take some time to change, I'm going to leave this until after the 2.2.0 release, as it's already somewhat overdue.

Seanny123:
@tbekolay given that Nengo 2.2.0 has be released, is this ready to be merged?

jgosmann:
Probably not? It seems references to SPA should be removed?

jgosmann:
I rewrote the tutorial to remove references to SPA. Please review it again.

jgosmann:
LGTM

PR #1134, opened 2016-07-25
Fix errors from flake8 upgrade

tbekolay:
**Description:**
Flake8 3.0 was just released, which caused a few lines to fail the static checks (the main reason for this seems to be that it parses comment lines more strictly now).

**Motivation and context:**
The lastest [TravisCI master build failed](https://travis-ci.org/nengo/nengo/jobs/147236675), which is :crying_cat_face:.

**How has this been tested?**
Upgraded `flake8` with `pip install --upgrade flake8`. Ran `flake8 nengo` locally. Passed!

It's a small diff, so the files tab.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/utils/builder.py
+++ b/nengo/utils/builder.py
@@ -177,8 +177,8 @@ def _create_replacement_connection(c_in, c_out):
     return c
 
 
-def remove_passthrough_nodes(objs, connections,  # noqa: C901
-        create_connection_fn=_create_replacement_connection):
+def remove_passthrough_nodes(  # noqa: C901
+        objs, connections, create_connection_fn=None):
     """Returns a version of the model without passthrough Nodes
 
     For some backends (such as SpiNNaker), it is useful to remove Nodes that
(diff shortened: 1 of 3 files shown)


jgosmann:
LGTM :cake:

nengo/utils/builder.py
```diff
@@ -177,8 +177,8 @@ def _create_replacement_connection(c_in, c_out):
     return c
 
 
-def remove_passthrough_nodes(objs, connections,  # noqa: C901
-        create_connection_fn=_create_replacement_connection):
```
hunse: Why was this not okay anymore?

tbekolay: I'm not sure if it was ever okay; the rule is basically you have to do one of

```python
def my_long_func(a, b,
                 c, d):
```
or
```python
def my_long_func(
        a, b, c, d):
```

I think the `# noqa` was being applied for all rules and wasn't just ignoring C901.

jgosmann: I'd assume that if you do a line break in the argument list, all lines have to start with the same indentation. That is either
```python
def remove_passthrough_nodes(objs, connections,
                             create_connection_fn=_create_replacement_connection):
```
or how it is done now.

But I'm not 100% sure.

jgosmann: @tbekolay was faster ... :dash:

hunse: And so changing the default to `None` and doing it in the function was just a side fix, and unrelated to flake8? That's fine, I'm just checking.

tbekolay: Oh yeah, I did that because without it it would break onto 3 lines and there was no way to make it look good.

PR #1135, opened 2016-07-29
Pickleable signals

hunse:
**Description:**

This PR makes `Signal`s pickle properly.

**Motivation and context:**

With the `copy` branch (#984), we can now start thinking about pickling whole `Model`s. I tried this out, but the simulator steps were getting generated in random orders after reloading the `Model`. I tracked this down to the fact that when we pickle signals, specifically those that are views on other signals, the pickling process (I was using the `dill` package, since it is more flexible) didn't maintain views between the `initial_value` arrays of signals. Since we rely on this in the `Signal.may_share_memory` function, that was failing.

This depends on #984.

**How has this been tested?**

Added a test which pickles a ``Model`` object and then unpickles it, ensuring that the simulation with both the original and re-created objects are the same.

It's quite short.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,9 @@ Release History
 
 **Fixed**
 
+- Fixed an issue in which signals could not be pickled,
+  making it impossible to pickle ``Model`` instances.
+  (`#1135 <https://github.com/nengo/nengo/pull/1135>`_)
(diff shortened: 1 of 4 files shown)


hunse:
This is ready for review.

PR #1136, opened 2016-08-02
Only allow one WarningCatcher block at a time.

### LINKED ISSUE #1011: Warning on open simulators test may fail

jgosmann:
**Description:**
This prevents multiple threads from activating a `WarningCatcher` at the same time. `WarningCatcher` relies on `warnings.catch_warnings` which is not thread-safe as it relies on global state. With this change it should be possible to use `WarningCatcher` safely in a threaded environment (e.g. pytest-xdist).

**Motivation and context:**
Fixes #1011.

**How has this been tested?**
Ran the tests. Only time will tell whether #1011 is indeed fixed (it seems to only occur on Travis-CI and even there only rarely).

:point_right: Files changed

**Types of changes:**

**Checklist:**

- [x] I have updated the documentation accordingly. (No changes neccessary?)

**Still to do:**

Do we need a changelog entry? It's not really frontend related ...

### DIFF
--- a/nengo/utils/testing.py
+++ b/nengo/utils/testing.py
@@ -185,13 +185,21 @@ class Logger(Recorder):
 
 
 class WarningCatcher(object):
+    lock = threading.Lock()
+
     def __enter__(self):
-        self.catcher = warnings.catch_warnings(record=True)
-        self.record = self.catcher.__enter__()
-        warnings.simplefilter('always')
+        self.lock.acquire()
+        try:
+            self.catcher = warnings.catch_warnings(record=True)
+            self.record = self.catcher.__enter__()
+            warnings.simplefilter('always')
+        except:
+            self.lock.release()
+            raise
 
     def __exit__(self, type, value, traceback):
         self.catcher.__exit__(type, value, traceback)
+        self.lock.release()
 
 
 class warns(WarningCatcher):


Seanny123:
This LGTM. I had to restart the TraivsCI test for some reason, but it seems to be passing now.

jgosmann:
What was the reason that it failed?

Seanny123:
Something about a `conda` command failing.

hunse:
Yeah, this looks fine. It would be nice if we didn't have to resort to locks, but I don't think this can ever cause a deadlock, and I think our warning catching blocks are also very small so there shouldn't be too much time when we're waiting on one.

A while ago, I added our `warns` function to `pytest`. We had always planned on using that in Nengo once the `pytest` release had been out long enough. If we do want to make that switch at some point, we'll have to figure out how to add in this multi-threading support. One option would be to actually submit a PR to `pytest`.

Seanny123:
+1 for submitting a pull request. I tried to look at the code to see where the locks should be put, but I couldn't figure it out.

jgosmann:
I can do that, because I already figured that part out when I tried to find out whether switching to `pytest.warns` would help. It probalby would be nice to write a code example that reproduces the problem somewhat reliably which could be the hard part.

jgosmann:
Upon furthor investigation, I came to the conclusion that this will not fix the problem ... and the Travis-CI build coveniently confirmed that for me.

It might still be worth to merge this because the thread-safety of the warnings catcher could still be a concern. But I can't create a condition where this in practice causes any problems. The main problem would probably be that the original warnings filter isn't restored properly, but that isn't really a problem because whenever we are interested in the warnings we create a new filter anyways.

Because of this, it is probably also much harder to convince people that the lock should be included in py.test.

I'll discuss further points in the original issue #1011 because they are not directly related to this PR.

tbekolay:
I'd prefer not to merge this if it doesn't fix the issue; better to do the monkey patching as discussed in #1011. OK with closing this @jgosmann?

jgosmann:
Yeah, I'm fine with closing it.

PR #1137, opened 2016-08-02
SPA compiler

jgosmann:
See PR #1014. Opened this PR to have the diff with respect to the right base branch and make the review process easier.


jgosmann:
This should be ready for review and is the next major PR (after hierarchical SPA modules) that needs to be merged before any other SPA issues can be addressed.

jgosmann:
Please note that the test failures seem to be due to a bug in our warnings catcher as discussed in #1011.

nengo/spa/spa_ast.py
```diff
@@ -0,0 +1,994 @@
...
+        tr = self.source.type.vocab.transform_to(self.type.vocab)
+        artifacts = self.source.construct(context)
+        return [a.add_transform(tr) for a in artifacts]
+
+    def evaluate(self):
+        tr = self.source.type.vocab.transform_to(self.type.vocab)
+        return SemanticPointer(np.dot(tr, self.source.evaluate().v))
+
+    def __str__(self):
+        return 'reinterpret({})'.format(self.source)
```
tcstewar: Typo: ```reinterpret``` should be ```translate```.  I'll do a fixup commit.

tcstewar:
I've added what may be a controversial commit to this.  Basically, I've added a ```strict``` flag that lets you indicate whether you want the strict behaviour (i.e. you must say ```state2 = translate(state1)``` if your two modules use different Vocabularies) or the old magic behaviour (i.e. you can get away with ```state2 = state1``` and it will magically try to translate between them).  

My reasoning for this is backwards compatibility with any existing models that people have made, including a few that happened during the summer school.  I don't want to modify master in such a way that those suddenly break -- I'd prefer to give a DeprecationWarning.  I definitely think that strict mode is better and what we should be doing going forward (i.e. there should only be strict mode and no option at all in the future ```nengo_spa``` repository), but it'd be nice to not break existing models, if we can.

There is also the option of not merging this PR with master, and just using it as the basis of the new ```nengo_spa```.  However, this PR improves so many things that I'd like to have it in master, if it doesn't break anything.  If that's not the plan, however (and I've forgotten what the decision was the last time this came up), then the commit I made should be dropped.  :)

jgosmann:
I'm fine with or without strict mode flag. Though, I might prefer the default to be `True` which would break existing models, but in an easily fixable way (just one added line to deactivate the strict mode and it could even say so in the exception). If this gets merged to master, I would be fine to set the flag to `False` there and to `True` in the new `nengo_spa` module.

tcstewar:
> I'm fine with or without strict mode flag. Though, I might prefer the default to be True which would break existing models, but in an easily fixable way (just one added line to deactivate the strict mode and it could even say so in the exception). If this gets merged to master, I would be fine to set the flag to False there and to True in the new nengo_spa module.

@jgosmann just reminded me that the plan was to not touch master with this stuff at all, and just use it as the basis for a new ```nengo_spa```, so in that case my commit should be removed.  I don't think we should have a strict mode flag in ```nengo_spa``` -- we should just force people to always be explicit.

jgosmann:
So it basically comes down to whether we want to merge this into master. Maybe something to discuss at the next dev meeting.

tcstewar:
> So it basically comes down to whether we want to merge this into master. Maybe something to discuss at the next dev meeting.

The more I think about it, the more I think it makes sense to just go with ```nengo_spa``` instead.  The fact that there are some great things in this PR will simply act as encouragement to transition over to ```nengo_spa```....  :)

tcstewar:
I really like how the current code makes a separate Network for all of the components needed to implement the Action rule.  However, there seems to be something slightly off about the logic.  In particular, it makes a Network to put things in if and only if the ```Effect``` is not a fixed effect.  This means that it doesn't make a dedicated subnetwork for this case:

```'dot(state1, state2) --> state3=A'```

Instead, the dotproduct Network it creates just floats out on its own.  

I'm not quite sure how to address this, as it's not clear to me how to detect whether or not we'll be creating artifacts during the construct call for ```self.condition```.  We can't just use ```self.condition.fixed``` since that'll also be False for ```dot(state1, A)```, which doesn't require artifacts.

**Comment by tbekolay** [comment]
Dev meeting update: we'll keep this out of master, merge into `new_spa` and iterate in the `nengo_spa` repo. Once `nengo_spa` is ready to go (1.0 release) we'll deprecate `nengo.spa` and point to it.

**Comment by jgosmann** [comment]
I pushed a commit to undo d13d0608ab4b5e2cbdf2805fa18854b6217d672c as we won't merge this into master and should enforce explicitness.

@tcstewar: Are you still reviewing this?

**Comment by tcstewar** [comment]
> @tcstewar: Are you still reviewing this?

The main outstanding problem for me here is that I'm not sure what to do about the issue I raised above (it doesn't make a subnetwork for the components created in an Action like ```'dot(state1, state2) --> state3=A'```).

In addition, I think it would be cleaner if there were separate subnetworks made for the LHS and the RHS of the Actions, rather than putting them both in the same network (after all, those parts never interact with each other).  That also means that the subnetworks will have shorter and more readable names in the nengo_gui (as it stands, things can get a little bit silly with the extremely long names created for the Actions).

Other than that, I'm very happy with this PR!

I probably won't have time to take a stab at implementing anything for those issues above until a week from now, but I can try then, if that would be helpful...

**Comment by jgosmann** [comment]
So, I split up the actions and moved the “left-hand side” (basically the gate) into the thalamus network.

> In particular, it makes a Network to put things in if and only if the Effect is not a fixed effect. This means that it doesn't make a dedicated subnetwork for this case:
> 
> 'dot(state1, state2) --> state3=A'
> 
> Instead, the dotproduct Network it creates just floats out on its own.
> 
> I'm not quite sure how to address this, as it's not clear to me how to detect whether or not we'll be creating artifacts during the construct call for self.condition. We can't just use self.condition.fixed since that'll also be False for dot(state1, A), which doesn't require artifacts.

I extended the concept of `fixed` to address it. Now it is a tri-state `staticity` attribute. It can be `FIXED` and `DYNAMIC` which are the states corresponding to the binary `fixed` attribute. The new `TRANSFORM_ONLY` state is inbetween, it denotes a node that changes over time, but can be implemented with a simple transform (no additional neural resources needed). This staticity information is propagated bottom-up like the `fixed` information. A binary operation on two `TRANSFORM_ONLY` nodes will become a `DYNAMIC` node because the two transforms cannot be merged into a single transform.

This now allows to put all objects related to a single condition into the same labeled network without creating empty networks for conditions where a simple transform is sufficient.

**Comment by tcstewar** [comment]
> I extended the concept of fixed to address it. Now it is a tri-state staticity attribute. It can be FIXED and DYNAMIC which are the states corresponding to the binary fixed attribute. The new TRANSFORM_ONLY state is inbetween, it denotes a node that changes over time, but can be implemented with a simple transform (no additional neural resources needed). This staticity information is propagated bottom-up like the fixed information. A binary operation on two TRANSFORM_ONLY nodes will become a DYNAMIC node because the two transforms cannot be merged into a single transform.

Nifty.  Looking much better, although there still seems to be cases where it behaves badly.  For example,

```
dot(state1,A)-dot(state2,B) --> ...
```

will still create a ```condition``` Network, but it'll be empty.

I wonder if, instead of doing this checking approach, it might make more sense to always create the Network, but then delete the Network if it is empty afterwards?  That way it'll always work, and we don't have this edge-case logic to worry about?

jgosmann:
I just fixed up the logic (I didn't realize that it has to differ for sums and products). But I could also see deleting network afterwards. It might be the better solution because it doesn't require reasoning about the “staticity” which is probably not a super obvious concept. But it also leaves the syntax tree without some information that might be useful?

tcstewar:
> But I could also see deleting network afterwards. It might be the better solution because it doesn't require reasoning about the “staticity” which is probably not a super obvious concept. But it also leaves the syntax tree without some information that might be useful?

Fair enough...  how about we leave it how it is for now, but if we in the future are adding things and it's a pain to do this tracking (or if we keep finding other edge cases), then we can look at doing the delete-the-network approach.

jgosmann:
Fine with me.

tcstewar:
I've added a quick commit that keeps the original parsed text that was used to create the Actions.  That way, when we create the Networks, we can use that text rather that building it ourselves with ```__str__```.

The reason for this is that in the GUI, it looks a bit odd to have ```((((A + B) + C) + D) +E)``` when I wrote ```A+B+C+D+E```.

tcstewar:
(Note: I'm not sure about the implementation of the commit I put above... there might be a nicer way to organize it -- the current implementation feels a bit like a hack to me, but I don't see a nicer way yet....)

jgosmann:
Yeah, it seems a bit like a hack at the moment. I also think this will only work for the top-level expression and not parts of it.

jgosmann:
> I also think this will only work for the top-level expression and not parts of it.

And probably we would need to write our own parser to change that.

tcstewar:
> > I also think this will only work for the top-level expression and not parts of it.
>
> And probably we would need to write our own parser to change that.

Yup, I don't think there's any way to do it for the lower-level expressions.  :(  But it is really nice to have the top-level networks in the GUI have a bit clearer syntax....

I'd definitely be willing to accept this PR without the proposal, btw, so it could be something that we don't do now and leave till later.

jgosmann:
It probably wouldn't be too hard to elimit unneccessary parentheses in the string. That could of course still give a different string from what the user entered, but might be better than what is currently done.

jgosmann:
I added an alternate proposal where the string representation is constructed with the minimal number of parenthesis by considering the operator precedence. This seems less hacky to me, bu won't give precisely the same strings as the user enters. It will also still change `A - B` to `A + -B`.

tcstewar:
> I added an alternate proposal where the string representation is constructed with the minimal number of parenthesis by considering the operator precedence. This seems less hacky to me, bu won't give precisely the same strings as the user enters. It will also still change A - B to A + -B.

Beautiful.  Looks good to me!

I can't think of anything else to complain about here, so I'm all set for this to be the basis of the new nengo_spa repository.  Thank you for doing this, @jgosmann !

tcstewar:
NOTE: As a reminder, even though this is marked as reviewed, it should NOT be merged into nengo master.  Rather, this is meant to be the basis for a new repo called ```nengo_spa```.

tbekolay:
> Rather, this is meant to be the basis for a new repo called nengo_spa

The PR is on the `new_spa` branch; I think the idea is to merge PRs into that branch (so we can clear them out of the queue) and when the time comes to make the new repo, I'll filter the `new_spa` history to include just the SPA stuff.

jgosmann:
I think this might still fail PEP8 ...
Also, does anyone want to give this a second look?

Seanny123:
It's failing PEP8. It's also failing Python 2.7, because of a test expecting `spa.SPA()` to throw a deprecation warning.

jgosmann:
> It's also failing Python 2.7, because of a test expecting spa.SPA() to throw a deprecation warning.

That's because of #1011 and won't be fixed as part of this PR.

jgosmann:
Though, I guess I could remove the deprecation warning as this will never go into nengo master.

jgosmann:
PEP8 failure should be fixed now.

PR #1141, opened 2016-08-09
Use monkey patch instead of catch_warnings.

### LINKED ISSUE #1011: Warning on open simulators test may fail

tbekolay:
**Description:**
Monkey patches `warnings.warn` and `warnings.warn_explicit` instead of using `warnings.catch_warnings` to determine if Nengo warns about something.

**Motivation and context:**
This should fix #1011 once and for all.

**How has this been tested?**
The test suite runs properly locally. pytest does contain [tests for `deprecated_call`](https://github.com/pytest-dev/pytest/blob/master/testing/test_recwarn.py), and we could write similar tests for our solution, but once we're sure that this works for us I'll also make a PR on pytest so that we can use `pytest.warns` and remove this stuff from our codebase. So, since the tests would be removed eventually, I feel like we can omit them (but if reviewers would like the tests included anyway, I can write them).

Read over the last bunch of comments in #1011, then the "files changed" tab here.

**Types of changes:**

**Checklist:**

- [x] I have included a changelog entry. (haven't, but I don't think it needs one)
- [x] I have added tests to cover my changes. (haven't, but see testing note above)

**Notes:**

I'm going to restart the TravisCI build for this branch a bunch of times to make sure it actually works, though even with that it's hard to guarantee it will never come up...

### DIFF
--- a/nengo/utils/testing.py
+++ b/nengo/utils/testing.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 
+import collections
 import inspect
 import itertools
 import logging
@@ -184,14 +185,44 @@ class Logger(Recorder):
             del self.handler
 
 
+RecordedWarning = collections.namedtuple(
+    'RecordedWarning', ['message', 'category', 'filename', 'lineno', 'module'])
+
+
 class WarningCatcher(object):
+
+    def __init__(self):
+        self.recorded = []
+
+    def warn_explicit(self, message, category, filename, lineno,
+                      module=None, registry=None, mod_globals=None):
+        self.recorded.append(RecordedWarning(
+            message, category, filename, lineno, module))
+        self.old_warn_explicit(
+            message, category, filename, lineno, module, registry, mod_globals)
+
+    def warn(self, message, category=None, stacklevel=1):
+        if isinstance(message, Warning):
+            category = type(message)
+        category = UserWarning if category is None else category
+        # warnings.warn uses sys._getframe and other mechanisms to get
+        # the filename, lineno, and module. We won't bother for now.
+        self.recorded.append(RecordedWarning(
+            message, category, None, None, None))
+        # We increase stacklevel by 1 to account for this function being
+        # in the stack trace.
+        self.old_warn(message, category, stacklevel + 1)
+
     def __enter__(self):
-        self.catcher = warnings.catch_warnings(record=True)
-        self.record = self.catcher.__enter__()
-        warnings.simplefilter('always')
+        self.old_warn_explicit = warnings.warn_explicit
+        self.old_warn = warnings.warn
+        warnings.warn_explicit = self.warn_explicit
+        warnings.warn = self.warn
+        return self
 
     def __exit__(self, type, value, traceback):
-        self.catcher.__exit__(type, value, traceback)
+        warnings.warn_explicit = self.old_warn_explicit
+        warnings.warn = self.old_warn
 
 
 class warns(WarningCatcher):
@@ -199,11 +230,11 @@ class warns(WarningCatcher):
         import pytest
         self._pytest = pytest
         self.warning_type = warning_type
+        super(warns, self).__init__()
 
     def __exit__(self, type, value, traceback):
-        if not any(r.category is self.warning_type for r in self.record):
+        if not any(r.category is self.warning_type for r in self.recorded):
             self._pytest.fail("DID NOT WARN")
-
         super(warns, self).__exit__(type, value, traceback)
 
 


nengo/utils/testing.py
```diff
@@ -184,14 +185,40 @@ def __exit__(self, type, value, traceback):
...
+
+    def warn(self, message, category=None, *args, **kwargs):
+        if isinstance(message, Warning):
+            category = type(message)
+        category = UserWarning if category is None else category
+        # warnings.warn uses sys._getframe and other mechanisms to get
+        # the filename, lineno, and module. We won't bother for now.
+        self.recorded.append(RecordedWarning(
+            message, category, None, None, None))
+        self.old_warn(message, category, *args, **kwargs)
```
jgosmann: Maybe the `stacklevel` argument should be incremented by one? (So that the warning does not refer to this line, but the line where `warn` was called.)

nengo/utils/testing.py
```diff
@@ -184,14 +185,40 @@ def __exit__(self, type, value, traceback):
...
+
     def __enter__(self):
-        self.catcher = warnings.catch_warnings(record=True)
-        self.record = self.catcher.__enter__()
-        warnings.simplefilter('always')
+        self.recorded = []
```
jgosmann: If the same WarningCatcher is used in two if blocks, it will only retain the warnings of the last block. I would probably expect that all warnings will be retained. So, maybe move this line to `__init__`?

nengo/utils/testing.py
```diff
@@ -184,14 +185,40 @@ def __exit__(self, type, value, traceback):
...
+
+        self.old_warn_explicit = warnings.warn_explicit
+        self.old_warn = warnings.warn
+
+        warnings.warn_explicit = self.warn_explicit
+        warnings.warn = self.warn
```
jgosmann: Isn't it best practice to `return self` at the end of `__enter__`?

jgosmann:
Nit pick about the commit message: It is not clear what “the previous solution” was, maybe add the commit hash? (Did we even merge the previous solution?)

tbekolay:
The previous solution in this case is what's in the parent commit, so people can look at the diff to see what's changed. I can fix up the stuff you mentioned after lunch today, or you can push fixups to do those things (they all sounds good to me)

jgosmann:
GitHub doesn't show me what the parent commit is, but I assume it's the latest commit in master which is "Connections can take points for functions"?

jgosmann:
Fixup commits LGTM. 🍰

drasmuss:
Also I've been restarting the travis job a bunch, and it hasn't failed yet.  Not proof, but at least a good sign.

tbekolay:
Pushed fixups to address Jan's comments (thanks!) and switched to explicit args instead of `*args, **kwargs` (as they were around the same length in both cases).

> I assume it's the latest commit in master which is "Connections can take points for functions"?

Yeah, that's right. But you shouldn't really have to look at the parent; if you do `git show 58649ce57be5ae57a30a7c7` it will show you the commit's diff, and what was removed is what I meant by the "previous solution".

jgosmann:
I see. You were referring to the previous WarningsCatcher implementation, not previous attempts to fix that (directly call `__del__` instead of the garbage collection, locking)

tbekolay:
> You were referring to the previous WarningsCatcher implementation, not previous attempts to fix that

Yep; perhaps "implementation" is more clear than "solution" in the commit message?

Also, how many TravisCI runs should we do until we're satisfied? Maybe as many as can fit into today, merge tomorrow if no fails?

jgosmann:
Yeah “implementation” sounds good to me.

Personally I would not bother running a lot of Travis-CI tests. Maybe check if this code piece from #1011 passes with the new WarningCatchers:
```python
import warnings

import pytest


def warn():
    warnings.warn(DeprecationWarning("test"))


@pytest.mark.parametrize('i', range(2))
def test_warning(i):
    if i % 2 == 0:
        warn()
    else:
        with pytest.warns(DeprecationWarning):
            warn()
```

tbekolay:
That snippet passes for me using `nengo.utils.testing.warns`, but not `pytest.warns`. Huzzah!

I also amended the commit message. Since I had to force push anyway, I fixed up all the commits also. I think it's ready to merge?

jgosmann:
👍

hunse:
LGTM

Issue #933, opened 2016-01-15
Lazy storage of eval_points and encoders


tcstewar:
While discussing #931, a point was raised that the eval_points actually take up a lot of memory, and we only keep them around just in case the user happens to access ```sim.data[ens].eval_points```.

Instead of storing all of that data, all we really need to store is the state of the rng and a reference to the Distribution that we're sampling from.  Then if someone accesses ```sim.data[ens].eval_points``` we just do the sampling and return the result.

This same trick could also be useful for ```sim.data[ens].encoders``` (since the only thing we actually need to have during simulation time is ```sim.data[ens].scaled_encoders```).

On the plus side, this change would be completely transparent to the user, and should save a lot of memory.

On the minus side, this would require a fair bit more complexity on the side of the BuiltEnsemble and BuiltConnection objects.

hunse:
Another minus is that the sampling takes considerable time, which is not necessary if the user knows they'll need the eval points and can tell us that.

The other main alternative is to just have an option to store the eval points or not. We could even have one option that does all the "extra" stuff, including encoders as well. Also, for device-based backends, they could save even more host memory by not keeping anything on the host once it's transferred to the device.

tcstewar:
> The other main alternative is to just have an option to store the eval points or not. We could even have one option that does all the "extra" stuff, including encoders as well. Also, for device-based backends, they could save even more host memory by not keeping anything on the host once it's transferred to the device.

I like the idea of having one option that just turns off all that extra stuff.  Maybe something like

```python
sim = nengo.Simulator(model, keep_data=False)
```

I'd prefer that to having separate options:

```python
sim = nengo.Simulator(model, keep_eval_points=False, keep_encoders=False)
```

just because that seems like it could proliferate to lots of parameter options.

tcstewar:
> Another minus is that the sampling takes considerable time, which is not necessary if the user knows they'll need the eval points and can tell us that.

I'm not too worried about that time cost, as it'll just be a one-off cost that only happens when someone accesses it, rather than the costs of either always having that stuff in memory (what we have currently) or having to remember to specify that extra parameter option to tell nengo to keep this stuff or not.

tbekolay:
I was thinking about this kind of thing when we originally were brainstorming about the probe system... I called them "static probes", but I think with the infrastructure we have now, this syntax makes sense to me:

```python
e = nengo.Ensemble(10, 1)
[...]

Issue #1145, opened 2016-08-17
API for data generated during build.


jgosmann:
We should define an API how information generated during the build (e.g., decoders, encoders, sample points) can be accessed. The reference builder currently provides `sim.model.params`, but not all backends provide `sim.model`.

#933 is related and should be considered together with this.

This sort of API is relevant for PRs as #995, #941 (even without deferrables), #1062, and #553.

tbekolay:
Actually the "correct" place to access these things is through `sim.data`; e.g. to get the decoders of a connection it's `sim.data[conn].weights`. But obviously this is not well documented or advertised, in part because #933 hasn't moved in a while.

jgosmann:
Does that work without probing and for all parameters?

tbekolay:
Yeah, we basically copy `sim.model.params` into `sim.data`.

jgosmann:
But there are things like the ensemble radius that are not in `sim.model.params`?

tbekolay:
Yeah, that's true, we could add them. In the builder refactoring that I started (but haven't worked on for a while) I add everything to `BuiltEnsemble` so that it's a second type of canonical representation for ensembles, and so on for other objects.

jgosmann:
Given the discussion in #933, I would propose the following:

All static parameters (in the sense they don't change during the simulation) on Nengo objects should be stored as attributes of `sim.data[obj]` where `obj` is the Nengo object. This includes sampled evaluation points, encoders, and `solver_info` on connections. This should be recommended for all backends (or they won't be feature complete and other Nengo things might not work, like plotting tuning curves or generic learning rules).

Access of these things via `sim.model` should be considered as obsolete (if it happens outside of the builder) because it is backend dependent.


I could potentially be convinced to go with [static probes](https://github.com/nengo/nengo/issues/933#issuecomment-172116510), because it matches our existing probe syntax nicely. But [there are some issues](https://github.com/nengo/nengo/issues/933#issuecomment-241233795) that would make the implementation more complex (and even without these issues it would probably be higher degree of complexity because the build of the static probes will be decoupled from the place where the evaluation points for example are generated).

PR #1146, opened 2016-08-22
Static probes

drasmuss:
**Description:**

This is a slapped together implementation of "static probes" (as discussed in https://github.com/nengo/nengo/issues/933#issuecomment-172116510).

It's basically just an alias for `sim.model.params`

```
a = nengo.Ensemble(...)
p = nengo.Probe(a, 'params')
...
assert sim.data[p] == sim.model.params[a]
```

I also stuck in the option to only keep the probe data from the most recent timestep.

**Motivation and context:**

I mainly did it so that I could see what it looked like using these probes to implement https://github.com/nengo/nengo/pull/553.

But I figure we can also use this as a way to flesh out what exactly a full implementation of this would look like (i.e. how would we want it to be different than this implementation?).

https://github.com/nengo/nengo/pull/553

**How has this been tested?**

I tried a few examples and it worked for those

It's actually quite a small change in terms of code.  The main thing to do is to think about what we actually want this feature to look like (compared to how it is implemented here).

**Types of changes:**

**Checklist:**

**Still to do:**

### DIFF
--- a/nengo/builder/probe.py
+++ b/nengo/builder/probe.py
@@ -100,6 +100,11 @@ def build_probe(model, probe):
         raise BuildError(
             "Type %r is not probeable" % type(probe.obj).__name__)
 
+    if probe.attr == 'params':
+        # don't add this probe to the model probe lists
+        model.params[probe] = model.params[probe.obj]
+        return
+
     key = probeables[probe.attr] if probe.attr in probeables else probe.attr
     if key is None:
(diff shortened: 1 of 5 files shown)


jgosmann:
This addresses #1145 and #933 (or is at least related to that discussion).

drasmuss:
Going to close this, since it's been sitting around for a while and it hasn't seemed like we've needed it.

PR #1147, opened 2016-08-24
Drop NumPy 1.6 support.

jgosmann:
**Description:**
Drops support for NumPy 1.6 by bumping version numbers in the requirements and removing it from the test martix.

**Motivation and context:**
The last NumPy 1.6 release is more than 4 years old. By not supporting it any longer some features will be easier to implement, namely most of the logic in #1016 would not be required. At the last dev meetings no objections were raised.

#1016

**How has this been tested?**
Hasn't been tested. I think the only thing worth to test would be whether I did not overlook any reference to NumPy 1.6, but no idea what else to do than grepping for `1.6`.

Files changed as usual, it might make sense to view the commits individually.

**Types of changes:**

- Version requirement bump (doesn't break any frontend or backend API, but might require an upgrade of the NumPy dependency)

**Checklist:**

**Still to do:**


jgosmann:
Under which heading would the changelog entry go? The usual categories don't really fit.

hunse:
This makes #1016 obsolete.

tbekolay:
> Under which heading would the changelog entry go? The usual categories don't really fit.

Behavioral changes I guess is the closest?

Seanny123:
LGTM.

hunse:
LGTM. I added a couple of minor fixups. I was looking through the Numpy 1.7 [release notes](https://github.com/numpy/numpy/blob/master/doc/release/1.7.0-notes.rst) and trying to see if there's anything else we can simplify. The only thing I noticed was they added `np.random.choice`, which we could use in our `Choice`, but I don't think it would simplify things much so I don't think it's necessary.

Seanny123:
Still LGTM btw.

tbekolay:
I'll give this a look over and merge this evening :+1:

PR #1148, opened 2016-08-24
Add options to presets.ThresholdingEnsemble.

jgosmann:
**Description:**
Adds two more options to the thresholding preset to control the width of the intercepts and the radius of the thresholding ensembles.

**Motivation and context:**
Sometimes it can be useful to fine tune these parameters.

none

**How has this been tested?**
The existing test for the preset still passes.

The usual place.

**Types of changes:**

**Checklist:**

**Still to do:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,7 +49,8 @@ Release History
   (`#1045 <https://github.com/nengo/nengo/pull/1045>`_)
 - Added ``ThresholdingPreset`` to configure ensembles for thresholding.
   (`#1058 <https://github.com/nengo/nengo/issues/1058>`_,
-  `#1077 <https://github.com/nengo/nengo/pull/1077>`_)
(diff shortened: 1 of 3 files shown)


Seanny123:
LGTM. I don't think this needs a changelog entry. I'm on the fence about whether we need a test to see if changing the radius actually works, but I'm leaning towards no, since it's really already covered by `ensemble` and `config` tests.

jgosmann:
Actually, I should write a test and make sure that this works as intended. I just remembered that eval_points get scaled by the radius (intercepts maybe too?).

jgosmann:
Regarding the changelog entry: I just added this PR to the existing entry. There wasn't a release wtih the preset yet.

tbekolay:
Documenting presets is the last thing on my todo for 2.2.0, so this should be in before 2.2.0 too. Pushed a fixup to add the defaults to the docstring. LGTM! Will merge once tests finish.

PR #1149, opened 2016-08-24
Alternative generic learning rule implementation

jgosmann:
**Description:**
Like #553 this implements a `GenericRule` for easy proto-typing of learning rules with Python functions.

This is the third variant and they differ mostly in how the input data is specified to the learning rules. All versions have an error or data input that is passed to the Python function implementing the rule. One connects to it by connecting to the learning rules. However, most learning rules depend not only on an error signal, but some aspects of the pre- and post-synaptic objects.
1. In the original implementation one would connect all of these things into the `data` input which then is a concatenated vector of those things. The reason why I dislike this:
   - It's hard to understand because `data[2:]` and `data[:2]` are not descriptive names. To me it seems to make it also more likely to introduce bugs and make it hard to find those.
   - Not everything can be connected from. For example, it is not possible to implement a learning rule based on the internal voltage of neurons.
   - Requires to pass in `model.params` to the learning rule function in some cases which is backend dependent.
2. The current implementation in `genericrule`/#553 extends probes. This seems better to me, but still has some problems.
   - This is tied to the discussion happening in #1145 and #933 about the general API to retrieve static values determined at build time. I think there should be one official way to access these. If we go with probes, the access via `sim.data`/`sim.model.params` should be deprecated, but that mean for the `tuning_curves` function one has to define a probe in the model, making the API slightly harder to use.
   - Probes for the `GenericRule` need to be build out of order. Technically not a real problem, but to quote @drasmuss "it just doesn't feel great."
   - The probes explicitly reference objects in the model which means that one `GenericRule` instance can only be used for one connection. My understanding is that other learning rule instances can be used in multiple connections.

Let's turn to the approach in this PR. Here the learning rule functions gets passed arguments  `pre` and `post` which are objects that allow to access all the required things on the pre- and post-synaptic objects. (Other objects in the model should not be relevant to the learning rule, but if, one can still resort to connecting these into a concatenated data signal.) For example one could do things like:

``` python
pre.get_input('out')  # Ensemble output
pre.get_input('neurons.out')  # Neuron output
pre.get_input('neurons.voltage')  # Neuron voltage
pre.built_params.encoders  # Encoders
pre.obj  # The pre Ensemble object
```

This doesn't depend on the backend. Other backends can create their own objects in an appropriate way providing this interface. (And in any implementation they have to implement some specific code for the `GenericRule` if they want to support it.)

This implementation also requires the user to write less front end code because most things can easily be accessed in the learning function. It might require some more complexity in the backend (but so do the static probes) and given that `GenericRule` is for quick proto-typing, I rather have that complexity in the backend where code gets probably better tested than having to deal with all that complexity when prototyping.

One slight problem might be that the names used depend on the signal names which are not really documented, but we could document general names like `in` and `out` as part of that interface the `pre` and `post` object need to provide. (Maybe it also makes sense to introduce some default that makes it possible to write `get_input('neurons')` instead `get_input('neurons.out'` to get a better correspondence to usual connections.) I think, the neuron signals correspond to the `probeable` names anyways so that we don't have to list each single one.

One final thing is that sometimes inputs to the learning rule should be filtered. This is were connections or probes have a slight advantage because they directly support this. Here I solved this with a `filters` argument to the generic learning rule that allows to specify the filters with a dictionary of the sort `{'pre.out': synapse1, 'post.neurons.voltage': synapse2}`.

Because the `GenericRule` does not reference any of the objects in the model, this implementation allows to assign the same instance to multiple connections.

**Motivation and context:**
See description.

Alternate implementation of #553.
Discussion in #1145 and #933 has some relevance to this.

**How has this been tested?**
The adapted tests from #553.

The first few commits have been extracted from #553, Look at the last two commits (either together or independently, first one is the basic implementation, second one adds filtering)

**Types of changes:**

**Checklist:**

**Still to do:**

- [ ] Come up with decent names. The current names for the classes added are quite horrible I think. I'm also not sure about the name `get_input`.
- [ ] Document this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,8 @@ Release History
   `#1077 <https://github.com/nengo/nengo/pull/1077>`_)
 - Tweaked ``rasterplot`` so that spikes from different neurons don't overlap.
   (`#1121 <https://github.com/nengo/nengo/pull/1121>`_)
+- Added ``GenericRule`` to allow for easy implementation of new learning rules.
+  (`#553 <https://github.com/nengo/nengo/pull/553>`_)
 
 **Bug fixes**
 
--- a/nengo/__init__.py
+++ b/nengo/__init__.py
@@ -24,7 +24,7 @@ from .node import Node
 from .neurons import (AdaptiveLIF, AdaptiveLIFRate, Direct, Izhikevich, LIF,
                       LIFRate, RectifiedLinear, Sigmoid)
 from .network import Network
-from .learning_rules import PES, BCM, Oja, Voja
+from .learning_rules import PES, BCM, Oja, Voja, GenericRule
 from .params import Default
 from .probe import Probe
 from .rc import rc, RC_DEFAULTS
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -2,7 +2,7 @@ import numpy as np
 
 from nengo.builder import Builder, Operator, Signal
 from nengo.builder.operator import DotInc, ElementwiseInc, Reset
-from nengo.connection import LearningRule
+from nengo.learning_rules import LearningRule
 from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError
 from nengo.learning_rules import BCM, Oja, PES, Voja
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -8,6 +8,7 @@ from nengo.exceptions import BuildError
 from nengo.learning_rules import BCM, Oja, PES, Voja, GenericRule
 from nengo.node import Node
 from nengo.synapses import Lowpass
+from nengo.utils.compat import is_number
 
 
 class SimBCM(Operator):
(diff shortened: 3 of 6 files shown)


jgosmann:
An example of how one can define the PES rule for weights (copied from the test):
```python
def regularized_pes(pre, post, dt, target, data):
    encoders = post.built_params.encoders
    gains = post.built_params.gain
    return (-dt / n * np.outer(
        np.dot(encoders, data) * gains,
        pre.get_input('neurons.out')) - target * reg)

pes = GenericRule(
    regularized_pes, learning_rate=1e-3, size_in=None, modifies='weights',
    filters={'pre.neurons.out': 0.005})
```

tbekolay:
I think this is a nice utility but given that there have been multiple implementations of it I'm a bit hesitant to put it in Nengo core as it's likely we'll want to change things once we play around with it a bit. Any objections to moving this to Nengo Extras?

tbekolay:
Cool, I'll look this over in more detail and separate it out into a Nengo extras PR, and if necessary a separate core PR.

drasmuss:
Going to close this PR since the decision was not to merge it into core (this doesn't preclude reopening a similar PR in nengo_extras).

PR #1150, opened 2016-08-25
Add docs for config system and presets

tbekolay:
**Description:**
This adds a new page to the user docs explaining the config system and the presets (currently just the `ThresholdingEnsembles` preset). Note that this tries not to overlap too much with the `config.ipynb` notebook, but links to it instead.

**Motivation and context:**
We discussed at the last dev meeting that these should be added to the docs before the 2.2.0 release.

**How has this been tested?**
I've built the docs with these changes. They build without warnings and look pretty deece.

Files changed tab.

**Types of changes:**

- Bug fix (non-breaking change which fixes an issue) -- sort of

**Checklist:**

- [x] I have updated the documentation accordingly. :wink: 
- [x] I have added tests to cover my changes. (nothing to test)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,6 +54,11 @@ Release History
 - Tweaked ``rasterplot`` so that spikes from different neurons don't overlap.
   (`#1121 <https://github.com/nengo/nengo/pull/1121>`_)
 
+**Documentation**
+
+- Added a page explaining the config system and preset configs.
+  (`#1150 <https://github.com/nengo/nengo/pull/1150>`_)
+
 **Bug fixes**
 
 - Fixed some situations where the cache index becomes corrupt by
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,3 @@
+code, .rst-content tt, .rst-content code {
+    white-space: pre-wrap;
+}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,15 +3,16 @@
 # This file is execfile()d with the current directory set
 # to its containing dir.
 
+import os
 import sys
 
 try:
     import nengo
     import sphinx_rtd_theme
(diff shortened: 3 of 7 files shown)


Seanny123:
LGTM.

jgosmann:
I found one sentence that I had trouble with understanding. I rewrote it and made it a bit more specific regarding the usage of object vs. class. I also moved the config documentation behind the network documentation because that seems to me like a more logical progression (especially because the config documentation mentions networks in a few places).

One thing that slightly irritates me is that all the headings in the same level of the navigation are nouns (e.g. "Networks"), where this one title is a whole phrase ("Setting parameters with Configs"). But shortening it to "Configs" or "Configurations" makes it less expressive.

So: LGTM :cake:

tbekolay:
Reword looks good, but I'd like to revert the move from before network docs to after. The network object itself is documented in `frontend_api.rst`, which comes first; the `networks.rst` doc only documents the specific networks that are included with Nengo. In terms of abstraction levels, I think presets are lower level than networks because networks use presets, and not the other way around, so I would put them first.

> One thing that slightly irritates me is that all the headings in the same level of the navigation are nouns (e.g. "Networks"), where this one title is a whole phrase ("Setting parameters with Configs").

This is a good point... I've been trying to think of a noun version of that title and nothing is popping out at me. Could switch it to just be "The config system" and rearrange the headings a bit... alternatively, can change the headings of the other sections to be more free-form. I think probably this is good enough for now, can revisit later.

tbekolay:
Pushed commits to revert and fix a line that was too long; if you're okay with it this way @jgosmann please squash and merge! :+1: :rainbow: :+1:

jgosmann:
> The network object itself is documented in frontend_api.rst, which comes first; the networks.rst doc only documents the specific networks that are included with Nengo.
Right!

PR #1151, opened 2016-08-26
Add progress bar for the build process.

### LINKED ISSUE #937: Print time spent on building model

jgosmann:
**Description:**
It adds a progress bar for the build process similar to the simulation progress bar.

**Motivation and context:**
Fixes #937. Gives you some indication of how long the build will take and you can look up how long it took after the fact. Like the simulation progress bar, this can be disabled by passing `progress_bar=False` to the `Simulator` or setting it in the nengorc file.

nada

**How has this been tested?**
Used the question answering example which takes some time to build (if the ETA is less than a second no progress bar will be shown). Works with and without `%load_ext nengo.ipynb`.

Files changed.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] This adds a `progress_bar` argument to the constructor. Not sure if we want that. Other proposals are welcome (maybe use the config system?).
- [x] The `progress_bar` arguments on the `run` function is not required anymore with the argument on the constructor, but maybe still helpful? (Are there situation where a different setting for build and run is desired?) Deprecate it or keep it?
- [x] Not sure what people think of the callback from the builder, but some sort of callback will be required and this was the path of least resistance and is pretty general (no dependence on the progress bar class itself).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,10 @@ Release History
 - Added ``copy`` method to Nengo objects. Nengo objects can now be pickled.
   (`#977 <https://github.com/nengo/nengo/issues/977>`_,
   `#984 <https://github.com/nengo/nengo/pull/984>`_)
+- A progress bar now tracks the build process
+  in the terminal and Jupyter notebook.
+  (`#937 <https://github.com/nengo/nengo/issues/937>`_,
+  `#1151 <https://github.com/nengo/nengo/pull/1151>`_)
 
 **Changed**
 
--- a/nengo/builder/builder.py
+++ b/nengo/builder/builder.py
@@ -87,6 +87,8 @@ class Model(object):
         self.time = Signal(np.array(0, dtype=np.float64), name='time')
         self.add_op(TimeUpdate(self.step, self.time))
 
+        self.build_callback = None
+
     def __str__(self):
         return "Model: %s" % self.label
 
@@ -116,7 +118,10 @@ class Model(object):
         obj : object
             The object to build into this model.
         """
-        return Builder.build(self, obj, *args, **kwargs)
+        built = Builder.build(self, obj, *args, **kwargs)
+        if self.build_callback is not None:
+            self.build_callback(obj)
+        return built
 
     def has_built(self, obj):
         """Returns true if the object has already been built in this model.
--- a/nengo/builder/network.py
+++ b/nengo/builder/network.py
@@ -6,13 +6,14 @@ import numpy as np
 import nengo.utils.numpy as npext
 from nengo.builder import Builder
 from nengo.network import Network
+from nengo.utils.progress import ProgressTracker
 
 logger = logging.getLogger(__name__)
 nullcontext = contextlib.contextmanager(lambda: (yield))
 
 
 @Builder.register(Network)  # noqa: C901
-def build_network(model, network):
+def build_network(model, network, progress_bar=False):
     """Builds a `.Network` object into a model.
 
     The network builder does this by mapping each high-level object to its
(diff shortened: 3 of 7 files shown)


hunse:
Thanks @mention-bot, you're so helpful!

I moved a comment. Otherwise LGTM. I haven't tested it out yet, though.

Seanny123:
This LGTM as well, although I haven:t tested it out either.

studywolf:
Works for me in the IPython notebook examples, but I don't see any progress bar in the command prompt. Is this just for notebooks?

jgosmann:
Do you see the progress bar for running the simulation?
Does the build take longer than a second?

studywolf:
yes and yes! i just made two 10,000 neuron populations and connected them up together, takes a little while (over 10-20 seconds) but i only see a progress bar for the simulation.

studywolf:
oh actually sorry it just popped up now in the terminal, but only after about 10-15 seconds, when it's at 60%

jgosmann:
Ok, that's possible because in the terminal the update is not asynchronous as in the Notebook, so it can only be updated between two build steps. Each object in the model constitutes one step and you only have about three objects (two ensembles, one connection, technically also a network and maybe synapses). The first ensemble takes a long time to build because it's so large and only after that the progress bar can detect that it takes a while and shows up.

studywolf:
i was wondering if that was the case; it is working properly for me, then! 
tried a bunch of notebook examples and scripts from the command prompt.

studywolf:
LGTM

studywolf:
:cake:

jgosmann:
Three people looked at the change, one person tested it. Is that good enought to set this to “reviewed”?

Seanny123:
I think so. @tbekolay can you merge?

tbekolay:
I'll take a look at this shortly, makes sense to get it in with 2.2.0 since it's waiting on some OCL stuff to settle down. Still needs a changelog entry, which I'll add when I look at it.

tbekolay:
Sorry for the delay, took a look at this. Made a squash commit that should probably be reviewed; I think I'll go ahead with 2.2.0 without this PR though, since it makes changes to `Simulator` meaning that it's likely to impact Nengo OCL (probably easy to fix, but not worth stopping imminent releases).

jgosmann:
Added more on top of that (IPython progress bar wasn't working, made it use the proper ellipsis character `&hellip;`).

tbekolay:
I rebased this branch and squashed the history. Tests all run, and I can confirm the build progress bar shows up in the terminal and notebook. I added a changelog entry also, so I think this is now good to go.

In the OP there are some questions I don't think have been discussed yet:

> The progress_bar arguments on the run function is not required anymore with the argument on the constructor, but maybe still helpful? (Are there situation where a different setting for build and run is desired?) Deprecate it or keep it?

I'm cool with keeping these, at least for now; it seems unlikely, but it is possible people might care about build time but not simulation time (maybe they're making one sim and running it a ton of times?) In any case, I'm happy to merge this as is and in the future revisit it.

> Not sure what people think of the callback from the builder, but some sort of callback will be required and this was the path of least resistance and is pretty general (no dependence on the progress bar class itself).

I'm not a huge fan of the callback, but I don't see a good alternative right now... plus the builder API is still not "official" anyhow, so again this is something we can revisit, possibly in #1189.

---

I'll give everyone a few hours to comment if desired; if I don't hear anything by the afternoon though, I'll go ahead and merge.

Issue #1154, opened 2016-08-30
Decoder cache should “touch” legacy.txt each time


jgosmann:
On Sharcnet it makes sense to write the cache files to the cluster-local `/scratch` (faster than `/home` or `/work`). The files in `/scratch` expire after 62 days and will be deleted. This is fine for normal cache files, but should not happen to `legacy.txt` because it means that the cache will be reinitialized discarding all other (not yet expired) cached files.

It's probably sufficient to “touch” (i.e., set the modification time to the current time) legacy.txt each time we instantiate a cache to prevent this.

tbekolay:
Seems too specific to a specific environment in my opinion... you could touch `legacy.txt` in your model to have the desired effect without affecting other super specific environments.

I'm not sure how long it takes to touch a file, but it's common for me to spin up a ton of simulators to test something in a small model; modifying the file system on each of those would add up. So given that, we could add some kind of check to only touch it if it's past a certain number of days, but then how many days, so then we make that configurable, and then it becomes a thing that is spread out over different parts of the code that we now have to maintain... as opposed to a one line change in each model you run on sharcnet.

jgosmann:
> we could add some kind of check to only touch it if it's past a certain number of days, but then how many days

Reading the modification time will probably take at least as long as touching the file, so just touching it would be faster. I don't think that this would performance largely, because if the cache is activated it has to access at least the cache index and an additional touch should not really make any difference.

I also don't think it's super specific to have the modification time updated. It's basically just a way to say this legacy information is correct as of *date*. And on some systems this fixes some problems (I assume other HPC clusters might have similar data storage).

tbekolay:
> on some systems this fixes some problems

I'm hesitant to add it in... it's hard to predict whether this will also cause some problems on some systems.

If this were the only possible solution to the issue, I would err on the side of including it, but since it's possible to fix this in a user model I say leave it. If it ends up appearing in a significant portion of user models across multiple users, then we could add it in, but for now this seems premature. Happy to hear other people's thoughts (I think ours are already pretty clear @jgosmann).

jgosmann:
> it's hard to predict whether this will also cause some problems on some systems.

I'm not sure what problems you expect. It's changing the modification time of a file we access anyways in a directory that needs to be writable anyways (and in some instances we even have to write the file in the first place). If that fails a lot of other things in the cache should fail too.

> but since it's possible to fix this in a user model I say leave it

[...]

PR #1155, opened 2016-08-30
Set ctime when accessing cache legacy file.

jgosmann:
**Description:**
Sets the modification time of the cache `legacy.txt` file when it is accessed (which happens once per build).

**Motivation and context:**
See #1154.

none

**How has this been tested?**
Used this script:

``` python
import nengo

def create_model(seed):
    with nengo.Network(seed=seed) as model:
        a = nengo.Ensemble(10, 1)
        b = nengo.Ensemble(10, 1)
        nengo.Connection(a, b)
    return model


for i in range(1000):
    with nengo.Simulator(create_model(i)) as sim:
        pass
```

Ran it once to initialize cache with corresponding decoders. Then ran it with master and measured the execution time, then ran it with this branch again measuring the execution time. Both need 11 to 12 seconds on my computer, this branch was at most 500ms seconds slower. Thus, I don't think there can be made a strong argument that this impacts performance negatively.

Of course I checked the output of `ls -l --time=ctime ~/.cache/nengo/decoders/legacy.txt` too to verify that the ctime is actually set (which it is).

**Types of changes:**

(either one of the above depending on your point of view)

**Checklist:**

- [x] I have updated the documentation accordingly. [I don't think any changes are necessary.]
- [ ] I have added tests to cover my changes. [I could write a test for this, if we agree on making this change.]

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -235,6 +235,7 @@ class DecoderCache(object):
         """Checks if the legacy file is up to date."""
         legacy_file = os.path.join(self.cache_dir, self._LEGACY)
         if os.path.exists(legacy_file):
+            os.utime(legacy_file, None)
             with open(legacy_file, 'r') as lf:
                 text = lf.read()
             try:


drasmuss:
> It's a one line change. duh

Try to keep in mind that the tone in these pull requests sets the tone of Nengo for anyone starting out as a new user or developer.  Perhaps it would be more helpful to provide a link to that one line, as a positive way of sending the same message.

jgosmann:
Sorry, it's just a bit annoying to fill this out each time when it's a short PR where you can't possible provide any more useful information than you get by clicking the “Files changed” link.

tbekolay:
If a section doesn't make sense to have in there, then just remove it, like we do for other section like "Still to do" or "Interactions with other PRs" etc

jgosmann:
Ok, removed it.

jgosmann:
For comparison, to achieve the equivalent effect in the model code the following code is necessary:
```python
import os
import os.path

dc = nengo.cache.DecoderCache()
if not dc.readonly:
    try:
        with nengo.utils.lock.FileLock(os.path.join(dc.cache_dir, 'legacy.lock')):
            legacy_file = os.path.join(dc.cache_dir, dc._LEGACY)
            if os.path.exists(legacy_file):
                os.utime(legacy_file, None)
    except nengo.exceptions.TimeoutError:
        pass
```

It requires knowledge about cache implementation details that could change in the future:

* The name of the lock file (`legacy.lock`)
* The legacy file filename which is at least provided as an attribute, but it's private.

There a some things in this piece of code that are easy to miss or to get wrong. (Actually I missed some of them initially and it took me a couple minutes of contemplation to notice.)

* Access to the legacy file should be in a lock in case another process removes or overwrites it.
* The lock can raise a `TimeoutError`.
* The legacy file might not exist (which would make the call to `os.utime` fail)

Those situations might seem unlikely to occur so that one would not need to guard against them. But in my experience such situations occur sooner or later when running huge amount of processes in parallel.

drasmuss:
Perhaps you could put that code snippet into `nengo.utils.cache`?

tbekolay:
Perhaps this serves as a good example of something that I think a lot about when writing Nengo code: there's a big difference between the kind of situational robustness that you need in library code (like Nengo) and the kind of quick fixes you can write in model code. I think that difference is a big part of why I'm hesitant to add things like this to the Nengo library.

Let's keep in mind the original problem: a file that the cache needs gets auto-deleted by some external process unless it gets touched. You don't control that external process.

----

If we try to solve this problem in Nengo itself, then you end up with something like the snippet above (copied here so you don't have to scroll up):

```python
import os
import os.path

dc = nengo.cache.DecoderCache()
if not dc.readonly:
    try:
        with nengo.utils.lock.FileLock(os.path.join(dc.cache_dir, 'legacy.lock')):
            legacy_file = os.path.join(dc.cache_dir, dc._LEGACY)
            if os.path.exists(legacy_file):
                os.utime(legacy_file, None)
    except nengo.exceptions.TimeoutError:
        pass
```

It looks like this because, within Nengo, we have to be robust to a few situations:

- Maybe the cache is readonly
- Maybe the legacy file is being used by another Python process/thread running Nengo
- Maybe the legacy file doesn't exist yet

And probably there are other situations that we haven't predicted that we will end up having to guard for in the future.

That's just how library code is. It has to be robust because people will use the library in ways we might not anticipate, so we're constantly trying to anticipate those situations and guard against them.

----

But let's consider the case in which you solve this problem for yourself in your model code. You know the exact situation you're dealing with.

- You know that you're not setting your cache as readonly.
- You know that the legacy file exists.
- You know that the file only needs to be touched once every 62 days.
- You know when you're running Nengo processes because you run them.

That means you have a ton of different ways to solve this problem without having to guard for all of the other ways that people might be using Nengo. A few ways off the top of my head:

- You queue up your jobs with a `queue_jobs.sh` script. Add

   ```
   touch "/scratch/.cache/decoders/legacy.txt"
   ```
   to that script so that it'll be touched before your Nengo jobs even start. This works unless you queue up more than 62 days worth of jobs.
- You have some access to the job queuing mechanism and can ask it if you're currently running any jobs. Add a small script that `cron` runs every day that touches the legacy file if you're not currently running any jobs.
- You don't care if it works every time, as long as it works at least once every 62 days, so at the top of your model script you do

  ```python
  try:
      os.utime(os.path.join(nengo.utils.paths.decoder_cache_dir, "legacy.txt", None)
  except:
      pass
  ```

And probably there are many other ways of doing it. I could foresee writing a little set of helper scripts for running Nengo models on sharcnet and putting these type of snippets there. Or writing a post on [the forum](https://forum.nengo.ai/c/hpc) with tips and tricks for running Nengo on sharcnet, which other people could add to and we could sticky.

Any and all of these possible solutions above mean that we don't have to write and maintain library code that is robust to a bunch of unpredictable situations. As much as possible, we should let specific things be handled in specific places.

----

The ugliness of these specific solutions is subjective, but since there are lots of possible solutions, I'm confident that you can find one that looks right for your situation. I totally understand the desire for your model scripts to look perfect, but it's all a big balancing act trying to deal with all these different levels of abstraction. I'm sure if I looked back at the hoop-jumping I put in my model scripts running Java/Jython Nengo from Python I would cringe! But I don't have to maintain those models ;)

Anyway, if you're on board with my reasoning and can find a solution that works for you in your model script, would you be okay closing this PR unmerged?

jgosmann:
> Perhaps you could put that code snippet into nengo.utils.cache?

I would rather not, because it adds a lot of complexity (actually duplicates it from `cache.py`) and adds an obscure function to the public API.

> If we try to solve this problem in Nengo itself, then you end up with something like the snippet above [...]
> It looks like this because, within Nengo, we have to be robust to a few situations:

We already have to be robust to all those things in the cache code.

Despite that #1154 is a quite specific and non-critical problem, I am of the opinion that the proposed one-line change is ok because:
* The complexity and additional code to be maintained is minimal.
* In my view it is extremely unlikely that it leads to problems. There are two things why the `utime` call could fail:
  1. The file does not exist. But this is already guarded against with an `os.path.exists` call which is within a `FileLock` so that no other Nengo process will remove the file in-between the two function calls. (Other processes could, but if other processes start messing with the cache in this way, then good luck. That can break all sorts of stuff already and doesn't seem like something we should worry about.)
  2. The process has no write permissions. But the whole block of legacy code operates under the assumption that it has write permissions (which should be fixed, see #1162, but is unrelated to this change)
* One could certainly make a vague claim that this change could lead to problems we can't foresee right now, but one could equally make the claim that not making this change could lead to problems we can't foresee right now. (And there is one “problem” right now that would be solved by the change!)
* While it solves a quite specific problem, it doesn't seem unreasonable to say that the cache should update the modification times on its “global” files (i.e. `legacy.txt` and `index`). It is happening for `index` anyways and that it is not happening for `legacy.txt` is just a side-effect of a particular implementation. Rewrite `legacy.txt` each time would be also a working implementation, but changes the modification time. Technically, the information of `legacy.txt` could be included in the `index` file which gets rewritten each time.

Your proposing instead of making this change to document this in some place like the forum. But I'm not sure that this is indeed a better solution:
* Documentation also needs to be maintained and kept up-to-date. This seems especially hard to do when it is not included with the main documentation. There is also limited support for testing documentation to find outdated information.
* People actually need to read the documentation instead of getting something that “just works as intended”.
* Admittingly, if they don't read the part to this particular problem it's not a big deal. But this less relevant documentation might obscure more relevant parts. Or maybe it intimidates users if it appears to be a lot of effort to get something running on Sharcnet due to the length of the documentation or some obscure code that should be placed in front of scripts (I doubt that most are interested in why that code might be necessary).
* It encourages people to rely on implementation details (`DecoderCache._LEGACY`). Shouldn't we make that then part of the public interface of the decoder cache?

> Anyway, if you're on board with my reasoning and can find a solution that works for you in your model script, would you be okay closing this PR unmerged?

I think I summarized my view in one place if it doesn't convince you, go ahead and close the PR (and issue). It doesn't make sense to keep it open if it is not going to get merged, does it? ;)

Though, I won't guarantee that I will document this because of the reasons above, it's more work, and I'm not sure what the best way to present this information would be.

-----------

The last part of this post is not as relevant to the discussion, but I believe some of your assumptions about how well I know the situation I'm dealing with are not correct.

> You know that you're not setting your cache as readonly.

In the past I changed the cache from enabled to disabled or readonly a bunch of times for various reasons. If I do this I don't want to change all models.

> You know that the legacy file exists.

I don't. It might just have gotten deleted after those 62 days or maybe I manually cleared the cache (did that before).

> You know when you're running Nengo processes because you run them.

On Sharcnet I submit jobs to a scheduler that will run them at some later time. So I don't exactly know when they will run.

> You queue up your jobs with a queue_jobs.sh script.

I'm actually using my own Python tool/library for that which is intended to be somewhat general. It doesn't seem to be the right place to fix this problem.

tbekolay:
> the information of legacy.txt could be included in the index file which gets rewritten each time.

I'd be OK with that change. Less files, presumably less code to maintain, solves your problem... win-win.

jgosmann:
> presumably less code to maintain

Don't get excited to early. It means changing the format of the cache index which is a bigger change and seems more likely to introduce new bugs (at least initialliy, it might indeed be less code in the end).

jgosmann:
Given that I have to think about #1162 it might actually as well scrap the legacy file and do it all in one file. But I have to carefully think about that.

jgosmann:
Superseded by #1164.

Issue #1156, opened 2016-08-30
Strange error


jgosmann:
With this model

``` python
with nengo.Network(seed=seed, label="Circular convolution") as model:
    v = np.random.randn(2, dimensions)
    v /= np.linalg.norm(v, axis=1)[:, None]

    model.input_a = nengo.Node(v[0])
    model.input_b = nengo.Node(v[1])

    model.cconv = nengo.networks.CircularConvolution(
        n_neurons=n_neurons, dimensions=dimensions)
    nengo.Connection(model.input_a, model.cconv.A)
    nengo.Connection(model.input_b, model.cconv.B)

    model.p_input_a = nengo.Probe(model.input_a, synapse=0.01)
    model.p_input_b = nengo.Probe(model.input_b, synapse=0.01)
    model.p_output = nengo.Probe(model.cconv.output, synapse=0.01)
```

I get this strange error:

```
Traceback (most recent call last):
  File "gosmann_frontiers2016/benchmarks/circ_conv.py", line 28, in <module>
    with nengo.Simulator(model) as sim:
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.5/site-packages/nengo/simulator.py", line 131, in __init__
    self.model.build(network)
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.5/site-packages/nengo/builder/builder.py", line 119, in build
    return Builder.build(self, obj, *args, **kwargs)
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.5/site-packages/nengo/builder/builder.py", line 211, in build
    return cls.builders[obj_cls](model, obj, *args, **kwargs)
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.5/site-packages/nengo/builder/network.py", line 91, in build_network
    model.build(conn)
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.5/site-packages/nengo/builder/builder.py", line 119, in build
    return Builder.build(self, obj, *args, **kwargs)
  File "/home/jgosmann/.virtualenvs/gosmann-frontiers2016/lib/python3.5/site-packages/nengo/builder/builder.py", line 211, in build
    return cls.builders[obj_cls](model, obj, *args, **kwargs)
[...]

PR #1157, opened 2016-08-30
Track offset in Signals

jgosmann:
**Description:**
Makes Signals track the offset instead of computing it from the initial value.

**Motivation and context:**
See #1156, including [this comment](https://github.com/nengo/nengo/issues/1156#issuecomment-243519135).

**How has this been tested?**
Added two unit tests. One is just dependent on the Signal and tests the offset value. The other one reproduces the original problem with the node input from #1156 (a bit simplified).

Probably best to understand the problem first, then to look at the changes in `signal.py`.

**Types of changes:**

**Checklist:**

- [x] I have updated the documentation accordingly. [no changes necessary]

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -102,6 +102,9 @@ Release History
 - Added a registry system so that only stable objects are cached.
   (`#1054 <https://github.com/nengo/nengo/issues/1054>`_,
   `#1068 <https://github.com/nengo/nengo/pull/1068>`_)
+- Nodes now support array views as input.
+  (`#1156 <https://github.com/nengo/nengo/issues/1156>`_,
+  `#1157 <https://github.com/nengo/nengo/pull/1157>`_)
 
 2.1.2 (June 27, 2016)
 =====================
--- a/nengo/builder/signal.py
+++ b/nengo/builder/signal.py
@@ -39,7 +39,8 @@ class Signal(object):
     # up in a model.
     assert_named_signals = False
 
-    def __init__(self, initial_value, name=None, base=None, readonly=False):
+    def __init__(self, initial_value,
+                 name=None, base=None, readonly=False, offset=0):
         self._initial_value = np.asarray(initial_value).view()
         self._initial_value.setflags(write=False)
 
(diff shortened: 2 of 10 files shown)


studywolf:
> Where should a reviewer start?
Probably best to understand the problem first, then to look at the changes in signal.py.

Dude, c'mon. This isn't conducive to a friendly / helpful environment.

jgosmann:
Honestly, it is not clear to me what is offensive about this sentence.

To clarify anyways: There are PRs where it is reasonable to dive directly into the diff. I don't think this is one of these PRs, it is probably easier to understand the changes after understanding the problem. That's why I suggest starting with that part first.

studywolf:
It's not offensive, it just comes across as sarcastic / curt. Fair enough the comment describing the problem (which explains it quite well!) is linked in the motivation, but just reposting it in that section would be much more friendly / useful: 

> So, I figured out what's causing this (but I'm not sure how to fix it yet):
An array view is used as initial_value for a signal (that is not a view). From that signal a signal view is created which creates a view of the initial_value which results in a view of view of an array. But Numpy collapses this hierarchy to just one view of a signal. When initializing the signal dict the view offset will thus be calculated relative to the base array, but the (first) view of the array will be used for the data array and thus the offset is invalid.

I get sometimes people can dive straight into the diff, but suggesting people understand the problem to help understand the solution isn't really helpful :stuck_out_tongue_winking_eye: It's just a tone thing too, eh? Emoji's help too :tada: :hand: :cake: :smiling_imp:

jgosmann:
I suppose, I can see that it comes across as curt. Sometimes filling in all that information feels to be more work than the actual changes. That's certainly warranted in some cases, but with straight-forward changes it starts to annoy me. Sure, I could delete redundant/unhelpful sections in those cases, but to me the template imposes a psychological barrier to do so.

Anyways, I'll try to better myself.

Seanny123:
LGTM.

tbekolay:
Sorry it took so long for me to get to this one! Looked it over now and played around with NumPy enough so I understand it ;) Looks great! I made a few fixup commits for minor style and to add an `__init__.py` to `builder/tests`, and then a commit to move the builder tests to where they probably should have been all along. If these look to you @jgosmann I'll merge.

jgosmann:
LGTM :cake:

tbekolay:
TravisCI is complaining because `flake8` recently updated with some new rules, so I'm going to sneak in a commit to fix those issues in this branch too.

PR #1158, opened 2016-08-30
Make PR template more concise.

jgosmann:
**Description:**
Given today's discussions, I propose some changes to the PR template to make it more concise.

There are two changes (in two separate commits):
* The description and motivation/context sections are sometimes overlapping, or sometimes it is better to start with the motivation before describing the changes. Consolidating them gives some more flexibility in this.
* The “Where should a reviewer start?” section is rarely useful. I went through a bunch of recent PRs and they pretty much all say “Files changed tab” (or something equivalent) or “read discussion X, than look at changes”. I found two PRs where actually some additional information was provided: #1112 and #1035. Given the rare use I could see completely removing that section, but for now I just inserted “Files changed.” as a default.

Files changed.

**Types of changes:**

- New feature (non-breaking change which adds functionality) :question: 

**Checklist:**

- [x] I have included a changelog entry. [not needed as no change to Nengo itself]
- [x] I have added tests to cover my changes. [nothing to test]

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,6 @@
-**Description:**
-<!--- What does this PR do? -->
-
 **Motivation and context:**
 <!--- Why is this change required? What problem does it solve? -->
-<!--- If it fixes an open issue, please link to the issue here. -->
+<!--- If it addresses an open issue, please link to the issue here. -->
 
 **Interactions with other PRs:**
 <!--- If this change depends on or conflicts with another PR please list it here. -->
@@ -13,9 +10,6 @@
 **How has this been tested?**
 <!--- Please describe in detail how you tested your changes. -->
 <!--- Reviewers will test your PR in at least this way. -->
-
-**Where should a reviewer start?**
-<!--- Indicate the best part to start looking to understand the changes. -->
 
 **How long should this take to review?**
 <!--- Please estimate if this PR is a quick, average, or lengthy PR. -->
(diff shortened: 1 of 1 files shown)


tbekolay:
I agree, there's a lot of overlap between the description and motivation and context, so I'm +1 on combining them. However, I've found that I prefer the motivation and context being front and center, because the description should follow naturally from that (and is already sort of hinted at in the title of the PR), so I swapped them in my fixup. You can see that a lot of my recent [commit messages](https://github.com/nengo/nengo/commits/master) start by talking about the context first... not that my commit messages are the gold standard or anything, it's just been a helpful thing I've been using to structure my messages lately.

I'm not sure that saying "Files changed" is particularly helpful since we talk about reading diffs at length in the [reviewing guidelines](https://github.com/nengo/nengo/blob/master/docs/contributing.rst#reviewing), so I instead moved the "where should a reviewer start" section to after the length estimation, and in the comments noted that this section should be removed unless you just said that the PR is lengthy (or average and complicated), like we do with the Interactions section.

Finally, since we're touching this file anyway and it's come up a few times, I added a new type of change, which is a Non-code change (which encompasses tests, docs, build scripts, and other meta things, like this PR!)

Thanks for taking another look at the template, it's hard to know what will end up being onerous / useful until we use a bunch. Definitely better with these changes!

jgosmann:
> However, I've found that I prefer the motivation and context being front and center, because the description should follow naturally from that (and is already sort of hinted at in the title of the PR), so I swapped them in my fixup.

Sounds good to me. :)

> so I instead moved the "where should a reviewer start" section to after the length estimation, and in the comments noted that this section should be removed unless you just said that the PR is lengthy

What I dislike about that is that it still requires some effort for the majority of (small) PRs. It would be nice if the amount of work filling out the PR is proportional to the size/complexity of the PR. I am happy to fill in that section when it's required for a complex PR, but it's bothersome to do it for every small PR with essentially the same statement over and over again. That's why I would like to have prefilled, so that I can just leave it in.

> I added a new type of change, which is a Non-code change

+1 but I can't see it in the changelog?

drasmuss:
Looks good (with @tbekolay's fixup)

drasmuss:
(I'll wait for @tbekolay to add in the Non-code change thing and then I can merge)

tbekolay:
Whoops! Pushed, merge away!

PR #1164, opened 2016-09-02
Refactor decoder cache.

### LINKED ISSUE #1154: Decoder cache should “touch” legacy.txt each time

### LINKED ISSUE #1162: Readonly cache might modify cache
Talking to @celiasmith I became aware of a potential problem with the current cache implementation. I haven't verified it yet, but from looking at the code it seems that the cache might perform write operations even when set to readonly.

In particular:
- `_remove_legacy_files` might remove files and write `legacy.txt`
- The cache index might be written back to the disk.

Both operations don't change the cache contents of an up-to-date cache, but just doing the write operation might be problem in some situations (e.g., readonly file system). With a non-up-to-date cache this would remove files and thus actually change the cache.

jgosmann:
**Motivation and context:**
This does a number of things:
- It saves the cache version information in the cache index file. This fixes #1154. It also simplifies things a little bit because only one file has to be read and we only have to take care of synchronizing the access to this one file.
- It splits up the cache index in two classes, one for read-only access and one for read/write access. The intent is to reduce if-statements by allowing to just switch out the implementation. It also structures the code in a way that makes it easier to check that the read-only class is not doing any writes The read/write class is shorter too because it only needs to implement the write and appropriate locking operations.
- The slit up in two classes makes the error handling with graceful degradation to read-only and then no cache at all easier.

The latter two changes fix #1162.

This PR would make #1155 unnecessary. #1155 is less likely to introduce new bugs because it introduces fewer changes, but in the long term this PR might be better as the general structure fo the code in improved.

**How has this been tested?**
All tests are still passing. I also ran the following script a couple of times alternating between this branch and the `master` branch to ensure that switching Nengo versions works.

``` python
import nengo

nengo.log('debug')

with nengo.Network(seed=1) as model:
    a = nengo.Ensemble(10, 1)
    b = nengo.Ensemble(10, 1)
    nengo.Connection(a, b)

    pass
```

I'd recommend starting bottom up. That is reviewing `CacheIndex` first which provides only readonly access. Then take a look at `WriteableCacheIndex` which provides write access and thus has to implement synchronized access. Hopefully, it becomes clear from those two classes how they are supposed to be used to then look at the changes in `DecoderCache`. Of particular interest in that class is the `__enter__` method that provides a graceful degradation from a writeable cache to a readonly cache if no write access could be obtained, and if even that fails it degrades to a deactivated cache (but should not crash).

**Types of changes:**

**Checklist:**

- [ ] I have included a changelog entry. [These are internal changes, do we need a changelog entry anyways?]
- [ ] I have added tests to cover my changes. [No new tests because no new feature, only refactoring.]

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -12,7 +12,7 @@ import warnings
 
 import numpy as np
 
-from nengo.exceptions import FingerprintError, TimeoutError
+from nengo.exceptions import CacheIOError, FingerprintError, TimeoutError
 from nengo.neurons import (AdaptiveLIF, AdaptiveLIFRate, Direct, Izhikevich,
                            LIF, LIFRate, RectifiedLinear, Sigmoid)
 from nengo.rc import rc
@@ -22,7 +22,7 @@ from nengo.solvers import (Lstsq, LstsqL1, Nnls, NnlsL2,
 from nengo.utils import nco
 from nengo.utils.cache import byte_align, bytes2human, human2bytes
 from nengo.utils.compat import (
-    int_types, is_string, pickle, replace, PY2, string_types)
+    int_types, is_string, iteritems, pickle, replace, PY2, string_types)
 from nengo.utils.least_squares_solvers import (
     Cholesky, ConjgradScipy, LSMRScipy, Conjgrad,
     BlockConjgrad, SVD, RandomizedSVD)
@@ -152,7 +152,7 @@ class Fingerprint(object):
         (tuple, check_seq),
         (list, check_seq)
     ] + [
-        (x, check_attrs) for x in SOLVERS + LSTSQ_METHODS + NEURON_TYPES
+        (_x, check_attrs) for _x in SOLVERS + LSTSQ_METHODS + NEURON_TYPES
     ])
 
     def __init__(self, obj):
@@ -171,6 +171,11 @@ class Fingerprint(object):
 
     @classmethod
     def supports(cls, obj):
+        """Determines whether ``obj`` can be fingerprinted.
+
+        Uses the `.whitelist`  method and runs the check function associated
+        with the type of ``obj``.
+        """
         typ = type(obj)
         in_whitelist = typ in cls.WHITELIST
         succeeded_check = typ not in cls.CHECKS or cls.CHECKS[typ](obj)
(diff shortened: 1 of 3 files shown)


jgosmann:
Should be ready for review now.

Seanny123:
I've been avoiding reviewing this PR, because I don't have a good understanding of the cache system. However, does anyone else other than @jgosmann totally understand the cache system? Should I dive in since no one else is likely to?

tbekolay:
> Should I dive in since no one else is likely to?

That would be great if you don't mind! :rainbow: :octopus:

nengo/cache.py
```diff
@@ -205,49 +314,80 @@ def __delitem__(self, key):
...
         with self._lock:
-            self._index = self._load_index()
+            try:
+                self._load_index()
+            except:
+                self._reinit()
+                self._index = {}
         return self
 
+    def _reinit(self):
```
Seanny123: Could you add a comment about what this is doing? What is it reinitializing specifically?

nengo/cache.py
```diff
@@ -272,45 +412,45 @@ class DecoderCache(object):
...
+                self.readonly = True
+                self._index = CacheIndex(self.cache_dir)
                 self._index.__enter__()
-        except TimeoutError:
-            warnings.warn(
-                "Decoder cache could not acquire lock and was deactivated.")
+                warnings.warn(
+                    "Decoder cache could not acquire lock and was to "
+                    "readonly.")
+        except:
```
Seanny123: Is it possible to have an alternative to this bare except?

nengo/cache.py
```diff
@@ -272,45 +412,45 @@ class DecoderCache(object):
...
             self.readonly = True
+            self._index = None
+            warnings.warn(
+                "Decoder cache could not acquire lock and was to deactivated.")
```
Seanny123: Do you mean "was deactivated"?

jgosmann: yes

nengo/cache.py
jgosmann: I don't think so. No matter what exception occurs, this function should fail gracefully. (Otherwise some random cache initialization failure would kill all of your simulation.)

tbekolay:
Took a look at this one. With the suggested squash etc commits it looks good to me! Here are my notes from my review -- the highlights I suppose, as I tried to put everything in commit messages.

- It wasn't clear to me that the cache index contained tuples of `(filename, start, end)`... it feels like it's trying to be generic and allow anything to be put in as value, but that doesn't actually work because we make assumptions in a few places that it's going to be a tuple (e.g., `self._index = {k: v for k, v in iteritems(self._index) if v[0] not in self._removed_files}`). Tried to clarify this and guard against incorrect input where possible...
- I was not aware that you can actually shove multiple objects into one pickle file! In the past I've just crammed everything into a big list and unpacked it, but that's only because I thought you could only store one object in there. Cool!
- One confusing thing here is when absolute paths are being used vs. when paths are relative to the cache_dir... in most cases they're relative, but that wasn't being handled correctly in `WriteableCacheIndex._removed_file`. I made it so that it works with relative or absolute, but that could be changed to require relative or something (but it needs to be clear what's happening).
- I think I addressed the comments @Seanny123 made so I removed the `needs changes` label...
- Found a few bugs that weren't caught by the test suite, so added tests for them and fixed the bugs (see commits for more details).

Also when this is rebased to master the flake8 error will go away. Waiting on @Seanny123 to finish his review before that though.

nengo/cache.py
```diff
@@ -327,40 +341,42 @@ def __getitem__(self, key):
...
+                for f in os.listdir(self.cache_dir):
+                    path = os.path.join(self.cache_dir, f)
+                    if path == self._lock.filename:
+                        continue
+                    if os.path.isdir(path):
+                        shutil.rmtree(path)
+                    else:
+                        os.remove(path)
                 self._index = {}
+                self._write_index()
```
jgosmann: I'm opposed to inlining this code. A function preferrably have one level of abstraction which is violated with the inlined code and makes it harder to understand. In the orginal version the code tells me immediatly: “try to load the index and if that fails reinitialize the cache" and I have only to look at the details of how the cache is reinitialized if that is actually relevant to me. With the inlined code it is not immediatly clear if all the code reinitializes the cache or whether part of it is maybe for some other error handling that is required in addition.

Especially with complex code like here, where there are a lot of things to keep in mind due to potential concurrent access, it seems helpful for me to try to keep to one level of abstraction per function. It makes it easier to reason what should happen in each function.

jgosmann:
Apart from the code inlining (see other comment), those changes look good to me.

nengo/cache.py
Seanny123: I also prefer the not-inlined version.

Seanny123:
I'm also fine with these changes and the PR as a whole, except for the inlined code, which I've also commented on.

nengo/cache.py
tbekolay: I disagree, but two against one so I'll put it back.

tbekolay:
Reverted the inlining and rebased etc! Will merge once CI finishes.

PR #1165, opened 2016-09-07
Remove code that was only there to support javaviz

jgosmann:
**Motivation and context:**
This removes code that was only there to support Javaviz. We are not using javaviz anymore and the code has become outdated so that it would fail if it ever would be executed. (The `VocabularyParam` now requires an additional argument.)

Supersedes #1005 that fixed the code before realizing that it is obsolete.

**How has this been tested?**
The same code has been removed in #1006. I have been using a branch based of those changes for quite a while now without problems.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/spa/__init__.py
+++ b/nengo/spa/__init__.py
@@ -11,5 +11,5 @@ from .pointer import SemanticPointer
 from .spa import SPA
 from .state import State
 from .thalamus import Thalamus
-from .utils import enable_spa_params, similarity
+from .utils import similarity
 from .vocab import Vocabulary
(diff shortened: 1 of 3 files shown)


hunse:
@mention-bot, you're so helpful!

I think this is okay. I tested it briefly with one of the built-in SPA examples in nengo_gui, and it was fine. If @tcstewar is fine with this, I think it's good to go.

Since the javavis stuff hasn't been working for a while, and I'm not sure if we ever officially really supported it, I think this is fine without a changelog entry.

tcstewar:
Nice.  Thank you for getting rid of this!  LGTM...

PR #1168, opened 2016-09-08
Add function to activate direct mode.

jgosmann:
**Motivation and context:**
Sometime I want to run networks in direct mode, but some parts use connections to or from neurons. In those cases it is not possible to set the top-level `neuron_type` to `Direct` because some ensembles need to still use neurons. This PR adds a function where direct mode is only activated for ensembles where this is possible.

Not sure whether this better fits into `nengo_extras`, but to me it seems like something pretty basic that one might need for many networks.

**How has this been tested?**
Added a unit test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,10 @@ Release History
 
 - Added documentation on config system quirks.
   (`#1224 <https://github.com/nengo/nengo/pull/1224>`_)
+- Added ``nengo.utils.network.activate_direct_mode`` function to make it
+  easier to activate direct mode in networks where some parts require neurons.
+  (`#1111 <https://github.com/nengo/nengo/issues/1111>`_,
+  `#1168 <https://github.com/nengo/nengo/pull/1168>`_)
 
 **Fixed**
 
--- a/nengo/utils/network.py
+++ b/nengo/utils/network.py
@@ -1,3 +1,4 @@
+import nengo
 from .magic import decorator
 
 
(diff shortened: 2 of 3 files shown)


tbekolay:
> it seems like something pretty basic that one might need for many networks

+1 for having it in Nengo! Handy thing to have around for sure.

nengo/utils/network.py
```diff
@@ -25,3 +26,31 @@ def add_one_2(self):
...
+    ----------
+    network : nengo.Network
+        Network to activate direct mode for.
+    """
+    requires_neurons = set()
+
+    for c in network.all_connections:
+        if isinstance(c.pre, nengo.ensemble.Neurons):
+            requires_neurons.add(c.pre.ensemble)
+        if isinstance(c.post, nengo.ensemble.Neurons):
```
tcstewar: These should be ```c.pre_obj``` and ```c.post_obj``` so that you can catch situations like ```nengo.Connection(a.neurons[0], b.neurons[0])```

tcstewar:
Nice...  yup, there's been a few times where I've wanted this sort of thing.  I've even tried a few times to implement something like this that will also add in a bit of saturation behaviour (by modifying the function on the Connections), but never got it to a state I was happy about.....

hunse:
LGTM too.

My inclination was to say this belongs in `nengo_extras`, but it's okay here, too.

CHANGES.rst
```diff
@@ -53,6 +53,10 @@ Release History
   `#1148 <https://github.com/nengo/nengo/pull/1148>`_)
 - Tweaked ``rasterplot`` so that spikes from different neurons don't overlap.
   (`#1121 <https://github.com/nengo/nengo/pull/1121>`_)
+- Added ``activate_direcd_mode`` function to make it easier to activate direct
```
Seanny123: Typo.

jgosmann:
Since I wrote this, it occured to me that this function should also handle learning rules which might require non-direct mode ensembles.

jgosmann:
Excluded learning rule ensembles from direct mode. This is now ready for a re-review.

PR #1169, opened 2016-09-09
Fixes for 2.2.0 release

tbekolay:
**Motivation and context:**
Was following our steps for releasing, and got some test failures for slow tests. Turns out our assumption that some of the SPA code was there just to support JavaViz was incorrect (see #1005, #1165) and caused the SPA examples to fail.

Since I was had to make that change for 2.2.0 anyway, I also did a bunch of other minor fixes that I've been queueing up for the next release... none of them are mission critical so not worth making a PR or changelog entry for, but hey, still worth doing. See b985ecc85565d75d585b4 for those details (happy to split those into separate commits if people think it'd be better that way).

**How has this been tested?**
Generated the documentation and ran all the slow tests.

**Types of changes:**

**Checklist:**

- [NA] I have updated the documentation accordingly.

**Notes:**

Once this is reviewed and merged, I'll cut the 2.2.0 release.

(sidenote: how do people feel about putting NA in there for not applicable? I could add it to the PR template in this PR if it feels good)

### DIFF
--- a/nengo/config.py
+++ b/nengo/config.py
@@ -390,13 +390,13 @@ class Config(object):
 
 
 class SupportDefaultsMixin(object):
-    """Mixin to support assigning ``Default`` to parameters.
+    """Mixin to support assigning `.Default` to parameters.
 
-    Implements ``__setattr__`` to do so. If the inheriting class overrides this
-    method, it has to call the mixins `__setattr__`.
+    Implements ``__setattr__`` to do so. If the inheriting class overrides
+    this method, it has to call the mixin's ``__setattr__``.
 
-    This mixin will also simplify the exception if the parameter value is
-    invalid given the `simplified` rc option is set
+    This mixin may simplify the exception depending on the value of the
+    ``simplified`` rc option.
     """
 
     def __setattr__(self, name, val):
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -337,13 +337,6 @@ class Connection(NengoObject):
         The slice associated with ``pre`` if it is an ObjView, or None.
     seed : int
         The seed used for random number generation.
-    size_in : int
-        Input size of the computed function, equal to ``pre.size_out``.
-    size_mid : int
-        Output size of the function ``len(function(np.zeros(size_in)))``,
-        equal to ``size_in`` if ``function is None``.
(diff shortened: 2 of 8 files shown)


jgosmann:
b985ecc LGTM
I'm unsure about e7b0edc. If we made the `vocab` argument of the `SPA.similarity` method mandatory, it wouldn't be necessary. (I actually always used `spa.utils.similarity`, so I never ran into this.)

> (sidenote: how do people feel about putting NA in there for not applicable? I could add it to the PR template in this PR if it feels good)

It breaks the markdown rendering somewhat. :/

tbekolay:
> If we made the vocab argument of the SPA.similarity method mandatory, it wouldn't be necessary.

Could be, we'd have to edit the SPA examples then, and yeah, I was hoping that we could just leave all that kind of stuff for after the split. Do you think it would be quick and easy? If so that's probably fine, but yeah, it feels like we wouldn't want to make API changes to SPA right now.

> It breaks the markdown rendering somewhat. :/

Yeah, that's kind of the point since in the [PR list](https://github.com/nengo/nengo/pulls) it lists how many boxes are checked and unchecked. By putting NA in there it removes it from that count... right now, I think we mostly ignore these checkboxes because it's valid for some of them to be unchecked sometimes. This way we are explicit that this checkbox shouldn't be checked (because it's not applicable in this PR). Not saying it's a perfect solution, but there have been a bunch of times that we haven't added changelog entries because we're mostly glossing over this list. Happy to hear other proposals, but also we should probably move that to a separate issue / PR!

jgosmann:
> Do you think it would be quick and easy?

Yes, it is quick and easy.

> If so that's probably fine, but yeah, it feels like we wouldn't want to make API changes to SPA right now.

That's why I'm unsure about it.

jgosmann:
Given the discussion from the dev meeting today, I'm fine with both commits.

hunse:
LGTM. No point in trying to fine-tune SPA stuff while it's still part of this repo, IMO, so just reverting back to the working behaviour makes sense to me.

tbekolay:
Cool cool, I'll do the release once someone merges (@hunse, mind doing the honors?)

PR #1173, opened 2016-09-13
Deprecate `sim.model.params` in favour of `sim.data`

jgosmann:
**Motivation and context:**
As discussed and decided in #1145 static data generated during the build (e.g. eval points, encoders) should be accessed via the `sim.data` attribute. This PR deprecates the access to `sim.model` (a `DeprecationWarning` will be issued when accessing the attribute).

One thing that might be controversial is whether there might be legit use cases to access `sim.model` (as it provides more than just `sim.model.params`). But to me it seems like the model does not necessarily belong in the public interface of the simulator and all tests were easily adjusted for the change.

Just deprecating `sim.model.params` would be more complicated because it should only be deprecated when accessed via `sim.model.params`, but not when accessed within the build process.

**How has this been tested?**
The first commit completely removes the access to `sim.model` to make all tests where it is used fail. All those places where fixed in the same commit. Thus all tests (including `--slow` tests) pass without the need to access `sim.model`. The second commit reintroduces `sim.model` with a deprecation warning and a test for this warning.

**Types of changes:**

**Checklist:**

**Still to do:**

In theory, it should be documented that backends should provide access to certain things via `sim.data`. But we do not have a general guide for backend developers yet, so I have no idea where I put this information.

We should also add some documentation on how to “probe” those static build data, but couldn't figure out a good place for that so far.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,12 +15,21 @@ Release History
 
    - Added
    - Changed
+   - Deprecated
    - Removed
    - Fixed
 
 2.2.1 (unreleased)
 ==================
 
+**Deprecated**
+
+- Access to ``nengo.Simulator.model`` is deprecated. To access static data
+  generated during the build use ``nengo.Simulator.data``. It provides access
+  to everything that ``nengo.Simulator.model.params`` used to provide access to
+  and is the canonical way to access this data across different backends.
+  (`#1145 <https://github.com/nengo/nengo/issues/1145>`_,
+  `#1173 <https://github.com/nengo/nengo/pull/1173>`_)
 
 
 2.2.0 (September 12, 2016)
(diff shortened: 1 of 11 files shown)


Seanny123:
LGTM. Let me know when you've made a changelog entry.

tbekolay:
This LGTM too :+1: Will merge once the changelog is pushed!

jgosmann:
Doing that now :)

jgosmann:
done :)

CHANGES.rst
```diff
@@ -21,6 +21,15 @@ Release History
 2.2.1 (unreleased)
 ==================
 
+**API changes**
```
tbekolay: In 2.2.0 + we're switching the changelog format to be Added/Removed/Changed/Fixed; removed probably makes sense right? Or do we want a deprecated section?

jgosmann: Deprecated would be more precise. Could also be changed. But I'm also fine wtih removed.
¯\_(ツ)_/¯

tbekolay: I'll go with deprecated then, no harm in more sections ;)

Issue #1076, opened 2016-05-30
Repeated sim.data[probe] accesses are very slow


arvoelke:
```
import time
import nengo

with nengo.Network() as model:
    x = nengo.Ensemble(1000, 1)
    p_neurons = nengo.Probe(x.neurons)
    
with nengo.Simulator(model) as sim:
    sim.run(10)

# Repeated probe accesses
start = time.time()
for m in range(x.n_neurons):
    sim.data[p_neurons][0, m]
print time.time() - start

# Only one probe access
start = time.time()
u = sim.data[p_neurons][0, :]
for m in range(x.n_neurons):
    u[m]
print time.time() - start
```

It takes 37 seconds on my computer to access a mere 1000 spikes. It looks like every time `sim.data[p_neurons]` is accessed the entire array is re-copied. And of course the same effect is seen when accessing various time points.

Perhaps we should be caching the read-only object?

PR #1175, opened 2016-09-21
Cache probe data on access.

jgosmann:
**Motivation and context:**
Repeated access to probes is slow (see #1076). This caches the read-only NumPy array to make it faster.

**How has this been tested?**
Ran this script adopted from @arvoelke in #1076 on master and this branch:
```python
from __future__ import print_function

import time
import nengo

    x = nengo.Ensemble(1000, 1)
    p_neurons = nengo.Probe(x.neurons)

    sim.run(10)

# Repeated probe accesses
start = time.time()
for m in range(x.n_neurons):
    sim.data[p_neurons][0, m]
print(time.time() - start)

# Only one probe access
start = time.time()
u = sim.data[p_neurons][0, :]
for m in range(x.n_neurons):
    u[m]
print(time.time() - start)
```
The repeated access block took 17s on master vs. 0.017s on this branch. The second single access block was reduced to 0.00017s from 0.017s. (I don't know what's up with all the 17s ... :ghost: )

The usual place: “Files changed”
Note that the cache does not need to use weak references because the `raw` dict in the `ProbeDict` class doesn't use weak references either. Thus, as long as the `ProbeDict` instance is around, there will always be a strong reference to each key.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/tests/test_simulator.py
+++ b/nengo/tests/test_simulator.py
@@ -53,6 +53,19 @@ def test_probedict():
     assert np.all(probedict.get("list") == np.asarray(raw.get("list")))
 
 
+def test_probedict_with_repeated_simulator_runs(RefSimulator):
+    with nengo.Network() as model:
+        ens = nengo.Ensemble(10, 1)
+        p = nengo.Probe(ens)
+
+    dt = 0.001
+    with RefSimulator(model, dt=dt) as sim:
+        sim.run(0.01)
+        assert len(sim.data[p]) == 10
+        sim.run(0.01)
+        assert len(sim.data[p]) == 20
+
+
 def test_close_function(Simulator):
     m = nengo.Network()
     with m:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,13 @@ Release History
 2.2.1 (unreleased)
 ==================
 
+**Changed**
+
+- Access to probe data via ``nengo.Simulator.data`` is now cached,
+  making repeated access much faster.
+  (`#1076 <https://github.com/nengo/nengo/issues/1076>`_,
+  `#1175 <https://github.com/nengo/nengo/pull/1175>`_)
+
 **Deprecated**
 
 - Access to ``nengo.Simulator.model`` is deprecated. To access static data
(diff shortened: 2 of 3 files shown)


Seanny123:
LGTM.

tbekolay:
Awesome, that's a huge speedup! :rainbow: 

Looking through the diff, will this give incorrect results in the following case?

```python
    sim.run(0.1)
    d = sim.data[probe]
    sim.run(0.1)
    assert d.shape != sim.data[probe].shape
```

If so, then I can see two possible solutions... either invalidate the cache in `Simulator.step`, or only enable the cache once we've closed the sim. Invalidating seems better, but since we call step a bunch, it might slow things down slightly...

jgosmann:
Good catch!
I pushed a test and fix for this. I opted for a slightly different solution where I check the length of the array and list in the probe dict. That's is still a quick operation on access, allows to do the caching before `sim.close()` and doesn't require any additional code in the `step` function.

tbekolay:
> I pushed a test and fix for this. I opted for a slightly different solution where I check the length of the array and list in the probe dict.

Nice, that's a better fix for sure! I'll take a closer look now, likely merge shortly.

nengo/tests/test_simulator.py
```diff
@@ -53,6 +53,19 @@ def test_probedict():
     assert np.all(probedict.get("list") == np.asarray(raw.get("list")))
 
 
+def test_probedict_with_repeated_simulator_runs(RefSimulator):
+    with nengo.Network() as model:
+        ens = nengo.Ensemble(10, 1)
+        p = nengo.Probe(ens)
+
+    dt = 0.001
+    with nengo.Simulator(model, dt=dt) as sim:
```
tbekolay: Note that this should be `RefSimulator`. I've fixed it locally.

jgosmann: I though of that when writing the function signature, but then I forgot when writing the actual code. :fearful:

tbekolay: Haha, force of habit :wink: :rainbow:

Issue #1176, opened 2016-10-03
NengoObject should error on iteration


hunse:
A `NengoObject` should raise an error if you iterate over it. Right now, something like this causes an infinite loop:

``` python
u = nengo.Node(1.)
for uu in u:
    print(uu)
```

This infinite loop will also occur if you try to multiply (or add, etc) `u` with a Numpy array. This can cause annoying bugs if one accidentally uses just `probe` instead of `sim.data[probe]`.

I recommend we just raise an error in the `__iter__` function for `NengoObject`.

jgosmann:
I think it might be better to check the dimensions in the `__getitem__` method or `ObjView` constructor and raise an `IndexError` when appropriate. This would give an early error when trying to do invalid slicing in connections and it should make iteration work as expected.

hunse:
Be my guest! Personally, I'm not sure if there's a lot of use for iterating over these things, but if you think it's worthwhile and not too difficult to implement and won't cause confusion, then go ahead. If that's something that can be done soon, then it can supercede #1177. Otherwise, I'd suggest going ahead with #1177 until iteration is implemented.

jgosmann:
Interstingly `ObjView` is already partially trying to do what I suggest. However, the code does not work as intended as it uses slices which do not raise `IndexError`s.

jgosmann:
Actually it is not entirely clear to me what the desired behaviour in some cases should be. Assume we have a three-dimensional ensemble `e = nengo.Ensemble(50, 3)` and let us walk through the different cases.

### Integer indexing

Valid indices like `e[0]` should obviously return the appropriate `ObjView`. Invalid integer indices like `e[4]` should raise an `IndexError` to be consistent with how Python indexing works in general and prevent infinite loops (currently no error is raised).

### Slice indexing

For slices in range like `e[1:3]`, `e[1:]`, `e[:2]`, `e[:]` again the appropraite `ObjView` should be returned. However, it is not clear to me what should happen with out of range slices like `e[1:5]`, `e[5:10]`, `e[:10]`, `e[-10:]`. Python's standard behaviour is to truncate the out-of-range part of the slice. But maybe we would prefer to raise an error instead? Maybe also because it is not clear in all cases how the truncation behaviour should work. Consider this somewhat contrived example:

```python
e = nengo.Ensemble(50, 2)
c = nengo.Connection(e[1:], post)
e.dimensions = 4
```

[...]

PR #1177, opened 2016-10-03
Iterating over NengoObject raises error

hunse:
**Motivation and context:**

Iterating over a `NengoObject` would cause infinite loops. See #1176.

**How has this been tested?**

Tested on a few toy examples. Now multiplying a NengoObject by a Numpy array, for example, causes an error.

**Types of changes:**

**Checklist:**

I don't think this needs additional tests. I also think it's fine without a changelog entry, though one wouldn't hurt (please go ahead).

The only possible thing to do would be to make sure we don't have any other objects in Nengo that this could cause an issue for. I already checked `Network` and it's fine.

### DIFF
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -63,6 +63,9 @@ class NengoObject(with_metaclass(NetworkMember, SupportDefaultsMixin)):
     def __setstate__(self, state):
         raise NotImplementedError("Nengo objects do not support pickling")
 
+    def __iter__(self):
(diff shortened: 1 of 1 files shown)


Seanny123:
Tests still seem to be failing on this pull request. Also, given that #1183 has been created and it targets the same bug, is this PR now obsolete?

tbekolay:
#1183 does indeed make this PR obsolete, as it's now possible to iterate over Nengo objects.

Issue #887, opened 2015-10-26
Create standard for networks with two inputs


Seanny123:
There are a lot of networks with two input nodes: `Compare`, `Bind`, `CircularConvolution` and `Product`. Currently they don't all use the same input standard. `Compare` use `inputA` and `inputB` and the others use `A` and `B`. None of these actually obey the typical python naming conventions, given that one uses camel-case, while the other starts with an upper case despite only being an attribute.

Should we change `Compare` to conform or should we change them all to actually be in line with normal Python?

jgosmann:
Just came across this and I would like to propose to potentially decide on this and include it in the 2.1 release. Because it might change the API of these networks, it doesn't seem like we would want to introduce these changes in a 2.1.1 release. So if we don't do it in 2.1, it probably won't happen before 2.2? Also, it seems to me benificial to have this standardized earlier than later.

tbekolay:
I'd rather leave it to 2.2, with the understanding that there will be a much shorter development cycle for 2.2 compared to 2.1 ;)

tcstewar:
I agree with putting it off till 2.2 and having a shorter development time.  :)

jgosmann:
ok :)

jgosmann:
> Compare use inputA and inputB and the others use A and B

It's even worse. Depending on how you connect to `Compare` you either have to use `inputA` and `inputB` or `A` and `B`. I feel using `A` and `B` is much more common in the code base and I'd vote for that variant (but I haven't gone through the code base and checked that my feeling is actually right).

jgosmann:
Another option would be to use lists (and potentiall dictionaries) to address the inputs as `inputs[0]` and `inputs[1]`. However, this might not be easily possible in the SPA action system and I have to check that.

jgosmann:
It seems to me that adding support for input lists is not straight-forward in the current SPA system (and I'm not likely to make the required changes). With the rewritten SPA compiler (#1014), however, it should be reasonably straight-forward. Unfortunately, the current plan is to not merge this into main Nengo, but to create a separate package and that will probably take a while to handle the logistics and add required features to nengo_gui.

jgosmann:
This question should also be answered for #431. Unlike other existing networks, the order of inputs matters in this case.

tbekolay:
We revisited this issue in the dev meeting, and decided to enumerate out possible options for standards to see if any of them stick as being the best.

It probably helps to also brainstorm about use cases. From what I've seen, there are three:

[...]

PR #1179, opened 2016-10-21
Rename network inputs.

jgosmann:
**Motivation and context:**
This makes the product and circular convolution network adhere to the newly defined standard for network inputs (#887) which should be named input_*. The old inputs are still available with a deprecation warning. This required turning the network creation function into class which in turn required to make the `net` argument obsolete. As discussed before this argument was rarely used and in almost all uses cases the new `**kwargs` passed to `nengo.Network` provide the same functionality.

For consistency I made the `net` argument also obsolete on other network creation functions. In some other networks I also introduced in `output` node to provide a common interface.

The new-spa system will require the standardized input naming.

**How has this been tested?**

Renamed the inputs and made sure that all tests still passed before re-introducing the deprecated names.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Re-introduce the deprecated names with a deprecation warning.
- [x] Update documentation with regards to the `net` argument.


jgosmann:
Adding the deprecated names with a deprecation warning is problematic. Usually I would use a property, but those networks are not real classes, but rather functions modifying a network and I can't add a property on a class instance.

tbekolay:
How about monkey patching `__getattr__` and raising the warning there?

jgosmann:
`__getattr__` has to be defined on the class, not the instance. That would mean monkey patching `nengo.Network` which then effects all instances.

tbekolay:
Hm, did not know that! So yeah, doesn't seem like there's a nice way to do it... the three not-as-nice ways I see are

1. Add something to `nengo.Network` to declare deprecated/renamed attributes and implement `__getattr__` there.
2. Have these functions return a `Network` subclass instead (and raise a warning noting the renaming if a network is passed in through the `net` kwarg).
3. Alias them and either don't warn (the deprecation will only be visible in the changelog) or always warn (will be visible, but annoying once the scripts are changed...)

I think I'm leaning toward 2 right now... but yeah, none of them are ideal.

jgosmann:
Another possibility:

4. Convert the functions into a sub-classes of `Network`. This would need changes to the `net` argument.

Actually, I'm not completely sure why we have the `net` argument. The two reasons I could come up with:

1. It reduces boilerplate (no super constructor call and replicating the three `Network` arguments). (There might be ways to do that in a sub-class.)
2. [To set configs on the network.](https://pythonhosted.org/nengo/examples/network_design_advanced.html#1-accept-a-network-argument)

Personally, setting configs that way on the network never occured to me before I looked it up in the documentation. I always create a `Config` object and use that, for example:

```python
cfg = nengo.Config(nengo.Ensemble)
cfg[nengo.Ensemble].radius = 2
with cfg:
    ea = nengo.networks.EnsembleArray(...)
```

This might be slightly more boilerplate than the network approach (because you have to define what classes to configure), but it seems that other people are using the same approach as we have been discussing to simplify this in issue #1043.

Also note, that certain networks like `EnsembleArray` and SPA networks/modules are subclasses of `Network`.

Changing all networks to classes would be definitely a large change and likely requires the recommendation in the documentation to be updated accordingly. I also assume, that functions vs. networks was discussed in detail before, so there might be more reasons for the function approach?

tbekolay:
So the idea for accepting a `net` argument comes from Matplotlib. Many pyplot functions (and functions in other libraries) accept an `ax` parameter so you can draw into a specific set of axes; if you don't pass in `ax` it will use the current axes, or make an axes object if there are no current axes. This was a more straightforward analogy back when we did have a default network (if you only had one network then you didn't have to use a `with network:` block), so maybe now it's not so useful. It also was not possible to do `with config:` when these functions were made, so that's another historical reason for it. Accepting the 3 `Network` arguments and passing them to `super` is a pain IMO, and in general I avoid APIs that require subclassing whenever possible, but if the consensus is to switch these over to subclasses I can deal with it.

jgosmann:
In Matpotlib it is no problem to plot multiple things on the same axis, but this is not true for Nengo networks:

```python
net = nengo.Network()
p1 = Product(..., net=net)
p2 = Product(..., net=net)
```

would leave me without access to the first product network, because `p1` and `p2` will be net.

I agree that passing the network arguments around is annoying and subclassing isn't nice from a user perspective.

tbekolay:
My attempt to summarize what we discussed at the lab meeting:

It's not clear if anyone uses the `net` argument on a regular basis -- **does anyone reading this use it?** Reply to this PR if so! Or give this comment a :+1:. If no one uses it, it's an easier decision to deprecate it and switch to subclassing.

The exact way to subclass is not clear at the moment... for the time being, the easiest resolution is to keep these functions mostly as-is (with the `net` argument deprecated?) and return a Network subclass instead of a Network.

Moving forward, it would be best to remove the disparity between the network functions and the network subclasses. There's not much choice but to subclass (since we want to add methods to the returned class -- see `EnsembleArray`) but having the three network arguments in the constructor leads to a lot of boilerplate code for each subclass. Would be great if we could figure out a way around that!

Seanny123:
The only time I've ever used it was when [I was making oscillatosr](https://github.com/Seanny123/nef-conceptors/blob/master/test_attrac_dmp.py#L28) and didn't know how to set the seed and the neuron type on them. Code shown below:

    osc = nengo.Network()
    osc.config[nengo.Ensemble].neuron_type = nengo.LIFRate()
    osc.config[nengo.Ensemble].seed = 0
    nengo.networks.Oscillator(0.1, 2*np.pi/period, 300, net=osc)

jgosmann:
@Seanny123's use case should not require the `net` argument:
```python
with nengo.Config(nengo.Ensemble) as osc_cfg:
    osc_cfg[nengo.Ensemble].neuron_type = nengo.LIFRate()
    osc_cfg[nengo.Ensemble].seed = 0
    nengo.networks.Oscillator(0.1, 2*np.pi/period, 300)
```

The only thing I used the net argument for is to change the network label:
```python
nengo.networks.Product(150, 2, net=nengo.Network(label="multiply"))
```
I assume that this can also be achieved without the `net` argument like this:
```python
prod = nengo.networks.Product(150, 2)
prod.label = "multiply"
```

One idea to reduce boilerplate when subclassing:
```python
def add_net_init_args(init_fn):
    def __init__(self, *args, label=None, seed=None, add_to_container=None, **kwargs):
        super(type(self), self).__init__(label=label, seed=seed, add_to_container=add_to_container)
        init_fn(self, *args, **kwargs)
    return __init__

class MyNet(nengo.Network):
    @add_net_init_args
    def __init__(self, d):
        with self:
            self.ea = nengo.networks.Product(150, d)
            # ...
```
There might be some corner cases where this is problematic (multiple inheritance?), but it would probably cover 99% of cases and in the remaining cases one could still resort to writing the required boilerplate.

jgosmann:
After writing this I was wondering why isn't `super(type(self), ...)` not a common idiom in Python and it turns out in case in subclasses the type would change to that subclass which would invoke the wrong super function. A better implementation would probably be:

```python
class add_net_init_args(object):
    def __init__(self, cls):
        self.cls = cls

    def __call__(self, init_fn):
        def __init__(inner_self, *args, label=None, seed=None, add_to_container=None, **kwargs):
            super(self.cls, inner_self).__init__(label=label, seed=seed, add_to_container=add_to_container)
            init_fn(inner_self, *args, **kwargs)
        return __init__

class MyNet(nengo.Network):
    @add_net_init_args(MyNet)
    def __init__(self, d):
        with self:
            self.ea = nengo.networks.Product(150, d)
            # ...
```

jgosmann:
Another option discussed in the dev meeting:
```python
class MyNet(nengo.Network):
    def __init__(self, d, **kwargs):
        super(MyNet, self).__init__(**kwargs)
        with self:
            self.prod = nengo.networks.Product(150, d)
            # ...
```

(Might be my favourite. Also note that both approaches still allow to use the full boiler plate when desired.)

tcstewar:
I could definitely see the ```**kwargs``` approach being okay for this situation (specifically, when subclassing Network), but still avoiding it in all other situations in nengo.  The available kwargs for Network are pretty easy to look up and are fairly standard, so it's not too bad to be opaque about this.  And it is a standard Python idiom.

tbekolay:
> I could definitely see the **kwargs approach being okay for this situation

Agreed, I think that's our best bet given the options we've been able to come up with. It's probably worth taking a look at `EnsembleArray` to see if there's still a reason to use `**` there; if not then this can be done for all network subclasses.

jgosmann:
Pretty much all that it does is
```python
self.config[nengo.Ensemble].update(ens_kwargs)
```
So the only reason to keep it would be convenience. (It would also a possibility to turn it into a single dict argument instead of a `**kwargs` argument.)

jgosmann:
This PR is almost ready for review. The only major thing is updating some documentation and deciding how to handle the `EnsembleArray`. With the changes in this PR networks usually take a `**kwargs` argument that is passed on to the `nengo.Network` constructor. In the `EnsembleArray` those arguments are explicitely listed and keyword arguments `**ens_kwargs` are used for the ensembles instead. That `**ens_kwargs` keyword argument would not be required as one could just change the config like this:
```python
with nengo.Config(nengo.Ensemble) as cfg:
    cfg[nengo.Ensemble].radius = 2.
    ea = nengo.networks.EnsembleArray(...)
```
But this is definitely more boiler plate than a simple `nengo.networks.EnsembleArray(..., radius=2.)` and the latter approach seems to be used a lot. So I am not sure whether we should deprecate `**ens_kwargs` for consistency with other networks. We could also make it a single dictionary argument instead of a `**kwargs` argument as a middle ground.

jgosmann:
Documentation updated, ready for review.

Seanny123:
LGTM.

jgosmann:
Another PR that would be great to get merged as nengo_spa sort of depends on it.

jgosmann:
Technically, this is a breaking change because of the removal of the `net` argument. So this might actually require a major release?

Seanny123:
Does this need to be reviewed again? It doesn't have any tags, so I'm a bit confused.

jgosmann:
It only got one review so far I believe? But it also won't be merged before Nengo 3.0 ...
Not sure what happened to the labels, they seem all to be gone.

tbekolay:
> Not sure what happened to the labels, they seem all to be gone.

The labels are redundant with Github's built-in reviewing system now, so I removed the labels to make sure everyone uses the reviewing system. The PR tab at the top has searching / filtering options to see what needs to be reviewed and such now, so there's no reason to keep the labels.

tbekolay:
Speaking of labels... I made a `breaking change` label so we can queue these up for the 3.0 release.

tbekolay:
In the recent dev meeting, we briefly discussed also making some attributes `Parameter`s in this PR (see https://github.com/nengo/nengo-spa/issues/181), and making this a non-breaking change by adding aliases to the old network attributes names. Making this note to remind myself to look into it, as it'll unblock this from being merged.

hunse:
Rebased to master.

I've also converted all the networks to classes. Personally, I think this is nicer, though the functionality is pretty much the same as the old function method. One advantage is that you can inherit from classes, so there's a bit more flexibility in terms of how people build off our networks. Are others on board with this?

Right now, I have the old `A` and `B` arguments in `Product` and `CircularConvolution` deprecated, as Jan had them. The other option, since this is 3.0, is to just get rid of them entirely. Thoughts?

I've kept this as separate commits for now, but I think for merging it can all be squashed into one commit (maybe we'd want to keep two, one for obsoleting net arguments and one for turning things into classes).

Seanny123:
@tbekolay you were the person who pushed to change the networks from object to functions originally. Do you remember why we made this choice? I feel like it had something to do with inheritance being tricky?

hunse:
Having them as objects, you don't need to use inheritance. You can do everything you would if they were functions.

For example, if you want to create your own integrator with a second feedback connection, you can do
```python
integrator = nengo.networks.Integrator(n_neurons=100, dimensions=1)
with integrator:
    nengo.Connection(integrator.ensemble, integrator.ensemble, synapse=0.1)
```
just as you could with a function. But you also have the option to inherit, if you want.

arvoelke:
I think @Seanny123 is asking why they were changed to functions in the first place? They used to be classes, but many years ago were switched to functions -- and now they are being switched back. There was a discussion that took place but I do not remember the link or if it even took place online (might have been at an early in-person dev meeting).

hunse:
It looks like the main reason (at least in the commit message) is that they seemed easier to read: d0878a061ae48e5d7e005e2d53c4167f3021d14d

The other aspect of this issue is that Jan changed some of the networks in this PR to classes (namely Product and CircularConvolution) so that we can deprecate the old inputs. If we want to deprecate these (which we do), then we need classes there. I think there's some sense in having consistency between networks, and also some sense in having the option to inherit if users want that. With the `net` argument gone, I feel like one of the main functional benefits of functions is gone.

Seanny123:
@hunse I agree completely. Thank you for taking the time to explain this reasoning to me.

tbekolay:
I believe we also discussed this at a dev meeting a year or two back that people wanted to standardize to network subclasses. I do still think that APIs shouldn't advocate for people to subclass their objects, but the consensus is pretty strong on this to make it standard, and in the absence of another way that people agree is better we'll make these all subclasses.

nengo/networks/tests/test_actionselection.py
```diff
@@ -46,3 +45,72 @@ def test_thalamus(Simulator, plt, seed):
...
+    )
+    for connection in bg.all_connections:
+        if connection.synapse is None:
+            continue
+        # Same synapses
+        assert isinstance(connection.synapse, nengo.Lowpass)
+        assert connection.synapse.tau in (0.002, 0.008)
+        # But solver should be different
+        assert isinstance(connection.solver, nengo.solvers.LstsqL2nz)
+    # Synapse should still not be in the config
```
hunse: Maybe "Ensure default synapse has been removed from config" to make it a bit clearer what we're testing?

nengo/networks/tests/test_actionselection.py
```diff
@@ -46,3 +45,72 @@ def test_thalamus(Simulator, plt, seed):
...
+        gaba_config=config_with_synapse,
+    )
+    for connection in bg.all_connections:
+        if connection.synapse is None:
+            continue
+        # Synapse changes
+        assert isinstance(connection.synapse, nengo.Alpha)
+        assert connection.synapse.tau == 0.1
+        # Solver should be the same
+        assert not isinstance(connection.solver, nengo.solvers.LstsqL1)
```
hunse: We should check that the correct synapse is still in the config at the end (it hasn't been overwritten or removed).

PR #1181, opened 2016-10-27
Factored out `sample` function

tbekolay:
**Motivation and context:**
The ensemble tuning function in #871 needed access to the builder's `sample` function. We're moving #871 to `nengo_extras`, but it's a good idea to move the `sample` function anyhow, as it's used in several places.

Originally @drasmuss moved this to `nengo.utils.builder`; I figured it's not bad to have in `nengo.dists`. Though, since it's not quite the same as `Distribution.sample` we should perhaps rename it to something? `sample_if_needed` or something?

Supercedes #871.

**How has this been tested?**
Ran the test suite locally. Doesn't require additional testing.

**Types of changes:**

**Checklist:**

**Still to do:**
- [x] We should decide if we're okay with having this in `nengo.dists`, and whether we should keep it named `sample` or something more explicit.
- [x] Changelog entry (once we decide the point above) since it maybe become part of the frontend API.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,10 @@ Release History
   in the terminal and Jupyter notebook.
   (`#937 <https://github.com/nengo/nengo/issues/937>`_,
   `#1151 <https://github.com/nengo/nengo/pull/1151>`_)
+- Added ``nengo.dists.get_samples`` function for convenience
+  when working with distributions or samples.
+  (`#1181 <https://github.com/nengo/nengo/pull/1181>`_,
+  `docs <http://pythonhosted.org/nengo/frontend_api.html#nengo.dists.get_samples>`_)
 
 **Changed**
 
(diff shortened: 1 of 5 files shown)


jgosmann:
How about `get_samples`? That name would agnostic to whether the samples are obtained by sampling or just returning existing samples.

tbekolay:
> How about `get_samples`?

Yeah, I can see that! What do you think about renaming the first argument to `dist_or_samples`? Seems to cover all the bases

jgosmann:
> What do you think about renaming the first argument to dist_or_samples?

I'm fine with that; I'm also fine with `dist`.

drasmuss:
Renaming and relocation both sound fine to me.

jgosmann:
LGTM :cake:

Seanny123:
LGTM.

tbekolay:
In order to check all the boxes above, I added a docstring, then realized we don't have the distributions anywhere in the docs. So I added them (and fixed up our current docstrings to print nicely) in ef5d221cced139790bec and added the docstring and a changelog entry. I think those don't warrant two re-reviews, so @jgosmann can you take a look at this and merge if they look good to you?

PR #1182, opened 2016-10-28
List probeable attributes on probe error

Seanny123:
**Motivation and context:**
It's annoying to look up documentation on an object you're trying to probe when you've made a typo. Consequently, I thought it would be a good idea to list the probeable attributes of an object when an error happens.

**How has this been tested?**
Ran this branch with the following code:

```
import nengo

    nd = nengo.Node([0])
    ens = nengo.Ensemble(100, 1)

    voja = nengo.Voja(post_tau=None, learning_rate=5e-2)
    conn_in = nengo.Connection(nd, ens, synapse=None,
                               learning_rule_type=voja)

    p_enc = nengo.Probe(conn_in.learning_rule, 'scaled encoders')
```

**Types of changes:**

**Checklist:**

This is such a small change it doesn't need a changelog entry, nor extra documentation in my opinion.

### DIFF
--- a/nengo/probe.py
+++ b/nengo/probe.py
@@ -28,8 +28,9 @@ class AttributeParam(StringParam):
                                 "into 'weights'. Probe 'weights' instead.",
                                 since="v2.1.0")
         if attr not in probe.obj.probeable:
-            raise ValidationError("Attribute %r is not probeable on %s."
-                                  % (attr, probe.obj),
+            raise ValidationError("Attribute %r is not probeable on %s.\n"
+                                  "Probeable attributes: %s"
+                                  % (attr, probe.obj, probe.obj.probeable),
                                   attr=self.name, obj=probe)
 
 


jgosmann:
Made a minor change; LGTM :cake:

Seanny123:
@bmorcos since you already know how probing works, would you mind quickly reviewing this PR?

nengo/probe.py
```diff
@@ -28,9 +30,10 @@ def validate(self, probe, attr):
                                 "into 'weights'. Probe 'weights' instead.",
                                 since="v2.1.0")
         if attr not in probe.obj.probeable:
-            raise ValidationError("Attribute %r is not probeable on %s.\n"
+            raise ValidationError("Attribute %r is not probeable on %s.%s"
```
tbekolay: Hm... I think the `\n` is preferable to `os.linesep` here... this is going to be printed to the terminal, and I don't think any terminal protocols require `\r`; if it's going to be printed to a file, Python automatically deals with the different line endings. From what I can google, it seems like `os.linesep` is only useful when reading/writing from files in bytes mode rather than ascii. We also don't do this in plenty of other cases, so even if it's wrong I would rather switch them all than just have this one switched.

jgosmann: Do you have a Windows command line to verify that `'\n'` works? I think back in the days it would print `'one\ntwo'` as
```
one
   two
```

jgosmann: You might be right. Apparently line endings are automatically converted for files read or written in text (as opposed to binary) mode. I assume the `print` uses a file descriptor in text mode, so that it should also convert line endings, but did not verify this.

tbekolay: I just verified this in Python 2 and 3 in a Windows command prompt; [`sys.stdout`](https://docs.python.org/2/library/sys.html#sys.stdout) is indeed a file object which is likely where the line endings get handled... in any case, they do appear to get handled so I'll revert :rainbow:

tbekolay:
LGTM too; I'll merge!

PR #1183, opened 2016-10-31
Make indexing and slicing consistent with Python.

jgosmann:
**Motivation and context:**
Refer to the discussion in #1176. Basically, Nengo was converting integer indices on Nengo objects into slice and thus no `IndexError` was raised for out of bound indices. One particular problem with this behaviour was that iteration over Nengo objects would produce an infinite loop.

With this PR integer indexing will raise the appropriate `IndexError` and slice indexing will truncate to match default Python behaviour. There is also array indexing which will behave like integer indexing. This changes make the iteration example from #1176 work without producing an inifinite loop.

**How has this been tested?**
Added a few more test cases to check out of range indices and corner cases like negative indices. Also run the infinite loop example from #1176 which did not produce an infinite loop anymore.

**Types of changes:**

Not entirely sure. I think of it as a bug fix, but in theory someone could have relied on the old truncating behaviour of integer indices, so the change could break existing models.

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,10 @@ Release History
 
 **Changed**
 
+- Integer indexing of Nengo objects out of range raises an ``IndexError``
+  now to be consistent with standard Python behaviour.
+  (`#1176 <https://github.com/nengo/nengo/issues/1176>`_,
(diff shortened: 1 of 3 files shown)


Seanny123:
LGTM.

Issue #1184, opened 2016-11-01
Error if node signal stride does not match dtype size


hunse:
The following code fails
```python
import numpy as np
import nengo

y = np.fft.ifft(np.arange(5)).real

with nengo.Network() as model:
    u = nengo.Node(y)
    up = nengo.Probe(u)

with nengo.Simulator(model):
    pass
```
with
```
  File "/home/ehunsber/workspace/nengo/nengo/builder/signal.py", line 254, in init
    dtype=x.dtype, buffer=self[signal.base].data)
ValueError: strides is incompatible with shape of requested array and size of buffer
```
This is because `y` has a stride of 16 (since it's a view into a complex data type), but the data type size is 8. This causes a problem somewhere.

This is easy to work around in this case (just copy `y` so it's contiguous, e.g. `y = np.array(y)`), but the error message is kind of cryptic.

jgosmann:
This seems like it might be a duplicate of #1156 which should be fixed by #1157.

jgosmann:
Unfortunately, #1157 does not fix this particular problem, but this might still be related to #1156.

hunse:
Hmm, I'm surprised. It definitely seems related to #1156. Can we just check if inputs are C contiguous, and if not make a C contiguous copy?

jgosmann:
I've got a rough idea what's going on now. I have to go to a seminar in a few minutes, but will explain later.

jgosmann:
1. An array `x` with 16-byte items is created.
2. A view `y` of `x` using only 8-byte of each item is created and used as node output.
3. The signal `s` is constructed from this view `y` inheriting the strides of 16 bytes (which is correct with respect to the base of the `initial_value`)
4. A signal view `v` is constructed from `s` which also inherits the 16 byte strides.
5. When the signals are initialized (i.e. the memory is allocated), `s` will be initialized by copying `y` (at least in this case, there might be cases where a view is used) which produces a new array `z` with 8-byte elements and a stride of 8.
6. The signal view `v` is initialized and will try to use `z` as base and at the same time 16-byte strides because that's what was previously correct.

In summary: Signal views are defined with respect ot their original base, but the base array maybe copied into a different memory layout on initialization which requires things like strides to be adjusted.

PR #1185, opened 2016-11-02
Make non-view/non-scalar signal data C contiguous

jgosmann:
**Motivation and context:**
If input to a model is provided as non-contiguous view with weird strides, building the model can fail (see #1184 for more details). This PR ensures that the `initial_value` of signals are contiguous where possible (we can't make scalars contiguous as some parts of the code require them to be 0d arrays, contiguous arrays are always at least 1d; we want to keep signal views as views).

#1156 which was fixed by #1157 falls into the same class of problems (views with weird offsets/strides in signals). Both fixes are required to fix both problems. Apart from some merge conflicts the changes should not influence each other.

**How has this been tested?**
Added a test based on the example code in #1184.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,12 @@ Release History
   (`#1191 <https://github.com/nengo/nengo/issues/1191>`_,
   `#1267 <https://github.com/nengo/nengo/pull/1267>`_)
 
+**Fixed**
+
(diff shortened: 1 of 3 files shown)


Seanny123:
I can't believe this would actually cause a bug. Good catch and fix! LGTM.

Issue #435, opened 2014-06-27
Get the total number of neurons


xchoo:
There was a feature in 1.4 where you could get the total count of the number of neurons in a network (or subnetwork). It would be helpful if you could do the same for 2.0 models.

tbekolay:
A `n_neurons` property now exists on `Network` (see #1186).

PR #1186, opened 2016-11-03
Added n_neurons property to Network.

tbekolay:
**Motivation and context:**
It's a pretty common thing to want to know how many neurons are in a particular network, as that's a metric people want to know when reviewing papers etc. I've given the one-liner to a few people over IM and it was asked for in #435, so might as well make it part of `Network`.

It should be noted that this PR also renames `EnsembleArray.n_neurons` to `EnsembleArray.n_neurons_per_ensemble`. Long, but we don't actually use it anywhere in tests or examples, so I think it won't matter much.

**How has this been tested?**
Added a short test for it (which passes).

**Types of changes:**

**Checklist:**

- [x] I have updated the documentation accordingly. (has a docstring, will show up in docs)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,7 +22,20 @@ Release History
 2.4.1 (unreleased)
 ==================
 
+**Added**
 
+- Added a ``n_neurons`` property to ``Network``, which gives the
+  number of neurons in the network, including all subnetworks.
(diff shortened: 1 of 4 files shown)


jgosmann:
LGTM. :cake:

Seanny123:
LGTM

PR #1189, opened 2016-11-07
Builder refactoring

tbekolay:
**Motivation and context:**
Several backends use bits and pieces of the builder. It's difficult for them to do this, because we did not design the builder anticipating that parts of it would be used by other backends.

The main goal of this PR is to split the build process into two parts.

In part 1, we construct a concrete model from the frontend model definition. Specifically, in part 1 we:

1. Sample all parameters that are set to distributions.
2. Solve for decoders.

And probably some other stuff once we start going through it all.

In part 2, we make signals and operators to implement the concrete model from part 1.

The end result of this split is another abstraction layer to make implementing Nengo backends easier (hopefully!)

A side goal of this PR is to make an explicit builder API that will be versioned. It's not clear yet whether the builder version will be tied to the Nengo version or if it will grow separately... probably best to keep them the same since they're in the same repo?

**How has this been tested?**

**Types of changes:**

**Checklist:**

**Still to do:**

Everything. Currently this PR is just a placeholder so that we can point other issues here for more discussion.


drasmuss:
Another wishlist item: it'd be nice to be able to register new build functions only for a specific backend.  So, e.g., I could register a new `Connection` builder for NengoDL, but if someone is building a `Connection` using the normal nengo Simulator (after importing NengoDL) they would still get the default `Connection` builder.

Right now I get around this by creating a custom subclass of the Nengo builder that I register the NengoDL build functions with, and then use that when building within the NengoDL simulator.  Which works, and isn't too bad, but it'd be even nicer if we could just have the `@Builder.register` decorator take an optional `backend=xyz` argument or something.

tbekolay:
Another wishlist item: a set of hooks for various things (sim reset, for example). This would allow several use cases that currently require subclassing `Simulator`.

tbekolay:
We have started planning out this refactoring internally at ABR and will do the refactoring as a series of related PRs, rather than as one big one, so I'm going to close this.

PR #1197, opened 2016-11-14
NeuronType functions handle array_like manually

tbekolay:
**Motivation and context:**
This is an alternative to #633, which achieved the same ends with manual `np.array` calls instead of a decorator. The consensus in dev meetings has been to go with the manual approach, so I wanted to make this to ensure that it still seems better to most people before closing #633.

Same functionality as #633, let's not merge both!

**How has this been tested?**
Tried this in a terminal:

```
>>> nengo.LIF().gain_bias([100, 120], [-0.5, 0.5])
(array([ 1.35549652,  5.36847925]), array([ 1.67774826, -1.68423963]))
```

Which would have failed previously. Didn't add any more tests, but I could if people feel strongly that they should be added.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -30,6 +30,23 @@ class NeuronType(FrozenObject):
     def _argreprs(self):
         return []
 
+    def current(self, x, gain, bias):
+        """Compute current injected in each neuron given input, gain and bias.
+
+        Parameters
+        ----------
+        x : (n_neurons,) array_like
+            Vector-space input.
+        gain : (n_neurons,) array_like
+            Gains associated with each neuron.
+        bias : (n_neurons,) array_like
+            Bias current associated with each neuron.
+        """
+        x = np.array(x, dtype=float, copy=False, ndmin=1)
+        gain = np.array(gain, dtype=float, copy=False, ndmin=1)
+        bias = np.array(bias, dtype=float, copy=False, ndmin=1)
+        return gain * x + bias
+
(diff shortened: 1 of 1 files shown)


nengo/neurons.py
```diff
@@ -42,18 +59,21 @@ def gain_bias(self, max_rates, intercepts):
...
         Returns
         -------
-        gain : ndarray(dtype=float64)
+        gain : (n_neurons,) array_like
             Gain associated with each neuron. Sometimes denoted alpha.
-        bias : ndarray(dtype=float64)
+        bias : (n_neurons,) array_like
             Bias current associated with each neuron.
         """
+        max_rates = np.array(max_rates, dtype=float, copy=False, ndmin=1)
```
tbekolay: Yep, that's right. I don't know if the `ndmin` is super necessary, but @hunse included it in his PR so I trust he did it for a reason!

hunse:
LGTM. I'm fine with this instead of #633

Seanny123:
This LGTM too. Should a test be made to check that `array_like` is handled properly or would that be silly to test?

tbekolay:
> Should a test be made to check that array_like is handled properly or would that be silly to test?

Not silly, but also not necessary. If you want to write it up then by all means! But I would say that if someone wants to merge this they shouldn't wait for the example.

jgosmann:
Two people said it looks good to them. Merge before the release, @tbekolay?

tbekolay:
OK, will do.

Issue #1201, opened 2016-11-16
Unhandled situation in new-spa action handling


tcstewar:
I just found an example that the new spa parser doesn't handle:

```python
import nengo
import nengo.spa as spa

D = 32

model = spa.SPA()
with model:
    model.vector = spa.State(D)
    model.scalar = spa.State(1)

    actions = spa.Actions(
        'dot(vector,A) + scalar --> vector=B'
        )

    model.bg = spa.BasalGanglia(actions)
    model.thalamus = spa.Thalamus(model.bg)
```

This worked in the old system (i.e. current master)

Here's the error:

```
Traceback (most recent call last):
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\page.py", line 219, in execute
    exec(compiled, code_locals)
  File "<nengo_gui_compiled>", line 61, in <module>
  File "c:\users\terry\documents\github\nengo\nengo\spa\thalamus.py", line 112, in __init__
    self.bg.actions.process()
  File "c:\users\terry\documents\github\nengo\nengo\spa\actions.py", line 227, in process
    self.process_new_actions(self.vocabs, *self.args, **self.kwargs)
  File "c:\users\terry\documents\github\nengo\nengo\spa\actions.py", line 232, in process_new_actions
    action.infer_types(self.construction_context.root_module, None)
  File "c:\users\terry\documents\github\nengo\nengo\spa\spa_ast.py", line 1022, in infer_types
    self.condition.infer_types(root_module, context_type)
  File "c:\users\terry\documents\github\nengo\nengo\spa\spa_ast.py", line 565, in infer_types
    self.lhs.infer_types(root_module, context_type)
  File "c:\users\terry\documents\github\nengo\nengo\spa\spa_ast.py", line 763, in infer_types
    self.lhs.type, self.rhs.type, self))
SpaTypeError: Incompatible types TVocabulary<32-dimensional vocab at 0x39b74908> and TVocabulary<1-dimensional vocab at 0x39b74e48> in dot product 'dot(vector, A)'.
```

**EDIT:** as @jgosmann points out below, this is desired behaviour

tcstewar:
And another unhandled case:

```python
import nengo
import nengo.spa as spa

model = spa.SPA()
with model:
    model.state1 = spa.State(16)
    model.state2 = spa.State(32)

    actions = spa.Actions(
        '0.5 --> state1 = translate(state2)*2'    # this works
[...]

PR #1202, opened 2016-11-16
New spa fixes

jgosmann:
**Motivation and context:**
Changes to address #1201.

* Change the base class of `Translate` and `Reinterpret` to allow it to be used in other operations like products.
* Don't consider the parent context for type inference in dot products. It shouldn't matter as the dot product returns a scalar anyways.
* Be more specific about which binary operations allow to combine `TVocabulary` and `TScalar`.

The last two points lead to better, more specific error messages.

**How has this been tested?**
* Added a test for multiplictaion with `Translate` and `Reinterpret`
* Used the examples from #1201 to inspect the error messages.

**Types of changes:**

**Checklist:**

@tcstewar, would you mind taking a look?

### DIFF
--- a/nengo/spa/spa_ast.py
+++ b/nengo/spa/spa_ast.py
@@ -792,7 +792,7 @@ class DotProduct(BinaryNode):
         return 'dot({}, {})'.format(self.lhs, self.rhs)
 
 
-class Reinterpret(Node):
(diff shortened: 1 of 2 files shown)


tcstewar:
This looks good to me.  :)

Issue #1204, opened 2016-11-17
Sorting of examples in slow and fast is incorrect when non-notebook files exist in folder


jgosmann:
I was observing that the test suite would fail on examples even though it had passed before. Turns out that the sorting in slow and fast notebooks screws up when there are files not ending in `.ipynb` in the example directories. Then it can get pretty random which examples are excuted and which are skipped.

tbekolay:
Hm... I think I checked for the `.ipynb` extension in some places in [test_examples.py](https://github.com/nengo/nengo/blob/master/nengo/tests/test_examples.py#L43), but not all of them... probably used too many list comprehensions there ¯\\\_(ツ)\_/¯

Also, workflow-wise, if you're planning to fix a bug, no need to make an issue reporting that bug beforehand... you can just make the PR fixing the bug. Having an issue and a PR (assuming the PR is relatively quick) means we have to remember to also close the issue, and it means that multiple people might work on it the same bug, which is a waste.

jgosmann:
Yeah, I know, but sometimes I don't have the time or motivation to immediately fix the bug and create an issue as a reminder for myself.

Issue #1203, opened 2016-11-17
Close figures created in tests.


jgosmann:
While working on new-spa I discovered this warning being produced by the tests (not sure if this is just for the Jupyter notebook examples or all tests):
```
/home/jgosmann/.pyenv/versions/3.5.2/lib/python3.5/site-packages/matplotlib/pyplot.py:516: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
  max_open_warning, RuntimeWarning)
```

It is probably a good idea to close all matplotlib figures after each test.

tbekolay:
So we do actually close all figures in the `__exit__` method of `Plotter` (see https://github.com/nengo/nengo/blob/master/nengo/utils/testing.py#L116). Originally I did this because if you don't, the unit tests will plot on top of one another in some cases. But it should also deal with this kind of warning -- my guess is that it's one particular example that is creating a ton of figures without closing any.

jgosmann:
Is the plotter used in the Jupyter notebook examples?

tbekolay:
[Yes, we use the `plt` funcarg specifically to close figures](https://github.com/nengo/nengo/blob/master/nengo/tests/test_examples.py#L58).

jgosmann:
[But is that actually used, if the `globals` and `locals` dict are empty?](https://github.com/nengo/nengo/blob/master/nengo/tests/test_examples.py#L67)

jgosmann:
Seems so. I think the problem might be that if `self.record` is `False`, [`self.plt` is assigned a `Mock()` object](https://github.com/nengo/nengo/blob/master/nengo/utils/testing.py#L89). Then instead of calling the actual Matplotlib close, [we call it on the mock object](https://github.com/nengo/nengo/blob/master/nengo/utils/testing.py#L96) which has no effect. In the Jupyter notebook the `plt` fixture is also not used, instead they import `matplotlib` and thus do the actual plotting instead of using the mock object.

jgosmann:
Not entirely sure, if what I said in the last comment is actually part of the problem, but foremost `plt.close` is never called (neither on matplotlib nor on the mock) as long as `self.record` is False.

tbekolay:
[...]

PR #1205, opened 2016-11-18
Fix example tests.

jgosmann:
**Motivation and context:**
Fixes for:
* #1203: Close figures created in examples properly to free some memory.
* #1204: Correctly ignore non-notebook files in example folder.

In case you are wondering why #1204 is a problem because we only have notebook files in those folders: I was editing some of those files in Vim for the new-spa branch (sometimes easier than opening the Jupyter notebook and making sure they are converted to the right format) and vim creates a temporary file `.filename.ipynb.swp` which then changed which tests were actually run.

**How has this been tested?**
Ran `py.test -s --strict -r w nengo/tests/test_examples.py` and got matplotlib's too many open figures warning, went away with these changes.

Create a file one of the folders containing slow example and observe which examples are run: `py.test -v nengo/tests/test_examples.py`. Without the fix slow and fast examples get messed up somewhere. For example creating a file `aaaa` in `examples/spa` will cause the `question_memory` example to run as fast test, though it should be in the slow example list.

**Types of changes:**

**Checklist:**

- [do we need one?] I have included a changelog entry.

### DIFF
--- a/nengo/tests/test_examples.py
+++ b/nengo/tests/test_examples.py
@@ -53,8 +53,8 @@ slow_examples.sort()
 fast_examples.sort()
 
 
-def assert_noexceptions(nb_file, tmpdir, plt):
-    plt.saveas = None  # plt used to ensure figures are closed, but don't save
+def assert_noexceptions(nb_file, tmpdir):
+    plt = pytest.importorskip('matplotlib.pyplot')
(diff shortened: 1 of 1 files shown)


tbekolay:
Looks great, thanks for the quick fix! Will merge after the second review.

Seanny123:
LGTM/

PR #1206, opened 2016-11-18
New spa overhaul

jgosmann:
**Motivation and context:**

This is an overhaul of the SPA system to introduce the config system and remove support for deprecated syntax that was still kept around with the SPA compiler.

Notable changes (I hope I won't forget anything important):
* The `dimensions` and `vocab` arguments have been unified into a single `vocab` argument. I could see renaming it to `vocab_or_dim`, but because it is so common it might be better to keep it short. I also became annoyed by having to state the dimensionality for every single `spa.State`. In most networks it is always the same. Thus, it is possible to set it once like so:
```python
with spa.Module() as model:
    model.config[spa.State].vocab = 64  # Use 64-dimensional default vocab, you could also set it to a specific vocab
    model.state1 = spa.State()
    model.state2 = spa.State()
    model.state3 = spa.State(32)  # Use another vocab for this state.
```
* It was common to have a pure Nengo network and then a corresponding SPA network. For some basic networks like `Product` this seems right because this should continue to be shipped with Nengo and the SPA networks tend to be not as general. But for the action selection networks `BasalGanglia` and `Thalamus` this seemed unneccssary. These should probably completely moved into SPA and the SPA module was just adding boilerplate to pass through all the parameters. So I merged the modules and networks. They do not exist as pure Nengo network anymore, but the SPA module can be used like a normal Nengo network (so there is no loss in functionality).
* Because of the hierarchical module support the syntax for inputs has changed to an attribute style assignment. Looks like this:
```python
with spa.Module() as model:
    model.state1 = spa.State(64)
    model.cmp = spa.Compare(64)
    model.stimulus = spa.Input()
    model.stimulus.state1 = 'A'
    model.stimulus.cmp.input_a = lambda t: 'A' if t < 0.5 else 'B'
```
* The way actions are build changed. First you create an Action object:
```python
actions = spa.Actions('a = b', 'dot(a, 'A') --> b = C')
```
This object is not tied to any model or module. It just parses the rules to construct the abstract syntax tree (AST) and could in theory be reused. To create the required Nengo objects within a module the `build` function is used:
```python
with spa.Module() as model:
    # create the required modules and assign the to model as attributes
    actions.build(model)
```
This might not look very Nengo-ish and I'm open to suggestions to make this look more Nengo-ish, but I want to keep the seperation of parsing and creating the actual Nengo objects. The `build` function will automatically create `BasalGanglia` and `Thalamus` as needed, but if you want to provide your own netwoks, you can pass them to the `build` function. Due to the config system, you can set `BasalGanglia` and `Thalamus` parameters also without creating your own instances. One more thing to note: the `build` function will assign the created objects to `model.bg` and `model.thalamus`. I'm not entirely convinced that this is the best thing to do (because it might overwrite used-defined attributes). But in general I have to figure out how to best provide access to the objects created by `build` and I'll reconsider this together.

There are a few things in here which should go into Nengo master like `ConfigParam` and I will separate those out into an extra PR, but this does not need to block this PR.

This PR is based on #1202.

**How has this been tested?**
Existing tests have been adjusted to use the new syntax and still pass.

**Types of changes:**

**Checklist:**

**Still to do:**

* Modules have a `vocabs` parameter which can be easily confused with the `vocab` parameter. Maybe we should rename `vocabs` to `default_vocabs`?
* There might be places where the documentation did not get updated, but after all the SPA refactoring it is probably necessary to go through all the documentation anyways (and to write a lot fo new documentation).
* `BasalGanglia` has an `input_synapse` (default: 0.002) parameter. This is only used by the automated builder because in the legacy SPA this value was used. When manually creating connections to the BasalGanglia the normal connection default of 0.005 would be used. How important is this value? To me it seems to slightly simplify things and make them more consistent if we just used the standard 0.005 value.
* I haven't touched the associative memory at all yet. That is a slightly bigger job that requires a separate PR because there are several other issues to take care of.


jgosmann:
@tbekolay: flake8 fails on Travis-CI, but not on my local system, and the failures seem to be unrelated to the changes in this PR.

tbekolay:
@jgosmann Travis-CI reinstalls `flake8` each time, so they're new checks from a recent release. `pip install --upgrade` to see them locally (they're fixed in master, so feel free to cherry-pick a3bdda7234165b5c)

jgosmann:
I was about to say that I just upgraded `flake8`, but then I realized that `tox` uses its own venv. m(

tcstewar:
Lots of changes here, but I like all of them, and the code looks good.  :)

As for the open questions:

> Modules have a vocabs parameter which can be easily confused with the vocab parameter. Maybe we should rename vocabs to default_vocabs?

I'd support that, but can be done in a separate PR

> There might be places where the documentation did not get updated, but after all the SPA refactoring it is probably necessary to go through all the documentation anyways (and to write a lot fo new documentation).

Makes sense to me.

> BasalGanglia has an input_synapse (default: 0.002) parameter. This is only used by the automated builder because in the legacy SPA this value was used. When manually creating connections to the BasalGanglia the normal connection default of 0.005 would be used. How important is this value? To me it seems to slightly simplify things and make them more consistent if we just used the standard 0.005 value.

This got discussed in #1213 and keeping the current system seemed to be the consensus.

> I haven't touched the associative memory at all yet. That is a slightly bigger job that requires a separate PR because there are several other issues to take care of.

Definitely.

PR #1207, opened 2016-11-18
Config improvements

jgosmann:
**Motivation and context:**
I'm proposing a number of changes to the config system. Each change is explained in a bit more detail in the commit message.

Some of these changes would be helpful for the new-spa overhaul, but none (except introducing the `ConfigParam`) is strictly required.

**How has this been tested?**
Mostly refactoring, so I made sure that the existing tests still pass. Added a test for `ConfigParam`.

I'd recommend to do this PR commit by commit as the changes are mostly independent.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/learning_rules.py
+++ b/nengo/learning_rules.py
@@ -7,12 +7,10 @@ from nengo.utils.compat import is_iterable, itervalues
 
 
 class ConnectionParam(NengoObjectParam):
-    def validate(self, instance, conn):
+    def coerce(self, instance, conn):
         from nengo.connection import Connection
-        if not isinstance(conn, Connection):
-            raise ValidationError("'%s' is not a Connection" % conn,
-                                  attr=self.name, obj=instance)
-        super(ConnectionParam, self).validate(instance, conn)
+        self.check_type(instance, conn, Connection)
+        return super(ConnectionParam, self).coerce(instance, conn)
 
 
 class LearningRuleType(FrozenObject):
@@ -305,15 +303,7 @@ class Voja(LearningRuleType):
 
 
 class LearningRuleTypeParam(Parameter):
-    def validate(self, instance, rule):
-        if is_iterable(rule):
-            for r in (itervalues(rule) if isinstance(rule, dict) else rule):
-                self.validate_rule(instance, r)
-        elif rule is not None:
-            self.validate_rule(instance, rule)
-        super(LearningRuleTypeParam, self).validate(instance, rule)
-
-    def validate_rule(self, instance, rule):
+    def check_rule(self, instance, rule):
         if not isinstance(rule, LearningRuleType):
             raise ValidationError(
                 "'%s' must be a learning rule type or a dict or "
@@ -325,3 +315,11 @@ class LearningRuleTypeParam(Parameter):
         if rule.modifies not in ('encoders', 'decoders', 'weights'):
             raise ValidationError("Unrecognized target %r" % rule.modifies,
                                   attr=self.name, obj=instance)
+
+    def coerce(self, instance, rule):
+        if is_iterable(rule):
+            for r in (itervalues(rule) if isinstance(rule, dict) else rule):
+                self.check_rule(instance, r)
+        elif rule is not None:
+            self.check_rule(instance, rule)
+        return super(LearningRuleTypeParam, self).coerce(instance, rule)
--- a/nengo/params.py
+++ b/nengo/params.py
@@ -5,8 +5,8 @@ import numpy as np
 
 from nengo.exceptions import (
     ConfigError, ObsoleteError, ReadonlyError, ValidationError)
-from nengo.utils.compat import (
-    is_array, is_integer, is_number, is_string, itervalues)
+from nengo.utils.compat import (int_types, is_array, is_integer, is_number,
+                                is_string, itervalues, string_types)
 from nengo.utils.numpy import array_hash, compare
 from nengo.utils.stdlib import WeakKeyIDDictionary, checked_call
 
@@ -109,8 +109,7 @@ class Parameter(object):
         return self.data.get(instance, self.default)
 
     def __set__(self, instance, value):
-        self.validate(instance, value)
-        self.data[instance] = value
+        self.data[instance] = self.coerce(instance, value)
 
     def __repr__(self):
         return "%s(%r, default=%s, optional=%s, readonly=%s)" % (
@@ -133,8 +132,29 @@ class Parameter(object):
     def set_default(self, obj, value):
         if not self.configurable:
             raise ConfigError("Parameter '%s' is not configurable" % self)
-        self.validate(obj, value)
-        self._defaults[obj] = value
+        self._defaults[obj] = self.coerce(obj, value)
+
+    def check_type(self, instance, value, type_):
+        if value is not None and not isinstance(value, type_):
+            if isinstance(type_, tuple):
+                type_str = " or ".join((t.__name__ for t in type_))
+            else:
+                type_str = type_.__name__
+            raise ValidationError("Must be of type %r (got type %r)."
+                                  % (type_str, type(value).__name__),
+                                  attr=self.name, obj=instance)
+
+    def coerce(self, instance, value):
+        if isinstance(value, DefaultType):
+            raise ValidationError("Default is not a valid value. To reset a "
+                                  "parameter, use 'del'.",
+                                  attr=self.name, obj=instance)
+        if self.readonly and instance in self.data:
+            raise ReadonlyError(attr=self.name, obj=instance)
+        if not self.optional and value is None:
+            raise ValidationError("Parameter is not optional; cannot set to "
+                                  "None", attr=self.name, obj=instance)
+        return value
 
     def equal(self, instance_a, instance_b):
         a = self.__get__(instance_a, None)
@@ -153,17 +173,6 @@ class Parameter(object):
         else:
             return id(value)
 
-    def validate(self, instance, value):
-        if isinstance(value, DefaultType):
-            raise ValidationError("Default is not a valid value. To reset a "
-                                  "parameter, use 'del'.",
-                                  attr=self.name, obj=instance)
-        if self.readonly and instance in self.data:
-            raise ReadonlyError(attr=self.name, obj=instance)
-        if not self.optional and value is None:
-            raise ValidationError("Parameter is not optional; cannot set to "
-                                  "None", attr=self.name, obj=instance)
-
 
 class ObsoleteParam(Parameter):
     """A parameter that is no longer supported."""
@@ -180,10 +189,11 @@ class ObsoleteParam(Parameter):
             return self
         self.raise_error()
 
-    def validate(self, instance, value):
+    def coerce(self, instance, value):
         if value is not Unconfigurable:
             # don't allow setting to anything other than unconfigurable default
             self.raise_error()
+        return value
 
     def raise_error(self):
         raise ObsoleteError(self.short_msg, since=self.since, url=self.url)
@@ -194,11 +204,9 @@ class BoolParam(Parameter):
 
     equatable = True
 
-    def validate(self, instance, boolean):
-        if boolean is not None and not isinstance(boolean, bool):
-            raise ValidationError("Must be a boolean; got '%s'" % boolean,
-                                  attr=self.name, obj=instance)
-        super(BoolParam, self).validate(instance, boolean)
+    def coerce(self, instance, value):
+        self.check_type(instance, value, bool)
+        return super(BoolParam, self).coerce(instance, value)
 
 
 class NumberParam(Parameter):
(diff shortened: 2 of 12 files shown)


Seanny123:
The last fixup is making things break on Python 3.

jgosmann:
Nope, that's just Travis-CI having trouble downloading Conda. I restarted the builds.

Seanny123:
LGTM. I especially like the new TypeParam and the reduction in boiler-plate code it causes.

jgosmann:
Hey, can we get this reviewed and merged? I'm working on something right now where this would be useful.

jgosmann:
> I don't like the changes to the validate. It feels wrong to modify the thing that you're validating inside the validate method. While it may result in slightly more lines of code, any modifications should be done in __set__ to make it clear where we are doing validation and where we are changing the thing that has been validated.

1. `NdarrayParam` is already using `validate` to cast the value prior to this PR. This isn't an argument to do this for all params, but there is a consistency argument to do it either for all params or none (i.e. change `NdarrayParam` to not use `validate` to cast the value).
2. In many cases the most robust validation is trying to convert the value to the desired type and see if it fails. Thus, we already do the type cast already in `validate`. Doing it in `__set__` again is double effort in CPU cycles and lines of code.
3. By using the value returned by `validate` it is less likely that the `validation` code gets out of sync with the `__set__` code which could lead to values that validate, but cannot be set.
4. When overwriting `__set__` it is easy to forget to call `validate`.
5. Currently, the code in `ConnectionFunctionParam.__set__` seems to do implement some part of the validation. This doesn't really make it clear where things are validated in my opinion. It also means, that calling `validate` does not tell me whether the value can actually be assigned/not assigned.
6. Slightly better support for inheritance. A deriving class can more easily extend how the casting is done because it can either modify the return value of the super `validate` or change the value first before passing it to the super `validate`. Might not be that relevant because we are not doing that sort of inheritance. I think I remember wanting to do this sort of inheritance, but don't know anymore for what (probably something that is obsolete now?).

> I'm also not a fan of the new TypeCheckedParameter base class. I feel like it adds a level of indirection (which hurts readability) without the usual benefit of reduced code (checking the type of a parameter is never more than a line or two).

To me the reduce cognitive load when implementing a new `TypeCheckedParameter` is more important than the decrease in readability. Right now I have to remember to also check for `None` and then (more importantly) what should go into the `ValidationError` error message (or find a different parameter that does that stuff to look it up). `TypeCheckedParameter` also ensures that the error message is consistent for different types. (`nengo_spa` implements a bunch of new parameter types, in #1307 I implemented a new parameter type, so this has been affecting me repeatedly).

jgosmann:
> Sorry for not making the changes myself, but the GUI refactoring is taking up my time at the moment.

Should be mostly a matter of dropping those two commits (if that's what we decide), so no worries.

tbekolay:
Re: `validate` and `__set__` changes, the original intention was to have `validate` validate, and `__set__` mutate if necessary. You've convinced me that we could put these together into one method instead (and discourage overriding `__set__`), but I don't think that method should be called `validate`. Any ideas? Right now what comes to mind is `cast` or `coerce`... Also, it occurs to me that since we raise `ValidationError` when things go wrong, it's still clear that this method is doing validation :+1:

> Right now I have to remember to also check for None and then (more importantly) what should go into the ValidationError error message (or find a different parameter that does that stuff to look it up). TypeCheckedParameter also ensures that the error message is consistent for different types.

How about adding a `check_type` method to the `Parameter` base class instead? Classes that would inherit from `TypeCheckedParameter` can call that method.

jgosmann:
> Any ideas? Right now what comes to mind is cast or coerce

I learn towards `coerce`, but `cast` is fine too. The exact definition of the terms seems to vary depending on the programming language. It seems that a cast is sometimes assumed to be explicit and a coercion to be implicit (when assigning a parameter it is implicit).

References:
http://stackoverflow.com/questions/1602122/casting-vs-coercion-in-python
http://stackoverflow.com/questions/8857763/what-is-the-difference-between-casting-and-coercing

> How about adding a check_type method to the Parameter base class instead?

Not sure. In an inheritance hierarchy it wouldn't be clear whether one of the base classes is already checking the type. It also feels a bit weird to have that method on the Parameter class to me, maybe it's more like a utility function, but that also has its cons. Overall it would definitely alleviate my main concern, so I would be fine with it.

tbekolay:
I'm implementing the changes discussed at the dev meeting now.

One thing I realized is that `coerce` happens to be a [Python builtin](https://docs.python.org/2/library/functions.html#coerce), so calling a method `coerce` shadows that builtin. However, the documentation lists this as a "Non-essential Built-in Function", and I had no idea it existed until today, so I'm inclined to think it's fine to repurpose the name.

If this is objectionable to anyone, speak now or forever hold your peace! If it matters, the alternative name, `cast`, is not a builtin.

jgosmann:
It seems that he `coerce` built-in has been removed in Python 3.

tbekolay:
OK, I pushed two commits that handle the issues I brought up in my review. @jgosmann, take a look at them and let me know if you have any issues with them.

I also noticed something while making these changes that I wanted to check with you @jgosmann. In `ConfigParam`, we [automatically make a Config if None is passed](https://github.com/nengo/nengo/pull/1207/files#diff-22d165ffd2524c330057bc1c5868c111R464). Is this an important part of the `ConfigParam`? It kind of goes contrary to some things we've typically kept as design constraints, which is that when you set something into a param, you should be able to get it back, at least to a certain extent (we do change some things, but those are important exceptions rather than the rule). It also makes the semantics of the `optional` flag confusing... Currently, it is used to say that `None` is a valid setting, which means that it isn't set. If `None` gets auto-coerced to a bare `Config`, then `None` is a valid setting regardless of the state of `optional`, and in both cases setting `obj.config = None` results in `isinstance(obj.config, Config) == True`.

Would it be acceptable to, instead, force `ConfigParam` to never be optional? Or just ensure that in the situations where you use `ConfigParam` in this manner you set `optional=False`?

jgosmann:
ba8a923 looks good to me.

The `optional` semantics are an important point and it would be great to fix this. I think the problem is that we cannot safely provide a default `Config` via the `default` argument because all class instances would share the same `Config` instance as default. Changing a parameter on one of them would change all. (It is basically the same reason why one should use `def fn(arg=None):` instead of `def fn(arg=[]):`.) I'm also not sure whether it makes sense to distinguish between no config set and an empty config set.

As reference: [this seems to be the only place where `ConfigParam` gets currently used.](https://github.com/nengo/nengo_spa/blob/5b0fa511ee53452224c9ecce9784012b801f4d09/nengo_spa/basalganglia.py#L147)

> Would it be acceptable to, instead, force ConfigParam to never be optional?

Potentially, would that still work with `Default`?

tbekolay:
> I think the problem is that we cannot safely provide a default Config via the default argument because all class instances would share the same Config instance as default.

Hm... yeah, this is a tough one because, unlike in other similar situations, we expect people to mutate `config` objects. We could make this a special case, similar to [NengoObjectParam](https://github.com/nengo/nengo/blob/master/nengo/base.py#L200), where we disallow setting a default (we force it to be `Unconfigurable`), then in `coerce`, we would do

```python
value = Config() if self.optional and value is Unconfigurable else value
```

By doing this, if the config is non-optional and one isn't passed, it stays as `Unconfigurable` and will have a validation error raised when trying to access it.

jgosmann:
That sounds good to me if I'm not overlooking anything.

tbekolay:
> Potentially, would that still work with `Default`?

I don't think `Network` is set up to support `Default`, only `NengoObject`.

jgosmann:
> I don't think Network is set up to support Default, only NengoObject.

But `nengo_spa.Network` is.

tbekolay:
Alrighty. I'll make the `ConfigParam` change.

tbekolay:
Ok, I'm looking at [`test_configparam`](https://github.com/nengo/nengo/pull/1207/files#diff-c2017afc21f73a014653594c703fc1f0R226) and I'm back to being confused... There, you set a default config, which would have the issues that you point out (it's easy to modify that config, which modifies all instances of the `Test` class). So is this behavior that you want or don't want?

jgosmann:
Not sure what I was thinking. 😕 I'd say go ahead and change the test.

tbekolay:
I can understand the desire to have a default config (like, a config that gets used as the default if none is explicitly set) because then you can have some classes set as configurable at the start -- we do this in `Network` [here](https://github.com/nengo/nengo/blob/master/nengo/network.py#L127). But with the current parameter system it's difficult to set that up, I think. Probably something to revisit later; for now I'll change the test :+1:

tbekolay:
I pushed a fixup with a `ConfigParam` that works as you described it, but it is pretty hacky.

Are you sure you need a `ConfigParam`? In #814 we removed passing configs to working memory newtorks based on discussion in #790. I thought the same was happening with `BasalGanglia`. Are the alternatives in #790 insufficient for the SPA system?

jgosmann:
> In #814 we removed passing configs to working memory newtorks based on discussion in #790.

This never happened for the `BasalGanglia` network, so the [`BasalGanglia` network in nengo_spa](https://github.com/nengo/nengo_spa/blob/master/nengo_spa/basalganglia.py) is still using configs. But looking at it again, it seems that those configs are pretty much only used for two things:
* Control ampa and gaba synaptic time constants. This could be equally well done by just using two `SynapseParams`.
* Give the ability to easily use different configs for ampa, gaba, and all other connections. By creating these connections in corresponding network they could be grouped and later be modified if desired.

So I guess I don't need `ConfigParam`. Especially not in core Nengo. (It is implemented in `nengo_spa` too to not depend on this particular branch.)

jgosmann:
@tcstewar, does this seem reasonable to you? You're more familiar with the basal ganglia network, I suppose.

tcstewar:
> @tcstewar, does this seem reasonable to you? You're more familiar with the basal ganglia network, I suppose.

Yup!  It looks like it's just that the `nengo.networks.actionselection.BasalGanglia` never got changed to using a parameter instead of the configs.  But I don't see a reason, as it's just the two `SpynapseParams`.

tbekolay:
Ok cool. Alright if I remove those commits then? I'll rebase, fix up the history, then ping this thread again for re-reviews.

jgosmann:
👍

tbekolay:
Made those changes, @jgosmann and @Seanny123 can you re-review?

jgosmann:
LGTM

nengo/params.py
```diff
@@ -130,8 +132,29 @@ def get_default(self, obj):
...
+            raise ValidationError("Must be of type %r (got type %r)."
+                                  % (type_str, type(value).__name__),
+                                  attr=self.name, obj=instance)
+
+    def coerce(self, instance, value):
+        if isinstance(value, DefaultType):
+            raise ValidationError("Default is not a valid value. To reset a "
+                                  "parameter, use 'del'.",
+                                  attr=self.name, obj=instance)
+        if self.readonly and instance in self.data:
```
Seanny123: I may have brought this up before, but when there's a bunch of chained errors, I feel like they should be `elif` instead of just `if`. I think the conclusion we came to last time was that an `elif` implies priority when there is none in this case?

nengo/learning_rules.py
```diff
@@ -7,12 +7,10 @@
 
 
 class ConnectionParam(NengoObjectParam):
```
Seanny123: This is also going to get removed in #1310, but I guess that's obvious enough that it isn't worth mentioning?

Seanny123:
I had a few nitpicky comments into regards to style, but I'm fine with this getting merged without addressing them. I really like how `coerce` combine the `validate` and `__set__` in all the subclasses, making further subclasses easier to write.

nengo/params.py
jgosmann: I think `elif` is for mutually exclusive cases? I don't think these are mutually exclusive, e.g., a parameter could be not optional and readonly at the same time.

nengo/learning_rules.py
tbekolay: Ah yeah, I forgot about that. This'll get merged first probably and the work's done so no harm in changing it.

nengo/params.py
tbekolay: Yeah, these aren't mutually exclusive, though it's true that the same thing will happen if they're `elifs`.

For me, the reason I prefer `if` statements here is because each check is an independent thing. Typically with `if`/`elif` chains, you have to keep in mind the previous conditions when looking at each condition. E.g.,

```python
if a:
    # do A things
elif b:
    # do B things
elif c:
    # do C things
```

The second condition is actually entered when `not a and b`; similarly the third condition is `not a and not b and c`. The ability to simplify the condition with `elif`s is convenient, and usually the logic is pretty clear. But in this case, it doesn't help the reader to have the `elif` because the previous if statements are irrelevant to the current one. Like, it sort of matters in that they will never be evaluated if the previous if statement is true, but you don't have to think about the previous ones when looking at an individual if statement.

Another reason why I prefer `if`s here is that it follows a typical and useful pattern:

```python
if thing.will_fail:
    # quickly exit this function because we're going to fail; either
    return None  # keep going because failure is ok
    raise Error  # stop because we can't continue
```

When I see that kind of thing, I know what's up without having to think much about it. In this case, most of `coerce` is a bunch of instances of this pattern, so I find it easier to scan than if they had been elifs. Either would act the same, though.

Seanny123:
I totally accept both of your reasoning and accept that it should not be
changed. Thanks for taking the time to explain this to me.

tbekolay:
:smile: Merging!

Issue #1144, opened 2016-08-16
Should networks and SPA modules stay together?


tbekolay:
Github recently made is possible to [change the base branch of a pull request](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request), so I've changed all of the open [SPA-labelled PRs](https://github.com/nengo/nengo/pulls?q=is%3Apr+label%3ASPA+is%3Aopen) to be based on the `new-spa` branch rather than `master`.

In doing so, however, I noticed a few cases in which the SPA PRs touch non-SPA code. Specifically:

1. #743 modifies `nengo/utils/numpy.py`
2. #810 modifies `nengo/networks/actionselection.py`
3. #991 modifies `nengo/networks/assoc_mem.py`
4. #1104 modifies `nengo/networks/circularconvolution.py`

I think the change in #743 is clearly something that should be split off into a separate PR, but the other three instances all touch the non-SPA versions of networks. So the question that I'd like people's opinions on is whether these networks should live in the new SPA module or whether they should remain in core Nengo?

tbekolay:
My thoughts: there was some talk previously about splitting off all or most of the networks in Nengo to another repo, so it would make sense to me that the underlying networks that SPA makes would move along with the modules to be in the new SPA repo. If we add some infrastructure in Nengo or Nengo GUI to detect installed networks, then the new SPA repo would use that mechanism like any other repo. So it seems to me that there aren't any significant downsides to moving them with the rest of the SPA module.

jgosmann:
I think we might have three sets of networks:

1. Core networks that should stay in core Nengo.
2. Extra networks that could be moved to an extra package (nengo_networks or nengo_extra or multiple packages).
3. SPA related networks.

Regarding 2. (extra networks) I wonder what would be the advantages or disadvantages of keeping those networks with core Nengo?

Regarding 3. (SPA networks)  it would probably make sense to move them to nengo_spa (even if they could be used without SPA modules, they will mainly used in that context).

Looking at our list of networks, I would them sort in the following way:

**Core**
CircularConvolution (not entirely sure, where to sort this; it is a pretty basic operation and used a lot, but mostly in the context of SPA, but not dependent on it)
EnsembleArray
Integrator
Oscillator
Product

**Extra**
InputGatedMemory

**SPA**
actionselection (BasalGanglia, Thalamus)
AssociativeMemory


One worry about too much fragmantation is that it makes it hard to discover existing networks. If they are all in one place, there's only one place to check in.

[...]

PR #1208, opened 2016-11-18
Change circular convolution scaling.

jgosmann:
**Motivation and context:**
The spaopt methods (which hopefully will be finally implemented as helper function for new-spa) require the scaling factor on the FFT transforms to be 1/sqrt(d) to preserve the distribution of vector lengths. This PR changes the scaling accordingly.

**How has this been tested?**
Tests still pass.

**Types of changes:**

- Bug fix (non-breaking change which fixes an issue) [not really a bug, but also not a new feature]

**Checklist:**

### DIFF
--- a/nengo/networks/circularconvolution.py
+++ b/nengo/networks/circularconvolution.py
@@ -49,6 +49,7 @@ def transform_in(dims, align, invert):
             tr[i] = row.real if i % 4 == 0 or i % 4 == 3 else row.imag
 
     remove_imag_rows(tr)
+    tr /= np.sqrt(dims)
     return tr.reshape((-1, dims))
 
 
@@ -66,9 +67,6 @@ def transform_out(dims):
 
     tr = tr.reshape(4*dims2, dims)
     remove_imag_rows(tr)
-    # IDFT has a 1/D scaling factor
-    tr /= dims
-
     return tr.T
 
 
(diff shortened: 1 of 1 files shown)


Seanny123:
LGTM. Maybe a comment somewhere explaining the vector length thingy? Or will that be documented elsewhere? Either way, I don't feel strongly enough about this to make it a blocking change.

nengo/networks/circularconvolution.py
```diff
@@ -195,8 +193,9 @@ def CircularConvolution(n_neurons, dimensions, invert_a=False, invert_b=False,
     with net:
         net.A = nengo.Node(size_in=dimensions, label="A")
         net.B = nengo.Node(size_in=dimensions, label="B")
-        net.product = Product(n_neurons, tr_out.shape[1],
-                              input_magnitude=input_magnitude * 2)
+        net.product = Product(
+            n_neurons, tr_out.shape[1],
+            input_magnitude=input_magnitude * 2. / np.sqrt(dimensions))
```
jgosmann: This changes an assumption underlying this network. Without the `sqrt(dimensions)` scaling each input dimension is regarded as uniformly distributed. With the scaling the input vectors are assumed to be unit length (times `input_magnitude`). It seems to be unlikely (or at least much rarer) that this network is used in cases where the first assumption holds true. I'm not entirely sure what the purpose of the 2 factor is.

jgosmann:
I have to reconsider the scaling on the `input_magnitude` in light of an improved spa-opt form. Also, it seems that there is some consensus (#1144) that the circular convolution network should move to `nengo_spa` which would allow us to make theses changes without worrying about backwards compatibility.

Issue #1109, opened 2016-06-23
Get object references for automagically created SPA networks


xchoo:
The current SPA system automagically creates networks (typically in the BG & cortical action systems). Unfortunately, adjusting the properties of these networks (e.g. to change the radii of the internal ensembles) is pretty tedius. Having a helper function (or some way of specifying the parameters you want to use for said connection) would be real great!

jgosmann:
That's something I want to take into consideration with #1014, #941, etc.
As discussed in the last dev meeting we probably won't implement any new features in the current SPA system and move the development to a new repository. But to make this new repository the official SPA nengo/nengo_gui#797 has to be resolved.

I added a SPA label for issues to easily find these issues that need to be considered once the new repository is created.

xchoo:
Ah yes. I was just making the feature request as a reminder. :smile:

jgosmann:
Please take a look at #1209. Does that seem sufficient?

jgosmann:
Done in #1209.

PR #1209, opened 2016-11-21
New spa get objects

jgosmann:
**Motivation and context:**
This PR allows it to access the objects created by the SPA compiler. It also ensures that the SPA compiler does not set (or overwrites) attributes on user created networks. These changes address #1109.

Normally one would use the config system to set parameters of automatically created networks:
```python
with spa.Module() as model:
    # create some modules
    with nengo.Config(spa.Bind, spa.BasalGanglia, spa.Thalamus) as cfg:
        cfg[spa.Bind].neurons_per_dimension = 500
        cfg[spa.BasalGanglia].n_neurons_per_ensemble = 200
        cfg[spa.Thalamus].route_inhibit = 10.
        actions = spa.Actions(...)
        actions.build()
```

Those changes, however, affect all modules created by the actions. Sometimes more fine-grained control is needed or one might want to access the created basal ganglia or thalamus objects. Thus, the `build` function returns three things:
```python
bg, thalamus, constructed = actions.build()
```

`bg` and `thalamus` are the `BasalGanglia` and `Thalamus` object. These also store related objects that get created, e.g. `bg.input_connections`, `thalamus.channels`, or `thalamus.gate_in_connections`. This things can either be lists in the order things were added or dictionianries indexed by the action index (depending on what information is available and whether there is a 1:1 correspondence). Finally, `constructed` is a dictionary of all other objects constructed by the `build` function. The keys are the AST nodes that generated specific objects and the values are lists of these objects. For example, for the rule `c = a * b` there will be a `Product` node in the syntax tree. Using that as key for `constructed` it will return a list containing a `spa.Bind` module and two connections into that module. Note that objects from child nodes will not be included.

For that indexing one needs the corresponding nodes from the syntax tree. The AST for each rule can be accessed on the `Actions` object either by integer indexing (for unnamed rules in order they were given and named rules in random (?) order) or string indexing (for named rules). Then one can navigate through the syntax tree to find the node to access the constructed objects one is looking for.

This is still somewhat tedious. My main goal here was to make it possible to somehow access each automatically created object without the need to iterate through all objects in the model which would be intermixed with non-automatically created objects. To me it seems a rare enough use case to leave it at this for now (given that most things can be done with the config system).

Based on new-spa-overhaul, PR #1206.

**How has this been tested?**
Added some tests.

**Types of changes:**

**Checklist:**


nengo/spa/actions.py
```diff
@@ -165,37 +164,28 @@ def __getitem__(self, key):
...
+    respectively.
 
     The keyword argument `vocabs` is special in that it provides a dictionary
     mapping names to vocabularies. The vocabularies can then be used with those
     names in the action rules.
     """
 
-    def __init__(self, *args, **kwargs):
-        super(Actions, self).__init__(BasalGanglia, Thalamus)
+    def __init__(self, *args, vocabs=None, **kwargs):
```
tcstewar: Python2 doesn't support this syntax.  :(  Here's an alternate implementation that I think does what you're intending here:

```python
        super(Actions, self).__init__()

        self.actions = []
        self.named_actions = {}

        self.construction_context = None

        self.parse(*args, **kwargs)

    def parse(self, *args, **kwargs):
        if kwargs.has_key('vocabs'):
            vocabs = kwargs['vocabs']
            del kwargs['vocabs']
        else:
            vocabs = None

        sorted_kwargs = sorted(kwargs.items())
        ...
```

tcstewar:
Other than the two comments above, this looks good to me!

jgosmann:
Cool, will fix those things and merge when I have time.

PR #1214, opened 2016-11-25
New spa vocab

jgosmann:
**Motivation and context:**
This reworks the `Vocabulary` and `SemanticPointer` class for the new-spa system. Highlights:

* `SemanticPointers` are readonly/immutable to prevent accidental changes. Each mathematical operation will return a new `SemanticPointer`.
* `Vocabulary.vectors` is readonly to prevent accidental modification.
* `Vocabulary` implements the dictionary interface because that's what it basically is.
* When creating a new `Vocabulary` instance it will be in `strict` mode by default. That means that `parse` will not automatically add new pointers and all pointers have to be added either by `add` or the new `populate` method. This is intended to prevent accidentally adding vectors (by typos or conversion between vocabs) and replaces the `readonly` functionality. But sometimes, when building simple models, we don't want to explicitely define all the vocabs beforehand. Because of that all automatically generated vocabulary instances will be non-strict and automtically add unknown vectors in `parse` operations.
* Removed a bunch of functionality from `Vocabulary` which I think doesn't belong there. It now mostly lives in `spa/utils.py`.

##### The populate method.
The populate method is an easy way to populate your vocabulary:
```python
v = spa.Vocabulary(64)
v.populate('A; B; C')  # Gives you three semantic pointers names A, B, C
v.populate('D = A + B')  # Gives you a semantic pointer named D which is defined as A + B (non unit length)
v.populate('E = (A + B).normalized()')  # Same as the previous line, but normalized
v.populate('F.unitary()')  # Gives a semantic pointer named F that is unitary.
v.populate('G = (2 * D + A).unitary()')  # Unitary pointer G based on 2 * D + A
```

###### spa.similarity
The `similarity(data, vocab)` function takes some 2d data (e.g., probed from a network, time points by dimensions) and compares it to all vectors in a vocabulary (returns times by #vectors array). `data` can now also be a one-dimensional vector or `SemanticPointer` in which case a 1d array is returned. `vocab` may also be a list of 1d arrays to compare to or a list of `SemanticPointer`s. The `Vocabulary.parse_n` method and `similarity` have been designed to play well together. So you can do something like:
```python
spa.similarity(data, vocab.parse_n('A', 'A * B', '(C + D).normalized() * E'))
```
That seems pretty useful to me and hopefully makes it easy to add support for arbitrary expressions to the Nengo GUI Semantic Pointer plot.

##### pairs
Vocabularies used to have an option to store all convolutional pairs of its elements. This seemed to be only relevant for similarity calculations and be a special case that the vocab shouldn't take care of in my opinion. But it's certainly helpful to include pairs in some plots. That's why there is the `pairs(vocab)` helper functions in `utils.py`. It takes a vocab and returns strings for all convolutional pairs:
```python
vocab = Vocabulary(64)
vocab.populate('A; B; C')
pairs(vocab)  # Returns {'A*B', 'A*C', 'B*C'}
```
To include those pairs in the similarity function:
```python
spa.similarity(data, vocab.parse_n(*pairs(vocab)))
spa.similarity(data, vocab.parse_n*(list(vocab.keys()) + pairs(vocab)))
```

Parsing all those vectors might take a while (that's why they were stored with the vocab before I assume), but you can store the result of the `vocab.parse_n` call if you need it in multiple places.

##### text
The `text` function stayed mostly the same, but uses `spa.similarity` now internally which allows it to accept arbitrary `terms` and avoids computing more similarities than needed.

##### prob_cleanup
Uses now `nengo.dists.CosineSimilarity` which should make it more accurate, but also introduces a dependency on SciPy. But it doesn't seem like a widely used function.

Technically this should be independent of the other SPA PRs, but because some updates to examples and tests are required it does not cleanly apply to the `new-spa` branch and I based the PR on `new-spa-overhaul`.

**How has this been tested?**
Test have been updated to new syntax and locations of function. Some additional tests were added.

**Types of changes:**

**Checklist:**


jgosmann:
Completed the PR description. ✍️

Seanny123:
FYI, flake8 is complaining and Python2 isn't liking the non-ASCII characters.

jgosmann:
> flake8 is complaining

That is unrelated to the SPA stuff.

> Python2 isn't liking the non-ASCII characters

Should be fixed now.

jgosmann:
@tcstewar, any prediction when you will have a look at this?

nengo/spa/pointer.py
```diff
@@ -150,35 +139,24 @@ def distance(self, other):
...
+        super(Identity, self).__init__(data)
 
-    def copy(self):
-        """Return another semantic pointer with the same data."""
-        return SemanticPointer(data=self.v)
 
-    def mse(self, other):
-        """Return the mean-squared-error between two vectors."""
-        return np.sum((self - other).v**2) / len(self.v)
+class AbsorbingElement(SemanticPointer):
```
tcstewar: Is this really an absorbing element?  convolution of an arbitrary vector with one of these will produce a scaled version of the absorbing element...  but I can't think of a better name.....

jgosmann: I think for our purposes it pretty much is. You get a scaled version only if the other vector is not unit length and every information about what the vector except its length will be lost.

nengo/spa/vocab.py
```diff
@@ -21,149 +26,109 @@ class Vocabulary(object):
...
+
+    def __iter__(self):
+        return iter(self._keys)
 
     def __getitem__(self, key):
-        """Return the semantic pointer with the requested name.
+        """Return the semantic pointer with the requested name."""
+        if not self.strict and key not in self:
+            assert '.' not in key
+            assert ';' not in key
```
tcstewar: Why these asserts in particular?  Things also go badly if someone does `vocab['A+B']` or any other combination like that.

jgosmann: `vocab['A+B']` will actually work. It might not be super useful because the only way to access the SP `'A+B'` is via the `__getitem__` method; it cannot be used in parsed expressions, but it doesn't break anything. Names containing `'.'` or `';'` (and looking at the code also `'='` which should probably be added) will lead to a `KeyError` because `populate` will split the name at those symbols, but `__getitem__` looks for the full name. I think I inserted the asserts to fail early before `populate` already modified the vocab in an unintended way.

jgosmann: I suppose after all there isn't really a reason to use `populate` because the parsing functionality of it isn't used if none of those symbols is in the name. We could just do a `self.add(key, self.create_pointer())` without the asserts.

tcstewar: > vocab['A+B'] will actually work.

Hmm, I get this:

```python
v = nengo.spa.Vocabulary(5)
v['A+B']
```

```
KeyError                                  Traceback (most recent call last)
<ipython-input-3-edeabdf48a9c> in <module>()
      1 v = nengo.spa.Vocabulary(5)
      2 
----> 3 v['A+B']

/home/tcstewar/github/nengo/nengo/spa/vocab.pyc in __getitem__(self, key)
    116             assert ';' not in key
    117             self.populate(key)
--> 118         return self.pointers[key]
    119 
    120     def __hash__(self):

KeyError: 'A+B'
```

jgosmann: It only works in non-strict mode (`spa.Vocabulary(d, strict=False)`) if you do not add that key beforehand.

jgosmann:
```
In [1]: v = spa.Vocabulary(5)

In [2]: v.add('A+B', v.create_pointer())

In [3]: v['A+B']
Out[3]: <nengo.spa.pointer.SemanticPointer at 0x7fcf0f5d7278>
```

tcstewar: Argh!  You're entirely right.  I'm an idiot.  :)

Issue #1217, opened 2016-11-28
Conda fails in Travis a lot


Seanny123:
Often the reason for our unit-tests failing is that one of the conda packages failing to download.

How silly of an idea would it be to host our own mirror of `miniconda` and the Conda packages we need to reduce the chances of the download failing?

tbekolay:
Yes, conda's flakiness is unfortunate. Hosting a mirror of miniconda and relevant packages would be an option, though might end up costing money so would hope to avoid it... another option might be to use TravisCI's [caching mechanism](https://docs.travis-ci.com/user/caching/); I'd have to look at it more closely, but you can cache [arbitrary directories](https://docs.travis-ci.com/user/caching/#Arbitrary-directories), so caching the directory with conda might give us what we want?

Seanny123:
[Some other people](https://github.com/broadinstitute/viral-ngs/blob/31d8e5ae993444dbb2979eb35440ad45a900d0b0/.travis.yml#L5) are already caching with conda, so I'll make a pull request to try that out and see if it works. While I'm at it, what do you feel about moving the scripts we currently have in `travis.yml` into their own `travis/` directory so they're easier to read and edit, like they do in the linked repository?

tbekolay:
> what do you feel about moving the scripts we currently have in travis.yml into their own travis/ directory

Yeah, I've seen this in a few places. I would consider it if it reduced code duplication (e.g., if some of the scripts were shared between TravisCI and Appveyor) or if our .travis.yml was super long, but at less than 100 lines I think I'd prefer to keep it as one file for now. If either of the those criteria changes, we can revisit.

jgosmann:
I wonder if this is still an issue. I can't remember seeing a TravisCI failure because of conda, but that might not mean anything.

tbekolay:
I agree, I haven't seen a failure related to Conda in quite some time (there was definitely a time when it happened a fair bit). Closing!

PR #1222, opened 2016-11-30
Improve Travis CI

Seanny123:
**Motivation and context:**
We could make Travis CI run faster and more reliably if we use caching, as discussed in #1217. Also, reduced the amount of Nengo history we were cloning from Git and make sure pip was installing in the user space.

**How has this been tested?**
Tested by trying to get the tests to pass on Travis

Start by reading the links associated the issue.

**Types of changes:**

**Checklist:**


Seanny123:
At this point, `conda` is failing to install `numpy` as a module for reasons I don't understand. I'm going to sleep on it and revisit this later. Alternatively, @tbekolay if you have any ideas, let me know.

tbekolay:
I pushed a squash commit putting it back into one file. I'm also not convinced we need the `CACHE_NAME` variables since we have other environment variables that should disambiguate the jobs (if I'm reading [this](https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices) correctly).

> conda is failing to install numpy

This happened because we use a conda environment (called test). It was being activated (`source activate test`) in the `install-conda.sh` script, not in `.travis.yml`. That would be fine if we sourced `install-conda.sh`, but we were executing it, which makes a subshell rather than executing it in the current shell.

Putting it back into one file has the nice side-effect of fixing that since we're not calling another script ;)

Seanny123:
I bow down to your awesome Travis and Conda skills.

Seanny123:
This LGTM. I don't know who technically "owns" this pull request, but I'm going to go ahead and say it needs review by one more person.

tbekolay:
So one thing that's a bit concerning to me is that the jobs aren't actually running faster at all; I've restarted them a few times and it seems like the cost to load things up from the cache is about the same as installing Miniconda. Without the speedup, I feel like we're opting for one suboptimal process over another; I don't mind restarting builds that fail for Conda reasons, but we could get into really weird situations with a cache where we use the wrong versions of things, which will be annoying to debug. What do you think @Seanny123?

Seanny123:
> we could get into really weird situations with a cache where we use the wrong versions of things, which will be annoying to debug

Actually, that's a good point. Does this script capture the case that a new version of packages are released?

tbekolay:
> Does this script capture the case that a new version of packages are released?

It does not right now; it could be made to do so by doing the `conda update --all` whether the conda is cached or not (TravisCI will re-cache any changes to cached files).

I think I'm still :-1: on using the cache as switching to it will mean that your install process isn't tested.

Seanny123:
> using the cache as switching to it will mean that our install process isn't tested

One possible solution to that is have a Travis CI environment that just tests the build process, but then I guess that only reduces the probability of the `conda` download failing, which doesn't seem ideal. In which case I think I'll close this pull request (without deleting the branch) and investigate making our own private mirror of `conda` using something as simple as a `cron` job and an sftp server.

Issue #579, opened 2014-12-08
Generate Poisson or regular spikes from a rate model


tbekolay:
This was briefly discussed in #533. The basic idea is to make a neuron model that takes in a rate based neuron model, and produces a spiking approximation of that firing rate by generating spikes at regular intervals, or from a Poisson process.

tcstewar:
I took a stab at doing this with Nodes in the past, so these functions might be useful:

Regular spiking (i.e. given a constant rate, this will produce regularly spaced spikes):

```python
class RegularSpiking(object):
    def __init__(self, size, dt=0.001):
        self.state = np.zeros(size)
        self.threshold = 1.0 / dt
        self.output = np.zeros(size)
    def __call__(self, t, x):
        self.output[:] = 0
        self.state += x
        spikes = np.where(self.state > self.threshold)
        self.output[spikes] = self.threshold
        self.state -= self.output
        return self.output
```

Poisson spiking:

```python
class PoissonSpiking(object):
    def __init__(self, size, seed, dt=0.001):
        self.rng = np.random.RandomState(seed=seed)
        self.dt = dt
        self.value = 1.0 / dt
        self.size = size
        self.output = np.zeros(size)
    def __call__(self, t, x):
        self.output[:] = 0
        p = 1.0 - np.exp(-x*self.dt)
        self.output[p>self.rng.rand(self.size)] = self.value
        return self.output
```

Neither of these do any sanity-checking on their inputs.  Negative inputs will do weird things to RegularSpiking (it should probably not allow ```state``` to go below zero).

I've got a notebook where I use these to convert rates into spikes:

https://github.com/tcstewar/testing_notebooks/blob/master/spike%20trains%20-%20poisson%20and%20regular.ipynb

![image](https://cloud.githubusercontent.com/assets/2380772/20292040/6eb0f0d6-aab8-11e6-83fe-557f5f05ef39.png)

tcstewar:
Here's a much more accurate poisson spike generator

```python
class PoissonSpikingExact(object):
    def __init__(self, size, seed, dt=0.001):
        self.rng = np.random.RandomState(seed=seed)
        self.dt = dt
        self.value = 1.0 / dt
        self.size = size
        self.output = np.zeros(size)
    def next_spike_times(self, rate):        
        return -np.log(1.0-self.rng.rand(len(rate))) / rate
    def __call__(self, t, x):
        self.output[:] = 0
        
        next_spikes = self.next_spike_times(x)
[...]

PR #1225, opened 2016-12-05
WIP: Implement RegularSpiking and PoissonSpiking

tbekolay:
**Motivation and context:**
See #579. The idea here is to generate spikes either at a regular rate, or with Poisson statistics, from a rate-based neuron type. So, if you do something like `nengo.RegularSpiking(nengo.LIFRate)`, you get something that is (very?) close to `nengo.LIF`.

This is still marked as a WIP because the `PoissonSpiking` implementation has poor accuracy and I'm not sure if that's a mistake in the implementation, or because they're bad. @tcstewar could you try out this branch and let me know what you think?

There are also some other things I want to look at in more detail. See the "Still to do" section below.

**How has this been tested?**
Real tests are TODO, but for now I've been playing with the new neuron types in the GUI to ensure that they build and run and look pretty normal. Right now, this looks right:

```python
import nengo

model = nengo.Network()
with model:
    stim = nengo.Node([0])
    a = nengo.Ensemble(n_neurons=50, 
                       dimensions=1, 
                       neuron_type=nengo.RegularSpiking(nengo.LIFRate()))
    nengo.Connection(stim, a)
```

But this looks strange / unstable:

```python
import nengo

model = nengo.Network()
with model:
    stim = nengo.Node([0])
    a = nengo.Ensemble(n_neurons=50, 
                       dimensions=1, 
                       neuron_type=nengo.PoissonSpiking(nengo.LIFRate()))
    nengo.Connection(stim, a)
```

**Types of changes:**

**Checklist:**

**Still to do:**

- [ ] Make sure the `PoissonSpiking` implementation is correct
- [ ] Use the Param system to store `base_type`
- [ ] Implement tests
- [ ] Add to `nl` / `nl_nodirect`

**Discussion points:**

One issue with testing this is that the neuron type requires a base neuron type to function (a neuron type from which to figure out the correct spike rate). That's fine for normal models, but currently the way our test suite works is to pass the neuron type to functions through a pytests fixture as a type, not as an instance. For each neuron type, we make an instance by just doing `nl()`, but that won't work for these composite types since they require something like `nl(some_other_nl())`. So there are two options.

1. Don't add it to `nl` / `nl_nodirect`; test them separately and specifically like we do for `AdaptiveLIF` and `Izhikevich`

2. Modify the testing infrastructure to take in instances rather than types. It used to be not possible to do this, since neuron types were mutable and that could cause annoying issues. But now neuron types are immutable, so we should be able to pass in instances. This, however, might be difficult to do through the command line, which is currently possible (but maybe could be done some other way?)

### DIFF
--- a/nengo/__init__.py
+++ b/nengo/__init__.py
@@ -21,8 +21,18 @@ from .config import Config
 from .connection import Connection
 from .ensemble import Ensemble
 from .node import Node
-from .neurons import (AdaptiveLIF, AdaptiveLIFRate, Direct, Izhikevich, LIF,
-                      LIFRate, RectifiedLinear, Sigmoid)
+from .neurons import (
+    AdaptiveLIF,
+    AdaptiveLIFRate,
+    Direct,
+    Izhikevich,
+    LIF,
+    LIFRate,
+    PoissonSpiking,
+    RectifiedLinear,
+    RegularSpiking,
+    Sigmoid
+)
 from .network import Network
 from .learning_rules import PES, BCM, Oja, Voja
 from .params import Default
--- a/nengo/builder/neurons.py
+++ b/nengo/builder/neurons.py
@@ -1,8 +1,17 @@
+import inspect
+
 import numpy as np
 
 from nengo.builder import Builder, Operator, Signal
 from nengo.neurons import (
-    AdaptiveLIF, AdaptiveLIFRate, Izhikevich, LIF, NeuronType)
+    AdaptiveLIF,
+    AdaptiveLIFRate,
+    Izhikevich,
+    LIF,
+    NeuronType,
+    PoissonSpiking,
+    RegularSpiking,
+)
 
 
 class SimNeurons(Operator):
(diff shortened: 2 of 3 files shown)


tcstewar:
> This is still marked as a WIP because the PoissonSpiking implementation has poor accuracy and I'm not sure if that's a mistake in the implementation, or because they're bad. @tcstewar could you try out this branch and let me know what you think?

That behaviour looks right to me.  Poisson spiking is *way* worse than regular spiking in terms of signal-to-noise.  But if you crank up the number of neurons to 2000 in this implementation then we get decent behaviour out of it, so I'm pretty sure it's right.

studywolf:
what are the next steps for moving this PR forward? interested in having a spiking ReLu available, unless it would be better to just implement that directly?

tbekolay:
There's a todo list in the original post which I think is still valid. I can prioritize this next week if y'all want.

arvoelke:
Note that the code in #1487 is a bit more up-to-date in terms of things like handling `amplitude`. It also uses `log1p`, and possibly some other improvements.

tbekolay:
I'll look at #1487 when I get around to this to combine the benefits of both (hopefully :crossed_fingers:)

hunse:
Done in #1609. I used this PR as a foundation for that, and double-checked that everything here is now there (in one form or another).

PR #1226, opened 2016-12-06
Add HebbError learning rule

hunse:
**Motivation and context:**

This is a very common type of learning in machine learning. I'm not sure if ``HebbError`` is a good name for it, and perhaps it has it's own name, but I've heard it described as a Hebbian rule on an error (i.e. weight updates are proportional to pre activities times the error in post activity space).

The PES rule is a type of this rule, where the error is the post error in vector space times the post encoders.

**How has this been tested?**

The unit test tests backprop-style learning on encoders in a network with fixed decoders. i.e. the decoders are generated randomly, and then the transpose of the decoders is used to propagate the output error back to adjust the encoders.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Test this and add some unit tests
- [ ] Maybe add a notebook demonstrating it

### DIFF
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -5,7 +5,7 @@ from nengo.builder.operator import DotInc, ElementwiseInc, Reset
 from nengo.connection import LearningRule
 from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError
-from nengo.learning_rules import BCM, HebbError, Oja, PES, Voja
+from nengo.learning_rules import BCM, DeltaRule, Oja, PES, Voja
 from nengo.node import Node
 from nengo.synapses import Lowpass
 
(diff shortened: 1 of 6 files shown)


hunse:
I just put this here because I was playing around with it. It hasn't really been tested yet.

The learning rule equation is: 
image

I'm curious what people think: Is this a useful type of learning rule to have?

Also, I think all of our learning rules should have their update equation in their docstring.

drasmuss:
Maybe just `DeltaRule` as a name?  That's basically what this is, with the assumption of a linear activation function (so no derivative term).

hunse:
I think `DeltaRule` is a good name. I've made the change.

I think this is ready for review. It still needs changelog entries, which I can do.

I added a commit at the end that allows LinearFilters to be combined. Maybe this should be part of a different PR, I just thought of it while I was doing this one and it's a pretty small addition.

jgosmann:
I took a look at this, but I have a hard time understanding the intuition behind this learning rule.

* What exactly is meant by `e_i is the input error in the postsynaptic neuron space`? Is this saying how much higher the postsynaptic neuron activity should be?
* So if the activity of a postsynaptic neuron is too low, the connection weights from presynaptic neurons get increased to increase the postsynaptic activity?
* Can you add some explanation of what the network is supposed to do? Especially why `decoders` is generated in that particular way is not clear to me and why those are used in the places they are used. What is the purpose of `y`?
* In more general terms it would be nice to have a notebook demonstrating the learning rule and giving and intuition in what sort of situations it can be used.

Somewhat aside: Do we have a policy on the inclusion of learning rules in core Nengo? It seems that there could be a vast number of learning rules that we could possibly include ...

hunse:
Moved to nengo/nengo_extras#42

Issue #766, opened 2015-07-02
Cannot supply array of eval_points to EnsembleArray


arvoelke:
```
...
  File "c:\users\aaron\ctn\nengo\nengo\networks\ensemblearray.py", line 55, in __init__
    self.config[nengo.Ensemble].update(ens_kwargs)
  File "c:\users\aaron\ctn\nengo\nengo\config.py", line 105, in update
    setattr(self, key, d[key])
  File "c:\users\aaron\ctn\nengo\nengo\config.py", line 56, in __setattr__
    param.validate(self, value)
  File "c:\users\aaron\ctn\nengo\nengo\dists.py", line 336, in validate
    dist = super(DistOrArrayParam, self).validate(instance, dist)
  File "c:\users\aaron\ctn\nengo\nengo\params.py", line 212, in validate
    "distribution instead." % attr)
ValueError: dimensions not yet initialized; cannot determine if shape is correct. Consider using a distribution instead.
```

Possible solutions?
- Force people to only use distributions with EnsembleArrays. This seems like an unreasonable limitation to me.
- Include the number of dimensions before the ensembles are configured, i.e.
  
  ```
  self.config[nengo.Ensemble].dimensions = ens_dimensions
  self.config[nengo.Ensemble].update(ens_kwargs)
  ```
  
  But this is hacky because this solution has to be repeated for all similar situations with other networks. Also, it's important to note that you can't just do `ens_kwargs['dimensions'] = ens_dimensions` otherwise you get undefined behaviour, because it relies on the order of dictionary iteration to update `dimensions` _before_ `eval_points`, which may or may not happen depending on the alignment of the sun with Neptune.
- Create a new `StaticDistribution` that always iterates through a fixed list, and then cast `eval_points` arrays to this distribution when the param is set (thus removing `DistOrArrayParam`)? I think this is a better long-term solution, assuming I'm not overlooking anything.

Note: Add a unit test for this.

jgosmann:
Has this been resolved with the merge of #1233?

jgosmann:
ping @arvoelke

arvoelke:
Yep!

```
import nengo
import numpy as np
with nengo.networks.EnsembleArray(3, 2, eval_points=np.zeros((3, 1))) as model:
    for ens in model.ensembles:
        print ens.eval_points
```

```
Samples(samples=array([[ 0.],
       [ 0.],
       [ 0.]]))
Samples(samples=array([[ 0.],
       [ 0.],
[...]

PR #1233, opened 2016-12-12
Add Samples distribution

### LINKED ISSUE #691: Specifying encoders to an EnsembleArray as a non-distribution may cause config error to occur.
If the non-distribution encoders are provided to an EnsembleArray, there is a possibility that the following error will occur:

``` python
ValueError: n_neurons not yet initialized; cannot determine if shape is correct. Consider using a distribution instead.
```

This occurs when the config update line

``` python
self.config[nengo.Ensemble].update(ens_kwargs)
```

updates the `encoders` parameter in `kwargs` before it has done so for `n_neurons`.

tbekolay:
**Motivation and context:**
Look, people want to pass in NumPy arrays to `EnsembleArray`. They just do! See #691 and and #766 for proof, it's a thing. In #766 @arvoelke suggested a "no-op" distribution, which no one at the dev meeting disliked, so here it is.

Closes #691 and #766.

**How has this been tested?**
New `EnsembleArray` test included :+1:

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Discussion point: right now I make a copy of the array passed into `Samples`. Should we?
- [x] Add to docs
- [x] Add changelog entry
- [x] Are there other places `Samples` would simplify code?

### DIFF
--- a/nengo/dists.py
+++ b/nengo/dists.py
@@ -311,6 +311,55 @@ class Choice(Distribution):
         return self.options[i]
 
 
+class Samples(Distribution):
+    """A set of samples.
+
+    This class is a subclass of `.Distribution` so that it can be used in any
+    situation that calls for a  `.Distribution`. However, the call to `.sample`
+    must match the dimensions of the samples or a `.ValidationError`
+    will be raised.
+
+    Parameters
+    ----------
+    samples : (n, d) array_like
+        ``n`` and ``d`` must match what is eventually passed to `.sample`.
+    """
+
+    samples = NdarrayParam('samples', shape=('...',))
+
+    def __init__(self, samples):
+        super(Samples, self).__init__()
+        self.samples = samples
+
+    def __repr__(self):
+        return "Samples(samples=%r)" % (self.samples,)
+
+    def sample(self, n, d=None, rng=np.random):
+        samples = np.array(self.samples)
+        shape = (n,) if d is None else (n, d)
+
+        if d is None:
+            samples = samples.squeeze()
+
+        if d is not None and samples.ndim == 1:
+            samples = samples[..., np.newaxis]
+
+        if samples.shape[0] != shape[0]:
+            raise ValidationError("Wrong number of samples requested; got "
+                                  "%d, should be %d" % (n, samples.shape[0]),
+                                  attr='samples', obj=self)
+        elif d is None and len(samples.shape) != 1:
+            raise ValidationError("Wrong sample dimensionality requested; got "
+                                  "'None', should be %d" % (samples.shape[1],),
+                                  attr='samples', obj=self)
+        elif d is not None and samples.shape[1] != shape[1]:
+            raise ValidationError("Wrong sample dimensionality requested; got "
+                                  "%d, should be %d" % (d, samples.shape[1]),
+                                  attr='samples', obj=self)
+
+        return samples
+
+
 class SqrtBeta(Distribution):
     """Distribution of the square root of a Beta distributed random variable.
 
--- a/nengo/networks/ensemblearray.py
+++ b/nengo/networks/ensemblearray.py
@@ -89,6 +89,10 @@ class EnsembleArray(nengo.Network):
 
         super(EnsembleArray, self).__init__(label, seed, add_to_container)
 
+        for param in ens_kwargs:
+            if is_iterable(ens_kwargs[param]):
+                ens_kwargs[param] = nengo.dists.Samples(ens_kwargs[param])
+
         self.config[nengo.Ensemble].update(ens_kwargs)
 
         label_prefix = "" if label is None else label + "_"
--- a/nengo/networks/tests/test_ensemblearray.py
+++ b/nengo/networks/tests/test_ensemblearray.py
@@ -2,7 +2,7 @@ import numpy as np
 import pytest
 
 import nengo
-from nengo.dists import Choice
+from nengo.dists import Choice, UniformHypersphere
 from nengo.exceptions import ValidationError
 from nengo.utils.compat import range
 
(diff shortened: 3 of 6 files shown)


nengo/dists.py
```diff
@@ -311,6 +311,43 @@ def sample(self, n, d=None, rng=np.random):
...
+
+    def __repr__(self):
+        return "Samples(samples=%r)" % self.samples
+
+    def sample(self, n, d=None, rng=np.random):
+        samples = np.array(self.samples)
+        shape = (n,) if d is None else (n, d)
+
+        if d is None:
+            samples = samples.squeeze()
```
arvoelke: What if `d is not None` and does not match the second dimension? Is this handled somewhere downstream?

nengo/dists.py
```diff
@@ -311,6 +311,43 @@ def sample(self, n, d=None, rng=np.random):
...
+    """
+
+    samples = NdarrayParam('samples', shape=('...',))
+
+    def __init__(self, samples):
+        super(Samples, self).__init__()
+        self.samples = samples
+
+    def __repr__(self):
+        return "Samples(samples=%r)" % self.samples
```
arvoelke: `... % (self.samples,)`, otherwise if `self.samples` is a tuple, then you will get a string formatting `TypeError`.

nengo/dists.py
```diff
@@ -311,6 +311,43 @@ def sample(self, n, d=None, rng=np.random):
...
+    ----------
+    samples : (n, d) array_like
+        ``n`` and ``d`` must match what is eventually passed to `.sample`.
+    """
+
+    samples = NdarrayParam('samples', shape=('...',))
+
+    def __init__(self, samples):
+        super(Samples, self).__init__()
+        self.samples = samples
```
arvoelke: Cast to `ndarray` here instead of in `sample`? This will then work if given a generator, and will also resolve the comment in `__repr__` indirectly.

nengo/networks/ensemblearray.py
```diff
@@ -89,6 +89,10 @@ def __init__(self, n_neurons, n_ensembles, ens_dimensions=1,
 
         super(EnsembleArray, self).__init__(label, seed, add_to_container)
 
+        for param in ens_kwargs:
+            if isinstance(ens_kwargs[param], np.ndarray):
```
arvoelke: This could be generalized to handle any iterable?

nengo/networks/tests/test_ensemblearray.py
```diff
@@ -238,3 +238,27 @@ def test_neuronoutput(Simulator, seed):
...
+            n_eval_points=eval_points.shape[0])
+
+    with Simulator(net) as sim:
+        pass
+    built = sim.data[ea.ea_ensembles[0]]
+    assert np.allclose(built.encoders, encoders)
+    assert np.allclose(built.max_rates, max_rates)
+    assert np.allclose(built.intercepts, intercepts)
+    assert np.allclose(built.eval_points, eval_points)
+    assert built.eval_points.shape == eval_points.shape
```
arvoelke: Include test for mismatched shape?

arvoelke:
For your discussion point `right now I make a copy of the array passed into Samples`, I also think a copy is fine. Better safe than sorry (risk/reward is too high).

nengo/dists.py
tbekolay: This is done automatically since `samples` is an `NdarrayParam`.

nengo/dists.py
tbekolay: As above, `samples` is always a numpy array here so we should be good. I considered the defensive tuple though, so I can slip it in there for kicks.

nengo/dists.py
tbekolay: Oh whoops, totally forgot to check that ;) Originally the check below checked the whole shape, but then I changed it and forgot to add in the check for the second dimension. Will update tomorrow.

nengo/networks/ensemblearray.py
tbekolay: Right yeah, I'll use `nengo.utils.compat.is_array_like`.

nengo/networks/ensemblearray.py
```diff
@@ -89,6 +89,10 @@ def __init__(self, n_neurons, n_ensembles, ens_dimensions=1,
...
+                ens_kwargs[param] = nengo.dists.Samples(ens_kwargs[param])
```
tbekolay: Yeah, so like we talked about in the dev meeting, the issue there is that then we lose the model construction-time validation of shapes. This isn't a simple no-op, it means that if people are accessing `ensemble.encoders` they will never get a NumPy array, which is almost always what they want if they're accessing it to do some math or some such. I see the benefit of this distribution for fixing this specific problem, but I think it's premature to switch a lot of core functionality over to it because of issues like this. Not saying we won't, but I'd prefer it in a separate PR. I'm happy to switch over other specific instances in which nothing changes to the user, but yeah, mostly I wanted to get this out quickly and haven't looked around for other places to do this.

nengo/networks/tests/test_ensemblearray.py
tbekolay: Will do! :+1:

nengo/networks/ensemblearray.py
jgosmann: > if people are accessing ensemble.encoders

How common is that? I think I never set `encoders` to something different than a distribution.

> I see the benefit of this distribution for fixing this specific problem

How feel you about leaving this block out of the PR? It requires a little bit more boilerplate on the user side, but also makes explicit what's happening. Also with this change `ensemble_array.ensembles[i].encoders` will never return a NumPy array. So if we worry about that in general, why not in the case of ensemble arrays?

tbekolay:
OK, pushed a fixup addressing @arvoelke's comments. I opted for testing mismatched shapes on the distribution itself, so added tests to `tests_dists.py` for that.

Also, for future PRs, could you read over our [reviewing guidelines](http://pythonhosted.org/nengo/contributing.html#reviewing) @arvoelke? We're trying to move away from reviews of the form "please change this" and instead having the reviewer make the change themselves. We're also not using Github's reviewing interface and opting for labels instead, though we might move to using that in the new year.

Re @jgosmann's comments, setting encoders to numbers is why Xuan raised #691 so people do it. As for not including the block in `EnsembleArray`, if we didn't include it then we couldn't close #691 or #766, so I'm not a fan of that.

One possible way around it would be for the `DistributionParam` to return a NumPy array when accessing the value of the distribution is a `Samples` instance, but that's definitely pretty magical so I'd want to sit on that for a while. Setting parameters is far more common than reading them, so while in the ideal case what you set is always what you get back, that's already not quite the case in all situations. It's also not at very common to access the individual ensembles in an `EnsembleArray`, so given all these factors, this half-measure seems okay for now and not worth blocking this PR which fixes two open issues. Happy to hear other perspectives though.

jgosmann:
> setting encoders to numbers is why Xuan raised #691 so people do it

This was not about *setting*, but *accessing* encoders. So how common is it for people to set the encoders to numbers and access them afterwards to do math on them? (And even if, one could easily keep a variable with those encoders around or query them from `Samples`.)

Why was it that we're updating the config and do not pass `**ens_kwargs` to the `Ensemble` constructor? This seems to me to have exactly the same effect except that NumPy arrays instead of distributions would work out of the box.

tbekolay:
> This was not about setting, but accessing encoders.

Sorry, I was replying to the comment "I think I never set encoders to something different than a distribution."

> Why was it that we're updating the config and do not pass `**ens_kwargs` to the Ensemble constructor?

To save a few lines of code, and so that if you add more ensembles to the ensemble array after it's created those parameters still apply. I brought this up as a possible resolution to these issues when we discussed it in the dev meeting but no one had an opinion on it, but yeah, we could do that instead.

Anyone else have thoughts?

jgosmann:
> To save a few lines of code,

Doesn't this add a line of code?

> so that if you add more ensembles to the ensemble array after it's created those parameters still apply.

To me this seems to be a pretty weird use case ... if no one is actually doing this or has any other arguments, I would favour passing `**ens_kwargs` to the `Ensemble`. Even though, the `Samples` distribution can still be useful in other situations.

jgosmann:
Added a small commit, apart from the discussion about the usage in `EnsembleArray` this LGTM.

tbekolay:
Pushed a commit for one comment I hadn't addressed yet. @arvoelke and/or @jgosmann, could you take another look at this?

jgosmann:
The commit looks fine to me.

tbekolay:
Sorry should have been more explicit... could you take a look at the whole PR, and give it a review approval if it looks good?

PR #1240, opened 2016-12-14
Test best practices in examples.

### LINKED ISSUE #1230: Test usage of best practices in examples
If not too complicated we should test that all examples creating a simulator invoke `%load_ext nengo.ipynb`. There might be other things worth check.

jgosmann:
**Motivation and context:**
To ensure that examples follow certain standards. So far only checking that `%load_ext nengo.ipynb` is used if a `Simulator` is created in the example.

Closes #1230.

**How has this been tested?**
Ran the new test, found an example violating the best practice and fixed the example.

**Types of changes:**

**Checklist:**

### DIFF
--- a/examples/learning/learn_associations.ipynb
+++ b/examples/learning/learn_associations.ipynb
@@ -70,7 +70,8 @@
       "import matplotlib.pyplot as plt\n",
       "%matplotlib inline\n",
       "\n",
-      "import nengo"
+      "import nengo\n",
+      "%load_ext nengo.ipynb"
      ],
      "language": "python",
      "metadata": {},
--- a/nengo/tests/test_examples.py
+++ b/nengo/tests/test_examples.py
@@ -82,25 +82,50 @@ def test_slow_noexceptions(nb_file, tmpdir):
     assert_noexceptions(nb_file, tmpdir)
 
 
+def iter_cells(nb_file, cell_type="code"):
+    from nengo.utils.ipython import load_notebook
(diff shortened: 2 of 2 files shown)


Seanny123:
LGTM.

tbekolay:
Made a squash commit to reduce some code duplication; will merge if it looks good to you @jgosmann!

jgosmann:
Not sure how I feel about not using regular expressions. If an accidental space finds it way into `nengo. Simulator(...)` or someone comments the extension (`#%load_ext nengo.ipynb`), these things would not be captured by the test.

arvoelke:
But a regex is also more error-prone, for instance the first expression you wrote matches the following:
```
re.search(r'\bnengo\s*.\s*Simulator\s*\(', '#nengo@Simulator()', re.MULTILINE)
```

Since the `\b` is probably not what you want, and the `.` matches anything. By the way, this is a good opportunity to plug a tool that a friend from UW created: https://www.debuggex.com/. It's pretty sweet (note it does a `match` starting from the triangle slider -- not a global `search`).

jgosmann:
True and after all following the best practices is not super critical. Let's go ahead with the substring matching then.

arvoelke:
Just throwing out another option: if we really want a rigorous test, we can run the notebook with a patched `Simulator` that registers any occurrence of its creation in a static flag. I do something similar [here](https://github.com/arvoelke/nengolib/blob/master/nengolib/tests/test_notebooks.py#L36) where I patch `pylab.show()` to issue a warning in my examples. Then just check the flag after the example is run. This could be combined with the existing examples tests.

Issue #1231, opened 2016-12-12
Warn about intercepts out of radius


jgosmann:
Especially the `Exponential` distribution (if one forgets to specify `high`) makes it easy to accidentally generate intercepts outside of the ensemble radius. This leads to flipped response curves.
![index](https://cloud.githubusercontent.com/assets/850157/21113559/13357bfc-c078-11e6-844c-a5af78d082c5.png)

I think Nengo should at least warn or even raise an exception when this is happening.

arvoelke:
This was also pointed out by #1212.

jgosmann:
Closing this as a duplicate.

Issue #1212, opened 2016-11-24
Degenerate tuning curves when intercept >= 1


arvoelke:
I'm not sure about the right way to handle this (my preference would be to at least emit a warning) or if it's been discussed before, but weird things happen to the tuning curves when the intercept is exactly 1.

```
import nengo
import pylab

for neuron_type in (nengo.neurons.LIFRate(),
                    nengo.neurons.Sigmoid(),
                    nengo.neurons.RectifiedLinear()):
    with nengo.Network() as model:
        x = nengo.Ensemble(3, 1, neuron_type=neuron_type,
                           intercepts=[0.9, 1, 1.1])

    with nengo.Simulator(model) as sim: pass

    u = np.linspace(-1, 1, 100000)
    acts = nengo.builder.ensemble.get_activities(sim.data, x, u[:, None])

    pylab.figure()
    pylab.title(str(x.neuron_type.__class__.__name__))
    for i, intercept in enumerate(sim.data[x].intercepts):
        pylab.plot(u * sim.data[x].encoders[i], acts[:, i], label=str(intercept))
    pylab.legend()    
    pylab.show()
```

For LIFRate, the curve becomes a heavy-side function, and for sigmoid / ReLU it disappears:

![download 1](https://cloud.githubusercontent.com/assets/5420057/20606491/469d6ca8-b23f-11e6-8d6d-197c73160486.png)
![download 2](https://cloud.githubusercontent.com/assets/5420057/20606493/47af1466-b23f-11e6-93f4-964e6d585569.png)
![download 3](https://cloud.githubusercontent.com/assets/5420057/20606494/48d773ec-b23f-11e6-9202-237fc0aaf9ab.png)

arvoelke:
Also, the maximum firing rate has an unexpected interpretation once the intercept is > 1 (it is no longer the maximum firing rate within [-1, 1], but rather the rate at 1). This is most visible for the ReLU case.

Issue #701, opened 2015-04-16
Can't specify only gain or bias


studywolf:
was trying to set the bias and nengo throws: 

```NotImplementedError: gain or bias set for <Ensemble (unlabeled) at 0x40a5610>, but not both. Solving for one given the other is not implemented yet.```

tbekolay:
Was wondering when someone would try to do that ;) I think that the math to figure out one given the other and the `intercepts` and `max_rates` is pretty straightforward algebra, but I haven't gone through it myself.

PR #1243, opened 2016-12-14
Raise exception on invalid intercepts.

jgosmann:
**Motivation and context:**
As noted in #1212 (and #1231) intercepts >= 1. give unexpected results. It seems unlikely that anyone is using this weird behaviour and I think it should give an error (but at least a warning). This PR adds an exception at build time.

**How has this been tested?**
Added a unit test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/ensemble.py
+++ b/nengo/builder/ensemble.py
@@ -8,6 +8,7 @@ from nengo.builder import Builder, Signal
 from nengo.builder.operator import Copy, DotInc, Reset
 from nengo.dists import Distribution, get_samples
 from nengo.ensemble import Ensemble
+from nengo.exceptions import BuildError
 from nengo.neurons import Direct
 from nengo.utils.builder import default_n_eval_points
 
@@ -94,6 +95,8 @@ def get_gain_bias(ens, rng=np.random):
     else:
         max_rates = get_samples(ens.max_rates, ens.n_neurons, rng=rng)
         intercepts = get_samples(ens.intercepts, ens.n_neurons, rng=rng)
+        if np.any(intercepts >= 1.):
+            raise BuildError("Intercepts need to be < 1.")
         gain, bias = ens.neuron_type.gain_bias(max_rates, intercepts)
 
     return gain, bias, max_rates, intercepts
--- a/nengo/builder/ensemble.py
+++ b/nengo/builder/ensemble.py
@@ -96,7 +96,10 @@ def get_gain_bias(ens, rng=np.random):
         max_rates = get_samples(ens.max_rates, ens.n_neurons, rng=rng)
         intercepts = get_samples(ens.intercepts, ens.n_neurons, rng=rng)
         if np.any(intercepts >= 1.):
-            raise BuildError("Intercepts need to be < 1.")
+            raise BuildError(
+                "Some intercepts in %r are >= 1, which results in flipped "
+                "response curves (more current elicits less activity). "
+                "Modify intercepts to be < 1." % (ens,))
         gain, bias = ens.neuron_type.gain_bias(max_rates, intercepts)
 
     return gain, bias, max_rates, intercepts
--- a/nengo/tests/test_ensemble.py
+++ b/nengo/tests/test_ensemble.py
@@ -4,6 +4,7 @@ import pytest
 import nengo
 import nengo.utils.numpy as npext
 from nengo.dists import Choice, Gaussian, UniformHypersphere
+from nengo.exceptions import BuildError
 from nengo.processes import WhiteNoise, FilteredNoise
 from nengo.utils.testing import warns, allclose
 
@@ -365,3 +366,13 @@ def test_noise_copies_ok(Simulator, nl_nodirect, seed, plt):
 
     assert np.allclose(sim.data[ap], sim.data[bp])
     assert np.allclose(sim.data[bp], sim.data[cp])
+
+
+@pytest.mark.parametrize('intercept', [1.0, 1.1])
+def test_raises_exception_for_invalid_intercepts(Simulator, intercept):
+    with nengo.Network() as model:
+        nengo.Ensemble(1, 1, intercepts=[intercept])
+
+    with pytest.raises(BuildError):
+        with nengo.Simulator(model):
+            pass


Seanny123:
LGTM!

tbekolay:
To check my understanding, am I right in thinking that this would also be a problem if intercepts are `<= -1`? If so I can make a fixup to do check that as well. I'm good with raising an exception; if anyone is relying on this behavior I would be extremely curious to see what their use case is.

jgosmann:
Actually, intercepts `<= -1` seem to work fine.

jgosmann:
I guess, we're taking 1 as the point of the max firing rate and any value below that is fine.

tbekolay:
Is this still true for scalar signals with negative encoders?

jgosmann:
Yes.

tbekolay:
> Yes.

Care to expand a bit...?

xchoo:
Technically, an error should be thrown as long as the intercepts are outside of `(-radius, radius)` (note the brackets, also, without the radius scaling, it should be `(-1, 1)`). If we want to allow `intercepts <= -radius`, then we should fix the code to allow for intercepts `>= radius` as well.

jgosmann:
> then we should fix the code to allow for intercepts >= radius as well.

I don't think we can fix the code that way because mathematically this is ill-posed.

1. Firing rates should increase as the dot product with the encoder increases.
2. We define the firing rate distribution for when the dot product is exactly 1 (“max_rates”).
3. The intercept defines the point were the firing rate crosses zero.
4. If we were to have an intercept larger than 1, we cannot fulfil conditions 1 and 2 at the same time.

Having intercepts <= -1, however, does make sense mathematically. Both condition 1 and 2 can be fulfilled without problems. Condition 3 will be a point outside of the radius, so the neuron will fire for all values in the radius, but with varying firing rate. I actually could imagine this to be useful in some circumstances.

xchoo:
> Firing rates should increase as the dot product with the encoder increases.
Not necessarily. This is the convention we have chosen, but in the case where the intercept > 1, this would not be the case. Mathematically, the tuning curve is defined with two points (any two points), and the LIF (or neuron of your flavour) equation. The encoders are not part of the equations.

However, this is not the point I am trying to make. Here, I'm trying to say that having intercepts >= 1 or <= -1 would not be in the "spirit" of the NEF, where everything (the core equations) is taken to be well-defined within the working range of (-1,1).

tbekolay:
> 1. Firing rates should increase as the dot product with the encoder increases.
> 2. We define the firing rate distribution for when the dot product is exactly 1 (“max_rates”).
> Having intercepts <= -1, however, does make sense mathematically. Both condition 1 and 2 can be fulfilled.

If the encoder is -1 and the intercept is -1.1, isn't that essentially the same case as having an encoder of 1 and an intercept of 1.1, which is what this thread is about?

jgosmann:
> Here, I'm trying to say that having intercepts >= 1 or <= -1 would not be in the "spirit" of the NEF, where everything (the core equations) is taken to be well-defined within the working range of (-1,1).

I think that everything is still well-defined with an intercept <= -1. In most cases one probably does not want to use intercepts <= -1 and I could see a warning because of that, but it seems less likely to accidentally end up with intercepts <= -1 than with intercepts >= 1.

> If the encoder is -1 and the intercept is -1.1, isn't that essentially the same case as having an encoder of 1 and an intercept of 1.1, which is what this thread is about?

No, the encoder tells you in which direction the tuning curve increases in the space of inputs. With an encoder of -1 and intercept of -1.1, you get a neuron firing with “max_rate” at -1 and decreasing in firing rate for inputs >= -1 until it intercepts at 1.1 (in the input space). With an encoder of 1 the tuning curve gets flipped (it now increases for larger values in the input space) and at the same time the point to determine “max_rate” is set to 1 (instead of -1). An intercept of 1.1 would be invalid in this case. But with an intercept of -1.1 we a neuron equivalent to the first case, if we flipped the sign on the input.

xchoo:
> I think that everything is still well-defined with an intercept <= -1. In most cases one probably does not want to use intercepts <= -1 and I could see a warning because of that, but it seems less likely to accidentally end up with intercepts <= -1 than with intercepts >= 1.

Technically, everything is well-defined within (-inf, inf). All of the equations still work. The anchor points chosen just determine which way the curves faces, and what the curves look like. What I'm trying to say (and repeating myself here), is the that the NEF "convention" is to use things defined within (-1, 1), not (-inf, 1).

jgosmann:
> The anchor points chosen just determine which way the curves faces, and what the curves look like.

Yes, and I think the anchor points should not define the direction of the curve. This should be done solely the by the encoder because that is what someone will change to affect the direction of the tuning curve. Under this assumption (that the direction of the tuning curve is fixed by the encoder), it is no longer well-defined for intercepts >= 1.

xchoo:
Once again, gotta reiterate, what I'm talking about here is just neuron space, excluding any assumptions of the encoders. If you want accept the case of having intercepts <= -1, then we should also accept the case of having the intercepts >= 1 (because mathematically, both cases are equivalent, all the math checks out, etc). 

If you want to limit the intercepts to < 1 because of the un-intuitive outcome of combining the encoders with a tuning curve where the intercepts are >= 1, then we are bringing in the NEF assumptions into the discussion (the encoders being purely an NEF construct), and that being the case, we should also limit the intercepts to > -1 to keep within the NEF conventions.

jgosmann:
I disagree, but this doesn't seem to lead anywhere right now. I'll think more about it and see if I change my mind or come up with new arguments.

jgosmann:
Just a remark: I think if we are talking about intercepts (as used in Nengo), we have to take NEF assumptions into account because the intercept is defined along the encoder (which is an NEF concept). If we wanted to ignore NEF concepts, we would have to talk about gains and biases.

xchoo:
Note: since all intercepts are defined along the encoders, you can ignore
the encoders in the calculations.


-------------------------
Xuan Choo
HP: (226) 339 3892

jgosmann:
I disagree. The encoder is doing a unit conversion in a sense (with a conversion factor of 1 Ampere). It converts from the “unit” of our input space to current. (You could also ascribe that function to the gain factor, but then the input to your neuron would still be in the “unit” of the input space and as such not independent of NEF concepts.)

xchoo:
Hmm. My point here is that we should either be mathematically consistent (i.e. take two points, and the neuron equation, and figure out what the shape of the tuning curve is without any preconception of what it  _should_ look like), or stick with the NEF convention (i.e. keep everything within (-1, 1)).

Personally, I'd rather we keep it mathematically consistent (who knows, maybe we'll find a neuron one day that decreases its firing rate as the input current increases), and just throw up a warning say "hey! You've defined stuff outside what the code is expecting, there may be unintended side effects"

xchoo:
As a side note: If we limit the intercepts to (-1,1), and someone wants intercepts < -1, it is still possible to do so by adjusting the intercepts, radius, and evaluation points appropriately. 

Imho, having one limit and not the other is (pardon the euphemism) half-arsed. 😜

jgosmann:
In abstract terms the gain and bias of a neuron are defined by two points *(x1, y1)* and *(x2, y2)* (given the remaining neuron parameters), where *x1* and *x2* are points along the encoder (thus related to the input space) and *y1* and *y2* are firing rates. I think what we should ensure is that for a given neuron type either *x1 < x2 <=> y1 < y2* or *x1 < x2 <=> y1 > y2*. This would account for neuron types where the firing increases and for neurons where it decreases with more input current. The important thing is it ensures that we get tuning curves all pointing in the same direction. (This is desirable because to flip the direction of the tuning curves the sign of the encoder should be used; flipping it via those anchor points is an indication that the user did not intend this to do.) So this all seems, mathematically consistent to me with the addition of preventing some behaviour that the user probably did not want.

Now, if we apply this to Nengo some values of these anchor points are fixed. The intercept is *(x1, 0)* and the other anchor point is *(1, max_rate)*. From above math it would follow that for a specific neuron type either *x1 < 1* or *x1 > 1* (the latter case would be for neurons that decrease in firing with increasing input current; we don't have those right now and there are probably a bunch of design decision in Nengo that would make these awkward to use anyways; thus I would ignore it for now).

One way to allow intercepts >= 1 would be to allow to set the point *x2* at which max_rate is defined. But this would still imply that there should be an exception for *x1 < x2*. (Or for *x2 > x1* for weird firing rate decreasing neurons.)

I think this reasoning is mathematically consistent. It is not about limiting the intercepts to a specific range, but ensuring that tuning curves are not accidentally flipped. The boundary of 1 just follows from where we fixed the max_rates.

arvoelke:
I agree with everything Jan said. And although I am just repeating him here, a way of framing this that might make Xuan happier with the (-inf, 1) asymmetry is that what we are really trying to do is enforce that, for the values `dot(x, e)` (i.e., the response curve) we have:
1. the minimum at -1
2. the maximum at 1
3. monotonic increase over (-1, 1)

And we "don't care" about the values outside of this range (note only the 2nd constraint is really 'enforced' by Nengo, but the 3rd is common to all neuron models that I'm aware of, and the 1st constraint follows from the 3rd).

Intercepts within (-1, 1) enforce that the minimum firing rate is zero. Having an intercept < -1 just means the minimum firing rate will be non-zero (it will lift up). This transition happens smoothly as the intercept is decreased, and in a way that is consistent with the neuron model.

But having an intercept at >= 1 brings in a bunch of inconsistencies. In a way you're trying to enforce that "the minimum required current must exceed the max current", where "max current" is the current which produces the desired `max_rate`. There will be a discontinuity (or sudden flipping) in the response curve as you transition the intercept across 1 in order to stay consistent with the model. But now the maximum firing rate is actually the minimum firing rate within `(-1, 1)` and the maximum firing rate is unbounded (e.g., the ReLU graph in #1212)! There isn't a way to fix this without damaging some other definition. This is really an artifact of the constraints mentioned at the top. Also note we can't just flip these constraints for intercepts > 1, because you would still get a mirror in the response curves (and still undefined discontinuity at 1).

Intercepts < -1 might be unheard of traditionally, but is certainly still a useful and well-defined case. As a concrete example, I can say with about 99% confidence that you can decode a constant function with a dozen spiking LIF neurons using intercepts < -1 with _way_ better accuracy than the same neurons but intercepts within (-1, 1). This should hold even after controlling for the average firing rate.

arvoelke:
This is kind of a different issue, but this made me realize that ReLU is kind of ill-conditioned for intercepts approaching 1 from the left-hand side (since the gain goes to infinity). Noise can easily push the representation to be above 1, in which case the firing rate will also explode. Formally, with probability `p = (1 - c)/2` we get an intercept `>= c = (1 - 2p)`, which has gain `max_rate / (1 - c) = max_rate / (2p)` which explodes with small probability. Note that other neuron models don't have this problem due to saturation (ignoring possible numerical issues). Kind of an odd case. This would be solved in theory if the decoder regularization were sensitive to gain (but it's not). It could also be solved by sampling outside the radius, adjusting the radius, capping the max firing rate, etc.

xchoo:
I don't think you understand the point I am trying to make. 

>  we are really trying to do is enforce that, for the current dot(x, e) the response curve has:
> 1. the minimum at -1
> 2. the maximum at 1
> 3. monotonic increase over (-1, 1)

My point being that given two known points, and an equation, you:
1. Don't need the minimum to be at -1
2. Don't need the maximum to be at 1
3. Don't need to have a monotonic increase over (-1, 1)

All of the assumptions that is being enforced seem to stem from the use of the 'max rate' term, as that implies that that point should be maximal in some way. What I'd say is that if we want to be mathematically rigorous, we shouldn't even be using the term 'max_rate', I'd rather the term 'rate_at_1'. Also, using the 'max_rate' term allows for the misconception that that is the true maximal firing rate of the neuron (when in fact, it is rather 1/tau_ref).

But, seeing as that doesn't seem to be the favourite option here, I'd rather limit the intercepts to between (-1, 1), not to be mathematically rigorous, but rather to stick with the NEF convention of having everything defined within the unit circle. If the user wants to define something outside of this norm, they had full well understand what they are doing to do so (and that involves known what the intercepts, "max_rates", evaluation points, etc, mean and how they affect the computation of the decoders). If the user knows how to manipulate these values correctly, you can get the same effect by appropriately biasing and shifting intercepts and evaluation points.

> Xuan, the way you talk, it that sounds like you are advocating removing the radius as well

Sure. Let's do that. The radius parameter has already caused enough problems in N-D space where each dimension is scaled differently. The naive solution is to scale the radius to the maximal value across all dimensions; the actual solution is to scale each dimension separately. So yeah, let's get rid of the radius parameter, and teach the users of Nengo the proper way to do it.

xchoo:
Side note: If you want to use the NEF convention of sticking the max rate at 1, but leave the intercepts between (-inf, 1) because "it'll be convenient to do fancy tricks with neurons", I'd tell you that convenience is the way bugs get into your code and you spend hours hunting them down because it didn't behave the way you wanted it to.

arvoelke:
> I don't think you understand the point I am trying to make.

To be clear in the motivation for my last post, I was trying to address some of the misconceptions that were still floating around, such as:

> If you want accept the case of having intercepts <= -1, then we should also accept the case of having the intercepts >= 1 (because mathematically, both cases are equivalent, all the math checks out, etc).

----------------------

> But, seeing as that doesn't seem to be the favourite option here, I'd rather limit the intercepts to between (-1, 1), not to be mathematically rigorous, but rather to stick with the NEF convention of having everything defined within the unit circle. 

My point is that having intercepts < -1 doesn't have anything to do with the NEF convention of the unit circle. All it does is raise the minimum firing rate. As you noted, the maximum firing for LIF is achieved at `+inf`, which is also outside the unit circle. In an analogous way, the minimum firing rate would now be achievable outside the unit circle, but this is not what matters (the bias current still keeps you centered at 0). 

I would change my mind if, for instance, something characteristically switched on/off about the representation once intercepts went <= -1, but this is not the case as it is for intercepts >= 1.

> I'd tell you that convenience is the way bugs get into your code and you spend hours hunting them down because it didn't behave the way you wanted it to.

Speaking specifics, it seems the feature of allowing intercepts < -1 (and > 1) has been a part of Nengo since 2.0 at least and nobody even noticed (?) until Jan and I were messing around with them. Furthermore, intercepts >= 1 is really the degenerate case. So this argument feels like fear-mongering to me when it comes to this specific case of intercepts < -1... On the contrary, since intercepts < -1 fall naturally out of the models and definitions that we use, I feel pretty strongly that we shouldn't be going to extra lengths to limit what the user can easily do just because the case seems alien to us. Instead, pedagogically we should arm our users with precise formulations and tools for understanding the consequences of their actions.

jgosmann:
> My point being that given two known points, and an equation, you:
> 
>     Don't need the minimum to be at -1
>     Don't need the maximum to be at 1
>     Don't need to have a monotonic increase over (-1, 1)
> 
> All of the assumptions that is being enforced seem to stem from the use of the 'max rate' term

[The assumption I'm basing my argument on is much more general and does not rely on the 'max rate' term:](https://github.com/nengo/nengo/pull/1243#issuecomment-267479795) The orientation of the tuning curves should not randomly switch. This has no assumption about monoticity or in which direction we haven an increase or decrease. I think @arvoelke assumptions effectively follow for all Nengo neuron types, but that isn't that relevant for the argument. I think @xchoo has not addressed that more general assumption (or I missed in-between other arguments). So, do you agree with that assumption, @xchoo?

If not, we should first discuss why or why not this is useful to enforce. If you agree with it, it should follow that intercepts should be < 1 (assuming normal neurons with monotonic increase and given our max_rates point at 1). Then the question to discuss would be whether it also follows that intercepts should be > -1 or whether there is another reason separate from this argument to enforce intercepts > -1.

> I'd tell you that convenience is the way bugs get into your code and you spend hours hunting them down because it didn't behave the way you wanted it to.

I could argue that allowing intercepts >= 1 (flipping the orientation of the response curve) is much more likely to lead to a bug than intercepts <= -1 (*not* flipping the orientation of the response curve). But I think that would be a less clear argument than by math, so I'd prefer not to.

tbekolay:
This thread has devolved from a decision about whether we should limit intercepts from `(-inf, 1)` or from `(-1, 1)` to an argument with more rhetoric about why people are wrong than comments trying to teach each other and help see each other's point of view.

Remember that we are all working together trying to make Nengo the best software package it can be. The best way you can do that is by explaining your mental models and thought processes clearly, kindly, and with respect for your audience. Everyone in this thread has a certain level of expertise across a wide variety of subjects. Anyone who is not respecting this diversity of knowledge will be asked to stop participating in these threads. Right now, the language being used here is not appropriate and must stop.

The positions of the people who have participated in this thread thus far are clear. If anyone else wants to voice their opinion, then please do. If you think that other people in this thread have gaps in their knowledge and must be educated, then please find a constructive way to do that (make a notebook, make an example, write some documentation, etc). If you absolutely must continue arguing, please do it elsewhere.

xchoo:
> The orientation of the tuning curves should not randomly switch.

Why not? It makes complete sense. There are two fixed points in the tuning curve calculation. The intercept, and the firing rate when `x == 1`. If you move the intercept beyond `x == 1`, then the mathematically, the tuning curve will switch. Anyone who understands the principle of how the tuning curve is derived will understand this.

> Speaking specifics, it seems the feature of allowing intercepts < -1 (and > 1) has been a part of Nengo since 2.0 at least and nobody even noticed (?) until Jan and I were messing around with them.

Not true. The problem with defining the intercepts beyond (-1, 1) has been in Nengo all the way from the good ol' Nengo 1.4. It's just that when we used it, we learned that defining the intercepts outside the ranges lead to unexpected (but if you sat down and thought about it, completely expected) behaviour. Hence, we informally informed folks that used Nengo to always stick with the NEF convention of doing everything within (-1, 1), and the option to specify intercepts outside that range was left to the "power users".

Like I mentioned before, my preferred option is to leave in the ability to specify intercepts beyond (-1, 1) and just throw up a warning telling users "hey! you're doing something unexpected. Expect unexpected results! 😄"

jgosmann:
> The orientation of the tuning curves should not randomly switch.
> 
> Why not?

I would like to answer that because it is indeed a point I did not provide a good argument for (mostly because I was assuming it to be somewhat self-evident, but that's clearly a mistake on my side and is not meant to demean anyone). But am I allowed, @tbekolay?

Apart from that I would really be interested in a concrete example were intercepts <= -1 caused a problem.

drasmuss:
Feel free to discuss it with Xuan offline, as Trevor said I don't think the debate in this thread has been productive (or would be made more productive by further debate)

drasmuss:
Also, comments such as "But am I allowed, @tbekolay?" are an excellent example of the kind of negative comments we are trying to avoid here.  Trevor is trying to keep nengo development a positive experience, please be respectful of that rather than sarcastic.

arvoelke:
I thought we were having a productive exchange, and saw positive intentions coming from everyone. I myself was realizing new things, and did not know that I (or anyone else for that matter) was insulting or disrespecting anyone in the process. I apologize for this lack of foresight. That said, frankly I feel insulted and censored by the above on behalf of us all. Xuan, Jan, and I are all adults that are capable of expressing our own feelings for ourselves. But since this level of emotional moderation is apparently necessary in order to maintain a positive experience for everyone else, I have decided to leave the nengo organization and developer team for the time being to make it clear where I draw the line. Regards, Aaron.

xchoo:
Just an update - I spoke to Jan offline about his intent behind this PR to get a better understanding of why we have an ideological difference in opinions about this matter. I think I can sum up his argument with:
"This PR is meant to impose an upper limit on the intercept values so as to prevent conceptual contradictions between the meaning of the intercept and the term 'max firing rate'"

On the basis of this, I concede that this PR makes sense, however, I would like to request that the PR description be updated to remove ambiguity in the description (e.g. "invalid (values)" is pretty ambiguous without the proper context), with the caveat that we discuss some of the issues (listed below) at a later time.

- The use of the term "max rate". It is my opinion that this is a misnomer prone to creating misunderstandings on how the code (should) work.
- The ability to define where (at what `x` value) the "max rate" value should be. This came up in offline discussions with Jan, and it would be nice to be able to specify y-intercept values outside of the `x == 1` condition.
- When and where to stick with the NEF convention of keeping everything within the unit circle.
- Whether or not a warning or an exception is appropriate for conditions like this where unexpected inputs produce unexpected results. 
- Related to the last point, what is the best way to indicate / educate the user as to how the code base works and what their intuitions & expectations of the code should be.

tbekolay:
I pushed a fixup with a longer error message; @jgosmann feel free to reword and I'll merge when you're happy with it.

> discuss some of the issues brought up above (and listed below) at a later time

Sounds great, #701 also deals with inflexibility in the gain / bias generation, so it would be worth thinking about how best to do that (anyone feel free to open a discussion issue with ideas).

jgosmann:
Please hold off for now. Checking for intercepts >= 1 might not be the right thing to do after all, but I have to figure out some math first.

jgosmann:
Ok, I think I finally figured things out and will post my reasoning here because
* it is somewhat different from what I said before
* leads to a different conclusion
* seems conceptually quite important to me

I will try to present the argument self-contained, so that it is not necessary to read the previous discussion.

# Theory

## Is max_rates a misnomer?

It can seem that way. If we look at the tuning curve of a neuron it will have firing rates > max_rate if we look at inputs > 1. (I will assume the radius to be 1 for the purpose of this discussion.)



So if we interpret max_rate as the *global* maximum firing rate, this is clearly a misnomer. I, however, propose that max_rate should be interpreted as the maximum firing rate within the radius. This gives us a box in which valid firing rates fall.



Note that this is different from saying max_rate is the firing rate that should be obtained at x=1.

## Determining tuning curves

For a standard LIF neuron and an intercept -1 < x < 1, this means we can orient the tuning curve either way.



To get a unique solution we need another constraint: The orientation of the tuning curve. The choice is somewhat arbitrary, but it is convenient to have the firing rate increase with the input current. Another way to express this is to say that the gain on the neuron has to be positive.

To summarize we get the following constraints on a tuning curve:

1. It does not exceed max_rate within the radius.
2. It is max_rate at some point within the radius (I did not explicitly mention this before, but otherwise we could scale the gain with any factor <= 1 and thus not having a unique solution.)
3. The gain is positive (equivalent to a fixed orientation of the tuning curve).

## Why are we using the point (1, max_rate) to calculate tuning curves?

Given the three constraints, the max_rate will be reached at 1 for monotonic tuning curves (this applies to all our current neuron types). It is important to highlight that we are not specifying the firing rate at x=1. Specifying the maximum firing rate over [-1, 1] is different and just works out to (1, max_rates) given the constraints. If we were to decide to constrain the gain to negative values, it would end up being (-1, max_rates). Or if we had a neuron with a non-monotonic tuning curve we would end up with a different x-value for max_rate.



(I believe Nengo currently does not correctly support such tuning curves.)

## Why are tuning curves with a flipped orientation an error?

Given the constraints there are two reasons:

* The gain becomes negative. (This is mostly equivalent to my previous argument that tuning curves should not flip, but the next reason is new.)
* The maximum firing rate within the radius will exceed max_rate!



## Are intercepts >= 1 a problem?

Actually they are not necessarily a problem. It depends on the neuron type and I will give an example where intercepts > 1 are fine. But for LIF neurons it turns out that an intercept >= 1 prevents us from fulfilling above conditions: Using a negative gain would violate the positive gain and the maximum firing rate within the radius. Using a random positive gain would be closer to the correct behaviour as the tuning curve does not flip and the max_rate is not exceeded, but we're not getting a value of max_rate within the radius, still violating one condition (and leaving the exact gain value unconstrained).

So while it is ok for LIF neurons and similar neurons types to check that intercepts are < 1, this is not the right thing to check in the general case. **What we should check is that the gain is positive.** This will probably become a bit clearer with the following example.

# Example: Sigmoid neurons

Sigmoid neurons are in interesting example because they don't have a real intercept. Instead the intercept specifies the inflection point.



Given a max_rate below half of 1/tau_ref (the maximum rate the neuron can produce, half of it is the firing rate at the inflection point), it is completely fine to have an intercept > 1. None of the constraints from above needs to be violated (in particular, we do not need to flip the tuning curve).



Actuallly, the intercept is required to be > 1 (for max_rate < 1/(2 * tau_ref))! Independent of wether max_rate is below or above the inflection point firing rate, solving for an invalid intercept will give a non-positive gain.

# Nengo has a bug when solving for sigmoid tuning curves

When specifying a max_rate below the inflection point firing rate the inflection point (“intercept”) will not be at the specified location. For example:

```python
    ens = nengo.Ensemble(1, 1, neuron_type=nengo.Sigmoid(), intercepts=[0.9], max_rates=[50])

    pass

xs = np.linspace(0, 2, 100)
plt.plot(*nengo.utils.ensemble.tuning_curves(ens, sim, xs))
plt.xlabel("x")
plt.ylabel("Firing rate [Hz]")
plt.show()
```
produces the inflection point at 1.1 and not 0.9 as specified:


But 0.9 is not a valid intercept value in this case. Let's try 1.1, which is valid:



Now the orientation of the tuning curve is flipped and the inflection point is at 0.9 instead of 1.1 where it is supposed to be.

# Summary of what I suggest

* Instead of checking the intercepts (as in this PR), we check that the gain is positive.
* We fix the gain and bias calculation of the sigmoid neurons to correctly treat intercepts > 1 and max rates below the inflection point.

# Acknowledgements

I would not have realized these things out without the discussion with @xchoo and a comment by @tcstewar about positive gains.

tcstewar:
Very nice description, @jgosmann ....  :)

And I very much like the underlying principle being that gains must be positive.  I think that's a fairly strong assumption of the NEF -- indeed, without that assumption encoders are no longer the ideal stimulus for a neuron (or the preferred direction vector), which is pretty much what we always describe them as.  So that seems like a much deeper theoretical commitment than the position of intercepts, especially given that for some neuron types (e.g. sigmoids) there's no such thing as an actual x-intercept, but we just use it to mean something else.

So I'm definitely in favour of this change, and I'm very glad the discussion led to this clarity.  

If we do take this approach, what should we do when negative gains are found?   Should we throw an exception, or should we give a warning and set them to zero?

jgosmann:
> If we do take this approach, what should we do when negative gains are found? Should we throw an exception, or should we give a warning and set them to zero?

I am strongly in favour of an exception because the behaviour becomes essentially undefined and you need to know the neuron type and arbitrary additional assumptions (gain set to 0 vs. flipped tuning curve) to derive the actual behaviour. Also because I like to quote the Zen of Python: “In the face of ambiguity, refuse the temptation to guess.” (So we should not guess whether the user wants those gains set to 0 or flipped tuning curves.)

An interesting question will be how to phrase the exception (or warning). While saying “You're gain is negative!” is technically correct, it is not helpful for most users. I think we have to include something to tell them that they have to check their intercepts.

tcstewar:
> Also because I like to quote the Zen of Python: “In the face of ambiguity, refuse the temptation to guess.”

:)  That sounds like a pretty good reason to me.

> An interesting question will be how to phrase the exception (or warning). While saying “You're gain is negative!” is technically correct, it is not helpful for most users. I think we have to include something to tell them that they have to check their intercepts.

Hmm..  good point.  Perhaps something like: ```The specified intercepts lead to neurons with negative gain. Please adjust the intercepts so that all gains are positive.  This is often achieved by reducing the maximum intercept value to below 1.```

jgosmann:
I opened a new PR #1248 superseding this one accounting for my new insights.

Regarding @xchoo discussion points:

> The use of the term "max rate". It is my opinion that this is a misnomer prone to creating misunderstandings on how the code (should) work.

> The ability to define where (at what x value) the "max rate" value should be. This came up in offline discussions with Jan, and we agree that it would be nice to be able to specify y-intercept values outside of the x == 1 condition.

These points have been addressed to some degree [in my previous comment](https://github.com/nengo/nengo/pull/1243#issuecomment-267787806). I don't think "max rate" is a misnomer if we think of it as the max rate within the radius. It might be necessary to document this better.

Using points (x, y) with firing rate y at x another approach to define the tuning curve, but I see it distinct from our current intercept/max_rate approach. I'm open to future discussions whether this alternate approach should be supported.

> When and where to stick with the NEF convention of keeping everything within the unit circle.

> Whether or not a warning or an exception is appropriate for conditions like this where unexpected inputs produce unexpected results.

> Related to the last point, what is the best way to indicate / educate the user as to how the code base works and what their intuitions & expectations of the code should be.

Where would be the best place to discuss these issues? In the forum?

Seanny123:
> > When and where to stick with the NEF convention of keeping everything within the unit circle.
> > Whether or not a warning or an exception is appropriate for conditions like this where unexpected inputs produce unexpected results.
> > Related to the last point, what is the best way to indicate / educate the user as to how the code base works and what their intuitions & expectations of the code should be.
> 
> Where would be the best place to discuss these issues? In the forum?

Sounds like a thorny enough issue that it might be worth just discussing during a dev meeting.

jgosmann:
This PR is superseeded by #1248.

PR #1245, opened 2016-12-16
Simulator changes (`nengo_deeplearning` compatibility)

drasmuss:
**Motivation and context:**
This is kind of a grab bag of issues I came across while working on updates to `nengo_deeplearning`.  No new features, just fixing some bugs/inconsistencies/unnecessary code.  If anyone wants I can split these into separate PRs, but they're all 2--3 line changes so I thought I'd help out our PR queue and keep it all together (at least to start).

**How has this been tested?**
Added unit tests for the bug fixes (to show that the bug is fixed).  For the parts where I took out unnecessary simulator operations, the test is just that all the unit tests continue to pass with that bit removed.

Go through one commit at a time, as each is independent.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/tests/test_builder.py
+++ b/nengo/tests/test_builder.py
@@ -4,7 +4,6 @@ import pytest
 import nengo
 from nengo.builder import Model
 from nengo.builder.ensemble import BuiltEnsemble
-from nengo.builder.operator import DotInc, PreserveValue
 from nengo.builder.signal import Signal, SignalDict
 from nengo.exceptions import ObsoleteError, SignalError
 from nengo.utils.compat import itervalues, range
@@ -123,30 +122,6 @@ def test_signal_values():
         two_d.initial_value = np.array([[0.5], [-0.5]])
     with pytest.raises((ValueError, RuntimeError)):
         two_d.initial_value[...] = np.array([[0.5], [-0.5]])
-
-
-def test_signal_init_values(RefSimulator):
-    """Tests that initial values are not overwritten."""
-    zero = Signal([0])
-    one = Signal([1])
-    five = Signal([5.0])
-    zeroarray = Signal([[0], [0], [0]])
-    array = Signal([1, 2, 3])
-
-    m = Model(dt=0)
-    m.operators += [PreserveValue(five), PreserveValue(array),
-                    DotInc(zero, zero, five), DotInc(zeroarray, one, array)]
-
-    with RefSimulator(None, model=m) as sim:
-        assert sim.signals[zero][0] == 0
-        assert sim.signals[one][0] == 1
-        assert sim.signals[five][0] == 5.0
-        assert np.all(np.array([1, 2, 3]) == sim.signals[array])
-        sim.step()
-        assert sim.signals[zero][0] == 0
-        assert sim.signals[one][0] == 1
-        assert sim.signals[five][0] == 5.0
-        assert np.all(np.array([1, 2, 3]) == sim.signals[array])
 
 
 def test_signaldict():
--- a/nengo/utils/connection.py
+++ b/nengo/utils/connection.py
@@ -107,3 +107,22 @@ def eval_point_decoding(conn, sim, eval_points=None):
     decoded = np.dot(activities, weights.T)
     targets = get_targets(conn, eval_points)
     return eval_points, targets, decoded
+
+
+def function_name(func):
+    """Returns the name of a function.
+
+    Unlike accesing ``func.__name__``, this function is robust to the
+    different types of objects that can be considered a function in Nengo.
+
+    Parameters
+    ----------
+    func : callable or array_like
+        Object used as function argument.
+
+    Returns
+    -------
+    str
+        Name of function object.
+    """
+    return getattr(func, "__name__", func.__class__.__name__)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,8 @@ Release History
 - The matrix multiplication example will now work with matrices of any size
   and uses the product network for clarity.
   (`#1159 <https://github.com/nengo/nengo/pull/1159>`_)
+- Fixed instances in which passing a callable class as a function could fail.
+  (`#1245 <https://github.com/nengo/nengo/pull/1245>`_)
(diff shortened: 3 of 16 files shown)


drasmuss:
@hunse I noticed something in this breaks `nengo_ocl`.  Here's the error:
```
File "d:\documents\nengo_ocl\nengo_ocl\simulator.py", line 293, in __init__
    self._plan.extend(self.plan_op_group(op_type, op_list))
  File "d:\documents\nengo_ocl\nengo_ocl\simulator.py", line 562, in plan_op_group
    return getattr(self, 'plan_' + op_type.__name__)(ops)
  File "d:\documents\nengo_ocl\nengo_ocl\simulator.py", line 633, in plan_SlicedCopy
    plans.append(plan_copy(self.queue, A, B, incs))
  File "d:\documents\nengo_ocl\nengo_ocl\clra_nonlinearities.py", line 212, in plan_copy
    assert (arr.shape1s == 1).all()
AssertionError
```

I believe it is this change https://github.com/nengo/nengo/pull/1245/commits/75db8beba02ebc54a0031c6312554f90e35f87b2#diff-d4a1ea7c740725a930af968ec8143735R350 where I changed the `target += delta*1` to `target += delta`.  Is the requirement that all copies be on vectors (not matrices) a hard one?  If so we can always go back to the `delta*1`, just thought I'd see if it's an easy fix or not.

hunse:
It's not hard to change. One question is how significant the performance drop would be if the data isn't contiguous. I'm also not sure if that performance drop is a run-time thing or compile-time. If it's run-time, then only data that's actually not contiguous would be slower to copy. If it's compile time, then everything could be slower, because the compiler doesn't know if the data's contiguous or not, so it has to assume not.

The simplest would be to make the change and run some large models (with the profiler on) and see if it makes a difference. I have a circular convolution example set up for benchmarking, so testing on that might be enough.

So the short answer is it's doable, though if the change can wait, or someone else wants to do it, then that's better for me.

drasmuss:
It's not an urgent change at all, since I can just patch the builder within `nengo_deeplearning` to change the  `ElementwiseInc` to a `SlicedCopy` just for that backend.  We only need to update it (if we do update it) in `nengo` if we want that efficiency improvement for all `nengo` builder backends.  And, like you say, if it makes the copy operation slower then we probably wouldn't want that in `nengo_ocl` anyway.

drasmuss:
I moved that bit into its own PR, so the remaining changes here should all be compatible with `nengo_ocl`.

nengo/tests/test_builder.py
```diff
@@ -125,30 +124,6 @@ def test_signal_values():
         two_d.initial_value[...] = np.array([[0.5], [-0.5]])
 
 
-def test_signal_init_values(RefSimulator):
```
Seanny123: FYI, for the next reviewer, this test is duplicated specifically by [this test](https://github.com/nengo/nengo/blob/master/nengo/tests/test_simulator.py#L140).

Seanny123:
LGTM, but I'm not super confident about the 2nd commit because I have a very limited familiarity with the builder. Hopefully, the next reviewer will assuage my fears.

jgosmann:
Changes look good to me, except for potentially a very minor comment I made inline and as Sean not being super confident about the `PreserveValue` commit. It definitely simplifies things a bit (and would marginally speed up things because less operators to iterate over). To me it seems that the no increment without set/update rule is to avoid hard to find bugs?

nengo/utils/connection.py
```diff
@@ -107,3 +107,21 @@ def eval_point_decoding(conn, sim, eval_points=None):
...
+    func : callable or array-like
+        object used as function argument
+
+    Returns
+    -------
+    str
+        name of function object
+    """
+
+    return getattr(func, "__name__", func.__class__.__name__)
```
drasmuss: `type(obj)` was what I had originally, but in Python 2.7 `type(func).__name__` just displays `instance`, whereas `func.__class__.__name__` gives you the name of the class.

drasmuss:
I believe that the main motivation for the no increment without set/update rule was that in the original design we didn't have our current system for setting initial values.  Signals were just empty containers with no value.  So the rule was supposed to check that the signal actually had a value before you tried to increment it.  But we've made some changes to the signal system since then, and now all signals start with initial values, so we don't need to be as worried about that.

nengo/utils/connection.py
hunse: I had assumed they were the same, and if your class inherits from `object` they are. For old-style classes though, the behaviour is as @drasmuss described.

Maybe I shouldn't have changed `x.__class__` to `type(x)` everywhere...

hunse: Looking at the commit though, I did it to help with some old types of objects. I wish I could remember what, exactly. https://github.com/nengo/nengo/pull/1010/commits/76f0a34571796e07fcd3be9db62f68e12cf007da

hunse: Maybe it was for MRO. For instances of old classes, `x.__class__.__mro__` does not exist, whereas `type(x).__mro__` does. However (and maybe I didn't realize this at the time I made the changes), `type(x).__mro__` just gives `(instance, object)` no matter what the actual inheritance is, so it's not useful. Does anybody know how to get the MRO of an old-style class?

Anyway, `x.__class__.__name__` seems generally better to me than `type(x).__name__`, so I'm fine to keep it here. We should also maybe consider reverting 76f0a34

hunse:
My impression was always that the set/update rule was for bug catching, but the initialization argument makes sense, too. It could have been a combination of the two. Anyway, I think we're fine without it, now.

This all LGTM

nengo/utils/connection.py
jgosmann: Somehow, missed the last bunch of comments and somewhat as side comment: I think we should in general use and encourage the usage of new-style classes (i.e. every class should inherit from `object`).

hunse: I agree, and I think most people do. I would prefer not to have a hard rule about that, though, and if we do, we would want to fail right away with an explicit check, not in some random place.

PR #1246, opened 2016-12-20
Change ElementwiseInc to copy in learning rules

drasmuss:
**Motivation and context:**
With the PreserveValue changes in #1245, it is no longer necessary to use an `inc` instead of a `set`.  So instead of doing an unnecessary elementwise multiplication by one, we can just directly add the learning rule delta to the target signal.

Based on #1245

Breaks `nengo_ocl`, which relies on all `SlicedCopy` ops being applied to vectors (whereas `delta` can be a matrix).

**How has this been tested?**
Tested by existing unit tests.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 from nengo.builder import Builder, Operator, Signal
-from nengo.builder.operator import DotInc, ElementwiseInc, Reset
+from nengo.builder.operator import DotInc, ElementwiseInc, Copy, Reset
 from nengo.connection import LearningRule
 from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError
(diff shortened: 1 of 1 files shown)


Seanny123:
This looks good to me, with the caveat that I still don't have a complete grasp of the builder. However, this seems to only be affecting an isolated part of it, so I would be astonished if it had widespread effects.

hunse:
It seems to me the only reason we're using `SlicedCopy` here is because it has `inc`, which is not a good reason. Ideally, in my mind, we'd finally get around to combining `Copy` and `SlicedCopy`, since `SlicedCopy` with no slices set is the same as `Copy`. I can look into that.

Currently, the way things are, this would actually slow down `nengo_ocl` I think. Multiplying by a constant is cheap, but copying something where you're using indices and don't know the memory layout (as in `SlicedCopy`) is expensive on a GPU.

It shouldn't be a huge deal to expand `nengo_ocl` to allow for copies on matrices.

jgosmann:
Could you check in nengo_ocl whether the copy is actually sliced and use faster code for non-scliced copies? The data needs to be “copied” (read and added to the target signal) in either case, doesn't it?

hunse:
Yes, that's possible. But I think it makes sense to combine `Copy` and `SlicedCopy`, and doing that now will make the code in `nengo_ocl` easier to write.

jgosmann:
So we are basically waiting on #1083.

hunse:
Hey, I forgot I'd done that! Yeah, I'd prefer to merge #1083 first. Alternatively, this could be merged first, as long as #1083 gets in before the next release.

drasmuss:
Rebased, post #1083

PR #1248, opened 2016-12-23
Raise exception on invalid gains.

jgosmann:
**Motivation and context:**
As noted in #1212 (and #1231) intercepts >= 1 give unexpected results for most neuron types. The matter is however a bit more complicated [as detailed in this comment](https://github.com/nengo/nengo/pull/1243#issuecomment-267787806). The correct generalization to all neuron types (in my opinion) is that non-positive gains give unexpected results. Thus, the core change of this PR is to check this and raise an exception when violated. There is a bunch of related changes, though:

* As detailed in the linked comment, the gain and bias calculation for the sigmoid neurons was incorrect giving flipped tuning curves for max. firing rates below the firing rate of the inflection point. This calculation was fixed. The new gain formula is based on the fact that the firing rate at the inflection point has to be 1/2 of 1/tau_ref.
* As also detailed in the linked comment, for max. firing rates below the firing rate of the inflection point, the intercept (i.e. inflection point) has to be > 1. The sigmoid inflection point was set to 250Hz while our default max_rates were 200 to 400Hz. Thus, depending on the sampled max_rates, the valid intercept range would switch. I propose with this PR to change sigmoid `tau_ref` to 0.0025 which brings the inflection point to 200Hz and the firing rate in the limit to 400Hz. This at least allows to use the sigmoid neuron type without an exception and changing any of the other ensemble defaults. But technically it is a breaking change! We could leave `tau_ref` at it's old value, but would likely need to adjust the max_rates in a lot of our tests.
* Changing the `tau_ref` changed the sigmoid tuning curves which required to fix a few other tests.

This PR supersedes #1243.

**How has this been tested?**
Added some tests for the sigmoid neuron and a test checking that invalid intercepts raise an exception.

**Types of changes:**

- [Without the `tau_ref` change: Bug fix (non-breaking change which fixes an issue)]

**Checklist:**

- [ ] I have updated the documentation accordingly. (Maybe some [this comment](https://github.com/nengo/nengo/pull/1243#issuecomment-267787806) should put somewhere in the documentation? Maybe we should add what intercepts are valid to the documentation of sigmoid neurons? In any case some feedback on the technical site of this PR would be nice before investing time in further documentation.)

### DIFF
--- a/nengo/builder/ensemble.py
+++ b/nengo/builder/ensemble.py
@@ -8,6 +8,7 @@ from nengo.builder import Builder, Signal
 from nengo.builder.operator import Copy, DotInc, Reset
 from nengo.dists import Distribution, get_samples
 from nengo.ensemble import Ensemble
+from nengo.exceptions import BuildError
 from nengo.neurons import Direct
 from nengo.utils.builder import default_n_eval_points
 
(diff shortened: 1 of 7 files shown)


nengo/builder/ensemble.py
```diff
@@ -95,6 +96,14 @@ def get_gain_bias(ens, rng=np.random):
         max_rates = get_samples(ens.max_rates, ens.n_neurons, rng=rng)
         intercepts = get_samples(ens.intercepts, ens.n_neurons, rng=rng)
         gain, bias = ens.neuron_type.gain_bias(max_rates, intercepts)
+        if gain is not None and (
+                not np.all(np.isfinite(gain)) or np.any(gain <= 0.)):
+            raise BuildError(
```
Seanny123: Would it be useful to link to a discussion of this problem? Maybe a forum post with the delightful doodles you made?

jgosmann:
Can we get this PR merged soon?
I just ran into this sort of problem again and an early failure would have been nice.

PR #1253, opened 2017-01-04
Rewrite associative memory.

jgosmann:
**Motivation and context:**
This PR introduces a rewritten associative memory for the `new-spa`.

Compared to previous implementations where the `AssociativeMemory` implemented all possible variants, I split up different possible implementations.

The core networks that can be used to implement something like an associative memory are found in `selection.py`. These are a simple array of thresholding ensembles (`ThresholdingArray`), a basic WTA (`WTA`) with lateral inhibition, and the independent accumulator (`IA`) network submitted to CogSci. All these core networks are standard Nengo networks because they don't operate on semantic pointers, but each dimension represents a potential choice. (I also wanted to repeatedly use such networks in situations where I didn't have semantic pointers.) This also leads to the question: **Do we want those networks potentially in core Nengo? Or do we keep them in nengo_spa?**

The `AssociativeMemory` is basically a wrapper around such a core network that adds the projections from and to the Semantic Pointer space and allows to use the network with other SPA modules. We pass only the class of the core network + kwargs to the AssociativeMemory to avoid the user to have to pass `add_to_container=False`. Also, for better usability for the end-user there are classes `ThresholdingAssocMem`, `WtaAssocMem`, and `IaAssocMem`.

Further comments:
* I am not setting a default value for the threshold. To me it seems better to explicitly force the user to specify the threshold.
* By default the identity function is applied to the outputs of `ThresholdingArray` and `WTA`. A step function might be more appropriate in some cases, but is easily added by providing the `function=lambda x: x > threshold` argument.
* I haven't implemented an `inhibitable` flag. One still has the option to create inhibitory connections to the core network and other SPA modules don't provide this option either (actually I would find it more useful on `spa.State` then on the `AssociativeMemory`). I plan to make a PR to core Nengo with a function that inhbits all ensembles in a network. It's a functionality I commonly need, even outside of SPA. That function could than also be used for all SPA modules.
* I haven't implemented the cleanup with double inhibition. It never made it into the Nengo master (but exists in @xchoo's `assoc_mem_v3` branch). I'm not sure whether we want to include it. It reduces the noise on the output in certain circumstances, but I haven't missed the functionality and it could always be added later.
* There is no non-SPA associative memory that uses vectors anymore. If you need that, you can use one of the core networks and set the transforms on the input and output connection yourself. It doesn't seem necessary to have just a network for that and chances are high if someone is using vectors, the SPA associative memory with a proper vocab can be used.

This PR supersedes #991.

**How has this been tested?**
Updated and adapted existing tests.

**Checklist:**

- [x] I have updated the documentation accordingly. [The associative_memory example probably needs to be updated, but all the SPA examples need to be updated at some point which will be a separete effort.]

**Still to do:**
- [x] documentation
- [x] drift diffusion network
- ~~cleanup with double inhibition?~~

### DIFF
--- a/examples/spa/associative_memory.ipynb
+++ b/examples/spa/associative_memory.ipynb
@@ -1,9 +1,9 @@
 {
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "nengodev",
    "language": "python",
-   "name": "python3"
+   "name": "nengodev"
   },
   "language_info": {
    "codemirror_mode": {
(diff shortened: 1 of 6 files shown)


jgosmann:
@tcstewar, @xchoo, @tbekolay Is anyone of you able to give this a rough look over to provide some feedback on the direction this PR is going?

xchoo:
Sorry Jan. I'm a little busy at the moment. I can look at it in maybe a month or so. :/

tbekolay:
I'll take a look at this tomorrow!

tbekolay:
Took a look at this! Not in any great detail, so I can't speak to the performance improvements or anything like that, but in terms of code organization, I think I would probably opt for multiple AssociativeMemory subclasses over passing in the network function and a set of args. From the user perspective,

```python
model.am1 = ThresholdingAssociativeMemory(..., threshold=0.5)
model.am2 = WtaAssociativeMemory(..., inhibit_scale=2.0)
```

seem clearer and easier to remember / document than

```python
model.am1 = AssociativeMemory(..., selection_net=ThresholdingArray, selection_net_args=dict(threshold=0.5))
model.am2 = AssociativeMemory(..., selection_net=WTA, selection_net_args=dict(inhibit_scale=0.5))
```

If the long names are an issue, I think shortening to `AM` is memorable enough (`ThresholdingAM`, `WTAAM` -- OK that one might not be the best).

I think it's possible to not repeat most of the logic using inheritance... but there would be a bit of repetition in the `__init__` method signature and in the `super(...).__init__` calls. But, the user doesn't have to see all of that, so I think a bit of repetition under the hood is worth it.

---

Another thing that I would caution against, since it was prevalent in Nengo 1.4 and ended up confusing me a lot when I used it, are the `LinearFunc`, `FilteredStepFunc` etc function that return functions. Maybe the requirements of SPA are different enough that it makes sense, but the reason we don't provide much of that even though it might be useful for `Connection.function` and `Node.output` is:

1. You can't predict all of the kinds of functions people will want to end up using;
2. If you try to predict all of them and add in functions that people want, the list of functions gets so long that it's more effort to look through the list of functions than it is to write your own function;
3. If you don't try to predict all of them and say "here's a small set of functions, aside from that make your own" you have to go through the effort of teaching people how to make functions anyway, so you might as well do that teaching earlier on and relieve people of the burden of remembering what functions exist and what functions they have to write themselves.

jgosmann:
> Not in any great detail, so I can't speak to the performance improvements or anything like that, but in terms of code organization

That's all I was asking for. :)

How do people feel about `ThresholdingAssocMem` and `WtaAssocMem` as a compromise between very long and very short names?

> Another thing that I would caution against, since it was prevalent in Nengo 1.4 and ended up confusing me a lot when I used it, are the LinearFunc, FilteredStepFunc etc function that return functions.

Seems fair.

jgosmann:
I added subclasses for the associative memory variants, but didn't change the implementation of the `AssociativeMemory` class. Is this better?

tbekolay:
> I added subclasses for the associative memory variants, but didn't change the implementation of the AssociativeMemory class.

Looks good to me!

jgosmann:
Cool! :)

jgosmann:
This should be ready for final review now, @tcstewar. I updated the PR description above.

There is one important question: Do we want the selection networks in core Nengo or do we want to keep them in nengo_spa?

tbekolay:
> Do we want the selection networks in core Nengo or do we want to keep them in nengo_spa?

I think moving them to `nengo_spa` makes sense.

jgosmann:
@tcstewar, any estimate when you will review this PR? It's the last one before moving SPA to a separate repository.

jgosmann:
The need to have nengo_spa in a separate repository became too pressing for me and I went ahead and moved the code over. This PR has been transferred to nengo/nengo_spa#1.

Issue #387, opened 2014-06-05
being able to specify how often a learning rule is applied


studywolf:
Right now in my own learning code I'm doing a hack where I have a counter and apply the learning rule once every N timesteps instead of every timestep, just to speed up the simulation, it might be nice to have this option for learning rules in general.

bmorcos:
Looking for some opinions on what is more intuitive. Let's say you want to apply a learning rule only **every 4 steps**, should you pass the parameter:
- As 4 (as in every 4 steps)
- As 1/4 (as in apply only 1/4 of the time)

Thoughts?

jgosmann:
To me every `dt` seconds seems the most intuitive because everything else is usually specified in seconds and would make the value independent of the simulator `dt`. However, the simulator `dt` needs to be a multiple of the learning rule `dt` and that condition could be hard to ensure, also because of floating point errors (`sim_dt % lr_dt == 0` might be false even though analytically it should be true).

tcstewar:
I definitely agree it should be specified in seconds, just like the Probe's ```sample_every``` argument.  And, indeed, that code also deals with the above problem of worrying about it being evenly divisible (or dealing with it if it isn't).  In that case, we just do:

```python
period = (1 if probe.sample_every is None else probe.sample_every / self.dt)
```

https://github.com/nengo/nengo/blob/master/nengo/simulator.py#L214

So that just rounds the number of steps to make it work.

bmorcos:
Another discussion regarding the PES learning rule implementation in `nengo/builder/learning_rules.py`. Right now the PES rule does not implement it's own operator class like the other learning rules do (ex. `class SimBCM(Operator)`); instead PES uses general purpose operators. I was led to believe this has to do with the `nengo_OCL` implementation. I am looking at implementing this _learn sometimes_ feature in the operator by doing something like this:

```
    def make_step(self, signals, dt, rng):
        pre_filtered = signals[self.pre_filtered]
        post_filtered = signals[self.post_filtered]
        theta = signals[self.theta]
        delta = signals[self.delta]
        alpha = self.learning_rate * dt
        period = (1 if self.apply_every is None else apply_every / self.dt)

        def step_simbcm():
            delta[...] = np.outer(
                alpha * post_filtered * (post_filtered - theta),
                pre_filtered)

        def step_simbcm_sometimes():
            if n_steps % period < 1:
                delta[...] = np.outer(
                    alpha * post_filtered * (post_filtered - theta),
                    pre_filtered)

[...]

PR #1254, opened 2017-01-04
Apply learning every n-th timestep

bmorcos:
**Motivation and context:**
Apply learning sometimes to help speed up simulation #387 

**How has this been tested?**
I made a small network in the GUI and compared normal learning to n-th step learning as I played with parameters (new `apply_every`, `dt`, `learning rate`). Added new tests and modified existing tests.

I rewrote the PES implementation, creating an operator class like the other rules, take a look at this to make sure I didn't break the normal use case (although it passes unit tests so it should be fine). Adding n-th step functionality to the other learning rules was fairly minor. 

**Types of changes:**
    - NOTE: new `SimPES` operator class breaks `nengo_ocl` [#131](https://github.com/nengo/nengo_ocl/issues/131) ,  non-breaking for `nengo`

**Checklist:**

**Still to do:**
- [x] Make an issue in `nengo_ocl`
- [ ] Make a PR in `nengo_ocl` to cover this change


nengo/tests/test_learning_rules.py
```diff
@@ -33,7 +33,9 @@ def _test_pes(Simulator, nl, plt, seed,
 
         conn = nengo.Connection(pre, post,
                                 function=function, transform=transform,
-                                learning_rule_type=PES(rate))
+                                learning_rule_type=PES(
```
Seanny123: Maybe define the rule outside of the `nengo.Connection` for the sake of readability?

nengo/tests/test_learning_rules.py
```diff
@@ -70,39 +72,45 @@ def _test_pes(Simulator, nl, plt, seed,
     assert not np.allclose(weights[0], weights[-1], atol=1e-5)
 
 
-def test_pes_ens_ens(Simulator, nl_nodirect, plt, seed):
+@pytest.mark.parametrize('apply_every', [None, 1e-5, 0.005])
```
Seanny123: Maybe test `apply_every=0` since it's a corner case and you might want to make it's behaviour explicit?

bmorcos: `0` should behave the same as `1e-5`, any value of `apply_every < dt` should be identical. I have tested `0` also and it's fine; any argument to use a test value other than `1e-5`... it is somewhat arbitrary and I have no attachment to it? (negative numbers should also behave the same but they are not valid and raise an exception)

nengo/builder/learning_rules.py
```diff
@@ -86,10 +97,30 @@ def make_step(self, signals, dt, rng):
         theta = signals[self.theta]
         delta = signals[self.delta]
         alpha = self.learning_rate * dt
+        period = (1 if self.apply_every is None else
```
bmorcos: Is there any difference using the ternary vs. normal or is this purely a preference/readability thing?

Maybe @tbekolay has an opinion on refactoring this? My understanding is that the operator must return instructions that get applied to the signals. Am I able to return something wrapped in a function?

tbekolay: Ternary vs. normal is just a preference thing; we use the ternary a lot to cut down on the number of lines. Using an if statement this would be 4 lines versus the 2 it is now. Of course, it can be easier to read 4 short lines than 2 dense lines, so it's a case-by-case thing. I think I prefer ternary here, but if statements would be good too.

As for the refactoring, I could see it being helpful and making the implementation a bit simpler. It would return the same thing, it would just move things around a bit. I can make a commit with that refactoring, or @Seanny123 can (making a commit with the change would be faster than trying to explain the change with words, I think).

Seanny123: Yeah, this is one of those cases where I should have made a commit. *bows apologetically*

I'll try and make one tonight.

bmorcos:
@Seanny123 should I take a stab at refactoring this or are you still interested in doing that?

Seanny123:
If you feel like refactoring it, go for it. I'm only going to be able to refactor this thing in February. Alternatively, you can get another reviewer to look at it and decide if the refactoring is necessary or not.

nengo/builder/learning_rules.py
```diff
@@ -278,6 +371,253 @@ def step_simvoja():
...
+            def step_simpesdecoders_sometimes():
+                if n_steps % period < 1:
+                    correction[...] = -alpha / n_neurons * error
+                    delta[...] = np.outer(correction, pre_filtered)
+
+                    # scale to compensate
+                    delta[...] *= period
+
+                else:
+                    delta[...] = 0
```
bmorcos: Do we want to set `correction[...] = 0` here also? That will make probing more intuitive as it holds `0` instead of the previous value. Is it worth the extra step?

nengo/builder/learning_rules.py
```diff
@@ -36,15 +37,23 @@ class SimBCM(Operator):
         The synaptic weight change to be applied, :math:`\Delta \omega_{ij}`.
     learning_rate : float
         The scalar learning rate, :math:`\kappa`.
+    apply_every : float or None
+        A scalar indicating how often to apply learning rule.
+    n_steps : Signal
+        The integer timestep which the simulator is executing.
```
jgosmann: The `n_steps` name suggests to me that this give some total number of steps, but from the docstring it seems it is just the step number? In that case I would just name it `step`, otherwise the docstring should be clarified.

nengo/builder/learning_rules.py
```diff
@@ -36,15 +37,23 @@ class SimBCM(Operator):
...
     Attributes
     ----------
+    apply_every : float or None
+        A scalar indicating how often to apply learning rule.
     delta : Signal
         The synaptic weight change to be applied, :math:`\Delta \omega_{ij}`.
     learning_rate : float
         The scalar learning rate, :math:`\kappa`.
+    n_steps : Signal
+        The integer timestep which the simulator is executing.
```
jgosmann: see above

nengo/builder/learning_rules.py
jgosmann: What is `correction`? Where does it come from? I don't see it in the code at all ...?

jgosmann: Ah, found it below. I'd say if `correction` is set to 0 it should also be scaled by the `period` to keep the average about the same. Though, I'm not sure whether we want to set it to 0 or not.

bmorcos: This is in the builder,
`correction = -learning_rate * (dt / n_neurons) * error` 
[This](https://github.com/nengo/nengo/blob/master/nengo/builder/learning_rules.py#L534) is the current implementation. My implementation is in the new `SimPES*` operators [here](https://github.com/nengo/nengo/blob/learn-sometimes/nengo/builder/learning_rules.py#L475).

jgosmann:
I'd also like to see refactoring to reduce code duplication.

tbekolay:
> I'd also like to see some refactoring to reduce code duplication.

Could you be more specific with this kind of comment? It's a tough thing for the PR author to do without some direction.

bmorcos:
>  I'd say if correction is set to 0 it should also be scaled by the period to keep the average about the same. Though, I'm not sure whether we want to set it to 0 or not.

It might more sense to scale `correction` directly instead of `delta`, that way probing `correction` will yield *true* values. Whether or not to reset to 0 is still a separate question I think.

Thoughts?

celiasmith:
Yeah I was wondering if that was a throw away 'it would be good in the future if...' observation, or if you think it should be done now before this is acceptable?  (If the latter, then maybe more direction would help Ben).

jgosmann:
Sorry, the refactoring was already discussed in previous comments, so it didn't seem necessary to me to reiterate this. Actually, I will take a stab it now because that is probably more efficient than textual explanations and should make things more clear.

Furthermore, I'm of the opinion that this should be done before merging. Once code is merged those things get forgotten and can decrease code quality slowly over time.

celiasmith:
@jgosmann Oh, sorry, my bad.  I didn't understand the context, thanks for pointing it out.  And also thanks for doing it! 🥇

jgosmann:
One concern that just occured to me: This adds `apply_every` to all learning rules which sort of forces every learning rule to implement it. That is fine for the given learning rules, but can be a problem when implementing new learning rules. Especially, if one might opt to use the standard operators to get NengoOCL support without a special implementation for OCL. Of course, one could ignore `apply_every` in such a learning rule, but it feels to me like one should not? Relatedly, it seems to me to force the implementation of all learning rules to use special operators, if they want to support `apply_every`.

jgosmann:
btw: I think this would be much easier to refactor in Ruby. Anyone interested in writing Nengo 3 in Ruby? ;) (I'm not being serious about this.)

bmorcos:
I think you mentioned adding this functionality to operators in the original issue (#387) too, but seems like it was largely overlooked... I imagine that would be a pretty serious undertaking too.

tbekolay:
> This adds apply_every to all learning rules which sort of forces every learning rule to implement it.

Yeah, @bmorcos and I discussed this and how you might implement it the other way at length. The way that operators are set up right now doesn't make it easy, as there aren't really any conditionals, so we can't have a particular operator run on some steps but not others. There are a few ways you could introduce this capability (e.g., have multiple lists of step functions that are applied at different rates, add something to every operator such that it can be run on some steps but not others), but none of those solutions seemed very satisfying, and even if one were satisfying it would affect too many things to be done in this PR.

So, this solution is a bit of a compromise, but the question isn't whether it's a compromise, but whether it's worth adding this feature now despite it affecting all learning rules, and requiring operators for each learning rule. If you don't think so, then mark your review as "Request changes" rather than "Approved".

jgosmann:
I added a commit with the refactoring. It might be a step towards implementing a second set of standard operators with intermittent execution that could easily be implemented in NengoOCL. That, however, is out of the scope of this PR.

I suppose when implementing a new learning rule that does not support `apply_every` the build function can raise an exception when it is used. So thinking of that it bothers me less to have it as a general argument for all potential learning rules.

nengo/builder/learning_rules.py
```diff
@@ -10,7 +10,88 @@
...
+
+    An intermittent operator allows for execution in less than each simulation
+    time step. To implement an intermittent operator derive from this class
+    and ensure that:
+
+    * In the constructor of the derived class you append to the operator lists
+      ``self.sets``, ``self.incs``, ``self.reads``, and ``self.updates`` or, if
+      you decide to overwrite them, include `step` in ``self.reads``.
+    * Set ``self.scaled`` to a list of signals that you want to scale by the
+      period length (``apply_every``) each time step.
```
bmorcos: `period = apply_every / dt` do we want to be that explicit in the docstring here?

nengo/builder/learning_rules.py
```diff
@@ -10,7 +10,88 @@
...
+    * Instead of `make_step` implement `make_concrete_step` in the same way
+      you would implement `make_step` normally. No special case code is needed.
+
+    Parameters
+    ----------
+    apply_every : float or None
```
bmorcos: This doesn't match the description in the learning rule operators, I think we should be consistent.  (I think yours is more clear than the one I wrote)

nengo/builder/learning_rules.py
```diff
@@ -10,7 +10,88 @@
...
+        A scalar indicating how often to apply the operator (in simulation
+        time). A value of ``None`` will apply it every time step.
+    step : Signal
+        Signal providing the integer timestep of the simulator.
+    tag : str or None, optional (Default: None)
+        A label associated with the operator, for debugging purposes
+
+    Attributes
+    ----------
+    apply_every : float or None
```
bmorcos: see above

nengo/builder/learning_rules.py
```diff
@@ -10,7 +10,88 @@
...
+        A scalar indicating how often to apply the operator (in simulation
+        time). A value of ``None`` will apply it every time step.
+    step : Signal
+        Signal providing the integer timestep of the simulator.
+    tag : str or None, optional (Default: None)
+        A label associated with the operator, for debugging purposes
+    scaled : list of Signals
+        Signals to be scaled by the period length (``apply_every``).
```
bmorcos: see above

nengo/builder/learning_rules.py
jgosmann: Yes, we should be precise, but maybe instead of mentioning ``apply_every``, just say “period length in simulation time steps”?

bmorcos: I suppose this won't be applied each time step either, as it's intermittent. How about just "Set ``self.scaled`` to a list of signals that you want to scale by the period length in simulation time steps"

bmorcos:
Sorry @jgosmann didn't see you assigned yourself this XD

jgosmann:
np

Seanny123:
@tbekolay was there anything else that needed to be resolved with this branch or is this ready to merge?

tbekolay:
> was there anything else that needed to be resolved with this branch or is this ready to merge?

Hm... I'm looking over this now and I'm not sure about the `IntermittentOperator`. I'll give it some more thought.

Seanny123:
@tbekolay after some thought, how do you feel now?

tbekolay:
After some discussion, we've decided to close this PR unmerged for the time being.

We still think that the idea behind this PR is good (#387 remains open) but the changes made in this PR are going to be very difficult to implement in some backends, particularly NengoDL, so we want to revisit the implementation at a later time (likely after builder refactoring, #1189) and make the change in NengoDL simultaneously to ensure that it can be done in both. We will definitely take parts of this PR when we revisit this feature, though, so your effort was not in vain @bmorcos!

PR #1261, opened 2017-01-27
Change DotInc to ElementwiseInc in PES rule

drasmuss:
**Motivation and context:**
As defined, the arguments to `DotInc` should be a matrix and a vector.  In this case it was being called with a scalar and a vector as input, which would more appropriately be an `ElementwiseInc`.  

The swap works fine in the reference backend, because `dot(scalar, vector) == scalar*vector` in numpy.  But in other backends, such as `nengo_ocl` or `nengo_dl`, there may be specialized implementations for `DotInc` that are optimized under the assumption that it will be called with a matrix.  Since there's no downside to switching this operation to an `ElementwiseInc` for the reference backend, we might as well in order to be consistent.

I've based this on #1245, the only new commit is f1c52e8431560ce3a2b0a4cbed6b7210937cce3e.  Technically it's independent, but it's in the same vein (I created a new PR for this one because #1245 has already been reviewed).

**How has this been tested?**
All unit tests continue to pass with the change.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -540,7 +540,7 @@ def build_pes(model, pes, rule):
                  else conn.pre_obj.size_in)
     lr_sig = Signal(-pes.learning_rate * model.dt / n_neurons,
                     name="PES:learning_rate")
-    model.add_op(DotInc(lr_sig, error, correction, tag="PES:correct"))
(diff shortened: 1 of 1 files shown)


bmorcos:
I rewrote the PES  as it's own operator in #1254, it no longer uses `DotInc` or `ElementwiseInc`, still working on `nengo_ocl` implementation. Not sure what this means for `nengo_dl`

Issue #495, opened 2014-10-06
Run PEP8 on iPython notebook examples


Seanny123:
Apparently, it's not that hard to run PEP8 [inside an iPython notebook](https://stackoverflow.com/questions/26126853/verifying-pep8-in-ipython-notebook-code) or even [after it's been exported](https://stackoverflow.com/questions/26205629/verifying-pep8-in-exported-ipython-notebook-code). It would be a good exercise for a noob to walk through the examples and fix some of the weird uses we have of reserved keywords. That being said, I am aware of #283 where it was concluded that some words were necessary.

mgserafi:
@Seanny123 , I see it's been 2 years since this PR was open. Have you tried running `pep8_magic` since then (the first link for running PEP8 inside a notebook)? I get a permission error when trying to use it.

jgosmann:
Does the error message give any other details? (like a filename or what kind of permission is required)

mgserafi:
When I have `%%pep8` in a cell, I get this error:

`INFO:pep8: ...\AppData\Local\Temp\tmp67jjo44b:1:1: E902 PermissionError: [Errno 13] Permission denied: '...\\AppData\\Local\\Temp\\tmp67jjo44b'`

I'm currently using Jupyter 4.1.0 and ipython 5.1.0

mgserafi:
Managed to get another method for running pep8 on notebooks to work. However, there don't seem to be anymore reserved keyword errors.
All the pep8 errors are based on spacing issues (i.e. blank lines with white spaces, trailing white spaces, missing spaces around operators, etc). Not sure if it's worth fixing all of these...

Seanny123:
So it turns out `pep8` doesn't catch using reserved keywords and that actually has to do with `pylint`. If you check the `.pylintrc` file, you see that we don't even warn on that. @tbekolay what do you think about changing the `.pylintrc` file to also catch use of built-ins?

That being said, I still think it's worth fixing the spacing issues. If this is tedious to do, there should be a way to just automatically fix it and review the results.

jgosmann:
> If you check the .pylintrc file, you see that we don't even warn on that.

[...]

PR #1265, opened 2017-02-13
PEP8 fixes to notebook examples

mgserafi:
**Motivation and context:**
Addresses #495, fixing PEP8 spacing issues. Doesn't address reserved keywords, as discussed in the issue.

**How has this been tested?**
Made sure examples still run without errors.

**Types of changes:**

**Checklist:**
- [X] I have read the **CONTRIBUTING.rst** document.
- [X] All new and existing tests passed.


tbekolay:
Very cool thank you! I'm hoping you did this with a tool and not by hand yes? If there's a tool for it, we should make it part of our automated testing so that the notebooks stay well-styled :+1: If you point me to it I can a separate PR for that, or if you want to add it to this PR feel free -- its inclusion won't block this PR in either case.

mgserafi:
Haha no no I didn't do it all by hand 😛 
I used a jupyter plugin called [code_prettify](https://github.com/jfbercher/code_prettify) which has an autopep8 thing for jupyter.

I'm not sure how to automate running it. I was running it within the notebook so I could see what changes it made. It would do a few things I didn't like such as remove the `;` and give errors when there's a `%` command.

tbekolay:
Cool cool, I'll take a look and think about how to automate running it in a future PR. Thanks!

examples/advanced/inhibitory_gating.ipynb
```diff
@@ -8,10 +8,11 @@
   {
    "cells": [
     {
-     "cell_type": "markdown",
+     "cell_type": "heading",
+     "level": 1,
      "metadata": {},
      "source": [
-      "# Nengo Example: Inhibitory Gating of Ensembles"
+      "Nengo Example: Inhibitory Gating of Ensembles"
```
jgosmann: Is there a reason for changing the markdown headings to cells of the the type "heading"? Current Jupyter version do not use the "heading" cell type anymore. While these cells will automatically converted, I would prefer to keep the markdown cells.

mgserafi: I don't think I explicitly changed this. It might've auto-changed when I opened it in jupyter v4 and converted back to v3?

jgosmann: I added anothor commit to undo these changes.

Issue #670, opened 2015-02-28
Add `plt.show()` to static example scripts


drasmuss:
Our notebook examples don't include `plt.show()`, since it isn't needed.  But it is needed in the static *.py files that are generated from the notebooks.  Right now, users run the example scripts and it seems like nothing happens, which is confusing (just had a new user ask me about this).

The easiest solution might just be to add it to the notebooks, assuming that doesn't cause any problems for ipython.  Or we could do some extra magic when the scripts are generated.

drasmuss:
We also need to strip out the ipython commands (e.g., `%matplotlib inline`), otherwise they cause errors when the script isn't run through ipython (another user reported error).

tbekolay:
We already strip out `%matplotlib inline`; see [this example](https://pythonhosted.org/nengo/examples/single_neuron.html)'s associated [Python script](https://pythonhosted.org/nengo/examples/single_neuron.py). For reference, this is done [here in `utils/ipython.py`](https://github.com/nengo/nengo/blob/master/nengo/utils/ipython.py#L192).

This could be improved to remove any line with `get_ipython` in it, though!

drasmuss:
Ah ok, the user was probably just copy-pasting from the online examples.

mgserafi:
How do you generate the static *.py files?

tbekolay:
> How do you generate the static *.py files?

The function that does this is [`export_py`](https://github.com/nengo/nengo/blob/master/nengo/utils/ipython.py#L171), which is called by the [Notebook directive](https://github.com/nengo/nengo/blob/master/nengo/utils/docutils.py#L50), which is what is used when

```rst
.. notebook::
```

is encountered in our documentation.

PR #1268, opened 2017-02-14
Updated generation of example scripts to include `plt.show()`

mgserafi:
**Motivation and context:**
Addresses #670. Added `plt.show()` to the end of the generated `.py` files and removed any lines that begin with `get_ipython()`.

None

**How has this been tested?**
Built documentation and made sure that:
- `plt.show()` was only added at the end of files that used `matplotlib.pyplot`
- `get_ipython()` is not in the generated script
- ran the generated scripts and made sure plots were generated

See above for how this has been tested.

**Types of changes:**

**Checklist:**
- [X] I have read the **CONTRIBUTING.rst** document.


jgosmann:
> Btw, try not to mention the issue number in your commit message and just describe what's being solved.

That's good practice for the first line, but I think putting a `Closes/Fixes/Addresses #...` below the first line is good practice as it will link the commit automatically to the issue and in case of `closes` and `fixes` will close the corresponding issue once the commit is merged to master.

mgserafi:
Ahh yeah, that's why I had it there, to auto close the issue. Should I go back and change the commit message?

jgosmann:
If it is not too much work, otherwise it can be done easily when merging.

tbekolay:
> Should I go back and change the commit message?

Yes please! To be explicit about why, the idea of the subject line is to get a sense of what the commit does when you're looking through a big long list of commit subjects. So, anything that references an issue isn't self contained so you don't know what it does without referring back to the issue.

tbekolay:
I have a few questions before I finalize my review!

1. What's the motivation behind removing all the lines that contain `get_ipython`? I think I'm OK with it, but curious if there was some example that was giving weird output before.

2. This is a general question to all, not just @mgserafi! My interpretation of #670 was to give the user the same experience as if they had run the script in the notebook. The way the notebook works is that for each cell, it will render out the current figure if there is a current figure, then close it. This means that you can do `plt.plot(x, y)` in a cell and it will render it without you having to do `plt.show()` or even `plt.figure()`. However, if we just add `plt.show()` to the bottom of the file, then you're going to get all of the plots overlaid on one figure if the notebook didn't have explicit `plt.figure()` calls throughout (and most of our notebooks do not do this).

    In my mind, this PR would add `plt.show()` to every cell that had a plotting statement in it. Unfortunately that's not actually a straightforward thing to determine, as if you're just looking at strings and you define a `plot_my_network` function in one cell and call it in another, you would put the `plt.show()` in the cell that defined the function rather than the one that called it. Doing that kind of code analysis doesn't seem to be the right thing to do here, but neither is popping up a plot with 20 lines that make no sense in the same plot.

    So, I think my concrete question is: Do we need to institute some kind of convention in our notebooks so that we explicitly call `plt.figure()` where necessary? Or is it a sufficiently rare case that people download the Python script and run it without Jupyter, so getting a confusing plot is better than no plot at all? Or is there some other solution that provides a good balance?

mgserafi:
> What's the motivation behind removing all the lines that contain `get_ipython?`

In #670, you mentioned that any line with `get_ipython` should be removed 😛 

> So, I think my concrete question is: Do we need to institute some kind of convention in our notebooks so that we explicitly call plt.figure() where necessary? Or is it a sufficiently rare case that people download the Python script and run it without Jupyter, so getting a confusing plot is better than no plot at all? Or is there some other solution that provides a good balance?

I don't know how often people download the python scripts but perhaps we should have `plt.figure()` where it's necessary. It's something I could just add to this PR.

tbekolay:
> In #670, you mentioned here that any line with get_ipython should be removed

Hahaa so I did 😊

jgosmann:
1. `get_ipython` needs to be removed because the script might run in a bare Python interpreter where it does not work.
2. I think the examples would ideally have a `plt.show()` for each plot distinct plot in the appropriate places because someone could decide to not use `%matplotlib inline` even in the notebook. Not sure about `plt.figure()` calls. I would consider it good style, but is also a bit of noise in case of the examples. We might want to test which of `plt.show()` and `plt.figure()` are required for `%matplotlib notebook`. That is definitely a reasonable choice a user could make for the plotting and I recall (but might be wrong) that it had some issues with multiple plots when not using those function calls in appropriate places.

jgosmann:
I should add, what I said about 2 is what I think the best solution would be given unlimited time/resources. If it is more feasible to go with something else, that might be fine too.

mgserafi:
>Not sure about plt.figure() calls. I would consider it good style, but is also a bit of noise in case of the examples. We might want to test which of plt.show() and plt.figure() are required for %matplotlib notebook.

From just playing around with some plots, you only need `plt.show()` to generate the plot. You can have both `plt.figure()` and `plt.show()` in the same plot but it behaves the same way.

`plt.figure()` is only necessary in each cell with a plot if `plt.show()` is called once at the end, to prevent overlaying of all the plots.

jgosmann:
`%matplotlib notebook` requires `plt.figure()` to not plot everything into a single figure. So it would be best to have `plt.figure()` for each plot. We could then do either do a `plt.show()` also for each figure to display those one after another in the scripts (one has to close one figure to see the next one) or we could put a single `plt.show()` at the end and pop up all the figures (in different windows) at the same time.

mgserafi:
I think it's best to have `plt.show()` at the end, so that users can have multiple plots open at the same time, in case they want to view more than one at a time.

mgserafi:
I added `plt.figure()` to the cells that needed it and kept a single `plt.show()` at the end of the example scripts.

Seanny123:
LGTM.

PR #1269, opened 2017-02-15
Fix float indexing in learn_associations

tbekolay:
**Motivation and context:**

Our [appveyor builds are failing on Python 3 64-bit](https://ci.appveyor.com/project/nengo/nengo/build/1.0.1039) :( I was able to reproduce the error locally; I'm not sure if the issue is a new version of NumPy or Python, but for whatever reason, dividing an int by and int somehow gets promoted to a float now, which is causing the error. This casts it to int _again_ to fix the error.

**How has this been tested?**
Fixes the broken test locally. We should wait for AppVeyor to finish before merging.

**Types of changes:**

**Checklist:**

### DIFF
--- a/examples/learning/learn_associations.ipynb
+++ b/examples/learning/learn_associations.ipynb
@@ -135,7 +135,7 @@
       "        raise ValueError(\"dt (%s) does not divide period (%s)\" % (dt, period))\n",
       "    def f(t):\n",
       "        i = int(round((t - dt)/dt))  # t starts at dt\n",
-      "        return x[(i/i_every)%len(x)]\n",
+      "        return x[int(i/i_every)%len(x)]\n",
       "    return f"
      ],
      "language": "python",


drasmuss:
I thought it was expected that `/` always produce a float now, regardless of the type of the inputs (and you have to use `//` if you want an int).

jgosmann:
* `//` is integer division
* `/` is float division in Python 3 (e.g., two ints divided by each other may produce a float)
* `/` is integer division in Python 2 by default, but becomes float division with `from __future__ import division`

I think either the explicit casting (as done here) or the `//` operator are acceptable fixes. I'm just wondering, why is this only failing on Windows?

tbekolay:
> I'm just wondering, why is this only failing on Windows?

No idea... but probably we should fix this before the upcoming release so reviews appreciated :+1:

PR #1270, opened 2017-02-15
Minor fixes for 2.3.1 release

tbekolay:
**Motivation and context:**
As part of the [Nengo release process](https://nengo.github.io/releasing.html) I look at all of the plots generated by the test suite. Noticed a few things that warranted fixing, so I also took the opportunity to tick a few other minor things off of my todo list. See the individual commits for details.

**How has this been tested?**
Ran all tests on Python 2 and 3, generated documentation and looked at rendered examples.

**Types of changes:**

**Checklist:**

- [NA] I have updated the documentation accordingly.


tbekolay:
> I think there is a reason because depending on whether SciPy is installed different code paths can be taken (for example in the BasalGanglia network). To get complete coverage information, both cases have to be run with coverage.

That would be true if there were different code paths within those specific code paths that differ in Python 2 and 3 (since we run Python 2 for coverage without SciPy). I don't think there are any such paths... and if there were, is it worth adding 5 minutes to all builds for those lines of coverage?

jgosmann:
True.

Another thing: Should we bump the Numpy version numbers? The latest stable release is 1.12.

tbekolay:
> Another thing: Should we bump the Numpy version numbers? The latest stable release is 1.12.

Good idea! Added that to the previous commit modifying .travis.yml (I wanted to rebase anyway so that the AppVeyor test would pass). I think this should be ready for review now.

jgosmann:
The NEF summary notebook is giving me some warnings (Python 3.5.2, Numpy 1.11.1):
```
/home/jgosmann/.virtualenvs/nengodev/lib/python3.5/site-packages/numpy/lib/shape_base.py:873: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  return c.reshape(shape_out)
```
But that might not be critical for the release.

tbekolay:
> The NEF summary notebook is giving me some warnings

Ah nice catch, thanks! This wasn't giving me an error running through `pytest -s` but it did in the notebook. I fixed it by using `//` instead of `/` in `np.tile` (see 5de370084334c4caa7a3d6559cb35582e10f7e25#diff-b6b9cc703d60ed0d0bb71c13093637a5R27).

Issue #1275, opened 2017-03-06
Non-deterministic behaviour despite seed being set.


jgosmann:
Using Nengo 2.3.1 the following will produce different results for `a` and `b` despite fixed seed:
```python
import nengo
import numpy as np

with nengo.Network(1) as model:
    ens = nengo.Ensemble(10, 1)
    inp = nengo.Node(np.sin)
    nengo.Connection(inp, ens)
    p = nengo.Probe(ens.neurons, 'spikes')

with nengo.Simulator(model) as sim:
    sim.run(.01)

a = sim.data[p]

with nengo.Simulator(model) as sim:
    sim.run(.01)

b = sim.data[p]
```

Example output for `a`:
```
array([[    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,  1000.,     0.,     0.,
            0.,  1000.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,  1000.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.]])
```

and `b`:
```
array([[    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,     0.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [    0.,  1000.,     0.,     0.,     0.,     0.,     0.,     0.,
            0.,     0.],
       [ 1000.,     0.,     0.,  1000.,     0.,     0.,     0.,     0.,
            0.,     0.],
[...]

PR #1277, opened 2017-03-08
Raise exception when passing seed as label.

jgosmann:
**Motivation and context:**
This helps issues like #1275 where one accidentally sets the label instead of the seed. @tcstewar mentioned that he made that mistake also twice in a weekend, so it seems to be not super uncommon and I don't see a good reason for not enforcing the proper types. If someone has an integer value and wants to set it as the label, it is still possible to manually convert it to a string which also makes the intend explicit.

none

**How has this been tested?**

Added a test.

**Types of changes:**

**Checklist:**


jgosmann:
There were even two tests where the seed was passed as label! 😲

PR #1278, opened 2017-03-08
Define pylint rules, fix pylint issues

jgosmann:
**Motivation and context:**
`pylint` can provide a number of useful warnings about problems in Python code. We already use it to check for missed super-constructor calls, but we should use it to check other things to. However, it also has a number of less useful warnings or disagrees with some of our code style, so it requires some configuration to be useful.

In this PR, I first enabled all warnings and then selectively disable warnings, locally disable warnings or fix the issue in the code. With some warnings it is not obvious if we want to keep them or not, so we can discuss this here.

List of questions (more might follow):
1. `unused-argument`: Do we want a warning about unused function arguments? There are bugs where one forgets to actually use a function argument, but a lot of times it is correctly unused (e.g. in interface classes) which requires to add disable comments to the affected functions. I have a slight tendency towards disabling the check.
2. `redefined-outer-name`: Sometimes we use variable names in functions (or function argument names) that are the same as a global name. This is somewhat bad practice as it makes the global name unaccessible and might cause confusion what is referred to. But sometimes this requires to change one of the the two variable names to something which is less intuitive or awkward to type and might not clear up the confusion. I have a slight tendencey towards disabling this check.
3. `no-self-use`: pylint warns when a instance method does not use `self` and could be written as `@classmethod` or `@staticmethod`. But for interfaces etc. this warning is sometimes wrong because the method is intended to be overwritten. I have a slight tendency to leave this check enabled and use comments to disable it locally where appropriate.
4. `redefined-variable-type` warns about when the type of a variable changes. This seems to me like a really good warning to have, but unfortunately all instances of it that I checked where false positives. pylint seems to get really easily confused if different classes from the same inheritance hierarchy are used in different branches of if-statements (supposedly a fixed issue, but apparently it is not). Because of that, I would vote for disabling the warning.
5. `too-many-branches` is a warning about the complexity of functions. I'm undecided on this one. It is good to provide some push for keeping functions simple enough, but sometimes nothing is won by splitting it up to get rid of such a warning. Maybe this decision when a function gets to complex is better left to the author and reviewers of the code.
6. `unused-imports` and maximum line length are checked by both pylint and pep8. Should both tools check it or just one? If both check it, incorrect warnings for individual lines have to be disabled for both tools, so I would vote for only one tool. I prefer pylint, because the line based disabling is more human readable (`pylint: disable=unused-import` vs. `noqa: F401`) and pylint supports a regex for lines that are allowed to exceed the maximum length (then long URL do not need to be prefixed with noqa).
7. `unexpected-keyword-arg`: Has some false positives in some tests because the `add_to_container` argument gets handled in the `__new__` of a metaclass. I'm leaning towards locally disabling the warning for these false positives.
8. `undefined-loop-variable` seems useful because it prevents usage of a loop variable that might be unitialized because no iterations of the loop where executed, but there is a number of false positives in our code where the code path after the for is only executed when an iteration has been executed or the number of iterations is known in advance. The warning is easily fixed by assigning a default value before the loop and in some cases this is the right thing to do, but in others it's just a bit of added noise and does not make the code any more correct.

Please comment if you have any opinions on those points.

none

**How has this been tested?**

todo

todo

todo

**Types of changes:**

todo

**Checklist:**

**Still to do:**

#### Discuss

- [ ] `redefined-builtin`

#### Separate PRs

- [x] `bad-super-call`
- [x] `dangerous-default-value`
- [x] `cyclic-import`
- [x] `deprecated-method`
- [x] `function-redefined`
- [ ] `unidiomatic-typecheck`
- [ ] `arguments-differ`
- [x] `cell-var-from-loop`
- [x] `signature-differs`
- [ ] `too-many-return-statements`
- [ ] `locally-disabled`
- [ ] `too-many-statements`
- [ ] `fixme`
- [ ] `duplicate-code`

#### Separate PRs cherry-picked from this PR
- [x] `anomalous-backslash-in-string`
- [x] `unused-import`
- [x] `bare-except`
- [x] `ungrouped-imports`
- [x] `import-error`
- [x] `reimported`
- [x] `wrong-import-order`
- [x] `bad-mcs-classmethod-argument`
- [x] `unnecessary-lambda`
- [x] `bad-whitespace`

#### This PR

- [x] `expression-not-assigned`
- [x] `nonexistent-operator`
- [x] `global-statement`
- [x] `undefined-loop-variable`
- [x] `undefined-variable`
- [x] `attribute-defined-outside-init`
- [x] `superfluous-parens`
- [x] `unused-variable`
- [x] `bad-continuation`
- [x] `no-member`
- [x] `pointless-statement`
- [x] `invalid-name`
- [ ] Get the number of pylint warnings to 0.

#### Single warning false positives
- [ ] `invalid-sequence-index`
- [ ] `not-callable`

#### Disable list
- [x] `eval-used`
- [x] `exec-used`
- [ ] `unused-argument`
- [ ] `redefined-outer-name`
- [ ] `no-self-use`
- [ ] `redefined-variable-type`
- [ ] `unexpected-keyword-arg`
- [ ] `undefined-loop-variable`

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -5,7 +5,7 @@ extension-pkg-whitelist = numpy
 [MESSAGES CONTROL]
 
 #disable = all
-disable = broad-except, missing-docstring, no-self-use, protected-access, redefined-outer-name, redefined-variable-type, redundant-keyword-arg, too-few-public-methods, too-many-ancestors, too-many-arguments, too-many-instance-attributes, too-many-branches, too-many-locals, unused-argument
+disable = arguments-differ, broad-except, deprecated-method, missing-docstring, no-self-use, protected-access, redefined-builtin, redefined-outer-name, redefined-variable-type, redundant-keyword-arg, too-few-public-methods, too-many-ancestors, too-many-arguments, too-many-instance-attributes, too-many-branches, too-many-locals, unidiomatic-typecheck, unused-argument
 enable = W0231
 
 
--- a/.pylintrc
+++ b/.pylintrc
@@ -11,7 +11,7 @@ enable = W0231
 
 [REPORTS]
 
-reports = no
+reports = yes
 
 
 [BASIC]
--- a/.pylintrc
+++ b/.pylintrc
@@ -5,7 +5,7 @@ extension-pkg-whitelist = numpy
 [MESSAGES CONTROL]
 
 #disable = all
-disable = arguments-differ, broad-except, deprecated-method, missing-docstring, no-self-use, pointless-statement, protected-access, redefined-builtin, redefined-outer-name, redefined-variable-type, redundant-keyword-arg, too-few-public-methods, too-many-ancestors, too-many-arguments, too-many-instance-attributes, too-many-branches, too-many-locals, unidiomatic-typecheck, unused-argument
+disable = arguments-differ, broad-except, deprecated-method, missing-docstring, no-member, no-self-use, pointless-statement, protected-access, redefined-builtin, redefined-outer-name, redefined-variable-type, redundant-keyword-arg, too-few-public-methods, too-many-ancestors, too-many-arguments, too-many-instance-attributes, too-many-branches, too-many-locals, unidiomatic-typecheck, unused-argument
 enable = W0231
 
 
--- a/.pylintrc
+++ b/.pylintrc
@@ -17,11 +17,13 @@ reports = yes
 [BASIC]
 good-names = E
 
-argument-rgx = ([A-Z]{1,2})|([a-z_][a-z0-9_]{0,30})$
-attr-rgx = [a-z_][a-z0-9_]{0,30}$
+argument-rgx = ([a-z_][a-z0-9_]{0,30})|([A-Z][a-zA-Z0-9]*(_[a-z0-9_]+)?)$
+attr-rgx = ([a-z_][a-z0-9_]{0,30})|([A-Z][a-zA-Z0-9]*(_[a-z0-9_]+)?)$
 class-attribute-rgx = ([A-Za-z_][A-Za-z0-9_]{0,30}|(__.*__))$
+class-rgx = [A-Z_][a-zA-Z0-9]*$
 const-rgx = (([a-zA-Z_][a-zA-Z0-9_]*)|(__.*__))$
-function-rgx = [a-zA-Z_][a-zA-Z0-9_]{2,30}$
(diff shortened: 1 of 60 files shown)


tbekolay:
Before you sink much more time in this, could you hold off until we discuss it at the dev meeting? I would consider adding all these pylint tests to be a relatively big job that touches a lot of code, which will make PR rebasing etc difficult down the line... It's also a lot of microdecisions about style and preference, and I would prefer to make most of these decisions.

jgosmann:
I'm not adding all of the tests. I'm looking at each one and then decide whether I think it should be enabled or disabled or needs further discussion (see list above). I also think we can have a more informed discussion if at least one person knows what changes each pylint warning would require.

I don't think that a lot of code (relative to the size fo the code base) has been touched so far because most things where small one line changes. But it's definitely a lot of files that have been touched.

jgosmann:
Dev meeting decisions about the questions above:
1–4: disable
5: enable
6: Done in #1289
7, 8: disable

jgosmann:
I think we need to rediscuss point 5 (`too-many-branches`). The idea was to replace flake8's McCabe complexity check (C901).  However, `too-many-branches` (and similar checks like `too-many-statements`) are a much simple measure. The `too-many-branches` check only finds one of the C901 warnings. List of options (some in-between things by just using a subset of pylint's complexity check would also be possible, focussing on the extremes here):
1. Keep flake8's C901 check, don't use pylint's complexity checks (`too-many-branches`, `too-many-statements`, `too-many-return-statements`)
2. Keep flake8's C901 check and enable pylint's complexity checks (they find a few things not detected by the McCabe complexity measure, but not sure about overall the helpfulness)
3. Disable flake8's C901 check and switch to pylint's complexity checks
4. Disable flake8's C901 check and use the [`pylint-mccabe` plugin](https://pypi.python.org/pypi/pylint-mccabe).

Seanny123:
@jgosmann did we ever discuss `too-many-branches` during a meeting? Also, would it be worth it to extract all the code change under `This PR` into a separate pull request while this one keeps track of all the stuff we're trying to change?

jgosmann:
> Also, would it be worth it to extract all the code change under This PR into a separate pull request while this one keeps track of all the stuff we're trying to change?

This is what I was doing until I got busy with other stuff ...

Seanny123:
From the meeting, we determined that `pylint` is good and we need to configure it so that `max_branches=10`.

jgosmann:
I extracted/newly created indivdiual PRs for pylint checks that I considered useful enough and will close this PR now.

PR #1282, opened 2017-03-27
Fix dangerous default values as reported by pylint.

jgosmann:
**Motivation and context:**
It is easy to forget that assigning lists or dictionaries as default argument is problematic because when the object gets changed, the default value of the function will change.

none

**How has this been tested?**
unit tests still pass

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/utils/tests/test_stdlib.py
+++ b/nengo/utils/tests/test_stdlib.py
@@ -164,8 +164,9 @@ def test_weakkeydict_setdefault(key=C(), value1="v1", value2="v2"):
     assert d[key] is value1
 
 
-def test_weakkeydict_update(in_d={C(): 1, C(): 2, C(): 3}):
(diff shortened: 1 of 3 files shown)


nengo/utils/tests/test_stdlib.py
```diff
@@ -164,8 +164,9 @@ def test_weakkeydict_setdefault(key=C(), value1="v1", value2="v2"):
     assert d[key] is value1
 
 
-def test_weakkeydict_update(in_d={C(): 1, C(): 2, C(): 3}):
+def test_weakkeydict_update():
     """This exercises d.update(), len(d), d.keys(), in d,  d.get(), d[]."""
```
jgosmann: I'm not sure why this was defined as an argument to the function in the first place ...

PR #1286, opened 2017-03-29
Give better error for non-finite values from nodes

### LINKED ISSUE #1178: Certain models may fail in direct mode
I'm trying to run Spaun in direct mode (just for benchmarking, I don't care whether the behaviour is replicated), but I am getting an exception. I didn't look into Spaun itself for the cause of this problem, but I have a minimal example that at least gives the similar exception.

Consider this network:

``` python
    e1 = nengo.Ensemble(50, 1)
    e2 = nengo.Ensemble(50, 1)
    nengo.Connection(e1, e2, function=lambda x: 1. / x)
```

Running this will work fine. Yes, there might be a division by 0 and the output may be way off, but it will run because the network is just producing spikes, filtering them, and multiplying them with some weights. No actual division is performed. But when I switch the first ensemble to direct mode like this:

``` python
e1.neuron_type = nengo.Direct()
```

This exception will be produced when running because an actual divison by 0 is performed which feeds `nan` values to the neuron code:

```
FloatingPointError                        Traceback (most recent call last)
<ipython-input-52-83cea7b88218> in <module>()
      1 with nengo.Simulator(model) as sim:
----> 2     sim.run(1.)

/home/jgosmann/.pyenv/versions/3.5.2/lib/python3.5/site-packages/nengo/simulator.py in run(self, time_in_seconds, progress_bar)
    259         logger.info("Running %s for %f seconds, or %d steps",
    260                     self.model.label, time_in_seconds, steps)
--> 261         self.run_steps(steps, progress_bar=progress_bar)
    262 
    263     def run_steps(self, steps, progress_bar=True):

/home/jgosmann/.pyenv/versions/3.5.2/lib/python3.5/site-packages/nengo/simulator.py in run_steps(self, steps, progress_bar)
    279         with ProgressTracker(steps, progress_bar) as progress:
    280             for i in range(steps):
--> 281                 self.step()
    282                 progress.step()
    283 

/home/jgosmann/.pyenv/versions/3.5.2/lib/python3.5/site-packages/nengo/simulator.py in step(self)
    290         try:
    291             for step_fn in self._steps:
--> 292                 step_fn()
    293         finally:
    294             np.seterr(**old_err)

/home/jgosmann/.pyenv/versions/3.5.2/lib/python3.5/site-packages/nengo/builder/neurons.py in step_simneurons()
     66 
     67         def step_simneurons():
---> 68             self.neurons.step_math(dt, J, output, *states)
     69         return step_simneurons
     70 

/home/jgosmann/.pyenv/versions/3.5.2/lib/python3.5/site-packages/nengo/neurons.py in step_math(self, dt, J, spiked, voltage, refractory_time)
    319         # set v(0) = 1 and solve for t to compute the spike time
    320         t_spike = dt + self.tau_rc * np.log1p(
--> 321             -(voltage[spiked_mask] - 1) / (J[spiked_mask] - 1))
    322 
    323         # set spiked voltages to zero, refractory times to tau_ref, and

FloatingPointError: invalid value encountered in true_divide
```

I think we should handle this more gracefully. But I'm not sure whether to convert all `nan` values in direct mode to some fixed number or provide an option to do that. Also note, that the exception right now is not helpful in providing a pointer what the actual problem is.

### LINKED ISSUE #1280: Useful error messages for representing infinity
If you accidentally create a Node that can output infinity, nengo gives a rather cryptic error message:

```python
import numpy as np
import nengo

model = nengo.Network()
with model:
    def problematic_func(t):
        return np.ones(1)/0
    n = nengo.Node(problematic_func)
    ens = nengo.Ensemble(n_neurons=100, dimensions=1)
    nengo.Connection(n, ens)
```

```

Building finished in 0:00:01.                                                   
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\page.py", line 496, in runner
    self.sim.step()
  File "c:\users\terry\documents\github\nengo\nengo\simulator.py", line 308, in step
    step_fn()
  File "c:\users\terry\documents\github\nengo\nengo\builder\neurons.py", line 68, in step_simneurons
    self.neurons.step_math(dt, J, output, *states)
  File "c:\users\terry\documents\github\nengo\nengo\neurons.py", line 326, in step_math
    -(voltage[spiked_mask] - 1) / (J[spiked_mask] - 1))
FloatingPointError: invalid value encountered in true_divide
```

It would be really nice if it at least indicated which Ensemble was receiving infinity or NaN as an input, as that can help identify where the problem is.

jgosmann:
**Motivation and context:**
This produces a slightly better error message when a node or direct mode ensemble produces non-finite values. By raising a `SimulationError` we make it clear that this is a problem that has to be solved on the user side and not some bug in the simulator (before one would get a FloatingPointError). The error message contains the name of the function (which might be `<lambda>` which is not that helpful, but definitely better than before if it is not a lambda).

I think for nodes it is correct to produce an error. In case of the direct mode ensemble with the function it would be nice if it would still run because it would run with non-direct mode ensembles. But to me it seems there is no logical value to set the result to and by not throwing an error it might seem like it works as intended, but does not. Like the Zen of Python says: “In the face of ambiguity, refuse the temptation to guess.”

I think we talked before about potentially implementing some sort of approximation of the neuron normalization in direct mode. This might solve this problem fully, but is also a much bigger jobs with some open questions.

Fixes #1280. Closes #1178.

none

**How has this been tested?**
added tests

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,11 @@ Release History
 
 **Changed**
 
+- We now raise an error when a ``Node`` or ``Direct`` ensemble
+  produces a non-finite value.
+  (`#1178 <https://github.com/nengo/nengo/issues/1178>`_,
+  `#1280 <https://github.com/nengo/nengo/issues/1280>`_,
+  `#1286 <https://github.com/nengo/nengo/pull/1286>`_)
 - We now enforce that the ``label`` of a network must be a string or ``None``,
   and that the ``seed`` of a network must be an int or ``None``.
   This helps avoid situations where the seed would mistakenly
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -625,9 +625,11 @@ class SimPyFunc(Operator):
             args = (np.copy(x),) if x is not None else ()
             y = fn(t.item(), *args) if t is not None else fn(*args)
             if output is not None:
-                if y is None:  # required since Numpy turns None into NaN
+                # required since Numpy turns None into NaN
+                if y is None or not np.all(np.isfinite(y)):
                     raise SimulationError(
-                        "Function %r returned None" % function_name(self.fn))
+                        "Function %r returned non-finite value" %
+                        function_name(self.fn))
                 try:
                     output[...] = y
                 except ValueError:
--- a/nengo/node.py
+++ b/nengo/node.py
@@ -44,6 +44,9 @@ class OutputParam(Parameter):
             output = npext.array(
                 output, min_dims=1, copy=False, dtype=np.float64)
             self.validate_ndarray(node, output)
+            if not np.all(np.isfinite(output)):
+                raise ValidationError("Output value must be finite.",
+                                      attr=self.name, obj=node)
             node.size_out = output.size
         else:
             raise ValidationError("Invalid node output type %r" %
--- a/nengo/tests/test_neurons.py
+++ b/nengo/tests/test_neurons.py
@@ -2,7 +2,8 @@ import numpy as np
 import pytest
 
 import nengo
-from nengo.neurons import NeuronTypeParam
+from nengo.exceptions import SimulationError
+from nengo.neurons import Direct, NeuronTypeParam
 from nengo.processes import WhiteSignal
 from nengo.solvers import LstsqL2nz
 from nengo.utils.ensemble import tuning_curves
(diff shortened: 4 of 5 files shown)


tbekolay:
Added a changelog entry; will merge once CI passes as long as no one has an issue with the changelog entry.

PR #1287, opened 2017-03-29
Fix contains on nengo.Config.

### LINKED ISSUE #1236: Wrong type in error message
I was doing `'solver' in cfg` when I meant to do `'solver' in cfg[nengo.Connection]` and got the following error exception:
```
/home/jgosmann/.pyenv/versions/3.5.2/lib/python3.5/site-packages/nengo/config.py in __getitem__(self, key)
    342         raise ConfigError(
    343             "Type '%(name)s' is not set up for configuration. Call "
--> 344             "configures('%(name)s') first." % {'name': type(key).__name__})
    345 
    346     def __repr__(self):

ConfigError: Type 'int' is not set up for configuration. Call configures('int') first.
```
Note that it says type 'int' even though I'm testing for a string. But even with the correct type in there, it might be possible to produce a better exception.

jgosmann:
**Motivation and context:**
This fixes the contains function on Config objects so that one can do things like `x in cfg`. Before this could produce cryptic exceptions.

Fixes #1236.

none

**How has this been tested?**
added a test

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/config.py
+++ b/nengo/config.py
@@ -329,6 +329,9 @@ class Config(object):
                               "current context to be '%s' but instead got "
                               "'%s'." % (self, config))
 
+    def __contains__(self, key):
+        raise TypeError("Cannot check if %r is in a config." % (key,))
+
     def __getitem__(self, key):
         # If we have the exact thing, we'll just return it
         if key in self.params:
--- a/nengo/tests/test_config.py
+++ b/nengo/tests/test_config.py
@@ -204,3 +204,12 @@ def test_instance_fallthrough():
     assert config[A].amount == 1
     assert config[inst1].amount == 2
     assert config[inst2].amount == 1
+
+
+def test_contains():
+    class A(object):
+        pass
+
+    cfg = nengo.Config(A)
+    with pytest.raises(TypeError):
+        A in cfg


tbekolay:
I'm a bit confused as to the use case for this... what kind of things would I be checking is contained in a config? For the normal Nengo config, for example, am I checking that `nengo.Ensemble` is in a config, and therefore I can set ensemble options, or am I checking that an ensemble instance is in a config, and therefore that it has been configured in some way? Although we are overloading the `__getitem__` here, I don't really see `config` objects as analogous to dicts... the classes that are configurable feel markedly different from the instances of those classes.

jgosmann:
Indeed, there is no reason to check if something in a config, but if you accidentally do (as happened to me in #1236), you would get a very confusing error.

jgosmann:
Note that with the implementation of `__getitem__` Python provides a default implementation of `__contains__` that goes through the integer indices to check for containment (at least it seems like that is happening).

tbekolay:
Ah, I see. I think I would prefer to raise an exception (`NotImplementedError` maybe? Suggesting the correct form in #1236?) than to return a confusing true or false. What do you think? If contentious we could talk about this at the next dev meeting!

jgosmann:
I think formally the implementation is correct and consistent with the behaviour of `__getitem__`. There might actually be a use case because it allows to check whether a class is set up to be configured with a given `Config` instance. But it is definitely not the most self-explanatory way of checking this, so it might be better to implement this in a different way anyways if needed.

It might also be possible to raise a `TypeError` for unexpected (non-class) types and do the contains check for class types. Maybe we should do that in any case in the `__getitem__` method ([the possiblitiy of raising `TypeError` is mentioned in the Python documentation](https://docs.python.org/2/reference/datamodel.html#object.__getitem__)), but that's somewhat separate as it doesn't affect the problem of the confusing exception for `__contains__` much.

I could also see raising an exception for all `__contains__` calls as you propose. But after looking into the documentation, I think that `TypeError` might be more appropriate.
* [Type error](https://docs.python.org/2/library/exceptions.html#exceptions.TypeError): “Raised when an operation or function is applied to an object of inappropriate type.”
* [NotImplementedError](https://docs.python.org/2/library/exceptions.html#exceptions.NotImplementedError): “In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method.” (`__contains__` is not supposed to be implemented by a derived class.)

Bottom line: None of the solutions appears obviously better than the others.

tbekolay:
Sure, `TypeError` sounds good to me. Regardless, feels like it's worth chatting about this in a higher bandwidth medium, so talking about this is on the dev meeting agenda for next week.

jgosmann:
Changed the PR to produce a TypeError instead as discussed at the dev meeting.

jgosmann:
Technically it's a bugfix? idk, I'm probably fine either way.

PR #1288, opened 2017-03-30
Add function to inhibit complete networks.

jgosmann:
**Motivation and context:**
I commonly want to inhibit complete networks (especially ensemble arrays and SPA states). This function simplifies that and give a more general solution to the `inhibt` option that was implemented for the associative memory networks, but is removed in PR #1253.

Somewhat related to #1253 which removes the `inhibit` option from associative networks, but no real interaction.

**How has this been tested?**
Added a test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/utils/network.py
+++ b/nengo/utils/network.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 import nengo
 from .magic import decorator
 
@@ -61,3 +63,28 @@ def activate_direct_mode(network):
     for e in network.all_ensembles:
         if e not in requires_neurons:
             e.neuron_type = nengo.Direct()
+
+
+def inhibit_net(pre, post, strength=2., **kwargs):
+    """Makes an inhibitory connection to all ensembles of a network.
+
+    Parameters
+    ----------
+    pre : 1-d nengo.Node or nengo.Ensemble
+        Scalar source of the inhibition.
+    post : nengo.Network
+        Target of the inhibition.
+    strength : float
+        Strength of the inhibition.
+    kwargs : dict
+        Additional keyword arguments for the created connections.
+
+    Returns
+    -------
+    list
+        Created connections.
+    """
+
+    return [nengo.Connection(
+        pre, e.neurons, transform=-strength * np.ones((e.n_neurons, 1)),
+        **kwargs) for e in post.all_ensembles]
--- a/nengo/utils/network.py
+++ b/nengo/utils/network.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 import numpy as np
 
 import nengo
--- a/nengo/utils/tests/test_network.py
+++ b/nengo/utils/tests/test_network.py
@@ -1,7 +1,9 @@
+import numpy as np
+from numpy.testing import assert_allclose
 import pytest
 
 import nengo
-from nengo.utils.network import activate_direct_mode
+from nengo.utils.network import activate_direct_mode, inhibit_net
 
 
 def test_withself():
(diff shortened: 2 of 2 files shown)


Seanny123:
All the Python2 tests are failing for weird reasons. They can't seem to find `np` properly? However, let me know when you've fixed that and I'll review it.

jgosmann:
The Python 2.7 problem should be fixed now.

tbekolay:
Hm... this is essentially a one-line function and 40-ish lines of documentation and tests, which feels off to me. It's also different from how we have done inhibition of networks in the past, as the function takes in a `pre` rather than creating a `Node` that can be connected to afterward. Not that one way is better than the other, but it's evidence that the function is not something that all Nengo users will want to do in the same way, so it feels out of place in Nengo core.

What do you think about making this an example in [`nengo_examples`](https://github.com/nengo/nengo_examples) instead? It could also live in nengo_extras, but it feels more like an example to me.

jgosmann:
> What do you think about making this an example in nengo_examples instead? It could also live in nengo_extras, but it feels more like an example to me.

I use such a function in pretty much every project, so it would be really nice to have something that is pip installable and importable. So an example wouldn't do it for me. Maybe nengo_extras, but this function feels more fundamental to me than `activate_direct_mode` (#1168) which is in core. Potentially nengo_spa because I mostly use this for SPA things where pretty much any basic unit is an ensemble array or some other sort of network, but it can be used for non-SPA things.

One thing I like about this function is that it immediately makes clear that something is an inhibitory connection without the need to parse a more complicated statement to find out that it as connection to neurons with a transform producing inhibition.

> It's also different from how we have done inhibition of networks in the past, as the function takes in a pre rather than creating a Node that can be connected to afterward.

It would be possible to return such a node from the function instead.

tbekolay:
> this function feels more fundamental to me than activate_direct_mode (#1168) which is in core

In this particular case, my opinion comes not from how fundamental it is, but how obvious the implementation is. #1168 is conceptually simple ("make this network use direct mode") but the actual implementation is tricky because you need to take into account neuron-to-neuron connections and connections with learning rules. Here, the concept is simple ("inhibit all the ensembles in the network") and the implementation is similarly simple; the only thing that a user might have to learn is that `all_ensembles` exists to iterate through ensembles in subnetworks (and that's helpful for them to know regardless).

In any case, this feels like a consensus decision, so if anyone has a preference among the options, register your thoughts in emoji reaction form:

- :+1: Nengo core
- :tada: Nengo example
- :heart: Nengo extras

If no clear consensus, we'll hash it out at the dev meeting.

jgosmann:
> but how obvious the implementation is

The frequency of how often I want such a function justifies it to me despite the obvious implementation. I suppose in a way it is a form of syntactic sugar. For example, it is also pretty obvious how to create a list with a for-loop instead of a list comprehension. Nevertheless, I assume that most people wouldn't want to remove list comprehensions from Python.

As for the vote: It would be nice to be able to do a ranked order ... I voted Nengo core, but would prefer Nengo extras over Nengo examples.

xchoo:
As a note, I do think such a function may be useful within Nengo core. I myself have a function of similar functionality within the Spaun codebase: https://github.com/xchoo/spaun2.0/blob/master/_spaun/configurator.py#L151

However, the implementation is different. Instead of requiring a pre and post to connect, it makes a node (named `inhibit`) that you can connect to.

jgosmann:
So given the discussion in #1301, is everyone ok to move this functionality to nengo_spa?

tbekolay:
> So given the discussion in #1301, is everyone ok to move this functionality to nengo_spa?

Sounds good to me. I think this could also go into `nengo_extras`, but for the time being no downside to putting it in nengo_spa.

PR #1289, opened 2017-03-31
Use pylint to check for unused imports.

jgosmann:
**Motivation and context:**
Instead of flake8 because flake8 misses some unused imports and
the by line disabling of the warning is less cryptic for pylint.
(Who knows what the error code F401 stands for from the top of their
head? Who knows the meaning of noqa? Apparently, it stands for “no
quality insurance”.)

none

**How has this been tested?**
pylint produced not warnings, tests still pass

**Types of changes:**

**Checklist:**

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -5,6 +5,7 @@ enable = anomalous-backslash-in-string,
          bad-super-call,
          dangerous-default-value,
          super-init-not-called,
+         unused-import,
 
 
 [REPORTS]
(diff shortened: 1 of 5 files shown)


jgosmann:
> I'd like to avoid # pylint: disable lines except for those that act as a TODO for things we should clean up eventually.

Would it make sense to mark these with TODO or FIXME in addition? To differentiate those from false positives that should not be cleaned up (and where the check is useful enough to not completely disable ti)?

More a general question, for this PR I'm fine with the `assert` approach.

tbekolay:
> Would it make sense to mark these with TODO or FIXME in addition?

Yeah, that sounds good. I wouldn't go back and change any `# noqa`s that are current in the code, though.

PR #1290, opened 2017-04-03
Fix bare excepts reported by pylint.

jgosmann:
**Motivation and context:**
A bare except also catches exceptions like SystemExit and KeyboardInterrupt
that should only handled in the outer most exception handler if at all.
Otherwise Ctrl-C might not work to exit a Python program as expected if it
happened to be pressed while in a bare exception handler. In most places,
I changed it to catch Exception which is pretty general, but seems to be ok
to me in those cases where an exception should never lead to Nengo exiting.
Within the cache I added logging statements to provide a bit more debugging
information because those exception can indicate problems, but are not
a failure we cannot recover from (and we should recover from it).

I explicitly disabled the broad-except warning because we have quite a number
where it seems acceptable to catch broadly. I hope that disallowing broad
excepts and thus forcing to specify exception types, everyone will think about
whether `Exception` is appropriate or a more narrow scope would be better.

none

**How has this been tested?**
tests still pass

**Types of changes:**

**Checklist:**

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,8 +1,9 @@
 [MESSAGES CONTROL]
 
-disable = all
+disable = all, broad-except
 enable = anomalous-backslash-in-string,
          bad-super-call,
+         bare-except,
          dangerous-default-value,
          super-init-not-called,
          unused-import,
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -300,7 +300,9 @@ class CacheIndex(object):
             with open(self.legacy_path, 'r') as lf:
                 text = lf.read()
             return tuple(int(x.strip()) for x in text.split('.'))
-        except:
+        except Exception:
+            logger.exception(
+                "Decoder cache version information could not be read.")
             return (-1, -1)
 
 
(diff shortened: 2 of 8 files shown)


tbekolay:
The appveyor test is still failing in a very odd manner... it fails in a `try` where the `except` appears to catch the right error. I'll try to reproduce this and debug in a Window machine tomorrow morning.

jgosmann:
The `ValueError` in `get_terminal_size` is [caught in Python 3.6](https://github.com/python/cpython/blob/3.6/Lib/shutil.py#L1082), but [not in 3.4](https://github.com/python/cpython/blob/3.4/Lib/shutil.py#L1072).

jgosmann:
I fixed it. The `terminal_size` constructor differs depending on whether Python provides it or we define it ourselves (as named tuple). Maybe that should be fixed? (but it will make our `terminal_size` definition more verbose)

jgosmann:
@tbekolay, this has been lying around for a while too. Maybe also a candidate to be included in the release?

jgosmann:
There's a bunch of other “pylint” PRs, but this one is not just stylistic, but fixes potential bugs (even though very specific and not very critical ones).

PR #1291, opened 2017-04-04
Fixes of pylint warnings related to imports

jgosmann:
**Motivation and context:**
Today's pylint PR: I'm handling a few more pylint warning types than just one in this PR because they are all related to imports. Each commit handles one specific warning type and I'm willing to discuss the decisions made in each individual commit. Most of the commits explain my reasoning in the commit message, except:

* `ungrouped-imports`/ e1b8e92f69b934cf9752c0955c4acac2893b0a36: In the `ipython.py` file this is hard to fix (and probably makes things less readable) because depending on the IPython version a number of different imports is required. Apart from that it seems ok to enable because there weren't really any false positives, but I don't feel strongly about it.
* `wrong-import-order`/ 7af4a5454bf55d2f6b469b5d111086f644c6aa1c: I don't feel strongly about this, it's mostly a stylistic issue. Again no false positives.

none

**How has this been tested?**
unit tests still pass

I'd recommend looking at each commit individually.

**Types of changes:**

Technically it is a breaking change because to fix cyclic imports I had to move two things around. They seem like things that are probably not used by any third party code, so this might be fine, but maybe we should put in fallbacks with deprecation warnings or raise an exceptions with information where things have moved. Or decide to not make those changes and accept the cyclic imports instead.

**Checklist:**


tbekolay:
In reviewing this, I rebased this branch to master and added a commit. However, I'm seeing weird behavior... as it is right now, there are no warnings raised. However, if I enable the `reimported` check (move it from the `disable` list to the `enable` list) then I get the following:

```
************* Module nengo
C: 16, 0: standard import "import logging" should be placed before "from .version import version as __version__" (wrong-import-order)
************* Module nengo.processes
C:  2, 0: standard import "import warnings" should be placed before "import numpy as np" (wrong-import-order)
************* Module nengo.ipynb
C: 27, 4: Imports from package IPython are not grouped (ungrouped-imports)
************* Module nengo.utils.connection
W:  7, 0: Reimport 'numpy' (imported line 5) (reimported)
************* Module nengo.utils.neurons
W:  6, 0: Reimport 'numpy' (imported line 4) (reimported)
************* Module nengo.utils.functions
C:  6, 0: standard import "import warnings" should be placed before "import numpy as np" (wrong-import-order)
************* Module nengo.utils.ensemble
W:  5, 0: Reimport 'numpy' (imported line 3) (reimported)
************* Module nengo.spa.tests.test_assoc_mem
C:  3, 0: Imports from package nengo are not grouped (ungrouped-imports)
************* Module nengo.tests.test_processes
C:  5, 0: standard import "import sys" should be placed before "import numpy as np" (wrong-import-order)
```

Ignoring the reimported warnings, there are now several instances of `wrong-import-order` and `ungrouped-imports` that somehow show up. I can't figure out why this is happening; the only thing I notice is that the other warning codes start with a `C` whereas `reimported` starts with `W`. No idea what functional effect that has though... any idea what's going on @jgosmann ?

jgosmann:
No idea, might be a bug. Maybe related to PyCQA/pylint#1702.

jgosmann:
With pylint 1.8.1 the import grouping error show up without the `reimported` check. I fixed them.

Seanny123:
FYI, TravisCI is still failing due to some `flake8` import errors.

PR #1292, opened 2017-04-07
Replace deprecated functions.

jgosmann:
**Motivation and context:**
Deprecated functions might get removed in future Python versions. We should avoid there usage. This PR activates the corresponding pylint warning and uses the newer replacements for deprecated functions where possible.

none, I believe

**How has this been tested?**
tests still pass

**Types of changes:**

**Checklist:**


Seanny123:
The changes look good to me, but cause `flake8` errors, because the comments make the lines too long. So I added a quick commit to fix that. Let me know if you think it's acceptable to move those `pylint-disable` comments.

nengo/ipynb.py
```diff
@@ -14,7 +14,8 @@
     from html import escape
 except ImportError:
     from cgi import escape as cgi_escape
-    escape = lambda s, quote=True: cgi_escape(s, quote=quote)  # pylint: disable=deprecated-method
+    # pylint: disable=deprecated-method
```
jgosmann: This might disable the warning for the whole file, but I have to double check.

jgosmann:
`pylint` also has a line length check that supports a regex for ignoring certain lines. That way it is for example possible to ignore overlong URLs without a `noqa` in front. I would enable the pylint check and disable the flake8 check.

This is not to say that it might not be better to fix the overlong lines in this PR. Those are somewhat ugly, but fixing it also changes the scoping of the disabling (the check could be reenabled after the line, but then you get two lines of clutter which is also annoying).

nengo/ipynb.py
jgosmann: It just applies to the block, so I think this is ok.

jgosmann:
Double checking with the [documentation](https://docs.pylint.org/en/latest/user_guide/message-control.html), I think the amendments are fine. :)

Seanny123:
I fixed a `pylint` error that I missed in my last commit. Now the static check is still failing on TravisCI, but for reasons I've already resolved in #1298 which will be resolved when this is rebased.

jgosmann:
I think that the failure might be related to this PR. The AppVeyor failure is unrelated, though.

PR #1304, opened 2017-04-20
Add summaries to API docs

tbekolay:
**Motivation and context:**
@celiasmith pointed out that the API docs pages are long and hard to navigate. This PR adds summaries to `frontend_api` and `networks` to (hopefully) make them easier to navigate.

When built, `networks` looks like this:

netsummary

And `frontend_api` looks like this:

api

That's only the amount that would fit on the screen, it goes on for a while more.

**How has this been tested?**
Built the docs to see how it renders.

**Types of changes:**

**Checklist:**

**Still to do:**

One thing we should discuss is whether the `frontend_api` summary should contain _all_ things that are listed on that page, or if we should omit things that users aren't likely to interact with very much, like base classes and `LearningRule`, Thoughts?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,6 +39,10 @@ Release History
   To get the number of neurons in each ensemble,
   use ``EnsembleArray.n_neurons_per_ensemble``.
   (`#1186 <https://github.com/nengo/nengo/pull/1186>`_)
+- The `Nengo modelling API document
+  <http://pythonhosted.org/nengo/frontend_api.html>`_
+  now has summaries to help navigate the page.
+  (`#1304 <https://github.com/nengo/nengo/pull/1304>`_)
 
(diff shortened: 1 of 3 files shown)


jgosmann:
Is it possible to exclude all the distributions from that table and have a separate table only for distributions?

tbekolay:
> Is it possible to exclude all the distributions from that table and have a separate table only for distributions?

Yep, all is possible. I can split it up so there's a summary per heading too, that might make sense (since the headings are shown in the left sidebar and are clickable to get to subgroups easily).

jgosmann:
> Yep, all is possible.

That would be great!

tbekolay:
Pushed a fixup separating the frontend_api summaries into separate sections; here's the distributions section (the others look similar):

dists

This seems like an improvement to me :+1:

PR #1306, opened 2017-04-22
Fixed SPA examples to stop using old modules

### LINKED ISSUE #1303: SPA examples in core still use Buffer instead of State
Even though the SPA implementation in core will be deprecated at some point, we should update the examples to use `spa.State` and not the old `spa.Buffer`.

Seanny123:
**Motivation and context:**
Updated the SPA examples used in the modules. Resolves #1303. I tested this by running all the notebooks.

**Types of changes:**

**Checklist:**


Seanny123:
Do I need to clear the kernel information somehow? Was that in the `nengo-devscripts`? I've read the previous conversations about this, but I don't remember what the conclusion was.

Seanny123:
@jgosmann this pull request is failing because I edited the examples under the assumption that they would be using the same standard as the new networks. This is the case in the `nengo_spa` repository. Consequently, should we change the spa implementation in this repository? Or does that not matter since we're going to [remove SPA anyways in 3.0](https://forum.nengo.ai/t/whats-planned-for-nengo-3-0/270?u=seanny123)?

jgosmann:
> Was that in the nengo-devscripts?

Yeah, `clear-outputs.py`.

> Consequently, should we change the spa implementation in this repository?

I suppose we could add the new input names in addition to the old ones ...

Seanny123:
I'm confused about what's happening to the future of SPA:
- is SPA being removed entirely from `nengo/nengo`?
- are we leaving the current implementation of SPA and only upgrading the
`nengo/nengo_spa`?

tcstewar:
> I'm confused about what's happening to the future of SPA:
> - is SPA being removed entirely from `nengo/nengo`?
> - are we leaving the current implementation of SPA and only upgrading the
> `nengo/nengo_spa`?

My understanding is that spa will stay in nengo/nengo for at least the next long while.  There is a possibility of removing it, but only when we have a full plugin system working for nengo_gui so that it's possible to separately install nengo_spa and have all the tutorials and graphs show up in the gui correctly.  But we have no idea how long that'll be, so in the meantime we can apply bug-fixes and cleanup to the nengo/nengo/spa.  But we shouldn't be adding new features to it.

jgosmann:
Is this still WIP?

Seanny123:
Yeah. I'm on vacation this week, but I'll fix it Monday.

Seanny123:
Yeah, I made the wrong changes and I'm only going to get access to my
laptop on Friday to fix everything again.

Seanny123:
And by "Monday", I meant Saturday. But I think I fixed it now? The diff looks ridiculous, but I removed the output with the devscript and the notebook cleanup didn't seem to do anything? Feedback appreciated.

Issue #427, opened 2014-06-23
Allow learning_rate to be a function of time


studywolf:
It would be nice to be able to pass in a function to the learning_rate parameter, so I can have explicit control of it as a function of time or some other variable in my network. e.g.

```
adapt = nengo.Ensemble(100, 1)
system = nengo.Ensemble(100, 1)
answer = nengo.Node(output=training_func)

error = nengo.Ensemble(1, 1)
nengo.Connection(system, error, transform=-1)
nengo.Connection(answer, error)

nengo.Connection(adapt, system, 
                               learning_rule=nengo.PES(error, learning_rate=training_func))
```

where I could define `training_func` as something like

```
def training_func(t):
    return np.exp(-t)
```

or some piecewise thing denoting training epochs or whatever.

As I'm typing this out it occurs to me you could also probably do something equivalent by just scaling the error inside an error node, especially if you want to make it a function of some variable in the network. So maybe that doesn't make sense, but for having learning rates as a function of time this way would probably be more intuitive to program.

Thoughts?

tbekolay:
We discussed this at the lab meeting and the consensus was that this is a good idea ;) And yes, is already possible in `nengo.Voja`, so it would be good to make it possible for all learning rules in a clear way.

We were all agreed that connecting to the learning rule will not work for all rules, as PES already allows you to connect to the learning rule, and it's also not clear when you connect to the learning rule that you're controlling the learning rate. So we want a more obvious solution, likely one that involve modifying `connection.learning_rule.learning_rate` or `connection.learning_rate`.

There wasn't an obvious way to do this, so we're going to use this issue to brainstorm ideas. I'll post a few that come to mind; please add your own, or add thoughts on the ideas I've put forth!

### Option 1: `LearningRuleType.learning_rate` accepts a number or a function of time

In this option, we modify `LearningRuleType` so that the `learning_rate` parameter can accept either a number (as is done now) or a function of time. The builder would have to be modified to run that function during the simulation.

Pros:

- Minimal changes to the frontend.
- Relatively easy to implement.

Cons:

- Cannot accept dynamic input from the simulation (i.e., from a node in the simulation).
- Difficult for some backends to support.

### Option 2: add a `learning_rate` object

[...]

PR #1307, opened 2017-05-03
Allow multiple error signals of arbitrary size for learning rules

jgosmann:
**Motivation and context:**
I'm currently working on a new learning rule that I call the Association Matrix Learning rule (AML). Essentially it learns associations from vectors u_i to v_i by learning an outer product transform matrix \sum_i v_i u_i^T. I am experimenting with different ways of implementing this learning rule (corresponding to different biological interpretations). In the two most general formulations, I end up either with a single d*d error signal or with two d-dimensional error signals. Currently, Nengo does not allow error signals of arbitrary size (d * d) or multiple error signals (note, that I cannot use slices into a single error signal because of the size limitations). To test these formulations of the learning rule, I extended Nengo to support multiple error signals of arbitrary size for learning rules.

There a still a few things to clean up and do to make this production quality code, but before I invest time in that, it would be good to know:

* Do we in general want the functionality for multiple error signals?
* How feel people about my choice of connection syntax? For learning rules with no or one error signal nothing changes, but if there are more error signal one would connect to named attributes of the learning rule. For example, `conn.learning_rule.error1` and `conn.learning_rule.error2` or `conn.learning_rule.cue` and `conn.learning_rule.target` (the names I chose for the AML). I think we had some related discussion in #553 and there was some concern that this is not consistent with how Nengo works in other cases. Though, I disagree because to me it is very similar to `ensemble.neurons`.
* Do we want to keep backwards compatibility for this? (It might make sense to disallow connections to `conn.learning_rule` and always require to use the name of the error signal.)

This is currently based on #1207 because it made the implementation easier and faster for me, but technically the code could be changed to not depend on that.

**How has this been tested?**

Unit tests still pass without changes (-> backwards compatible). Tested arbitrary size error signal and multiple error signals with my AML implementations that I have somewhere else. If this is polished up to production quality, unit tests need to be added accordingly.

Not ready for review yet. Just looking for feedback if this is a desirable feature.

tbd

**Types of changes:**

tbd

**Checklist:**

**Still to do:**

- [ ] PES is still constructing it's own error signal I believe instead of using the signal constructed in `build_learning_rule`.
- [ ] `params.py` might not be the best place for `FrozenMap`

### DIFF
--- a/nengo/base.py
+++ b/nengo/base.py
@@ -211,7 +211,8 @@ class NengoObjectParam(Parameter):
             NengoObject,
             ObjView,
             nengo.ensemble.Neurons,
-            nengo.connection.LearningRule
+            nengo.connection.LearningRule,
+            nengo.connection.ErrorSignal,
         )
         if not isinstance(nengo_obj, nengo_objects):
             raise ValidationError("'%s' is not a Nengo object" % nengo_obj,
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -365,6 +365,15 @@ def build_learning_rule(model, rule):
 
     conn = rule.connection
 
+    # --- Set up error signals
+    for err_sig in rule.error_signals:
+        error = Signal(
+            np.zeros(err_sig.size_in), name="learning_rule:" + err_sig.name)
+        model.add_op(Reset(error))
+        model.sig[err_sig]['in'] = error  # error connection will attach here
+    if len(rule.error_signals) == 1:
+        model.sig[rule]['in'] = model.sig[next(iter(rule.error_signals))]['in']
+
     # --- Set up delta signal
     if rule.modifies == 'encoders':
         if not conn.is_decoded:
@@ -385,7 +394,7 @@ def build_learning_rule(model, rule):
                 np.zeros((post.n_neurons, pre.n_neurons)), name='Delta')
         else:
             delta = Signal(
-                np.zeros((rule.size_in, pre.n_neurons)), name='Delta')
+                np.zeros((conn.post.size_in, pre.n_neurons)), name='Delta')
     else:
         raise BuildError("Unknown target %r" % rule.modifies)
 
--- a/nengo/config.py
+++ b/nengo/config.py
@@ -15,6 +15,7 @@ http://nbviewer.ipython.org/urls/gist.github.com/ChrisBeaumont/5758381/raw/descr
 import inspect
 import sys
 
+import nengo
 from nengo.exceptions import ConfigError, ValidationError
 from nengo.params import Default, is_param, iter_params
 from nengo.rc import rc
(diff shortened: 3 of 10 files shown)


drasmuss:
> note, that I cannot use slices into a single error signal because of the size limitations

Could you elaborate a bit on that?  You're saying that if you have two `d` dimensional error signals, you cannot create a `2*d` dimensional error signal and slice into that right?  What are the size limitations that prevent that?

jgosmann:
The size of the error signal is controlled via `error_type` which is restricted to:
* `'none'`: no error signal at all
* `'scalar'`: one-dimensional error
* `'decoded'`: decoded dimensionality of the pre-population
* `'neuron'`: not implemented

Furthermore, 10faba5cda19c06e8141b9e3cb285af0746f3ccb did not work for me [because of this line](https://github.com/nengo/nengo/pull/1307/files#diff-d4a1ea7c740725a930af968ec8143735L388). Not sure if it would work with that one line change added.

drasmuss:
Ah I see, you were saying that it isn't possible in the current Nengo, not that it isn't a possible solution.

I'd have to check why `10faba5cda19c06e8141b9e3cb285af0746f3ccb` wasn't working.  It was passing all unit tests at the time, so I suspect it just needs to be updated due to some refactoring.

tbekolay:
Anyone looking at this PR should also give a look at the discussion starting at https://github.com/nengo/nengo/issues/427#issuecomment-255850317 where something similar to this is discussed.

The solution in this PR is something like my Option 2 (Terry's Option 2 or 3). As I mentioned in that discussion, these options do something very similar to what `Neurons` and `LearningRule` do now. So, it seems to me like we have enough potential use cases that we should think a bit more about whether these things can be generalized.

The thing I worry about is that this PR adds `ErrorSignal`, a neuron type will want to add `PhasicDopamine`, `TonicDopamine`, maybe a synapse will want a `ReuptakeInhibitor` signal, etc etc. If we implement each one separately as it comes up, supporting them on multiple backends is going to be very challenging. It's going to be challenging regardless, but if we want to have some hope of doing it, they need to have some things in common.

Is this something we should chat about at the next dev meeting? Or should someone write a concrete NEP that we can discuss?

jgosmann:
> Is this something we should chat about at the next dev meeting?

Potentially, I already put it on the agenda.

> Or should someone write a concrete NEP that we can discuss?

Potentially too, but I'm not sure whether I will be able to do that at the moment. For now I just needed a first pass implementation to test a few things.

jgosmann:
Dev meeting decision:
* Implement arbitray `size_in` now.
* Revisit multiple inputs after builder refactoring.

drasmuss:
I can take care of fixing up 10faba5cda19c06e8141b9e3cb285af0746f3ccb to resolve the problem you identified, I'll do that tomorrow.

PR #1309, opened 2017-05-08
Release file locks on termination signals.

jgosmann:
**Motivation and context:**
If a Nengo process is terminated abnormally by a signal other than SIGINT (Ctrl+C), e.g. with the kill command, file locks acquired by the cache will not be released. That prevents all future Nengo instance from acquiring writing to the cache until the file lock in manually removed. This PR adds code to install signal handlers to release file locks before such an abnormal termination.

While this solves one problem, it may introduce another problem. Modifying signal handlers is a global operation. The PR makes some effort to call previously installed handlers (or the default handler) after releasing the file locks. However, the user could set signal handlers after the file lock handler has been installed and overwrite it unknowningly (but argubly that leaves us off not any worse than without this PR). More critically, the user could change the signal handling to a non terminating signal handler before the file lock handler is installed. In that case, the file lock might be released too early. Though, for most of the signals this would be somewhat bad coding practice as they are intended to terminate the program. SIGUSR1 and SIGUSR2, however, are sometimes used for things other than program termination (e.g. some versions `dd` will output a progress report on `SIGUSR1`). Thus, the file lock handler will only be installed for these two signals if the default handler (that terminates) has not been changed.

There might be an argument to reduce the number of signals that the file lock handler is registered for. SIGABRT, SIGILL, SIGFPE, SIGSEGV should normally not occur and are the indication of quite a severe problem, probably with the Python interpreter. It is not clear to me whether in those case the signal handler can still properly run at all. SIGUSR1 and SIGUSR2 are not necessarily used for termination (see above). The following three signals, however, should be included in my opinion if we install a signal handler at all:
* SIGHUP: Get send when the parent process exits. For example, when the user closes the terminal that is running the Nengo instance.
* SIGTERM: When killing the Nengo process with `kill`, it is the default signal that is sent.
* SIGQUIT: This can be sent with a keyboard short cut (it is different from SIGINT send with Ctrl+C, because by default it will produce a core dump, maybe there is an argument to not install the handler for it to not change the core dump)

Another potential issue is that all signal handler manipulations have to be done in the main thread. In theory, it would be possible to import Nengo not in the main thread, but only in child threads which would make the code fail.

none

**How has this been tested?**
Unit tests still pass, I tested parts of the code overwriting the signal handling manually. It would be nice to have unit tests for it, but not sure how to test this as it seems problematic to globally mess with the signal handles within the unit tests.

Helpful references:
* [signal man page](http://man7.org/linux/man-pages/man7/signal.7.html)
* [Invoking the default signal handler](http://stackoverflow.com/questions/6015498/executing-default-signal-handler)

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/utils/lock.py
+++ b/nengo/utils/lock.py
@@ -38,7 +38,7 @@ def _release_locks_handler(sig, frame):
 # signals will slightly differ depending on the platform (i.e. Windows does
 # not support some signals). We do not include SIGINT which is translated to
 # an exception by Python by default and SIGKILL for which no signal handler
-# can be registered. This also assumes that these signals are not “mis-used”
+# can be registered. This also assumes that these signals are not "mis-used"
 # for things that will not terminate the process (because in that case we
 # did not want to release the file locks). Because SIGUSR1 and SIGUSR2 are
 # commonly used as non-terminating signal, we only register the signal handler
--- a/nengo/utils/lock.py
+++ b/nengo/utils/lock.py
@@ -1,9 +1,65 @@
 import errno
 import os
 import os.path
+import platform
+import signal
 import time
+import weakref
 
 from nengo.exceptions import TimeoutError
+
+
+_locks = weakref.WeakSet()
+
+
+def _prepend_sig_handler(sig, fn):
+    current_handler = signal.getsignal(sig)
+    if current_handler == signal.SIG_DFL:
+        def prepended_handler(sig, frame):
+            fn(sig, frame)
+            h = signal.signal(sig, signal.SIG_DFL)
+            os.kill(os.getpid(), sig)
+            signal.signal(sig, h)
+    elif current_handler is None or current_handler == signal.SIG_IGN:
+        prepended_handler = fn
+    else:
+        def prepended_handler(sig, frame):
+            fn(sig, frame)
+            current_handler(sig, frame)
+    signal.signal(sig, prepended_handler)
+
+
+def _release_locks_handler(sig, frame):
+    for lock in _locks:
+        lock.release()
+
+
+# Build up a list of signals that terminate the process. The list of available
+# signals will slightly differ depending on the platform (i.e. Windows does
+# not support some signals). We do not include SIGINT which is translated to
+# an exception by Python by default and SIGKILL for which no signal handler
+# can be registered. This also assumes that these signals are not “mis-used”
+# for things that will not terminate the process (because in that case we
+# did not want to release the file locks). Because SIGUSR1 and SIGUSR2 are
+# commonly used as non-terminating signal, we only register the signal handler
+# if they still use the default (terminating) handler.
+_termination_signals = [
+    signal.SIGILL,
+    signal.SIGABRT,
+    signal.SIGFPE,
+    signal.SIGSEGV,
+    signal.SIGTERM,
+
+]
+if platform.system() not in ('', 'Windows'):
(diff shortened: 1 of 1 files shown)


jgosmann:
@tbekolay, do you want to take a look at this before the release? Might need more discussion though ...

tbekolay:
I'll definitely look at #1336, but I'm not super comfortable installing signal handlers without knowing a bit more about the implications of it... I can understand doing it as an application, but it feels weird for a library to do it. I'm not super sure where to look for guidance though.

jgosmann:
Yes, I completely agree with you. Definitely feels weird and I'm not sure if it's the right thing to do. It solves one problem, but might introduce other problems. Also, I have no idea where to look for guidance.

tbekolay:
We discussed this at today's dev meeting. A few issues were brought up, some of which have already been discussed here (but I will include for posterity), and include:

1. It feels like a library should not be affecting global interpreter state on import.
2. This is attempting to recover from a pretty tricky situation, and is one that many other programs will also not recover from properly (e.g. Python itself; if it did we wouldn't have to think about it). There's probably good reasons they don't do this.
3. It's a deep rabbit hole to find more and more esoteric OS issues and patch them in Nengo. This might be acceptable if it were in our core routines, but for the cache, the worst case is that a build takes longer than usual, so we should spend less code trying to anticipate potential issues and patching them, and perhaps more in making it easy to wipe the cache and start over.

All in all, I find it difficult to convince myself to go this route. I could see it in Nengo GUI, or some other application that already deals with signal handlers, but not in Nengo core. I'll leave the PR open for another week or so in case there are other points we did not consider, but will close it then if the majority aren't swayed.

jgosmann:
> It feels like a library should not be affecting global interpreter state on import.

I agree, but at the same time I feel that a library also should not affect other global state with side-effects like not releasing a lock on premature termination.


I happen to run into non-deleted lock files quite often on Compute Canada clusters. Sometimes I kill jobs myself, sometimes they get killed by the scheduler for exceeding the time limit. While it is not a huge deal for me to delete that file, it might not make for a good user experience for someone less familiar with the Nengo internals. I don't think I will sway anyone on this particular PR (and I understand the concerns), so I thought about alternative solutions that we should discuss:

1. We could try to detect whether a file lock is from process that does not exist any longer. One way to do so would be to write the PID of the creating process into the lock file and then check whether a process with the ID is still running. However, I am not sure whether there is a portable way to get a process list under Python. So this might require different code for Windows and Linux or a third party library like psutil.
2. Maybe trying to to use a single code base for the file locking is the wrong approach. Unix provides system calls like `fcntl` and `flock` for file locking that should clear the lock automatically when the file descriptor is closed (which the OS should take care off in case of a SIGTERM or SIGKILL). Windows also has system calls for file locking (that I am not yet familiar with). This clearly requires two different implementation for Unix and Windows.
3. Use the [`portalocker`](https://pypi.python.org/pypi/portalocker) module that uses the respective system calls to implement file locking. This would add another dependency, but would also keep us from reinventing the wheel.

tbekolay:
I'm warming up to the idea of using `portalocker`; we originally had a design goal of no dependencies except `numpy`, but I think the Python packaging ecosystem has improved significantly since then, so it might be worth having another discussion about allowing a few dependencies, as long as they're not difficult to install. I'll add it to the next dev meeting agenda!

jgosmann:
We decided to use a vendorized copy of portalocker at the dev meeting.

PR #1311, opened 2017-05-09
Pre post errors

hunse:
**Motivation and context:**

Some learning rules I'm working on have errors in the space of the pre object or post object. This allows for vector error signals with those number of dimensions.

It also adds a feature to combine LinearFilter synapses, and fixes a bit of the synapse documentation.

**How has this been tested?**

I've tested this with my new learning rules, and it does what I need. Also, it's replacing the ``neurons`` option for learning rules, which was unimplemented.

The three commits are all independent, so start wherever you want. Really they could be separate PRs, but they're all so small that it doesn't seem worth it.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -580,8 +580,10 @@ class LearningRule(object):
             return (self.connection.post_obj.ensemble.size_in
                     if isinstance(self.connection.post_obj, Neurons) else
                     self.connection.size_out)
-        elif self.error_type == 'neuron':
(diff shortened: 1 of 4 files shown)


jgosmann:
This conflicts with #1310 which might also cover your use case?

hunse:
These changes are now split between #1312 (first two commits) and #1310 (last commit).

PR #1312, opened 2017-05-10
Linearfilter combine

hunse:
**Motivation and context:**

It can be helpful to combine two LinearFilters, for example to see what the effects of passing through two subsequent synapses would be.

**How has this been tested?**

Added a little unit test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,9 @@ Release History
 - Added documentation on improving performance.
   (`#1119 <https://github.com/nengo/nengo/issues/1119>`_,
   `#1130 <https://github.com/nengo/nengo/pull/1130>`_)
+- Added ``LinearFilter.combine`` method to
+  combine two ``LinearFilter`` instances.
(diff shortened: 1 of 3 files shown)


arvoelke:
FYI if you want to try this out in another way for point of comparison, you can currently do this using the multiplication operator in `nengolib`, i.e.,

```
from nengolib import Lowpass
nengo.Connection(pre, post, ..., synapse=Lowpass(0.1)*Lowpass(0.2))
```

which is also equivalent to using `DoubleExp(0.1, 0.2)`. This can be done in all versions of `nengo>=2.1.0`.

Seanny123:
Any particular reason why this should be in Nengo core vs. extras?

hunse:
It adds to a class that's in Nengo. I'm not exactly sure how it would work
in nengo_extras.

hunse:
Added a changelog entry.

PR #1313, opened 2017-05-10
Solver building

hunse:
**Motivation and context:**

This allows other backends (e.g. nengo_ocl) to more easily extend this backend. The first commit makes solvers something that are built, so that a backend can define custom code for solvers. The second commit allows the Model builder to be changeable; this allows a backend to have it's own Builder subclass and add build functions to this without affecting the core Nengo Builder. The final commit allows other arguments to be passed through the decoder solver caching system without being cached. These are arguments that don't affect the end result of the decoder solver, but are useful to have access to. Here, I use it to pass the connection so that we can have a useful error message; in nengo_ocl, I use it to pass the OpenCL queue so that I can run things in OpenCL (see nengo/nengo_ocl#142).

**How has this been tested?**

Mostly it's covered by the current tests. I added a test for the zero-activities error.

I would go through the commits in order, though they are roughly independent.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/builder.py
+++ b/nengo/builder/builder.py
@@ -63,10 +63,11 @@ class Model(object):
         or for the network builder to determine if it is the top-level network.
     """
 
-    def __init__(self, dt=0.001, label=None, decoder_cache=NoDecoderCache()):
+    def __init__(self, dt=0.001, label=None, decoder_cache=None, builder=None):
         self.dt = dt
         self.label = label
-        self.decoder_cache = decoder_cache
+        self.decoder_cache = (NoDecoderCache() if decoder_cache is None
+                              else decoder_cache)
 
         # Will be filled in by the network builder
         self.toplevel = None
@@ -87,6 +88,7 @@ class Model(object):
         self.time = Signal(np.array(0, dtype=np.float64), name='time')
         self.add_op(TimeUpdate(self.step, self.time))
 
+        self.builder = Builder() if builder is None else builder
         self.build_callback = None
 
     def __str__(self):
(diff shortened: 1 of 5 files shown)


hunse:
> There is no way to guarantee that the solver result does not depend on it.

No, that's up to the user.

> Passing in the connection is also somewhat redundant as `solver` and `neuron_type` could be queried on the connection object which at least would reduce the length of the argument list.

Yes, except we want to cache `solver` and `neuron_type`. We don't want to cache the whole connection. Another good example of this is when I pass the OpenCL queue to the solver in nengo/nengo_ocl#142. This doesn't affect the results of the solver at all, but it's needed to run things on the GPU.

jgosmann:
> Yes, except we want to cache solver and neuron_type. We don't want to cache the whole connection.

There is no need to base the cache key on the whole connection even if it is used as argument to the function. What is used to determine the cache key is [specified here](https://github.com/nengo/nengo/pull/1313/files#diff-43ac82ff393157ab7572f01eda98f1b3R656).

hunse:
That's fair.

I'm open to other suggestions if you'd like to code up something that will work for here and for nengo_ocl. This is just the way that I found to do it.

jgosmann:
I would remove the `solver` and `neuron_type` argument and replace it with a mandatory `conn(ection)` argument. That should work for nengo_ocl, shouldn't it? It might make it a little bit less apparent what is used for the cache key. I'm fine with that if we're mixing arguments that are included and arguments that are not included anyways. I would be interested what other people think.

hunse:
> That should work for nengo_ocl, shouldn't it?

No, because I need to pass the queue as well, and that's not part of the connection.

jgosmann:
> No, because I need to pass the queue as well, and that's not part of the connection.

I wasn't intendending to remove the `uncached_kwargs`.

hunse:
Oh, gotcha. Yeah, that would work fine, then.

jgosmann:
Added a commit to do that.

Seanny123:
This still LGTM.

Issue #948, opened 2016-02-11
How should we deal with advanced indexing?


tbekolay:
Currently, we sort of support NumPy-style [advanced indexing](http://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#advanced-indexing) of objects in the context of connections. I say sort of because there are currently bugs related to it (#947) and it only shows up a few times in our unit tests (e.g., [here](https://github.com/nengo/nengo/blob/master/nengo/tests/test_connection.py#L369)), so we are likely not testing all cases.

There are a few directions we could go with advanced indexing, so it's worth discussing which one we want to take before implementing things.
1. Support all the advanced indexing cases.
   
   In addition to lists of integers, there's also boolean arrays and a few other weird cases. We'd have to implement tests and make sure they work across different backends. This may be possible to do with minimal changes to backends (possibly through a helper function that gives an equivalent network with only basic indexing). However, it might also require lots of changes for some backends; note that in NumPy, for example, advanced indexing always triggers an array copy, rather than constructing a view.
   
   Pros: Shorter / simpler code for certain use cases; less surprise for those experienced with NumPy.
   
   Cons: More complicated backend code; more things we need to teach users about.
2. Explicitly drop advanced indexing support and require only basic indexing.
   
   There shouldn't be loss in the type of models that one can implement here, just a difference in how those models are written. Connections with advanced indexing can be replaced with one or more connections with basic indexing.
   
   Pros: Simpler frontend and backend Nengo code; simpler for other backends to implement.
   
   Cons: Breaks some existing user scripts; disappointing for experienced NumPy users that want to use advanced indexing.
3. Support advanced indexing, but internally transform to basic indexing equivalents.
   
   This keeps the pros of the above approaches while easing the burdens on backends. This could be done either automatically and silently at the frontend (modifying the model as connections with advanced indices are created), or it could be done explicitly through a utility function that we will instruct backend designers to call in their `Simulator.__init__`.
   
   Pros: Pros of above.
   
   Cons: Magical. May confuse users when the mapping from advanced index to basic indices results in slow simulations (due to many additional connections, for example).

hunse:
4. Support some types of advanced indexing and not others.

Specifically, I'm thinking of extending our support for lists or arrays of indices (i.e. by fixing #947), but not allowing boolean arrays. What are the "few other weird cases" you were referring to, @tbekolay? We could either allow or disallow these as we see fit.

[...]

Issue #947, opened 2016-02-11
Multiple identical indexed slices are ignored


tcstewar:
Nengo allows you to make Connections by giving a list of dimension indices

```
Connection(a[[2,3]], b[[1,7]])
```

However, if you have the same index multiple times for the post part of the Connection, all but the last one are ignored:

```
Connection(a[[2,3]], b[[1,1]])
```

This ignores the mapping from dimension 2 to dimension 1, and just does the mapping from dimension 3 to dimension 1.

Here is a complete example:

```
import nengo
model = nengo.Network()
with model:
    stim = nengo.Node([-1, 0.5])
    a = nengo.Ensemble(n_neurons=100, dimensions=1)

    nengo.Connection(stim[[0, 1]], a[[0, 0]])
```

This model only sends the value 0.5 to the Ensemble `a`, and ignores the other dimension.  

I believe the correct behaviour would be to send both dimensions to a (i.e. equivalent to a transform of [[1, 1]]), but another alternative would be to throw an error saying that indexed slicing doesn't support this case.

hunse:
Ooh, yeah, that's tricky. The problem is in SlicedCopy, because it's doing a copy, it will copy the first dimension, and then overwrite it with the second dimension. Maybe an error would be best for now, until we can think of a nice way to get this to work.

One option for implementing it might be to have SlicedCopy zero everything in the target buffer and then do an increment. It might be a minor performance hit, but acceptable I think. We could mitigate the performance hit by only doing this if one of the slices is a list of indices.

hunse:
Looking at this again, this doesn't even work as you would think in Numpy. If you do
```python
a = np.array([-1, 0.5])
b = np.array([0.])
b[[0, 0]] += a[[0, 1]]
print(b)
```
you get ``0.5`` (just the last value), not ``-0.5`` as you would expect if both values had been added in. So just making the Copy increment the target value (as I suggested above) won't solve the problem.

This Numpy behaviour is going to make it difficult to do in Nengo without some special-case code that could really slow things down (because I think we'd have to loop over the indices). If anyone can figure out a fast Numpy expression to do this indexing operation, then post it here and we can go from there.

tcstewar:
The more I think about it, the more I'm tempted to just raise an Exception if someone does this....  I can't think of a situation where I'd really want to do this, and if numpy doesn't even produce the expected behaviour, then I feel justified in just not letting people do it.....  :)

hunse:
[...]

PR #1317, opened 2017-05-25
Fix PES learning with sliced post

tbekolay:
**Motivation and context:**
@celiasmith was having issues doing learning on a connection in which the `post` was an `ObjView`. This was failed when using a weight-based solver because we used the full encoder matrix. This PR adds a test for that case, and fixes the builder to slice the encoder matrix when the post object is an `ObjView`.

One discussion point:

Currently, this slices the `post` by slicing into the `encoders` signal. However, in `builder/connections.py`, we have a `slice_signal` helper function that is required for advanced indexing. I seem to recall that we talked about disallowing advanced indexing for Nengo objects, but I think it is still possible. Are we allowing advanced indexes?

**How has this been tested?**
Added a unit test which fails without this PR.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,6 +39,9 @@ Release History
 
 **Fixed**
 
+- Fixed an issue in which the PES learning rule could not be used
+  on connections to an ``ObjView`` when using a weight solver.
+  (`#1317 <https://github.com/nengo/nengo/pull/1317>`_)
 - The progress bar that can appear when building a large model
   will now appear earlier in the build process.
   (`#1340 <https://github.com/nengo/nengo/pull/1340>`_)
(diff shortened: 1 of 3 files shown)


jgosmann:
[NumPy doc on advanced indexing for reference](https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#advanced-indexing)

jgosmann:
There is already some discussion about advanced indexing in #948. There is also at least one bug related to to post slices with advanced indexing (#947) which might be relevant to this PR which uses the post slice.

It might be good to come to a decision before on advanced indexing before reviewing/merging this PR. Or we can go ahead with the review process (as it fixes at least one use case), but should then ensure that we revisit this code once we properly allow or disallow advanced indexing (because this code fragment is not related to the rest of the indexing code, I fear that it might be forgotton).

tbekolay:
I added #948 to the dev meeting agenda for next week. Let's review this one in isolation for now since it fixes what I think is a bug, and the actual fix is only two lines (the rest is tests).

Issue #1318, opened 2017-06-06
Misleading error when a function returns None


ikajic:
When forgetting to return a value from a function (e.g., omitting `return` or `return None`), a misleading `ValueError` message is returned:
```
ValueError: array must not contain infs or NaNs
```

Minimal working example to reproduce the error:
```python
import nengo

model = nengo.Network()
with model:
    a = nengo.Ensemble(n_neurons=50, dimensions=1)
    b = nengo.Ensemble(n_neurons=50, dimensions=1)
    
    def f(x):
        x

    nengo.Connection(a, b, function=f)
```

Possible better error (or warning) message would be to inform the user that `None` is returned.

tbekolay:
Can confirm this happens! It occurs during the decoder solving process (i.e., the error is raised by Scipy, not Nengo itself). I'll make a PR to raise a better error.

PR #1319, opened 2017-06-06
Raise BuildError if conn function returns None

tbekolay:
**Motivation and context:**
In #1318 @ikajic reported that we don't raise a nice error if someone forgets to return a value from a connection function (which is a common thing to forget, especially when moving from a lambda to a named function). This raises a nice error when that happens.

**How has this been tested?**
Added a test that the error gets raised.

**Types of changes:**

**Checklist:**

**Discussion points**

This error gets raised at build time, not a model creation time. This is because running a connection function might not be something the user expects when defining the model. It already gets run at build time, so this keeps the user's expectations the same. We could do this at model creation time though, if the consensus prefers that.

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -76,7 +76,11 @@ def get_targets(conn, eval_points):
     else:
         targets = np.zeros((len(eval_points), conn.size_mid))
         for i, ep in enumerate(eval_points[:, conn.pre_slice]):
-            targets[i] = conn.function(ep)
+            out = conn.function(ep)
+            if out is None:
+                raise BuildError("Building %s: Connection function returned "
+                                 "None. Cannot solve for decoders." % (conn,))
+            targets[i] = out
 
     return targets
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -55,6 +55,9 @@ Release History
   <http://pythonhosted.org/nengo/frontend_api.html>`_
   now has summaries to help navigate the page.
   (`#1304 <https://github.com/nengo/nengo/pull/1304>`_)
+- The error raised when a ``Connection`` function returns ``None``
+  is now more clear.
+  (`#1319 <https://github.com/nengo/nengo/pull/1319>`_)
 
 2.4.0 (April 18, 2017)
 ======================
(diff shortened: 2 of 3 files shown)


nengo/builder/connection.py
```diff
@@ -76,7 +76,11 @@ def get_targets(conn, eval_points):
     else:
         targets = np.zeros((len(eval_points), conn.size_mid))
         for i, ep in enumerate(eval_points[:, conn.pre_slice]):
-            targets[i] = conn.function(ep)
+            out = conn.function(ep)
+            if out is None:
+                raise BuildError("Building %s: Connection function returned "
```
ikajic: The error message is good! I am just wondering if those memory locations (and other meta-data about Ensembles) is really necessary, e.g.,:
```BuildError: Building <Connection from <Ensemble (unlabeled) at 0x7fe738f619d0> to <Ensemble (unlabeled) at 0x7fe738f61b90> computing 'f'>: Connection function returned None. Cannot solve for decoders.```

I find they obfuscate the message, but if this is standard in error reporting for Nengo, then it's probably better to leave it as it is.

tbekolay: The memory locations go away if the pre and post of the connection have labels. I agree that they're not necessary for most purposes, but they are useful for debugging... might be worth making them something that show up in debug mode only?

ikajic: > The memory locations go away if the pre and post of the connection have labels.

Cool, I didn't know connections can have labels! :)

> might be worth making them something that show up in debug mode only?

Yeah, (if it's not a hassle to implement) I'd say so. The error msg that appears when connections have labels is way clearer to me.  I don't think we used `label` keyword once during the summer school tutorials so I am guessing someone new to Nengo will most likely end up with that long error, which is more useful to devs than regular users.

jgosmann: Personally I prefer to have the memory addresses in the error message for unlabeled objects because it is the only thing maybe allows me to identify the object. Adding a “debug mode” might make the error message to appear simpler, but it will require users to know about the debug mode and it is likely to introduce more complexity to the code base.

tbekolay: I'll make a PR for it so we can discuss it in more depth there.

PR #1320, opened 2017-06-06
Print where affected objects are in the network hierarchy in BuildErrors

jgosmann:
**Motivation and context:**
BuildErrors often reference Nengo objects that caused the build failure. Unfortunately, it is usually hard to identify these objects in the model code if they do not have unique labels and we cannot provide the line number where the objects where defined. (I ran into this problem today while refactoring one of my models.) This PR adds information about the network hierarchy for the objects mentioned in the error message. While these could still all be unlabeled objects, it is much more common to have a network labeled than an other Nengo objects that are usually mentioned in the error message. It also tells you how deep in the hierarchy the object. Thus, I think this can make it significantly easier to identify the offending object.

At the current stage, this is primarily a proof of concept a working example to discuss. If there is consensus, that this is a worthwhile addition, I will think about adding a unit test and other code improvement (suggestions are also welcome).

This is somewhat related to [discussion in](https://github.com/nengo/nengo/pull/1319#discussion_r120403498) #1319 whether to remove the memory addresses from the error messages. While this is technically independent, the added information could be less useful without a unique identifier (because you have to cross-reference the object from the object message to the network hierarchy display).

**How has this been tested?**
So far only manually with this script:
```python
import nengo

with nengo.Network(label='m1') as m1:
    n = nengo.Node(lambda t: t)

with nengo.Network(label='m2') as m2:
    e = nengo.Ensemble(10, 1)
    nengo.Connection(n, e)

with nengo.Simulator(m2) as sim:
    pass
```
The output is:
```
Building finished in 0:00:01.
  File "test.py", line 10, in <module>
    with nengo.Simulator(m2) as sim:
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/simulator.py", line 157, in __init__
    raise err
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/simulator.py", line 154, in __init__
    self.model.build(network, progress_bar=self.progress_bar)
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/builder/builder.py", line 123, in build
    built = self.builder.build(self, obj, *args, **kwargs)
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/builder/builder.py", line 219, in build
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/builder/network.py", line 115, in build_network
    model.build(conn)
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/builder/builder.py", line 123, in build
    built = self.builder.build(self, obj, *args, **kwargs)
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/builder/builder.py", line 219, in build
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/builder/connection.py", line 216, in build_connection
    model.sig[conn]['in'] = get_prepost_signal(is_pre=True)
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/builder/connection.py", line 207, in get_prepost_signal
    [conn, target])
nengo.exceptions.BuildError: Building <Connection from <Node (unlabeled) at 0x1033022b0> to <Ensemble (unlabeled) at 0x1058e1b70>>: the 'pre' object <Node (unlabeled) at 0x1033022b0> is not in the model, or has a size of zero.
Related objects:
    <Connection from <Node (unlabeled) at 0x1033022b0> to <Ensemble (unlabeled) at 0x1058e1b70>>
        in <Network "m2">
    <Node (unlabeled) at 0x1033022b0>
        (not in model)

```
Note that this makes it also evident that in this case the problem is the object not being in the model and not zero size.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,9 @@ Release History
 - Added ``amplitude`` parameter to ``LIF`` and ``LIFRate``,
   which scale the output amplitude.
   (`#1325 <https://github.com/nengo/nengo/pull/1325>`_)
+- A ``BuildError`` caused by the reference ``Simulator`` will print more
+  information about the object involved to aid identifying the root cause.
+  (`#1320 <https://github.com/nengo/nengo/pull/1320>`_)
 
 **Changed**
 
--- a/nengo/builder/builder.py
+++ b/nengo/builder/builder.py
@@ -213,7 +213,8 @@ class Builder(object):
                 break
         else:
             raise BuildError(
-                "Cannot build object of type %r" % type(obj).__name__)
+                "Cannot build object of type %r" % type(obj).__name__,
+                [obj])
 
         return cls.builders[obj_cls](model, obj, *args, **kwargs)
 
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -93,7 +93,8 @@ def build_linear_system(model, conn, rng):
         raise BuildError(
             "Building %s: 'activites' matrix is all zero for %s. "
             "This is because no evaluation points fall in the firing "
-            "ranges of any neurons." % (conn, conn.pre_obj))
+            "ranges of any neurons." % (conn, conn.pre_obj),
+            [conn, conn.pre_obj])
 
     targets = get_targets(conn, eval_points)
     return eval_points, activities, targets
@@ -130,7 +131,8 @@ def solve_for_decoders(conn, gain, bias, x, targets, rng, E=None):
         raise BuildError(
             "Building %s: 'activities' matrix is all zero for %s. "
             "This is because no evaluation points fall in the firing "
-            "ranges of any neurons." % (conn, conn.pre_obj))
+            "ranges of any neurons." % (conn, conn.pre_obj),
+            [conn, conn.pre_obj])
 
     decoders, solver_info = conn.solver(activities, targets, rng=rng, E=E)
     return decoders, solver_info
@@ -217,11 +219,13 @@ def build_connection(model, conn):
         if target not in model.sig:
             raise BuildError("Building %s: the %r object %s is not in the "
                              "model, or has a size of zero."
-                             % (conn, 'pre' if is_pre else 'post', target))
+                             % (conn, 'pre' if is_pre else 'post', target),
+                             [conn, target])
         if key not in model.sig[target]:
             raise BuildError(
                 "Building %s: the %r object %s has a %r size of zero."
-                % (conn, 'pre' if is_pre else 'post', target, key))
+                % (conn, 'pre' if is_pre else 'post', target, key),
+                [conn, target])
 
         return model.sig[target][key]
 
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -589,7 +589,8 @@ def build_pes(model, pes, rule):
         local_error = correction
     else:
         raise BuildError("'pre' object '%s' not suitable for PES learning"
-                         % (conn.pre_obj))
+                         % (conn.pre_obj),
+                         [conn.pre_obj])
 
     # delta = local_error * activities
     model.add_op(Reset(model.sig[rule]['delta']))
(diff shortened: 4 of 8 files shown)


tbekolay:
I like this idea! Build errors are for sure tricky to debug.

I find the implementation a little hard to understand... a few thoughts that come to mind (not fully fleshed out, but figured a faster vague response is preferable):

- Are there that many situations where there are more than one related object? In the example posted, we could get the pre object from the connection if we have a handle to it. Usually when I run into a build error I will run it through ipython and `%debug` when I get the exception, which makes it easy to get the pre object. Maybe that's not easy in other workflows though.

- The method of getting the location of the object in the network hierarchy seems rather awkward to me... at least, it doesn't seem as though there should need to be changes to `simulator.py`. I also would think that the logic in `store_containment_info` should be put elsewhere as it seems a useful thing to be able to locate an object given the top level network.

jgosmann:
> I find the implementation a little hard to understand...

Yeah, I'm not completely happy with it ...

> Usually when I run into a build error I will run it through ipython and %debug when I get the exception, which makes it easy to get the pre object.

I run models often not interactively or with a debugger. It is always nice if the error give all necessary information without changes to the code and an additional run.

> it doesn't seem as though there should need to be changes to simulator.py

I think we need to do this somewhere at the “top level”. That means either in the Simulator or the Builder. Given the builder code if would have been more awkward because we would have to check in the build function whether it is building the top-level network.

But I would be happy to not change Simulator if there's a better solution.

> also would think that the logic in store_containment_info should be put elsewhere

Yeah, I could definitely see putting it somewhere else. Do you have a concrete suggestion?

jgosmann:
I added some documentation and a unit test.

I'm willing to make further changes with regard to the issues raised by @tbekolay, but would like concrete suggestions.

tbekolay:
After some discussion, we decided not to merge this PR. It is definitely difficult to debug many build errors, but this PR helps mostly in the situation where you have a label on the network, but not on the individual objects. I assume that this happens a fair bit in SPA models, but not so much in other models, where it's common for either all objects to have meaningful labels or none of them. We would encourage projects like SPA to give meaningful default labels to objects within network (as we try to do in Nengo core with [EnsembleArray](https://github.com/nengo/nengo/blob/master/nengo/networks/ensemblearray.py#L113)) to aid in debugging instead.

That all said, we are in the planning stages for a relatively large builder refactoring (#1189) and we will definitely have as an explicit goal of that refactoring to make it easier to debug build errors.

jgosmann:
A few points:

* I'm not sure if this is actually that specific to SPA models (I can't remember the actual situation where I encountered this). Because Nengo SPA sort of generates much of the actual Nengo objects, those models should usually built without problems (if there is no bug in Nengo SPA). It seems more likely to me that this is much more of a problem when constructing large, complicated models yourself. (Though, I might have encountered it while developing and debugging Nengo SPA?).
* I don't think that it is a reasonable assumption that either all or none of the objects have meaningful labels. Naming every object is quite a bit of effort, and often it is sufficiently helpful to name the key networks (e.g. to identify them in the GUI) ... as long as there is not build error at least.
* Giving meaningful names in Nengo SPA is easier said that done and closely intertwined with what the GUI displays as names. See nengo/nengo-gui#1000 for more details on that discussion and PR nengo/nengo-gui#1002 proposing a concrete change that like so often with Nengo GUI PRs hasn't gotten any feedback in months. (And without proper feedback it seems wasted effort to me to invest further time into this issue on the Nengo SPA side.) Another issue with default labels in Nengo SPA is that they could either just describe the specific object (maybe including its parent as context) which isn't really that useful for debugging build errors because many object can have the same name (I think this also applies to EnsembleArray); or they could be labeled with the complete SPA expression they compute, but this leads to very long names which aren't that useful in the GUI either (also the build error will likely effect a single ensemble which again won't include the information about its parents).

That being said, as stated in the initial comment, this was only a proof on concept and not production ready anyways. I'm curious how you intend to solve this problem in the builder refactoring.

PR #1323, opened 2017-06-09
Fix test for ensemble tuning curves (numpy 1.13 compatibility)

drasmuss:
The test was actually broken, but the error was hidden by a deprecation warning.  But that warning becomes an error in numpy 1.13, so we need to fix this for compatibility.

It might be good to make a bugfix release with this soonish, as right now if you do a fresh install of `nengo` it comes with `numpy` 1.13, and the unit tests don't pass.

Also, if we wanted to drop support for any old numpy versions, now might be a good time to do it.

**How has this been tested?**
Fixed one unit tests, and updated travis-ci tests to include numpy 1.13.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ notifications:
 
 env:
   global:
-    - NUMPY="1.12"
+    - NUMPY="1.13"
     - SCIPY="false"
     - COVERAGE="false"
     - STATIC="false"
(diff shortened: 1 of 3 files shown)


jgosmann:
> It might be good to make a bugfix release with this soonish, as right now if you do a fresh install of nengo it comes with numpy 1.13, and the unit tests don't pass.

I don't think it is super critical as it only affects the tests (that are not run by most user), but I'm also totally fine with a release.

PR #1324, opened 2017-06-09
Ensure probe cache is cleared on reset

hunse:
**Motivation and context:**

This fixes a bug, where probes were returning cached values after a simulator reset, if the simulator had been run for the same amount of time (see test).

**How has this been tested?**

I added a unit test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -59,6 +59,11 @@ Release History
   is now more clear.
   (`#1319 <https://github.com/nengo/nengo/pull/1319>`_)
 
+**Fixed**
+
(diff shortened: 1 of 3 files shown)


drasmuss:
(Note, the failing appveyor tests are due to the problem addressed in #1323, so should be fine when that is merged in)

hunse:
> That must have been an interesting bug to track down ;)

Unfortunately I haven't been keeping up with Nengo dev, otherwise I think it would have been easier. But yes, it was very fun :stuck_out_tongue_closed_eyes:

PR #1326, opened 2017-06-15
Fix Connection transform validation

drasmuss:
It should raise an error if `transform=None` (since `optional=False`), but it wasn't because it was being cast to `np.array(None)`.

**How has this been tested?**
Added a new test, existing tests pass.

**Types of changes:**

**Checklist:**


tbekolay:
FYI Appveyor failure will be fixed when rebased to current master (fixed in #1323).

PR #1329, opened 2017-06-20
Add --seed pytest cmd option.

### LINKED ISSUE #1328: Add pytest seed command line argument
If I'm not overlooking something, the only way to change the seed used in tests is to change it in `conftest.py`. This is pretty inconvenient when using the `seed` fixture in tests of other packages (like `nengo_spa`) and also somewhat prohibitve to quickly go through a number of seeds to make sure that tests will not randomly fail for some seeds. I think a custom pytest command line argument should be added to allow to set the seed.

jgosmann:
**Motivation and context:**
This allows to set the seed for tests from the command line. This makes it quicker to try different seeds as `conftest.py` doesn't need to be edited. In particular, this is useful for libraries based on Nengo that use part of the test infrastructure like nengo_spa.

Closes #1328.

none

**How has this been tested?**
manually

**Types of changes:**

**Checklist:**

- [n/a] I have included a changelog entry. (It's not really user facing is it? But I can add an entry if desired.)

### DIFF
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -60,6 +60,9 @@ def pytest_configure(config):
         ntypes = config.getoption('neurons')[0].split(',')
         TestConfig.neuron_types = [load_class(n) for n in ntypes]
 
+    if config.getoption('seed_offset'):
+        TestConfig.test_seed = config.getoption('seed-offset')[0]
+
     TestConfig.compare_requested = config.getvalue('compare') is not None
 
 
--- a/nengo/tests/options.py
+++ b/nengo/tests/options.py
@@ -22,3 +22,5 @@ def pytest_addoption(parser):
     parser.addoption(
         '--slow', action='store_true', default=False,
         help='Also run slow tests.')
+    parser.addoption('--seed-offset', nargs=1, type=int, default=0,
+                     help="Specify offset of the seed values used in tests.")


tbekolay:
I worry a bit about calling this `--seed`: Python has a hash seed, `random` and `numpy.random` have seeds, and `--seed` will not affect any of those. `--nengo-seed` doesn't really work since Nengo doesn't have a global seed. `--seed-offset` or `--seed-mod` is probably the closest to what it is... unless anyone has other ideas for names?

Seanny123:
How about `--testing-seed`?

jgosmann:
I'm ok with pretty much any name. Of the proposed names I like `--seed-offset` and `--testing-seed` best.

tbekolay:
I'm not super happy with any of them, but I think I prefer `--seed-offset` to `--testing-seed` because `pytest --testing-seed 0` seems about the same as `pytest --seed 0` to me (though it's not a strong preference).

Seanny123:
What's the difference between `pytest --seed` and `testing-seed`? Would the difference be communicated by naming the argument `nengo-test-seed`?

tbekolay:
> What's the difference between pytest --seed and testing-seed? Would the difference be communicated by naming the argument nengo-test-seed?

Yeah, we're trying to figure out the best way to communicate what's happening here. What's actually happening is:

- the `seed` pytest fixture (which is defined by Nengo, but not really a Nengo thing) is being modified
- the `seed` is determined by hashing the name of the test function, then adding a constant (normally 0)
- the command-line option would modify the constant that is added to the hashed function name

I don't like `--seed` because it makes it seem like all that is happening is that you're setting the seed that is passed in through the `seed` fixture. `--testing-seed` is about the same to me. `--seed-offset` communicates that better, but isn't quite as memorable or as descriptive as would be ideal (but it might be the best that we can do).

drasmuss:
What about an option like `--stochastic` that would just pass a random number in for all the seeds instead of generating one through hashing (since that's really the desired behaviour/use-case).

jgosmann:
I still want things to be deterministic. If a test fails for a particular seed, I want to be able to reproduce that failure with the same seed.

Seanny123:
Given my new understanding of this question, I prefer `--seed-offset`, even
if it isn't very memorable.

jgosmann:
So, shall I change it to ``--seed-offset``?

tbekolay:
> So, shall I change it to --seed-offset?

Sounds good.

jgosmann:
done ✔️

PR #1334, opened 2017-07-12
Add functions to compute max_rates and intercepts

drasmuss:
**Motivation and context:**

Each neuron type has a function that computes gains and biases given max_rates and intercepts.  This adds a function to each neuron type that does the inverse, computing max_rates and intercepts from gains and biases.

I don't love the name `max_rates_intercepts`, given the awkward double `_`, but I went with that for symmetry with `gain_bias`.

**How has this been tested?**
Added a new unit test, see `test_neurons.py:test_gain_bias`

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -1,6 +1,7 @@
 from __future__ import division
 
 import logging
+import warnings
 
 import numpy as np
 
@@ -88,7 +89,7 @@ class NeuronType(FrozenObject):
             J_max += 10
             J = np.linspace(-J_max, J_max, J_steps)
             rate = self.rates(J, gain, bias)
-        J_threshold = J[np.where(rate <= 1e-16)[0][-1]]
+        J_threshold = J[np.where(rate <= 0)[0][-1]]
 
         gain = np.zeros_like(max_rates)
         bias = np.zeros_like(max_rates)
@@ -108,6 +109,37 @@ class NeuronType(FrozenObject):
             bias[i] = J_top - gain[i]
 
         return gain, bias
+
+    def max_rates_intercepts(self, gain, bias):
+        """Compute the max_rates and intercepts given gain and bias.
+
+        Note that this default implementation is very slow! Whenever possible,
+        subclasses should override this with a neuron-specific implementation.
+
+        Parameters
+        ----------
+        gain : (n_neurons,) array_like
+            Gain associated with each neuron. Sometimes denoted alpha.
+        bias : (n_neurons,) array_like
+            Bias current associated with each neuron.
+
+        Returns
+        -------
+        max_rates : (n_neurons,) array_like
+            Maximum firing rates of neurons.
+        intercepts : (n_neurons,) array_like
+            X-intercepts of neurons.
+        """
+
+        max_rates = self.rates(np.ones_like(gain), gain, bias)
+
+        x_range = np.linspace(-1, 1, 101)
+        rates = np.asarray([self.rates(np.ones_like(gain) * x, gain, bias)
+                            for x in x_range])
+        last_zeros = np.maximum(np.argmax(rates > 0, axis=0) - 1, 0)
+        intercepts = x_range[last_zeros]
+
+        return max_rates, intercepts
 
     def rates(self, x, gain, bias):
         """Compute firing rates (in Hz) for given vector input, ``x``.
@@ -168,6 +200,10 @@ class Direct(NeuronType):
         """Always returns ``None, None``."""
         return None, None
 
+    def max_rates_intercepts(self, gain, bias):
+        """Always returns ``None, None``."""
+        return None, None
+
     def rates(self, x, gain, bias):
         """Always returns ``x``."""
         return np.array(x, dtype=float, copy=False, ndmin=1)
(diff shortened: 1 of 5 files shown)


nengo/neurons.py
```diff
@@ -89,6 +89,38 @@ def gain_bias(self, max_rates, intercepts):
...
+        intercepts : ndarray(dtype=float64)
+            X-intercepts of neurons.
+        """
+
+        max_rates = self.rates(np.ones_like(gain), gain, bias)
+
+        x_range = np.linspace(-1, 1, 101)
+        rates = np.asarray([self.rates(np.ones_like(gain) * x, gain, bias)
+                            for x in x_range])
+        last_zeros = np.maximum(np.argmax(rates > 1e-16, axis=0) - 1, 0)
```
jgosmann: I wonder if this should use `rates >= np.finfo(rates.dtype).tiny` instead of `rates > 1e-16` (which seems arbitrary]?

nengo/neurons.py
```diff
@@ -260,6 +311,14 @@ def gain_bias(self, max_rates, intercepts):
         bias = 1 - gain * intercepts
         return gain, bias
 
+    def max_rates_intercepts(self, gain, bias):
+        """Compute the inverse of gain_bias."""
+        intercepts = (1 - bias) / gain
+        max_rates = 1.0 / (self.tau_ref - self.tau_rc * np.log1p(
+            1.0 / (gain * (intercepts - 1) - 1)))
+        max_rates = np.nan_to_num(max_rates)
```
jgosmann: Under what conditions do the `max_rates` contain non-finite values? And if they do, wouldn't it be better to keep them that way and have the caller deal with it if they want to convert them to finite values?

nengo/neurons.py
```diff
@@ -89,6 +89,38 @@ def gain_bias(self, max_rates, intercepts):
...
+    def max_rates_intercepts(self, gain, bias):
+        """Compute the max_rates and intercepts corresponding to the given
+        gain and bias.
+
+        Note that this default implementation is very slow! Whenever possible,
+        subclasses should override this with a neuron-specific implementation.
+
+        Parameters
+        ----------
+        gain : ndarray(dtype=float64)
```
drasmuss: We do in [gain_bias](https://github.com/nengo/nengo/pull/1334/files#diff-7d7af92e3e5ffca4d32c566c721ccbc4R33), which is the parallel to this function.  We could remove it from that one as well, but it should probably be the same.

nengo/neurons.py
```diff
@@ -89,6 +89,38 @@ def gain_bias(self, max_rates, intercepts):
...
+            Gain associated with each neuron. Sometimes denoted alpha.
+        bias : ndarray(dtype=float64)
+            Bias current associated with each neuron.
+
+        Returns
+        -------
+        max_rates : ndarray(dtype=float64)
+            Maximum firing rates of neurons.
```
drasmuss: Yeah I would say the place to document what we mean by "max_rate" is in the `Ensemble.max_rates` docstring (which does contain a good deal more information).  I could copy that docstring to here as well though.

nengo/neurons.py
drasmuss: `finfo.tiny` seems a bit too small to me (e.g., ~1e-300 for float64).  We could just say `rates > 0` if we only want to know if the value is positive, the idea here is that there is some point that the firing rate is "low enough" that we'll call it 0.  But yeah, I don't have a good reason for picking `1e-16`, other than that is the cutoff we were already using in `gain_bias`.  We could adjust both of those if we think that's too high/low.

nengo/neurons.py
drasmuss: `max_rates` will be non-finite if `-1 < gain * (intercepts - 1) - 1 < 0`.  That can happen if `intercepts >= 1` and `gain` is some smallish value, or if intercepts are in the normal (-1,1) range it only happens if `gain` is 0.  In either of those cases (`intercepts >= 1` or `gain==0`) the firing rate will be zero, so converting `nan` to 0 seemed reasonable.  

I probably wouldn't just return `nan` and let the caller deal with it, since it may not be obvious to them why they're getting `nans`.  We could detect those conditions and raise a warning/exception though, if people think that is preferable to 0.

nengo/neurons.py
jgosmann: I would go with `rates > 0` if there isn't a reason to be concerned about floating point rounding errors (but I'm not sure about that part).

nengo/neurons.py
jgosmann: I have to think more about these things, but at the moment it seems to me:
* Intercepts should never be >= 1 (*) or if they are it should be treated as an error (not sure if that means an exception or non-finite values). Compare #1248, #1212, #1231, #1243 (especially [this comment](https://github.com/nengo/nengo/pull/1243#issuecomment-267787806)).
* The gain should never be <= 0 or if it is it should also be treated as an error. Potentially I could see returning max_rates of 0 for a gain of 0, but keep the intercepts as `nan` (because there isn't a defined intercepts and returning 0 seems to me misleading, while `nan` at least indicate that something's weird). Or maybe an exception is the better solution.

(*) This only applies to LIF neurons (not sigmoid neurons), but this function only applies to LIF neurons.

nengo/neurons.py
jgosmann: That sounds like a good place to me.

nengo/neurons.py
jgosmann: idk, it just caught my attention as being unusual.

drasmuss: I removed the dtype from all the docstrings in this class

nengo/neurons.py
drasmuss: Changed it to 0 instead of 1e-16 here and in `gain_bias`

nengo/neurons.py
drasmuss: Left the values as `nan` and added a warning

drasmuss:
Added some fixups to address Jan's comments.

PR #1337, opened 2017-07-18
No memoize

jgosmann:
**Motivation and context:**

I was changing the transforms on the connections of `CircularConvolution` in place. This had the effect of not only changing a single instance, but messing with all `CircularConvolution` instances (including future ones). It took me quite a while to figure out what was going on: [The `memoize` property on the input transform](https://github.com/nengo/nengo/blob/master/nengo/networks/circularconvolution.py#L23) caches the matrix to be reused by all circluar convolutions (of given dimension). This PR removes the memoize property from the input transform

This is only relevant in very specific cases, but then again it doesn't seem like the memoize optimization will give a large speed up.

Because `memoize` was only used in one more place where it also doesn't seem very necessary, I removed `memoize` and related code completely in a second commit. It does not make that much sense to keep that code around and maintain if it is not really used and if we ever need it back, it can be retrieved from the VCS/git.

But I'm willing ok with not merging either or both commits if there is some opposition.

nengo_spa needs a similar PR as it copied the circular convolution implementation.

**How has this been tested?**

existing tests

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/networks/circularconvolution.py
+++ b/nengo/networks/circularconvolution.py
@@ -6,7 +6,6 @@ import nengo
 from nengo.exceptions import ValidationError
 from nengo.networks.product import Product
 from nengo.utils.compat import range
-from nengo.utils.magic import memoize
 
 
 def circconv(a, b, invert_a=False, invert_b=False, axis=-1):
@@ -20,7 +19,6 @@ def circconv(a, b, invert_a=False, invert_b=False, axis=-1):
     return np.fft.ifft(A * B, axis=axis).real
 
 
-@memoize
 def transform_in(dims, align, invert):
     """Create a transform to map the input into the Fourier domain.
 
@@ -83,7 +81,6 @@ def remove_imag_rows(tr):
         tr = tr[(i == 0) | (i > 3)]
 
 
-@memoize
 def dft_half(n):
     x = np.arange(n)
     w = np.arange(n // 2 + 1)
--- a/nengo/utils/magic.py
+++ b/nengo/utils/magic.py
@@ -263,41 +263,6 @@ def decorator(wrapper):
     return FunctionWrapper(wrapper, _wrapper)
 
 
-def callable_decorator(callable_wrapper):
-    """Decorates callable decorators.
-
-    Similar to @decorator, except applied to anything that returns a wrapper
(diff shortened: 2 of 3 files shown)


Seanny123:
What case was this optimization for and how did you determine it's effect was minimal?

jgosmann:
> What case was this optimization for and how did you determine it's effect was minimal?

To some degree I can only speculate about the first part. The decorator made the first appearance in d988d19f01fa7891a822ad2abd9d2693be425270 by @arvoelke, but it isn't discussed in the commit message. If multiple circular convolutions are constructed, it will only create the the convolution transform once which might give some improvements in runtime and memory consumption. However, it seems that we rarely construct more than a few circular convolutions in a model and constructing that matrix for reasonable dimensionalities (e.g. up to at least 1024 dimensions) does not take very long (though constructing the circular convolution network takes a while, but the bottle neck seems to be somewhere else). Also, this transform matrix will not affect memory consumption that much. A 1024x2052 will consume about 16MB (plus some Python/NumPy overhead).

jgosmann:
> What case was this optimization for and how did you determine it's effect was minimal?

To some degree I can only speculate about the first part. The decorator made the first appearance in d988d19f01fa7891a822ad2abd9d2693be425270 by @arvoelke, but it isn't discussed in the commit message. If multiple circular convolutions are constructed, it will only create the the convolution transform once which might give some improvements in runtime and memory consumption. However, it seems that we rarely construct more than a few circular convolutions in a model and constructing that matrix for reasonable dimensionalities (e.g. up to at least 1024 dimensions) does not take very long (though constructing the circular convolution network takes a while, but the bottle neck seems to be somewhere else). Also, this transform matrix will not affect memory consumption that much. A 1024x2052 will consume about 16MB (plus some Python/NumPy overhead).

jgosmann:
Feel free to run your own tests. I made the PR also to encourage others to test out the effect of this change.

arvoelke:
The memoization was being done prior to that commit (where I just moved it into an explicit decorator). It traces back to the very first commit on the file (7d9e9beabfeed1943c1d6e50346e6b7ee013714c). I remember being unsure at the time if it was worth keeping around, but there was a lot of higher-priority stuff going on in development. I am fine with the changes.

PR #1344, opened 2017-08-11
Add link to doc build instructions to readme.

### LINKED ISSUE #1332: Document how to build documentation
For someone who doesn't know Sphinx (me), it's not at all obvious how to go about building the documentation for myself (i.e. the command ``sphinx-build -b html docs docs-built``). This could be especially frustrating if you don't have internet access.

In the README documentation section, can we put a brief blurb about how to build it? If the instructions need to be more lengthy, we could perhaps make a docs/README file that goes into more detail.

One problem I encountered once I figured out how to build was the warning ``UserWarning: Node.js 0.9.12 or later wasn't found`` followed by the exception ``PandocMissing: Pandoc wasn't found``. I assumed ``pip install -r requirements-docs.txt`` would get me good to go, instead I also had to ``apt-get install nodejs`` (or ``apt-get install pandoc``). We should probably document this additional requirement somewhere.

Also, is it possible to build everything but the examples? They take a long time.

jgosmann:
**Motivation and context:**

This adds a link to the build instructions for the documentation to the README to make it easier to find.

Fixes #1332.

none

**How has this been tested?**
documentation change only

**Types of changes:**

**Checklist:**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,6 +72,15 @@ Documentation
 Usage and API documentation can be found at
 `<https://www.nengo.ai/nengo/>`_.
 
+To build the documentation yourself, run the following command::
+
+    python setup.py build_sphinx
+
+This requires Pandoc to be installed,
+as well as the packages in ``requirements-test.txt`` and ``requirements-docs.txt``.
+For more details, `see the Developer Guide
+<https://www.nengo.ai/nengo/contributing.html#how-to-build-the-documentation>`_.
+
 Development
 ===========
 
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,4 @@
+comment: false
+coverage:
+  status:
+    project: false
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -72,6 +72,10 @@ Documentation
 Usage and API documentation can be found at
 `<https://pythonhosted.org/nengo/>`_.
 
+To build the documentation yourself, `follow the instructions in the Developer
```
hunse: Here, I would say that the documentation can be built with `python setup.py build_sphinx`, and requires the packages in requirements-test.txt and requirements-docs.txt, as well as Pandoc. That gives readers a quick one-sentence summary, and then the link can provide more detailed information.

Really, the most important part is having the build command written here so that I can refer to it easily. Even if we just have that along with the link, I'll be happy.

hunse:
I've added a commit enacting my proposed changes. Let me know what you think.

jgosmann:
Fine with me.

PR #1346, opened 2017-08-11
Document synapse=None.

### LINKED ISSUE #1273: Improve documentation on the synapse parameter in nengo.Connection
In order to skip synaptic filtering, it is possible to just pass

    synapse = None

as a parameter to the Connection object constructor. However, this is not explicitly mentioned in the documentation for that parameter.

jgosmann:
**Motivation and context:**
Adding documentation to explain what `synapse=None` means. @tcstewar, the SpiNNaker backend adds always a time step, doesn't it?

Fixes #1273.

none

**How has this been tested?**
documentation change only

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -285,9 +285,17 @@ class Connection(NengoObject):
     post : Ensemble or Neurons or Node or Probe
         The destination object for the connection.
 
-    synapse : Synapse, optional \
+    synapse : Synapse or None, optional \
               (Default: ``nengo.synapses.Lowpass(tau=0.005)``)
         Synapse model to use for filtering (see `~nengo.synapses.Synapse`).
+        If *None*, no synapse will be used and information will be transmitted
+        without any delay (if supported by the backend---some backends may
+        introduce a single time step delay).
+
+        Note that at least one connection must have a synapse that is not
+        *None* if components are connected in a cycle. Furthermore, a synaptic
+        filter with a zero time constant is different from a *None* synapse
+        as a synaptic filter will always add a delay of at least one time step.
     function : callable or (n_eval_points, size_mid) array_like, \
                optional (Default: None)
         Function to compute across the connection. Note that ``pre`` must be


mundya:
Hey! The SpiNNaker backend does always add a timestep.

tcstewar:
> @tcstewar, the SpiNNaker backend adds always a time step, doesn't it?

Sort of, but that's not quite the full story.  SpiNNaker's weird.  The first thing it tries to do is compile out passthrough nodes.  It doesn't get rid of all of them (some of them are helpful for dealing with bandwidth limitations), but almost all the time that there's a long chain of passthrough nodes with None as the synapse (for example, the circular convolution network), it'll collapse those down so only one (or possibly two) remain.

Of those remaining synapse=None synapses, it treats them exactly like it treats synapse=0 synapses.  Which is to say, it doesn't add a filter, and just adds a communication channel....  and on SPiNNaker a communication channel is literally a communication channel, in that the two ends of it are on different cores, and it sends packets from one to the next.

*However*, SpiNNaker a) is asynchronous and b) only guarantees communication time < 3ms.  So that means that the synapse=None connection might act exactly like a synapse=None, or a synapse=0, or even a synapse=z*(-3).  The place and route algorithm will try to place the ends of the communication channel near each other, so I'd say it'll often be synapse=None or synapse=0....

tcstewar:
(and hi @mundya !   Good to see you!   Did I screw anything up in my description above? )

jgosmann:
@tcstewar, that is a lot of detail and too much to include in the docstring. But I think the docstring captures that the behaviour of `synapse=None` can differ in weird ways on different backends. Do you agree?

tcstewar:
> @tcstewar, that is a lot of detail and too much to include in the docstring. But I think the docstring captures that the behaviour of synapse=None can differ in weird ways on different backends. Do you agree?

I think the docstring captures things well, and I'm happy with it.  As with pretty much everything in the nengo spec, the model is meant to be a request to the backend as to what to implement, and the backend will do the best job it can, but pretty much nothing is guaranteed.  So backends might do all sorts of things, not just adding one time step.  Heck, on any analog chip backend, there probably won't even be the possibility of not having a synapse, so it might just fill it in with the fastest possible synapse on that hardware....

arvoelke:
> or even a synapse=z**(-3)

By the way... if you are wondering where this notation is coming from, check out this documentation:
 [nengolib.synapses.DiscreteDelay?highlight=z](https://arvoelke.github.io/nengolib-docs/nengolib.synapses.DiscreteDelay.html?highlight=z) and [nengolib.signal.z](https://arvoelke.github.io/nengolib-docs/nengolib.signal.LinearSystem.html#nengolib.signal.z), which you can use in your Nengo models.

This follows the "standard" interpretation of `z` as a discrete time-shift operator (ctrl+F ["Time shifting"](https://en.wikipedia.org/wiki/Z-transform#Properties) for some mathematical details). 

</shameless plug> :)

mundya:
Hi @tcstewar, @jgosmann - the following will again be too much information for a docstring but might be useful if you're thinking about SpiNNaker:

> The first thing it tries to do is compile out passthrough nodes. It doesn't get rid of all of them (some of them are helpful for dealing with bandwidth limitations), but almost all the time that there's a long chain of passthrough nodes with None as the synapse (for example, the circular convolution network), it'll collapse those down so only one (or possibly two) remain.

I think that, after the nengo_spinnaker build process, there will never be more than one passthrough node-like thing between any pair of non-passthrough-Nodes. The passthrough node-like things ("interposers" in nengo_spinnaker terms) are out of phase with the rest of the nengo_spinnaker executables, so they should be invisible to the simulation provided that (a) the machine is relatively well synchronised (b) there isn't significant latency added by the network. (See Figure 5.15, p.120 of my thesis).

> Which is to say, it doesn't add a filter, and just adds a communication channel.... and on SPiNNaker a communication channel is literally a communication channel, in that the two ends of it are on different cores, and it sends packets from one to the next.

This is true even for a recurrent connection, i.e., packets will be sent from a core to itself. In this case you can guarantee there will a one time step delay since the outgoing packet won't be processed until at least the next simulation time step.

PR #1347, opened 2017-08-11
Document more CosineSimilarity relationships.

### LINKED ISSUE #1218: Distribution for single dimension of UniformHypersphere(surface=False)
As was discovered in https://github.com/nengo/nengo_extras/issues/34, the following two samples have the same distribution:

```
CosineSimilarity(d+2).sample(n) === UniformHypersphere(surface=False).sample(n, d)[:, 0]
```

this is analogous to the (currently documented) connection that the following two samples also have the same distribution:

```
CosineSimilarity(d).sample(n) === UniformHypersphere(surface=True).sample(n, d)[:, 0]
```

The latter is useful for projecting encoders onto one dimension (see the docstring for `CosineSimilarity`), while the former is useful for projecting evaluation points onto one dimension (see https://github.com/nengo/nengo_extras/issues/34). Both of these distributions have a role in optimizing intercepts for higher dimensions.

(Interestingly, this implies that:
```
UniformHypersphere(surface=True).sample(n, d+2)[:, 0] ===
UniformHypersphere(surface=False).sample(n, d)[:, 0]
```
which is really weird!!)

Since it took a bunch of work from multiple people to prove this (deceptively simple) relationship, we should consider making `CosineSimilarity(d+2)` into its own distribution, and/or documenting the above connection.

jgosmann:
**Motivation and context:**
Document further relationships of the CosineSimlarity distribution to the UniformHypersphere distribution. Also reference tech report with further details.

Fixes #1218.

none

**How has this been tested?**
documentation change only

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/dists.py
+++ b/nengo/dists.py
@@ -503,11 +503,25 @@ class CosineSimilarity(SubvectorLength):
 
     This is also equivalent to the distribution of a single coefficient from a
     unit vector (a single dimension of ``UniformHypersphere(surface=True)``).
+    Furthermore, ``CosineSimilarity(d+2)`` is equivalent to the distribution of
+    a single coordinate from points uniformly sampled from the d-dimensional
+    unit ball (a single dimension of
+    ``UniformHypersphere(surface=False).sample(n, d)``). These relationships
+    have been detailed in [Voelker2017]_.
 
     This can be used to calculate an intercept ``c = ppf(1 - p)`` such that
     ``dot(u, v) >= c`` with probability ``p``, for random unit vectors ``u``
     and ``v``. In other words, a neuron with intercept ``ppf(1 - p)`` will
     fire with probability ``p`` for a random unit length input.
+
+    .. [Voelker2017]
+       `Aaron R. Voelker, Jan Gosmann, and Terrence C. Stewart.
+       Efficiently sampling vectors and coordinates from the n-sphere and
+       n-ball. Technical Report, Centre for Theoretical Neuroscience,
+       Waterloo, ON, 2017
+       <http://compneuro.uwaterloo.ca/publications/voelker2017.html>`_,
+       `doi:10.13140/RG.2.2.15829.01767/1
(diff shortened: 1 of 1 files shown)


nengo/dists.py
```diff
@@ -503,12 +503,27 @@ class CosineSimilarity(SubvectorLength):
 
     This is also equivalent to the distribution of a single coefficient from a
     unit vector (a single dimension of ``UniformHypersphere(surface=True)``).
+    Furthermore, ``CosineSimilarity(d+2)`` is equivalent to the distribution of a
+    single coordinate from points uniformly samples from the d-dimensional unit
```
arvoelke: sampled

nengo/dists.py
```diff
@@ -503,12 +503,27 @@ class CosineSimilarity(SubvectorLength):
...
     and ``v``. In other words, a neuron with intercept ``ppf(1 - p)`` will
     fire with probability ``p`` for a random unit length input.
 
+    .. [Voelker2017]
+       `Aaron R. Voelker, Jan Gosmann, and Terrence C. Stewart.
+       Efficiently sampling vectors and coordinates from the n-sphere and
+       n-ball.
+       Technical Report, Centre for Theoretical Neuroscience, Waterloo, ON,
+       2017
+       <https://www.researchgate.net/publication/312056739_Efficiently_sampling_vectors_and_coordinates_from_the_n-sphere_and_n-ball>`_,
```
arvoelke: line too long?

jgosmann: Yes, but I don't think that's fixable. because it is in docstring it is also awkward to add a `# noqa` comment. I actually vote to disable the pyflakes line length check, but enable the it in pylint. pylint supports to define a regex for lines that are allowed to exceed the maximum length which could be configured to allow long URLs. I suppose I'll do a separate PR for that. (related: #1278)

jgosmann: Should be fixed after a rebase to master (which has #1360 merged). But should be verified after the rebase.

Issue #1308, opened 2017-05-07
Building examples in documentation fails with recent Jupyter notebook versions.


jgosmann:
This is the exception I get ([note that I had to remove the conversion to the warning to get the actual backtrace](https://github.com/nengo/nengo/blob/master/nengo/utils/docutils.py#L105) which should probably fixed to):
```
# Sphinx version: 1.5.5
# Python version: 3.5.3 (CPython)
# Docutils version: 0.13.1 release
# Jinja2 version: 2.9.6
# Last messages:
#   Running Sphinx v1.5.5
#   loading pickled environment...
#   not yet created
#   loading intersphinx inventory from http://docs.scipy.org/doc/numpy/objects.inv...
#   intersphinx inventory has moved: http://docs.scipy.org/doc/numpy/objects.inv -> https://docs.scipy.org/doc/numpy/objects.inv
#   building [mo]: all of 0 po files
#   building [html]: all source files
#   updating environment:
#   10 added, 0 changed, 0 removed
#   reading sources... [ 10%] examples/associative_memory
# Loaded extensions:
#   sphinx.ext.todo (1.5.5) from /Volumes/Home/blubb/.virtualenvs/nengodev/lib/python3.5/site-packages/sphinx/ext/todo.py
#   sphinx.ext.viewcode (1.5.5) from /Volumes/Home/blubb/.virtualenvs/nengodev/lib/python3.5/site-packages/sphinx/ext/viewcode.py
#   nengo.utils.docutils (unknown version) from /Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/utils/docutils.py
#   alabaster (0.7.10) from /Volumes/Home/blubb/.virtualenvs/nengodev/lib/python3.5/site-packages/alabaster/__init__.py
#   sphinx.ext.intersphinx (1.5.5) from /Volumes/Home/blubb/.virtualenvs/nengodev/lib/python3.5/site-packages/sphinx/ext/intersphinx.py
#   numpydoc (unknown version) from /Volumes/Home/blubb/.virtualenvs/nengodev/lib/python3.5/site-packages/numpydoc/__init__.py
#   sphinx.ext.mathjax (1.5.5) from /Volumes/Home/blubb/.virtualenvs/nengodev/lib/python3.5/site-packages/sphinx/ext/mathjax.py
[...]

Issue #1087, opened 2016-06-10
Rethinking the Jupyter progress bar.


jgosmann:
The Jupyter widget API changed quite a bit over the last couple of versions, but seems to get more stable now that they moved it to [its own package](https://ipywidgets.readthedocs.io/en/latest/index.html). We should take another look at the current recommended way to implement and [distribute such widgets](https://ipywidgets.readthedocs.io/en/latest/examples/Custom%20Widget%20-%20Hello%20World.html#More-advanced-uses:-Packaging-and-distributing-Jupyter-widgets) and see whether it makes sense to adopt the recommended approach for Nengo.

jgosmann:
The Jupyter progress bar will crash the notebook when used with the 5.1.0 notebook release. After that the page needs to reloaded to restore communication between the notebook and kernel. This might be due to the way we inject the JavaScript into notebook.

Currently we are trying to support a number of notebook versions. It would be nice if we could drop support for older versions to reduce code complexity (and testing burden) with the next minor release. After all it is nice, but non-essential feature (there is still the ASCII progress bar or the possibility to use an older Nengo version if desired).

Should we also do bugfix release to at least not break the notebook (fallback to the ASCII bar)? It is probably not obvious to the average user why the notebook breaks if they happen to use the Jupyter progress bar. On the other hand, the workaround is easy once known (just don't do `%load_ext nengo.ipynb`).

Thoughts, @tbekolay?

Also, should we rename `nengo.ipynb` to match the rebranding? Not sure, what a good name would be, though. Also, the notebooks also still use the `ipynb` extension.

tbekolay:
I'm fine dropping support for older versions (people can revert to an earlier released version that works with older notebook versions if they're annoyed), and yeah, if the current version is crashing the notebook I say we remove it for the time being. If a PR for that gets merged we can do a quick release.

jgosmann:
I can't figure out which older versions actually work because the versioning in the Jupyter eco system is a mess. There is a multitude of potentially relevant packages (`jupyter`, `jupyter-client`, `jupyter-core`, `notebook`, `widgetsnbextension`, `ipywidgets`), all with different current version and I can find no documentation what set of versions is supposed ot give an environment in which the widgets work. Even if I were able to figure out what versions work and which do not, versions on the Jupyter server side are probably relevant and these cannot be queried in the kernel (`%load_ext`  is executed in the kernel context).

jgosmann:
[...]

PR #1349, opened 2017-08-16
Use nbsphinx to build examples for documentation

jgosmann:
**Motivation and context:**
As noted in #1308 it is not possible to build the examples for inclusion in the documentation with current versions of Jupyter. While the solution so far was to install older versions, this should eventually be fixed and as I have the same issue with the documention of Nengo SPA, I decided to spend some time on this.

@tbekolay pointed me towards [nbsphinx](https://nbsphinx.readthedocs.io/en/0.2.14/) which allows to include Jupyter notebooks into Sphinx documentation with good integration with the Sphinx theme. However, Sphinx does not allow to have source files outside of the documentation folder (for security reasons I believe), but our examples folder lives outside the documentation folder. I am opposed to changing that, and I think @tbekolay agrees. So other potential solutions are:

* The PR spatialaudio/nbsphinx#33 which adds a RST directive that allows to include notebooks from arbitrary paths. Unfortunately it doesn't seem like this PR will get merged any time soon (if ever).
* A symlink, but this does not work on Windows (before 10 without elevated privileges). Also, this leads to problems with images (spatialaudio/nbsphinx#49).
* Copy all the notebooks which can be done automatically by sphinx through `conf.py`. This is the solution I chose in this branch because it was pretty much the path of least resistance. It seems like sphinx manages, despite the copied notebooks being overwritten in each run, to just re-excute notebooks that have been changed.

Another issue (which might also exist in the current system, but I haven't verifiied it): The Jupyter notebooks are executed by a Jupyter kernel that is independent from whatever virtualenv is active. It seems like a bad idea to just use the default kernel that might have the wrong version of Nengo installed (in particular true for my environment). Thus I added some code to conf.py to ask for which kernel to use (and the possibility to fix it via an environment variable).

This PR is still work-in-progress (see to do list below), but at this stage it would be good to get an idea if the general direction seems good.

ctn-waterloo/devscripts#6 changes the clearoutput script to add metadata to the notebooks that is required to make syntax highlighting with nbsphinx work. I ran the script on all existing examples to add that metadata.

**How has this been tested?**

build the docs and looked at through them

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] There are some warnings about non-included files (`examples/REAME.rst`, `examples/usage/exceptions.ipynb`, `examples/usage/strings.ipynb`). Figure out whether these should indeed be skipped in the documentation (and silence these warnings) or include them in documenation in an appropriate way.
- [x] In some notebooks input part of code cells has been collapsed. This does not work in nbsphinx. nbsphinx does support hiding complete cells, i.e. input and output, but in most cases we want to preserve the output. While a commonly requested feature, it doesn't seem likely that separate hiding of input and output will get implemented any time soon. Not sure what exactly to do about this or whether this rules out nbpshinx completely. It might also be possible to adapt our `hide_input` function to work in Jupyter notebook and the Sphinx output (at least the Javascript gets included).
- [x] The progress bar widget doesn't work and instead a warning will be included. It seems that this is because we are not properly installing the widget Javascript in the recommended way. This is something that we probably should fix (#1087), but involves some other issues. Also, I think it isn't very useful to include the progress bar in the documentation (what does it matter how long that example took on someone else's computer?). But currently I don't see a way to disable it when building the notebooks for the documentation.
- [x] Disable cache as it might introduce warnings into the documentation that should not be there.


tbekolay:
> Sphinx does not allow to have source files outside of the documentation folder (for security reasons I believe), but our examples folder lives outside the documentation folder. I am opposed to changing that, and I think @tbekolay agrees.

This was my original reason to investigate these types of options and make the PR I made for nbsphinx, but as time's passed I have to admit that I'm not as opposed to moving the notebooks to the `docs` directory as I used to be. I think the main argument to have an `examples` folder at the root level was to make it easily discoverable when people go to https://github.com/nengo/nengo/; however, I think it's going to become increasingly common for users to go to https://www.nengo.ai/nengo/ for help and not GitHub. For that reason, I think it might be a good time to 1) upgrade the notebooks to v4 and 2) move them to the `docs` folder. I added that to the next dev meeting agenda as something to discuss.

> The Jupyter notebooks are executed by a Jupyter kernel that is independent from whatever virtualenv is active.

Hm, I've never experienced that behavior before... when I'm in my Python 2 environment, I get:

```
$ jupyter kernelspec list
Available kernels:
  python2    /home/tbekolay/.virtualenvs/nengo/share/jupyter/kernels/python2
```

and in my Python 3 environment:

```
$ jupyter kernelspec list
Available kernels:
  python3    /home/tbekolay/.virtualenvs/nengo3/share/jupyter/kernels/python3
```

What do you get?

> In some notebooks input part of code cells has been collapsed.

I'm sure that it can be adapted, but also I'm not sure how strongly I feel about these input cells being collapsed in the first place... perhaps we should remove the `hide_input` function?

> The progress bar widget doesn't work and instead a warning will be included.

Indeed, we should get rid of the progress bar in the docs... I have a few thoughts about how this might be done but would have to play around with a bit...

jgosmann:
> I think the main argument to have an examples folder at the root level was to make it easily discoverable when people go to https://github.com/nengo/nengo/;

In Nengo SPA I moved the examples folder into `nengo_spa` which allows to install them as package date and extract them to a user-defined location with `python -m nengo_spa extract-examples <dst>`. This doesn't work will with moving them to the docs folder. :/

My `jupyter kernelspec list` output (independent of virtualenv, all my virtualenvs are using Python 3 at the moment):
```
Available kernels:
  bleeding-edge          /home/jgosmann/.local/share/jupyter/kernels/bleeding-edge
  cogscie17-decide       /home/jgosmann/.local/share/jupyter/kernels/cogscie17-decide
  imem                   /home/jgosmann/.local/share/jupyter/kernels/imem
  kajic-frontiers2016    /home/jgosmann/.local/share/jupyter/kernels/kajic-frontiers2016
  nengodev               /home/jgosmann/.local/share/jupyter/kernels/nengodev
  python3                /home/jgosmann/.local/share/jupyter/kernels/python3
```
(and I can switch the virtualenv in a notebook by changing the kernel)

> perhaps we should remove the hide_input function?

I'm probably ok with that.



> Indeed, we should get rid of the progress bar in the docs... I have a few thoughts about how this might be done but would have to play around with a bit...

Can you roughly describe your thoughts?

tbekolay:
> moved the examples folder into nengo_spa which allows to install them as package date

Makes sense!

> jupyter kernelspec list output

Hmm, do you not install `jupyter` in your virtualenvs? Seems like your virtualenv / jupyter setup is very different from mine... not sure if these kinds of environmental differences should end up in a shared repo.

> Can you roughly describe your thoughts?

I'm not sure how Sphinx works exactly, but the first thing I'd try is manipulating `nengo.rc` to turn off progressbars in the `conf.py` script. But maybe Sphinx or nbsphinx uses subprocesses for some part of the build, which would skirt the `nengo.rc` settings in that Python process, so if that's the case, then It might be possible to spit out a `nengorc` file in whatever directory the docs are being built from, and then remove it after the build (if there's already a `nengorc` file in that folder, then probably warn about the progress bars, leave it as is, and continue).

jgosmann:
My Jupyter kernels reside [in the user location](http://jupyter-client.readthedocs.io/en/latest/kernels.html#kernel-specs) which is not related to the virtualenv, whereas yours are in the “environment” location. Whether jupyter is installed in the virtualenv affects only the process managing the kernels (usually I start it without an active virtualenv if I don't want to test something with a specific Jupyter version). I think we can't rely on the default kernel being the correct one.

If we don't want the user to manually select the kernel, it might be possible to automatically install a kernel with the virtualenv's Python.

> But maybe Sphinx or nbsphinx uses subprocesses for some part of the build

I think Sphinx launches independent Jupyter kernels which are unaffected by config manipulations (but I might be wrong).

> , then It might be possible to spit out a nengorc file in whatever directory the docs are being built from, and then remove it after the build

I think this might actually work. Especially if we copy the examples, the file doesn't need to be removed.

jgosmann:
> then It might be possible to spit out a nengorc file in whatever directory the docs are being built from

That works indeed, I added a commit. :)

jgosmann:
I think this should now be ready for review. I made `hide_input` work in both the documentation and Jupyter. But I'm also still open to removing it either as part of this PR or a different PR.

In Nengo SPA I had some incompatiblitly with the numfig feature and numpydoc (nengo/nengo_spa@b61b770374d0389be921bd083c69e5d921011a58) and replaced numpydoc with `sphinx.ext.napoleon` (which ships with Sphinx and supports both numpydoc and Google style docstrings). It could be nice to do the same here for consistency. However, in the Nengo documentation it introduces some other issues:
* It doesn't play nice with the defaults. It changes the formatting a little bit and puts the argument data types into parenthesis, so that we always get weird looking parenthesis within parenthesis. This is somewhat easy to fix with a regex search and replace. (Though one could argue that the data type field isn't the right place for that information. According to the NumPy docs it is supposed to go into the description string if not taken from the function signature.)
* The formatting of the attribute list takes up way too much space (sphinx-doc/sphinx#2115). This is mostly fixed by the `napoleon_use_ivar = True` config option, but introduces a new problem:
* With `napoleon_use_ivar = True` sphinx tries to link the attribute names to random stuff it shouldn't be linked to (sphinx-doc/sphinx#2549). This won't be fixed before Sphinx 1.7.

For those reasons it seems best to stick with numpydoc in core Nengo for now.

jgosmann:
One way to properly show the default values in the function signature would be to add docstrings like this one to the `__init__` methods:
```python
class Node(NengoObject):
    # ....
    def __init__(self, output=Default, size_in=Default, size_out=Default,
                 label=Default, seed=Default):
        """__init__(output=None, size_in=None, size_out=None, label=None, seed=None)"""
        # ...
```
But I'm not sure if we want to do that:
* It adds a little bit more boilerplate to the code.
* Having the defaults in the function signature might not be the most convenient location.
* It might be misleading because the actual values will depend on the config (but not sure if it is really more misleading then the current notation of defaults).

tbekolay:
We discussed this at the dev meeting today and were generally in favor of moving the examples to the `docs` directory, and generally discourage people from viewing the notebooks through Github (as they're unevaluated). The main downsides to this are that:

1. It may not be obvious how to edit the notebooks, and
2. it may not be clear how to get the notebooks locally to evaluate them live.

To deal with these issues, we will 1) either manually add some text in the examples folder to say where all the examples live in the repo, and 2) make sure that the docs still include a link to download the `.ipynb` file. That might require a change in nbsphinx, but I've touched that code before and can likely make the PR if it's not already possible.

Seanny123:
> To deal with these issues, we will
>
> 1) either manually add some text in the examples folder to say where all the examples live in the repo, and 

So, just adding a `README.md` into the examples folder? 

> 2) make sure that the docs still include a link to download the .ipynb file.

As in, adding this as a header to the rendered/uploaded docs?

Also, why not do both of these things?

tbekolay:
> So, just adding a README.md into the examples folder?

No, the text would have to be in the rendered docs, as in the ideal case users would never need to know about the organization of the source tree and mostly interact with Nengo by pip installing and looking through the docs.

> As in, adding this as a header to the rendered/uploaded docs?

Ideally this would be done automatically in the conversion from notebook to doc page. With nbsphinx, the only way to add this as a header would be to include it in the notebook itself, which feels weird.

Seanny123:
Given this new information, I vote for the second option of adding a header to the rendered page, which would understandably require a modification to `nbsphinx`.

jgosmann:
I added a commit that changes the “View page source” link at the top right of normal documetation pages to a “Download example as Jupyter notebook” link for the examples. This is achieved by two things:

* Overwriting the relevant part of the template. This unfortunately involved copying a ~30 line block from the original template. The templates support inheritance, but are not fine grained enough to just change that one link with overwriting a bunch of other links (e.g. to GitHub) that could be included (tough we don't include any of those links).
* A hook into the end of the sphinx build process that copies the examples (they are actually already copied for the page source link, but a .txt extension is added. Copying them again seems to be the only way to get rid of that extension.

Let me know what you think of these changes.

jgosmann:
Rebased, apparently the documentation theme has changed. With the guzzle theme the template modification actually requires less boilerplate, so that is nice. :)

I think the only remaining thing is to add some documentation pointing to the new location of the examples folder. However, I am not sure where that should go and what it should say, @tbekolay. To me it doesn't seem particularly obscure to have the examples in the docs folder and I don't think we have any other parts of the folder structure documented (which doesn't mean that it isn't a good idea).

jgosmann:
I mentioned that it would be nice to automatically extract parameter defaults from the source code and I think it might be possible with a small Sphinx extension, but that is probably something for a separate PR.

Seanny123:
Note: TravisCI is failing for reasons unrelated to this PR.

Seanny123:
`requirements-docs.txt` is missing the Sphinx Guzzle Theme, but that should be fixed by rebasing/merging.

Seanny123:
Building docs on Windows 10, I get the error `Pygments lexer name 'ipython3' is not known`.

jgosmann:
I feel like I saw that error before ... can you try upgrading jupyter, notebook, ipython, and pygments?

jgosmann:
Apparently this might be an Anaconda bug, see spatialaudio/nbsphinx#24 and ContinuumIO/anaconda-issues#1430.

The `ipython3` lexer is provided by the IPython package and not pygments, so it won't be listed in the pygments documentation.

Seanny123:
That makes sense because I was using Anaconda when I got the aforementioned error. When I tried to build the notebooks on my lab desktop, it told me I was missing `ipywidgets`. Should this be added to `requirements-docs.txt`? Other than that little nitpick, this LGTM.

jgosmann:
Hm, `ipywidgets` should not be required. It might be the `nbsphinx` runs the `%load_ext nengo.ipynb` and that this requires `ipywidgets`. But those lines will get removed with PR #1380.

Seanny123:
That is precisely what's happening. If those lines are removed in the
following pull request, then I'm fine with this one.

Seanny123:
Is this PR still ready to merge? Just wanted to check that we were just waiting for a final review.

jgosmann:
Skimming over the comments, the only remaining issue seems to be:

> I think the only remaining thing is to add some documentation pointing to the new location of the examples folder. However, I am not sure where that should go and what it should say, @tbekolay. To me it doesn't seem particularly obscure to have the examples in the docs folder and I don't think we have any other parts of the folder structure documented (which doesn't mean that it isn't a good idea).

So, I suppose it is indeed just waiting for a final review.

tbekolay:
OK I've rebased this now and fixed issues from the rebase; if there are no objections to the current state of this PR I'll compress the history a bit and merge after lunch.

jgosmann:
There was code for selecting the appropriate kernel when building the notebooks with nbsphinx in the PR (see discussion above), but that seems to have completely vanished? 😮 

I still think it should be included, though I don't care as much about Nengo core as I usually don't have to build the documentation. But the same code seems to have gone missing from Nengo SPA too (or was it never in there?) which right now prevents me from building the docs with the correct kernel/virtualenv. 😞

jgosmann:
Luckily I found a local copy of the code.

tbekolay:
Yeah, I removed that from this core PR, but I didn't touch anything in Nengo SPA.

tbekolay:
As for why I removed it, it just seemed out of place in `conf.py`. It's a local environment thing, so it seems as though it should be handled in the local environment, not in `conf.py`. I am able to build these docs without that snippet, as is Travis CI.

jgosmann:
But it cannot be handled without modifying `conf.py` (I believe) because `nbsphinx_kernel_name` needs to be set.

tbekolay:
You can pass that in to `sphinx-build` with the `-D` flag

jgosmann:
Does that somehow work with `python setup.py build_sphinx` (which is the command in our documentation)?

tbekolay:
I'm not sure, but it's definitely easier to update the documentation than have that in `conf.py`

jgosmann:
Makes sense.

I rescued the code [into a Gist](https://gist.github.com/jgosmann/c5ad9a36b7c4c96a1d4bb48ce7884841) in case I or someone else wants to refer to it at a later point.

PR #1052, opened 2016-05-02
Run only simulator tests for non-reference sim.

jgosmann:
Addresses nengo/nengo_ocl#65

PR #1350, opened 2017-08-16
Report the number of runnable tests.

### LINKED ISSUE #1082: Report total number of runable tests
For backends, there are a lot of frontend tests that get skipped because they're not relevant. @jgosmann suggested reporting this in #1052.

@jgosmann:

> If we go with the skip solution. Maybe we can count how many tests should be run to get the complete set of relevant tests (e.g. excluding compare tests or frontend tests) and print that number? That might be a lot easier.

@tbekolay:

> Are you suggesting something that prints a more detailed summary of the tests collected to the console right after collection? Something like `Running 637/655 frontend tests` etc? That would be great

@jgosmann:

> Something like (can still work on the exact formulation):
> `Total of XXX runnable tests (excluding compare tests).`
> For compare tests:
> `Total of XXX runnable tests (excluding non-compare tests).`
> For alternate simulator:
> `Total of XXX runnable tests (excluding compare tests and front-end tests).`
> 
> And maybe a line if all runnable tests have been run (easier to check for than comparing the number of tests):
> `All runnable tests were run.`

jgosmann:
**Motivation and context:**

This adds a line (or more) to the pytest output indicating the number of total runnable tests and how many have run. It also adds lines explaining why some tests are not runnable. For example compare and non-compare tests are mutually exclusive and frontend tests aren't run for non-reference backends. But these will still be counted in the number of skipped tests which means that number will never be zero and thus cannot be used as an indicator whether the complete test suite ran.

I'm not entirely sure about the best way to present this information and the wording can probably be still improved. But at least the harder (?) part of writing the logic is done.

Closes #1082.

none

**How has this been tested?**

Manually by running pytest with different arguments.

**Types of changes:**

**Checklist:**

**Still to do:**

- [ ] better wording maybe?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,9 @@ Release History
 - We now warn that the progress bar is not supported in Jupyter Notebook <5.
   (`#1428 <https://github.com/nengo/nengo/pull/1428>`__,
   `#1426 <https://github.com/nengo/nengo/issues/1426>`__)
+- Added information about number of runnable tests to pytest summary
+  (`#1082 <https://github.com/nengo/nengo/issues/1082>`_,
+  `#1350 <https://github.com/nengo/nengo/pull/1350>`_)
 
 **Changed**
 
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -337,7 +337,48 @@ def pytest_runtest_setup(item):  # noqa: C901
                 pytest.xfail(reason)
 
 
+def determine_run_stats(terminalreporter):
+    non_runnables = {
+        "compare tests not requested",
+        "frontend tests not run for alternate backends",
+        "RefSimulator not overridden",
+        "Simulator not overridden"
+    }
+    n_ran = 0
+    n_runnables = 0
+    for key in terminalreporter.stats:
+        if key != '':
+            if key == 'skipped':
+                for report in terminalreporter.stats[key]:
+                    reason = report.longrepr[-1]
+                    if reason.startswith('Skipped: '):
+                        reason = reason[9:]
+                    if reason not in non_runnables:
+                        n_runnables += 1
+            else:
+                n = len(terminalreporter.stats[key])
+                n_ran += n
+                n_runnables += n
+    return n_ran, n_runnables
+
+
 def pytest_terminal_summary(terminalreporter):
+    n_ran, n_runnables = determine_run_stats(terminalreporter)
+    if n_ran == n_runnables:
+        line = "Ran all {} runnable tests.".format(n_ran)
+    else:
+        line = "Ran {} of {} runnable tests.".format(n_ran, n_runnables)
+    terminalreporter.write_sep("=", line, bold=True)
+    if TestConfig.compare_requested:
+        terminalreporter.writer.line(
+            "Non-compare tests run only without --compare option.")
+    else:
+        terminalreporter.writer.line(
+            "Compare tests require --compare option.")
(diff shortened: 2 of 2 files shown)


jgosmann:
Fixed the pyflakes error (complexity of the function) and added a changelog entry. Ready for review. If the reviewer has suggestions for how to improve the wording, I'm happy to incorporate those. Otherwise it is good enough for me.

Seanny123:
One TravisCI test seems to be failing. Will that be fixed by brining this branch up to date?

jgosmann:
Those failures seem to be unrelated to this PR and only happen for NumPy 1.7 for which support was dropped in the latest release. So I assume it will be fine once rebased.

tbekolay:
I rebased this branch then tried running

```
pytest -k ensemble
```

The output (relevant to this PR) that I get is

```
========================================== test session starts ==================================
platform linux2 -- Python 2.7.15, pytest-3.6.3, py-1.5.4, pluggy-0.6.0
rootdir: /home/tbekolay/Code/nengo, inifile: setup.cfg
collected 1053 items / 929 deselected                                     
...
==================== Ran all 1065 runnable tests. =============================
Compare tests require --compare option.
...
=============== 124 passed, 929 deselected, 12 warnings in 32.94 seconds ===============
```

The `Ran all 1065 runnable tests` is confusing since only 124 tests actually ran due to the filtering, and it's also not super clear why 1065 tests are listed rather than the 1053 noted above. If the difference is that the compare tests make up the remaining 12 tests, I would have expected something like `Collected 1053 tests of 1065 total tests` and then `12 compare tests not run (use the --compare option)` or something like that. But I'm not sure that the discrepancy is because of the compare tests, as when I do the compare tests I still get 1053 collected...

I'm also a bit concerned that we're overloading people with numbers here because in the snippet above we see `1053 items / 929 deselected`, then `Ran all 1065` then `124 passed etc` which is a lot.

In #1082 we seem to be talking about how many tests are collected, rather than run (though we say "run" there). So I think maybe a better place to report these numbers is right after the `collected 1053 items / 929 deselected` line... maybe something like

```
========================================== test session starts ==================================
platform linux2 -- Python 2.7.15, pytest-3.6.3, py-1.5.4, pluggy-0.6.0
rootdir: /home/tbekolay/Code/nengo, inifile: setup.cfg
collected 1053 items / 929 deselected
12 compare tests skipped (pass --compare to run them)
88 example tests skipped (--noexamples passed)
...
```

Actually the information printed with the `-rs` option ([example](https://github.com/nengo/nengo/pull/1052#issuecomment-224781634)) is pretty close to what I think I would want to see. But that information isn't available until after the tests run (as opposed to right after collection) so it might not be feasible.

Another option would be to deselect the tests rather than skipping them. It seems as though in #1052 we were collecting the tests, then removing them, which is definitely strange. We switched to skipping them, but I wonder if it is possible instead to deselect them after collection. See https://github.com/nengo/nengo/pull/1052#issuecomment-224407418 for a summary of that. I think if deselecting is possible, and especially if we can print the deselection reason, that would be preferable.

Let me know what you think @jgosmann so we can figure out how to proceed with this PR.

jgosmann:
Sounds reasonable. I had an initial look into the pytest documentation/source code and think that the deselection and generation of messages is possible. I'll probably take a stab at implementing it tomorrow and it will presumably a new PR (rendering this one obsolete).

tbekolay:
Sounds great; if you open a new PR then close this one and I'll look at the new one right away.

Issue #649, opened 2015-02-08
how to load in weights from a previous trial


studywolf:
Heyoo just wondering how to go about setting the connection weights of a connection? 

Or whatever it is I need to do to be able to learn a set of weights, write those to file, then regenerate the model to resume from the last point of learning. Is there an example of this somewhere?

drasmuss:
What I did was create a custom decoder solver that loads in weights from file.  Something like this:
```python
class decoder_setter:
    def __init__(self, decoder_file):
        # load in decoders

    def __call__(self, A, Y, rng=None):
        return self.decoders, []
```
then you just pass that to the connection that's doing the learning, and it will initialize it with the weights you loaded, and continue on from there.  Like
```python
learn_conn = nengo.Connection(a, b, learning_rule=nengo.PES(...), decoder_solver=decoder_solver(filename), ...)
```

However, that is just a hack I put together, and it was several iterations of Nengo ago, so there may be a more elegant way to do it now.

studywolf:
thanks dan! I started with your hack and then talked with @hunse, had to modify it a bit for the new code, and actually getting the decoder values post learning was a pain in the ass. So here's the code I used to hack it but we need a way for noobs to do this too. Possibly related to #207 

```
import numpy as np
import nengo
from nengo.networks import EnsembleArray
import nengo.solvers
import matplotlib.pyplot as plt
nengo.log()


decoders = np.zeros((100,1))

class decoder_setter(nengo.solvers.Solver):
    def __init__(self, decoders, weights=False):
        self.weights = weights
        self.decoders = decoders

    def __call__(self, A, Y, E=None, rng=None):
        return self.decoders, [] 

decoder_solver = decoder_setter(decoders)

for ii in range(10):
    model = nengo.Network(seed=1)
    with model: 

        context = nengo.Node(output=np.sin)#[0,]*dim)
        pop1 = nengo.Ensemble(100, dimensions=1)
        pop2 = nengo.Ensemble(100, dimensions=1)
        error = nengo.Ensemble(1, dimensions=1, neuron_type=nengo.Direct())
        answer = nengo.Ensemble(1, dimensions=1, neuron_type=nengo.Direct())

        nengo.Connection(context, pop1)
        # desired signal
        nengo.Connection(context, error, function=lambda x: x**2)
        # actual signal
        nengo.Connection(pop2, error, transform=-1)
        # error signal
        err = nengo.Connection(error, pop2, modulatory=True)
        # learned connection
[...]

Issue #608, opened 2015-01-02
Add ability to save and load connection weights for learned connections


Seanny123:
In Dan's Hierarchical Reinforcement Learning stuff, there's a piece of code that allows you to save connections weights and load them again for a connection that acts as the "state memory" of the network. This is useful for testing something that breaks at a certain state, since it allows you to revisit the exact point of failure again and again. Is this already built into Nengo and if it isn't, should this be something that should be included?

hunse:
This is very possible with current Nengo: just grab the weights after the simulation, save them, and then load them and use them as the initial weights next time you build the network.

It would be a good idea to at least a) show how to do this in a notebook, maybe b) add some helper functions to make this easier, or c) go the whole nine yards and build it right into Nengo (though I'm not quite sure how this would work).

drasmuss:
It is nice to be able to save the weights during the simulation (since often part of the reason you're saving weights is to prevent losing everything if the simulation crashes).  You could do that right now by taking over the simulation loop yourself (i.e., manually calling `step`, with weight saving interspersed).  This is a couple times now though where I've thought it would be really useful if the `Node` function could somehow get a handle to the `model` object whenever it is called.  It's kind of weird, because when you're designing the `Node` function you'll have to be working with objects that don't exist yet, and it kind of breaks the design/simulation barrier we've constructed.  So I wouldn't want to make it a front end feature, but maybe something a bit hidden for advanced users?

tcstewar:
> This is a couple times now though where I've thought it would be really useful if the Node function could somehow get a handle to the model object whenever it is called.

I tend to do this either by using a default value for a function argument or by defining a class

```python
model = nengo.Network()
with model:
    def function(t, model=model):
        print t, model
    node = nengo.Node(function)
```

or

```python
class MyNodeFunction(object):
    def __init__(self, model):
        self.model = model
    def __call__(self, t):
        print t, self.model

model = nengo.Network()
with model:
    node = nengo.Node(MyNodeFunction(model))
```

I think that approach is better than adding special-case advanced features to the API, but it'd be good to have an example of doing it...

drasmuss:
Sorry, I meant the simulator Model class, not Network.  That's what you need if you want access to the online connection weights and things like that.

tcstewar:
Ah.  Good point.  hmm, well, the first approach still works if you just rely on global variables, but it feels a bit ugly to me:

```python
import nengo

model = nengo.Network()
with model:
    def function(t):
[...]

PR #1352, opened 2017-09-06
Added 'NoSolver' solver

bmorcos:
**Motivation and context:**
It has come up a few times that it would be nice to set decoders or weights directly. More specifically, loading from previous runs as mentioned in #649, #608. This doesn't facilitate saving and loading per se, but does allow the user to set decoders or weights even if they do not have access to `pre.neurons`. This can be used to help load and save decoders or weights as outlined in nengo/nengo_extras#35.

**How has this been tested?**
Added a new test, all current tests passed.

Just look at the commit.

**Types of changes:**

**Checklist:**

**Still to do:**
- [ ] Add some error checking on `values` to give better error messages for shape mismatch

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,12 @@ Release History
 2.5.1 (unreleased)
 ==================
 
+**Added**
+
+- Added a ``NoSolver`` solver that can be used to manually pass in
+  a predefined set of decoders or weights to a connection.
+  (`#1352 <https://github.com/nengo/nengo/pull/1352>`_)
+
 **Changed**
 
 - Learning rules can now have a learning rate of 0.
--- a/docs/frontend_api.rst
+++ b/docs/frontend_api.rst
@@ -204,6 +204,7 @@ Decoder and connection weight solvers
    nengo.solvers.Nnls
    nengo.solvers.NnlsL2
    nengo.solvers.NnlsL2nz
+   nengo.solvers.NoSolver
 
 .. autoclass:: nengo.solvers.Solver
    :special-members: __call__
@@ -227,3 +228,5 @@ Decoder and connection weight solvers
 .. autoclass:: nengo.solvers.NnlsL2
 
 .. autoclass:: nengo.solvers.NnlsL2nz
+
+.. autoclass:: nengo.solvers.NoSolver
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -12,7 +12,7 @@ from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError, ObsoleteError
 from nengo.neurons import Direct
 from nengo.node import Node
-from nengo.solvers import Solver
+from nengo.solvers import NoSolver, Solver
 from nengo.utils.compat import is_iterable, itervalues
 
 built_attrs = ['eval_points', 'solver_info', 'weights', 'transform']
@@ -162,6 +162,18 @@ def slice_signal(model, signal, sl):
 @Builder.register(Solver)
 def build_solver(model, solver, conn, rng, transform):
     return build_decoders(model, conn, rng, transform)
+
+
+@Builder.register(NoSolver)
+def build_no_solver(model, solver, conn, rng, transform):
+    activities = np.zeros((1, conn.pre_obj.n_neurons))
+    targets = np.zeros((1, conn.size_mid))
+    E = np.zeros((1, conn.post_obj.n_neurons)) if solver.weights else None
+    # No need to invoke the cache for NoSolver
+    decoders, solver_info = conn.solver(activities, targets, rng=rng, E=E)
+    weights = (decoders.T if conn.solver.weights else
+               multiply(transform, decoders.T))
+    return None, weights, solver_info
 
 
 @Builder.register(Connection)  # noqa: C901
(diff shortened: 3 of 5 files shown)


bmorcos:
I'm wondering if we want to add warnings on high values too maybe?

arvoelke:
Yay! I've needed this before, as has @psipeter.

But, I am wondering about the name `PassThrough`. In signal processing, [passthrough](https://en.wikipedia.org/wiki/Passthrough) has a pretty specific meaning. For instance, the `D` matrix in a state-space model (linear system) is often called the passthrough matrix [1, p. 427], and [I use this terminology](https://arvoelke.github.io/nengolib-docs/nengolib.signal.LinearSystem.html?highlight=passthrough) in nengolib.

Might I suggest a name like `StaticWeights`, `FixedWeights`, or `PreSolvedWeights`?

Also, note that the variable `self.value` will change from `None` before building to something else afterward. This may be unexpected if the user (for whatever reason) is accessing `conn.solver.value` both before build and after build. Most objects in Nengo appear to be frozen (immutable) unless there's some need otherwise. More importantly, this will cause a bug if the same solver instance is used for multiple connections. It's certainly not common to see, but it's technically legal to create a single solver instance (`solver=Passthrough()`) and then pass this same `solver` to multiple connections. Currently, your solver would change `self.value` when the first connection is solved, providing a shape that will not necessarily match the `A` (or `Y` or `E`) needed for the second connection. Removing the mutation should fix both issues.

[1] Seeler, K.A. _System Dynamics: An Introduction for Mechanical Engineers_. Springer New York, 2014.

Seanny123:
At the risk of bike-shedding, how about `InitialWeights` instead of `PreSolvedWeights`?

bmorcos:
I think I like `InitialWeights` as these aren't necessarily fixed/static. 

>Most objects in Nengo appear to be frozen (immutable) unless there's some need otherwise.

We don't have access to `n_neurons`, `dimensions`, or `post.n_neurons` until `__call__` which dictates the shape. I thought it would be nice to default to zeros for empty args, but we don't have to do this.

>Currently, your solver would change self.value when the first connection is solved, providing a shape that will not necessarily match the A (or Y or E) needed for the second connection

This is a tricky one I think... if we are using `None` to default to zeros, then it would be nice to have one instance create all the correct shapes. For the broader use of setting weights, you couldn't use the same instance of the solver unless the connection shape was identical anyways. Maybe we just add an error or warning about this?

tbekolay:
This is a hard one to name for sure... one other possibility to throw out there is `NoSolver`, since one important aspect of this class is that no decoder solving will be happening. I think I prefer that to `IntialWeights`, which is confusing in the context of `nengo.Connection`; e.g.:

```python
nengo.Connection(pre, post, solver=InitialWeights(myweights))
nengo.Connection(pre, post, solver=NoSolver(myweights))
```

I also think `Passthrough` makes some sense in the `nengo.Connection` context too, but `NoSolver` is probably better. But yeah, in general I think that a term having a definition in some other context is a good thing (and not a dealbreaker) as that general idea can translate to this context (as long as it's not confusing, which I don't think it is in this case).

> I thought it would be nice to default to zeros for empty args, but we don't have to do this.

I think the comment about immutability can be easily resolved by returning all zeros without setting `self.value`. There's no reason to change `self.value` in the `__call__` method; all that matters is what is returned.

bmorcos:
That commit addresses a couple points, I think we still need to talk about error checking here though. if there is a shape mismatch the current error is not very obvious:

```Traceback (most recent call last):
  File "/home/ben/Git/nengo/nengo/simulator.py", line 156, in __init__
    self.model.build(network, progress_bar=self.progress_bar)
  File "/home/ben/Git/nengo/nengo/builder/builder.py", line 123, in build
    built = self.builder.build(self, obj, *args, **kwargs)
  File "/home/ben/Git/nengo/nengo/builder/builder.py", line 218, in build
  File "/home/ben/Git/nengo/nengo/builder/network.py", line 118, in build_network
    model.build(conn)
  File "/home/ben/Git/nengo/nengo/builder/builder.py", line 123, in build
    built = self.builder.build(self, obj, *args, **kwargs)
  File "/home/ben/Git/nengo/nengo/builder/builder.py", line 218, in build
  File "/home/ben/Git/nengo/nengo/builder/connection.py", line 264, in build_connection
    tag="%s.weights_elementwiseinc" % conn))
  File "/home/ben/Git/nengo/nengo/builder/operator.py", line 554, in __init__
    A.initial_value, X.initial_value, Y.initial_value, self.tag)
  File "/home/ben/Git/nengo/nengo/builder/operator.py", line 499, in reshape_dot
    % (tag, A.shape, X.shape, Y.shape))
nengo.exceptions.BuildError: shape mismatch in <Connection from <Ensemble (unlabeled) at 0x7f997cc1fc50> to <Ensemble (unlabeled) at 0x7f997cc2a908>>.weights_elementwiseinc: (2, 2) x (10,) -> (2,)
```

To reiterate again what @arvoelke brought up, we might want to at least warn if the same solver instance is used multiple times?

arvoelke:
> To reiterate again what @arvoelke brought up, we might want to at least warn if the same solver instance is used multiple times?

Your changes at a glance look to fix this now, so a warning shouldn't be necessary. A legitimate use-case would be using the config system to set initial weights to zero for each ensemble in an `EnsembleArray`, and from what I can see there is no longer a potential issue here (perhaps a unit test is in order?). 👍  There may still be some more error checking needed for the case when values are supplied however. Haven't given this thought.

PR #1355, opened 2017-09-11
Piecewise param

AllenHW:
**Motivation and context:**
piecewise() function accepts a dictionary of time points and corresponding function/constants. We want to turn this into be a process. Thus we need a new parameter type(s) used to define a piecewise function.

Previous PR used two NdArrays, tp and yp to define a piecewise function. This limits users from defining a piecewise function that includes non-constant subdomains (i.e. sin() from 0.5 < t < 1). The following are two options I considered:

Option 1: use NdArray for tp, and add a parameter NumericalSeries, which represents a list of either numerical constants OR callable functions whose output dimentions are the same
Option 2 (chosen):  Add a dictionary parameter who keys are time points and values are either callable functions or numerical constants

Both parameters would be specific to piecewise process. I chose option 2 to maintain previous interface of piecewise() function

This change includes changes from https://github.com/nengo/nengo/pull/1100. Addition of PiecewiseDataParam is included.

**How has this been tested?**
Added new unit tests andran them

Please have a look at the first PR. Then, have a look at PiecewiseDataParam to see if the logic makes sense, as well as the choice of such parameter type.

**Types of changes:**

**Checklist:**

**Still to do:**

### DIFF
--- a/nengo/processes.py
+++ b/nengo/processes.py
@@ -8,6 +8,7 @@ from nengo.exceptions import ValidationError
 from nengo.params import (
     BoolParam, DictParam, EnumParam, NdarrayParam, NumberParam)
 from nengo.synapses import LinearFilter, Lowpass, SynapseParam
+from nengo.utils.compat import is_number, iteritems, itervalues
 
 
 class WhiteNoise(Process):
@@ -264,6 +265,46 @@ class PresentInput(Process):
         return step_presentinput
 
 
+class PiecewiseDataParam(DictParam):
+    """Piecewise-specific validation for the data dictionary.
+
+    In the `.Piecewise` data dict, the keys are points in time (float) and
+    values are numerical constants or callables of the same dimensionality.
+    """
+
+    def coerce(self, instance, data):
+        data = super(PiecewiseDataParam, self).coerce(instance, data)
+
+        size_out = None
+        for time, value in iteritems(data):
+            if not is_number(time):
+                raise ValidationError("Keys must be times (floats or ints), "
+                                      "not %r" % type(time).__name__,
+                                      attr='data', obj=instance)
+
+            # figure out the length of this item
+            if callable(value):
+                try:
+                    value = np.ravel(value(time))
+                except:
+                    raise ValidationError("callable object for time step %.3f "
+                                          "should return a numerical constant"
+                                          % time, attr='data', obj=instance)
+            else:
+                value = np.ravel(value)
+                data[time] = value
+            size = value.size
+
+            # make sure this is the same size as previous items
+            if size != size_out and size_out is not None:
+                raise ValidationError("time %g has size %d instead of %d" %
+                                      (time, size, size_out),
+                                      attr='data', obj=instance)
+            size_out = size
+
+        return data
+
+
 class Piecewise(Process):
     """A piecewise function with different options for interpolation.
 
@@ -303,7 +344,8 @@ class Piecewise(Process):
     data : dict
         A dictionary mapping times to the values that should be emitted
         at those times. Times must be numbers (ints or floats), while values
-        can be numbers, lists of numbers, or numpy arrays of numbers.
+        can be numbers, lists of numbers, numpy arrays of numbers,
+        or callables that return any of those options.
     interpolation : str, optional (Default: 'zero')
         One of 'linear', 'nearest', 'slinear', 'quadratic', 'cubic', or 'zero'.
         Specifies how to interpolate between times with specified value.
@@ -313,16 +355,17 @@ class Piecewise(Process):
 
     Attributes
     ----------
+    data : dict
+        A dictionary mapping times to the values that should be emitted
+        at those times. Times are numbers (ints or floats), while values
+        can be numbers, lists of numbers, numpy arrays of numbers,
+        or callables that return any of those options.
     interpolation : str
         One of 'linear', 'nearest', 'slinear', 'quadratic', 'cubic', or 'zero'.
         Specifies how to interpolate between times with specified value.
         'zero' creates a plain piecewise function whose values change at
         corresponding time points, while all other options interpolate
         as described in `scipy.interpolate`.
-    tp : ndarray
-        Specified time points.
-    yp : ndarray
-        Values corresponding to the time points, ``tp``.
 
     Examples
     --------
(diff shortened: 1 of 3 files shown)


tbekolay:
I'm :+1: on the param in general; it's not very long so it doesn't add much extra complexity.

nengo/params.py
```diff
@@ -350,6 +350,35 @@ def coerce(self, instance, value):
...
+
+    def coerce(self, instance, data):
+        data = super(PiecewiseData, self).coerce(instance, data)
+
+        self.interpolatable = True
+        output_length = None
+        for time, value in data.iteritems():
+            if not is_number(time):
+                raise ValidationError("Keys must be times (floats or ints), not %r"
+                                      % type(time).__name__, attr='data')
```
tbekolay: Should also pass `obj=instance` here.

nengo/params.py
```diff
@@ -350,6 +350,35 @@ def coerce(self, instance, value):
...
+
+            # figure out the length of this item
+            if callable(value):
+                value = value(0.1)
+            length = np.asarray(value).size
+
+            # make sure this is the same length as previous items
+            if length != output_length and output_length is not None:
+                raise ValidationError("time %g has %d items instead of %d" %
+                                      (time, length, output_length), attr='data')
```
tbekolay: `obj=instance` here too

nengo/processes.py
```diff
@@ -4,7 +4,7 @@
 from nengo.base import Process
 from nengo.dists import DistributionParam, Gaussian
 from nengo.exceptions import ValidationError
-from nengo.params import BoolParam, DictParam, NdarrayParam, NumberParam
+from nengo.params import BoolParam, DictParam, EnumParam, NdarrayParam, NumberParam, PiecewiseDataParam
```
tbekolay: This line is too long, should be < 80 characters.

nengo/processes.py
```diff
@@ -260,3 +260,129 @@ def step_presentinput(t):
...
+                raise ValidationError(
+                    "To interpolate, Scipy must be installed",
+                    attr='interpolation', obj=self)
+
+            if self._output_length() != 1:
+                raise ValidationError(
+                    "Interpolation is only supported for 1d output",
+                    attr='interpolation', obj=self)
+
+            if not self._interpolatable():
```
tbekolay: I think I would warn and switch to no interpolation here rather than raising a validation error.

nengo/processes.py
```diff
@@ -260,3 +260,129 @@ def step_presentinput(t):
...
+                raise ValidationError(
+                    "Cannot interpolate on nonconstant sub-domains",
+                    attr='interpolation', obj=self)
+
+        super(Piecewise, self).__init__(
+            default_size_in=0, default_size_out=self._output_length(), **kwargs)
+
+
+    def _interpolatable(self):
```
tbekolay: As mentioned above, this should be tracked in the param rather than here :+1:

nengo/processes.py
```diff
@@ -260,3 +260,129 @@ def step_presentinput(t):
...
+        for time, value in self.data.iteritems():
+            if callable(value):
+                return False
+        return True
+
+    def _output_length(self):
```
tbekolay: This could also be a property of the param

nengo/processes.py
```diff
@@ -260,3 +260,129 @@ def step_presentinput(t):
...
+                return 0
+
+    For times before the first specified time, it will default to zero (of
+    the correct length). This means that the above can be simplified to:
+
+        Piecewise([0.5, 0.75, 1], yp = [[1], [-1], [0]])
+
+    Parameters
+    ----------
+    tp : time points of the function where values are specified
```
tbekolay: This will have to be updated, as `tp` and `yp` are no longer attributes of the param.

nengo/tests/test_processes.py
```diff
@@ -334,3 +334,185 @@ def test_present_input(Simulator, rng):
...
+        tp, yp = zip(*data.items())
+        with pytest.raises(ValidationError):
+            process = nengo.processes.Piecewise(tp, yp)
+            assert process
+
+
+    def test_function(self, Simulator):
+        f = self.run_sim({0: np.sin, 0.5: np.cos}, 'zero', Simulator)
+
+        assert np.allclose(f[0], [np.sin(0.001)])   # t = 0.001
```
tbekolay: As I did in the other other tests, we should opt for passing in things like `t == 0.001` rather than passing in `0` and commenting the actual time (which is more understandable to the reader).

nengo/tests/test_processes.py
```diff
@@ -334,3 +334,185 @@ def test_present_input(Simulator, rng):
...
+        assert np.allclose(f[249], [np.sin(0.25)])   # t = 0.25
+        assert np.allclose(f[498], [np.sin(0.499)])  # t = 0.499
+        assert np.allclose(f[499], [np.cos(0.5)])    # t = 0.5
+        assert np.allclose(f[749], [np.cos(0.75)])  # t = 0.75
+        assert np.allclose(f[999], [np.cos(1.0)])   # t = 1.0
+        assert np.allclose(f[1499], [np.cos(1.5)])  # t = 1.5
+
+
```
tbekolay: Inside a class, PEP8 says to leave one blank line between methods, not more than one.

nengo/tests/test_processes.py
```diff
@@ -334,3 +334,185 @@ def test_present_input(Simulator, rng):
...
+        assert np.allclose(f[0], [0]) # t = 0.001
+        assert np.allclose(f[499], [1]) # t = 0.5
+        assert np.allclose(f[999], [0]) # t = 1.0
+
+        f = self.run_sim({0.001: 0, 0.5: 1, 0.75: 0.5, 1.0: 0}, 'cubic', Simulator)
+        assert np.allclose(f[0], [0]) # t = 0.001
+        assert np.allclose(f[499], [1]) # t = 0.5
+        assert np.allclose(f[749], [0.5]) # t = 0.75
+        assert np.allclose(f[999], [0]) # t = 1.0
+
```
tbekolay: Additionally, PEP8 says no extra newlines at the end of a file.

nengo/processes.py
```diff
@@ -371,31 +411,39 @@ def __init__(self, data, interpolation='zero', **kwargs):
         self.interpolation = interpolation
 
         super(Piecewise, self).__init__(
-            default_size_in=0, default_size_out=self.yp[0].size, **kwargs)
+            default_size_in=0, default_size_out=self._shape_out()[0], **kwargs)
+
+    def _interpolatable(self):
```
AllenHW: I had initially thought of tracking this in the ``PiecewiseDataParam`` class, but didn't know if we could retract any info from the descriptor class.

tbekolay: Ah you're right, it is difficult to access attributes of descriptor classes. Never mind!

tbekolay:
I like the ravel change, and setting a property on the instance is done in some Nengo objects... but I think if we're going to set it on the instance, we should add a `size_out` param to it and some documentation that it will get set automatically, which seems like more effort than keeping the property. While it does seem that the property does a lot, it's really only making an iterator (should be very cheap) and getting the first item (also cheap) so I'ma revert that one, though it would have been nice to not recompute the `size_out`.

AllenHW:
Thanks for the changes! It looks neater now 😄 

Didn't know we could modify the instance from the descriptor....that's actually pretty cool. I was also thinking it'd be to nice to compute things once in the descriptor, but I'll go along with @tbekolay

tbekolay:
> I'll go along with @tbekolay

PR #1356, opened 2017-09-13
Allow learning_rate=0

drasmuss:
**Motivation and context:**
Often when working with a learning model it is helpful to quickly disable learning by setting the learning rate to 0.  Currently we don't allow this in Nengo, which isn't the end of the world but just makes things a little awkward.  I think there is pretty minimal downside to allowing it; it seems unlikely that users will accidentally set the learning rate to zero when they do actually want learning.

**Types of changes:**

**Checklist:**

- [n/a] All new and existing tests passed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,11 @@ Release History
 2.5.1 (unreleased)
 ==================
 
+**Changed**
+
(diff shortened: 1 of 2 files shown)


studywolf:
this has come up for me a bunch of times would be very handy

Issue #1354, opened 2017-09-10
Zero-division when simulating for 0 seconds


arvoelke:
```
with nengo.Network() as model:
    with nengo.Simulator(model) as sim:
        sim.run(0)
```

```
...
c:\Users\Aaron\Anaconda2\lib\site-packages\nengo\utils\progress.pyc in eta(self)
    112         float
    113         """
--> 114         if self.progress > 0.:
    115             return (
    116                 (1. - self.progress) * self.elapsed_seconds() / self.progress)

c:\Users\Aaron\Anaconda2\lib\site-packages\nengo\utils\progress.pyc in progress(self)
     88         float
     89         """
---> 90         return min(1.0, self.n_steps / self.max_steps)
     91 
     92     def elapsed_seconds(self):

ZeroDivisionError: division by zero
```

Also, strangely enough, a negative simulation time will silently go through (this will simulate for zero time-steps), as will `sim.run(0.0006)` (this will simulate one step, despite being less than `dt`), but not `sim.run(0.0005)` (this will give the same error as above).

PR #1357, opened 2017-09-13
Raise an error if sim is run for < dt seconds

tbekolay:
**Motivation and context:**
If you pass a number less than `dt` to `sim.run`, we might raise a `ZeroDivisionError` or we might do something even weirder; see #1354. This raises an explicit error if `sim.run` is called with a value less than `dt`.

**How has this been tested?**
A unit test is included.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -11,7 +11,7 @@ from nengo.builder import Model
 from nengo.builder.optimizer import optimize as opmerge_optimize
 from nengo.builder.signal import SignalDict
 from nengo.cache import get_default_decoder_cache
-from nengo.exceptions import ReadonlyError, SimulatorClosed
+from nengo.exceptions import ReadonlyError, SimulatorClosed, ValidationError
 from nengo.utils.compat import range, ResourceWarning
 from nengo.utils.graphs import toposort
 from nengo.utils.progress import ProgressTracker
@@ -272,10 +272,19 @@ class Simulator(object):
     def run(self, time_in_seconds, progress_bar=None):
         """Simulate for the given length of time.
 
+        If the given length of time is not a multiple of ``dt``,
+        it will be rounded to the nearest ``dt``. For example, if ``dt``
+        is 0.001 and ``run`` is called with ``time_in_seconds=0.0006``,
+        the simulator will advance one timestep, resulting in the actual
+        simulator time being 0.001.
+
+        The given length of time must be positive. The simulator cannot
+        be run backwards.
+
         Parameters
         ----------
         time_in_seconds : float
-            Amount of time to run the simulation for.
+            Amount of time to run the simulation for. Must be positive.
         progress_bar : bool or `.ProgressBar` or `.ProgressUpdater`, optional \
                        (Default: True)
             Progress bar for displaying the progress of the simulation run.
@@ -285,10 +294,19 @@ class Simulator(object):
             For more control over the progress bar, pass in a `.ProgressBar`
             or `.ProgressUpdater` instance.
         """
+        if time_in_seconds < 0:
+            raise ValidationError("Must be positive (got %g)"
+                                  % (time_in_seconds,), attr="time_in_seconds")
+
         steps = int(np.round(float(time_in_seconds) / self.dt))
-        logger.info("Running %s for %f seconds, or %d steps",
-                    self.model.label, time_in_seconds, steps)
-        self.run_steps(steps, progress_bar=progress_bar)
+
+        if steps == 0:
+            warnings.warn("%g results in running for 0 timesteps. Simulator "
+                          "still at time %g." % (time_in_seconds, self.time))
+        else:
+            logger.info("Running %s for %f seconds, or %d steps",
+                        self.model.label, time_in_seconds, steps)
+            self.run_steps(steps, progress_bar=progress_bar)
 
     def run_steps(self, steps, progress_bar=None):
         """Simulate for the given number of ``dt`` steps.
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,0 @@
-comment: false
-coverage:
-  status:
-    project: false
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -48,5 +48,5 @@
 - [ ] I have run the test suite locally and all tests passed.
 
 **Still to do:**
-<!--- If this is a work in progress, note below what you stil plan to do. -->
+<!--- If this is a work in progress, note below what you still plan to do. -->
 <!--- Use the task list syntax `- [ ]` so that progress can be tracked. -->
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,10 @@ Release History
   (`#947 <https://github.com/nengo/nengo/issues/947>`_)
 - Learning rules can now have a learning rate of 0.
   (`#1356 <https://github.com/nengo/nengo/pull/1356>`_)
+- Running the simulator for zero timesteps will now issue a warning,
+  and running for negative time will error.
+  (`#1354 <https://github.com/nengo/nengo/issues/1354>`_,
+  `#1357 <https://github.com/nengo/nengo/pull/1357>`_)
 
 **Fixed**
 
--- a/nengo/tests/test_simulator.py
+++ b/nengo/tests/test_simulator.py
@@ -8,7 +8,7 @@ from nengo.builder import Model
 from nengo.builder.ensemble import BuiltEnsemble
 from nengo.builder.operator import DotInc
 from nengo.builder.signal import Signal
-from nengo.exceptions import ObsoleteError, SimulatorClosed
+from nengo.exceptions import ObsoleteError, SimulatorClosed, ValidationError
 from nengo.utils.compat import ResourceWarning
 from nengo.utils.testing import warns
 
@@ -274,3 +274,14 @@ def test_probe_cache(Simulator):
         ub = np.array(sim.data[up])
 
     assert not np.allclose(ua, ub, atol=1e-1)
+
+
+def test_invalid_run_time(Simulator):
+    net = nengo.Network()
+    with Simulator(net) as sim:
+        with pytest.raises(ValidationError):
+            sim.run(-0.0001)
+        with warns(UserWarning):
+            sim.run(0)
+        sim.run(0.0006)  # Rounds up to 0.001
+        assert sim.n_steps == 1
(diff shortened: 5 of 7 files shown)


tcstewar:
I like this, but the learning_rate=0 PR (#1356) makes me wonder whether we should just run for 0 time steps if someone asks for `t<dt` .  This would be consistent with asking to run for 0.0024s with a `dt=0.001`, which just runs for 2 time steps.

jgosmann:
I think this might require some more discussion how the the `run` method should behave with “weird” values.

* Values smaller than 0 should clearly produce a reasonable exception. (We don't support going back in time.)
* To me nothing seems wrong with `sim.run(0)`. I would expect this to run 0 timesteps.
* `sim.run(t)` with `0 < t < dt` is a weird case. It would not run a timestep, but that might be surprising.
* What about `sim.run(n * dt + t)` with `0 < t < dt`? In that case there is also a time period that no time step is run for. Why should this not produce an error if `sim.run(t)` does? (But due to rounding errors it might be troublesome to expect `t` to be exactly divisible by `dt`.)

It would be possible to allow `0 < t < dt` and advance the simulation only when multiple calls to `run` accumulate to another time step. This seems to make sense to me, but adds to the implementation complexity and I'm worried about floating point errors accumulating when doing this a lot.

So the best might indeed be to drop what does not constitute a full time step without an error, even though it might be surprising (should probably be documented)?

tbekolay:
I'm OK on letting `run(0)` or `run(0.0)` silently do nothing, but I think the case of running for `0 < t < dt` is likely to indicate that something's gone wrong; e.g., they're doing a while loop running up to some time and accidentally used a stepsize less than `dt`. I agree that the accumulation that @jgosmann mentions would introduce unnecessary complexity and will be difficult to do robustly given floating point roundoff. So I think that case should result in an error.

As to the `n * dt + t` case, while I can see it being surprising, I think that the vast majority of use cases are to run for a large chunk of time (much larger than `dt`). Additionally, we make available both `sim.trange` and `sim.time` as ways of asking the simulator how far it's run, so I think we've done our due dilligence to make sure people are aware of how the simulator's handled time.

jgosmann:
> I think the case of running for 0 < t < dt is likely to indicate that something's gone wrong

Would it be fine to just produce a warning instead of failing hard with an exception? I might want to argue that `0 < t < dt` should not be counted as an error, if `n * dt + t` isn't either. But I'm fine with a warning as a compromise and would save my energy if we agree on that. A warning would give the user basically the same information, but if the user actually has a reason to use `0 < t < dt` (e.g., because it is determined in some calculation), they do not need to write special case code to avoid running the simulator with that value.

tbekolay:
Hm. I could probably be convinced to warn instead of error out on `0 < t < dt`. Anyone else have a preference? If no other thoughts, I'll make a commit switching to warning tomorrow morning.

arvoelke:
I agree with Jan that it seems unnatural to allow `t == 0` and `t >= dt` but disallow `0 < t < dt`. This makes me visualize the parameter as a line across the non-negative reals with a hole in the open interval `(0, dt)`. It feels more natural to explain / document that "simulation times must be non-negative, but we'll warn you if you effectively requested zero time steps".

tbekolay:
I'm in the process of changing it to a warning now, but one thing I see that we do is _round_ the value to the nearest timestep, rather than cutting off any values between `dt` steps. So, with a `dt` of 0.001, if the user asks us to run for 0.0014 seconds, we do 1 step, but for 0.0015, we do two steps.

This has been here from the start, so I don't super remember if it was discussed; if I had to guess, we were worried about instances in which float rounding errors would mean that we were asking for something like `0.00099999999`. But my older and wiser eyes find this weird.

So, vote on this comment for what to do:

:+1: Keep it as is, seems fine
:laugh: Switch to using `floor` instead of `round` so we only run for fully requested `dt`s, and not more
:tada: Make a separate issue about it, discuss at dev meeting

tbekolay:
Pushed a commit that warns instead of errors. Reviews welcome!

jgosmann:
> Reviews welcome!

I think we should decide on the handling of the rounding issue. (I'm actually fine with either `round` or `floor`, but why change without reason ...)

tbekolay:
> I think we should decide on the handling of the rounding issue. (I'm actually fine with either round or floor, but why change without reason ...)

IMO that shouldn't block this PR, since the rounding isn't being changed (save for the `0.5*dt <= t < dt` case, which would previously run for 1 step, which was surprising to @arvoelke in #1354).

jgosmann:
But it might influence the exact warning message or the exact cases when a warning should be given. Right now a warning is given if `time_in_seconds < self.dt`, but this will run a timestep and produce a warning if `time_in_seconds = self.dt / 2. + epsilon`. I don't think a warning should be produced when a timestep is actually run (but if we decide for `floor` no timestep would be run and there should be a warning).

tbekolay:
> this will run a timestep and produce a warning if time_in_seconds = self.dt / 2. + epsilon

It warns and then returns, so no timestep is run. Though that is a behavior change from what is in master.

jgosmann:
Righto, missed that when skimming the code. I think whatever we do it should be consistent. If `dt / 2 + epsilon` runs no timestep and `1.5 * dt + epsilon` runs 2 timesteps, it is not consistent.

Also, it might make sense to return after the `logger.info` call to log the requested running duration and actual number of timesteps run? Checking wether the number of timesteps is <= 0 might even make the code shorter and be actually independent of wether we change from `round` to `floor`?

tbekolay:
Alright. We'll delay until we discuss at the next dev meeting.

tbekolay:
I updated this PR as per the discussion at the recent dev meeting. Please review!

nengo/simulator.py
```diff
@@ -285,10 +291,18 @@ def run(self, time_in_seconds, progress_bar=None):
...
             or `.ProgressUpdater` instance.
         """
-        steps = int(np.round(float(time_in_seconds) / self.dt))
-        logger.info("Running %s for %f seconds, or %d steps",
-                    self.model.label, time_in_seconds, steps)
-        self.run_steps(steps, progress_bar=progress_bar)
+        try:
+            steps = int(np.round(float(time_in_seconds) / self.dt))
+        except ZeroDivisionError:
+            steps = 0
```
jgosmann: This can only produce a `ZeroDivisionError` when `dt == 0.`. In that case an exception should be raised because that is indeed an invalid value and running the simulator will never succeed. Though this might not be the place to check it (`Simulator.__init__` or `Model.__init__` might be better places and `ValueError` or `ValidationError` are probably better errors).

I could also see not checking `dt` at all because I don't think we have any guideline when to check argument ranges and when not? (We do it whereever the parameter system is involved by means of the assignment, but probably not for most other arguments.)

nengo/simulator.py
```diff
@@ -285,10 +291,18 @@ def run(self, time_in_seconds, progress_bar=None):
...
+
+        if steps <= 0:
+            warnings.warn("0 timesteps requested. Simulator still at time %g."
+                          % (self.time,))
```
jgosmann: It actually might make sense to raise an exception for `steps < 0` because running backwards is clearly wrong. (But not for `steps = 0` which might be intentional as running a Simulator for 0 steps should be a no-op. Though it could indicate a problem which warrants the warning.)

It might make sense to include the requested `time_in_seconds` in the message? That might help to see why 0 timesteps where requested ... (e.g., did I pass an actual 0s or was the duration > 0s, but too short?)

nengo/simulator.py
```diff
@@ -285,10 +291,18 @@ def run(self, time_in_seconds, progress_bar=None):
...
+        else:
+            logger.info("Running %s for %f seconds, or %d steps",
+                        self.model.label, time_in_seconds, steps)
```
jgosmann: Would it make sense to log all requested runs? That might be helpful debug information if someone wonders why the simulator is not advancing as much as it should (though it might be redundant with the warning).

nengo/simulator.py
jgosmann: The warning type should be a `NengoWarning`, maybe even a specific subtype to make it easy to silence the warning in programs where it is expected.

nengo/simulator.py
tbekolay: Right... I did this because I figured the zero division happened here, but looking back at the original issue (#1354) the zero division happens in the progress bar when running for 0 timesteps. I guess I'll remove this, since now we only call `run_steps` if `steps > 0` but I guess I'll also fix the progress bar stuff...

In general I think we should always fail fast if we can.

nengo/simulator.py
tbekolay: No one mentioned this at the dev meeting and, in general, people were in favor of not stopping things, bothering the user, or changing existing behavior... I would be OK adding a different warning for negative values, but I feel like I would want other people to weigh in before adding an exception, and this PRs already dragged on for some time...

We raise lots of warnings that aren't `NengoWarning`. Unlike with exceptions, it's already easy to filter out warnings by module (see [filterwarnings](https://docs.python.org/3.5/library/warnings.html#warnings.filterwarnings); it has a `module` argument)... I can see the advantage of doing it, but not so much if we're not doing it consistently right now.

nengo/simulator.py
tbekolay: You can set up the logger to [log all warnings](https://docs.python.org/3/library/logging.html#logging.captureWarnings), but yeah I don't have a strong preference either way...

nengo/simulator.py
jgosmann: > No one mentioned this at the dev meeting and, in general, people were in favor of not stopping things, bothering the user, or changing existing behavior...

I think we didn't discuss `steps < 0` explicitely and were more focussed on the rounding issue? Definitely wouldn't mind other people's opinion on this. I think it can be quite important to get the details right of what is an actual error condition, just a warning, or silently passes. But I am also somewhat traumatized by Matlab where too many things just silently pass (and various other software where an error is raised when things should silently pass).

I'm fine with leaving the warning type as `UserWarning` for now. It seems more discussion of what the best practices in that regard is required, but that doesn't need to block this PR as it isn't consistent in the code base right now anyways.

tbekolay:
Can we get a quick poll on the issue of warning or erroring on passing negative numbers to `run`? Add emoji reactions as below and I'll go with the majority opinion.

:laughing: Treat negative numbers the same as 0 (warn that no steps will be run)
:tada: Raise an exception for negative numbers

tbekolay:
Slight consensus for raising an exception. I updated the PR as a result, please re-review.

PR #1359, opened 2017-09-20
Set rng explicitly when caching solver.

### LINKED ISSUE #1358: Custom solver __call__ must have default arguments
Minimal example:

```
import nengo

class MySolver(nengo.solvers.Solver):
    def __call__(self, *args, **kwargs):
        return super(MySover, self).__call__(*args, **kwargs)
    
with nengo.Network(seed=0) as model:
    nengo.Connection(nengo.Ensemble(1, 1), nengo.Ensemble(1, 1), solver=MySolver())
    nengo.Simulator(model)
```

```
...
c:\Users\Aaron\Anaconda2\lib\site-packages\nengo\cache.pyc in cached_solver(conn, gain, bias, x, targets, rng, E, **uncached_kwargs)
    653             except TypeError:
    654                 args, _, _, defaults = inspect.getargspec(conn.solver.__call__)
--> 655             args = args[-len(defaults):]
    656             if rng is None and 'rng' in args:
    657                 rng = defaults[args.index('rng')]

TypeError: object of type 'NoneType' has no len()
```

I'm assuming that it's expecting the signature `def __call__(self, A, Y, rng=None, E=None)` which has default arguments. My case doesn't have any since I delegate the `__call__` to a sub-solver.

jgosmann:
**Motivation and context:**
As noted in #1358 trying to inspect the default arguments has two problems:

* It prevents the usage of the ``(*args, **kwargs)`` idiom to write a wrapper for solvers.
* The solvers don't specify the actual RNG being used as default argument, but `None` and retrieve `np.random` later. So the RNG state was not saved anyways when relying on the default arguments (but I believe the builder always sets the `rng` argument, so that this bug did not ocurr in practice).

Instead of inspecting the defaults, `E` is assumed to be always `None` by default (any other default would seem quite weird to me) and `rng` is set to `np.random` if `None`.

By setting the RNG explicitly, all cached solvers will depend on the RNG state. This means that decoders are only reused if the RNG state matches even though the solver might be independent of the RNG. Unfortunately, there is no way to know whether the solver will use the RNG and this issue did exist before.

Fixes #1358.

none

**How has this been tested?**
Unit tests still pass. I did not test the original issue reported in #1358, but as the failing code was completely removed, it should be resolved. Not sure if we should add a regression test. It seems weird to write a test that tests that the code isn't doing certain things and it seems unlikely that #1358 is reintroduced. I also did not add a test for the hashing of the RNG state because it seems to be quite awkward to test (without gaining much). But I could probably be swayed to add tests.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -640,15 +640,12 @@ class DecoderCache(object):
         """
 
         def cached_solver(conn, gain, bias, x, targets,
-                          rng=None, E=None, **uncached_kwargs):
+                          rng=np.random, E=None, **uncached_kwargs):
             if not self._in_context:
                 warnings.warn("Cannot use cached solver outside of "
                               "`with cache` block.")
                 return solver_fn(conn, gain, bias, x, targets,
                                  rng=rng, E=E, **uncached_kwargs)
-
-            if rng is None:
-                rng = np.random
 
             try:
                 key = self._get_cache_key(conn.solver,
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -2,7 +2,6 @@
 
 import errno
 import hashlib
-import inspect
 import logging
 import os
 import shutil
@@ -648,15 +647,8 @@ class DecoderCache(object):
                 return solver_fn(conn, gain, bias, x, targets,
                                  rng=rng, E=E, **uncached_kwargs)
 
-            try:
-                args, _, _, defaults = inspect.getargspec(conn.solver)
-            except TypeError:
-                args, _, _, defaults = inspect.getargspec(conn.solver.__call__)
-            args = args[-len(defaults):]
-            if rng is None and 'rng' in args:
-                rng = defaults[args.index('rng')]
-            if E is None and 'E' in args:
-                E = defaults[args.index('E')]
+            if rng is None:
+                rng = np.random
 
             try:
                 key = self._get_cache_key(conn.solver,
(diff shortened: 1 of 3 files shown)


nengo/cache.py
```diff
@@ -648,15 +647,8 @@ def cached_solver(conn, gain, bias, x, targets,
...
-                args, _, _, defaults = inspect.getargspec(conn.solver)
-            except TypeError:
-                args, _, _, defaults = inspect.getargspec(conn.solver.__call__)
-            args = args[-len(defaults):]
-            if rng is None and 'rng' in args:
-                rng = defaults[args.index('rng')]
-            if E is None and 'E' in args:
-                E = defaults[args.index('E')]
+            if rng is None:
+                rng = np.random
```
tbekolay: Can I ask if there's a reason to specify the default this way rather than in the arg list? I know putting mutable things as a default can be problematic, but in this case `np.random` is mutable but also global so it would be the same were it mutated.

jgosmann: I am just following what's done in [`solvers.py`](https://github.com/nengo/nengo/blob/master/nengo/solvers.py), but I was wondering the same thing.

tbekolay: I say change it both places :+1:

jgosmann: Can do! Tagging @hunse who changed it to what it is now in 63c21040fd7f5ac139a18efe552ef882d44a5f59 in case we overlooked something.

hunse: Why did I change this... that's a good question! It might have been so that all the solvers would have the same default ``rng=None``, whether they actually use an ``rng`` or not, and maybe I thought it kept the function signature a bit cleaner, too. But I'm fine making the change suggested here.

jgosmann: Changed in both places. :)

PR #1361, opened 2017-09-21
Fixes bug where multiple identical indexed slices are ignored

AllenHW:
**Motivation and context:**
As per https://github.com/nengo/nengo/issues/947, we have a bug where advanced indexing works when the indices are different but does not work when same index repeats multiple times. Eric found the api .at(), and I use it to fix the bug. 

**How has this been tested?**
I added a test *test_set_weight_solver* in *tests/test_connection.py*.

**Types of changes:**

**Checklist:**

**Still to do:**


AllenHW:
Some of the tests are failing but can't reproduce them locally because I don't have the same numpy version (I should get virtualenv..). Will look into it

AllenHW:
It looks like the issue was that for numpy 1.8, .at() does not take Ellipsis as a slice. So np.add.at(np.array([1,2]), Ellipsis, np.array([1,1])) doesn't work. Hence I had to add a silly branch statement where if the slice is Ellipsis we fall to the previous method

nengo/builder/operator.py
```diff
@@ -388,7 +388,10 @@ def make_step(self, signals, dt, rng):
 
         if inc:
             def step_copy():
-                dst[dst_slice] += src[src_slice]
+                if dst_slice == Ellipsis:
+                    dst[dst_slice] += src[src_slice]
```
AllenHW: dst_slice == Ellipsis if no slice is provided. So we'd think that dst = dst[dst_slice]
Yet,  *dst += src[src_slice]* sometimes fails because "dst referenced before assignment" because of some weird numpy internal

nengo/builder/operator.py
```diff
@@ -388,7 +388,10 @@ def make_step(self, signals, dt, rng):
 
         if inc:
             def step_copy():
-                dst[dst_slice] += src[src_slice]
+                if type(dst_slice) == type(Ellipsis):
```
AllenHW: type(dst_slice) == type(Ellipsis) if no slice is provided. So we'd think that dst = dst[dst_slice]
Yet, dst += src[src_slice] sometimes fails because "dst referenced before assignment" because of some weird numpy internal

jgosmann: I think `isinstance(dst_slice, Ellipsis)` is usually preferred for type checks (except when derived types should be explicitly excluded which is very rare).

jgosmann: Also, the slices should not change during the simulation. That should allow to move the if-branching outside of the `step_copy` function. That might have some performance impact, but no idea whether that's significant.

I'm also curious if there is a performance impact replacing `+=` with `at`.

tbekolay: I agree with Jan, the slices won't change during the simulation so the if branch should be outside the `step_copy` function. The `step` functions returned in the builder are what are actually run when you do `sim.run`, so they should be written to be as fast as possible.

Additionally, it should only be necessary to use `at` when `dst_slice` or `src_slice` are lists (or numpy arrays) of integers, correct? You could check for that also before defining the step function so that we only use `at` when we need to.

nengo/tests/test_connection.py
```diff
@@ -613,6 +613,45 @@ def test_slicing_function(Simulator, plt, seed):
...
+        sim.run(0.4)
+
+    t = sim.trange()
+    a_data = sim.data[a_probe]
+    b_data = sim.data[b_probe]
+    c_data = sim.data[c_probe]
+    d_data = sim.data[d_probe]
+
+    plt.plot(t, a_data)
+    assert np.allclose(a_data[100:], [0], atol=0.3)
```
tbekolay: These tolerances feel pretty big to me; I'll look at the plots when I rereview, but try a larger synapse value on the probes to see if the decoded value is more stable?

nengo/tests/test_connection.py
```diff
@@ -613,6 +613,45 @@ def test_slicing_function(Simulator, plt, seed):
     assert np.allclose(w, y, atol=0.1)
 
 
+def test_advanced_indexing(Simulator, plt, seed):
+    N = 100
+
+    model = nengo.Network()
```
tbekolay: You should pass the seed into the Network (`Network(seed=seed)`) so that the test becomes deterministic; otherwise, we'll get random parameters each time and potentially random test failures! :smile:

nengo/tests/test_connection.py
```diff
@@ -613,6 +613,45 @@ def test_slicing_function(Simulator, plt, seed):
...
+    d_data = sim.data[d_probe]
+
+    plt.plot(t, a_data)
+    assert np.allclose(a_data[100:], [0], atol=0.3)
+    plt.plot(t, b_data)
+    assert np.allclose(b_data[100:], [2], atol=0.3)
+    plt.plot(t, d_data)
+    assert np.allclose(c_data[100:], [-1,1], atol=0.3)
+    plt.plot(t, d_data)
+    assert np.allclose(d_data[100:], [1,1], atol=0.3)
```
tbekolay: This test is great, thanks!

We mentioned in the dev meeting that it would also be nice to support boolean arrays for indices. Have you tried testing with boolean indices? If not, it might be worth trying it out to see if it works... if it doesn't, it makes sense to implement it in another PR, but if it just works then it'd be good to add another test for that here!

nengo/builder/operator.py
hunse: If we really want to be specific, the only time we need ``at`` is when ``dst_slice`` is a list or ndarray with repeated indices. We could check for this with something like ``len(np.unique(dst_slice)) < len(dst_slice)``.

hunse:
I did a test, and ``at`` seems to be considerably slower than ``+=`` (at least three times slower, and much more if ``dst_slice`` is a slice or ``Ellipsis`` or something). So I think we should only use ``at`` if ``dst_slice`` has repeated indices.

Here's a script if anyone wants to play around.

```python
import timeit
import numpy as np

m = int(1e6)

if 0:
    n = m
    inds = Ellipsis
elif 0:
    n = int(0.9*m)
    inds = slice(5, 5+n)
else:
    n = int(0.9*m)
    inds = np.random.permutation(m)[:n]
    # inds = np.sort(inds)

a = np.random.uniform(-1, 1, size=n)
b = np.zeros(m)

t_inc = min(timeit.repeat(
    'b[inds] += a', 'from __main__ import *; b[:] = 0', repeat=3, number=1))
t_at = min(timeit.repeat(
    'np.add.at(b, inds, a)', 'from __main__ import *; b[:] = 0', repeat=3, number=1))

print(t_inc, t_at)
print(t_at/t_inc)
```

nengo/builder/operator.py
```diff
@@ -388,7 +388,10 @@ def make_step(self, signals, dt, rng):
...
+                    dst[dst_slice] += src[src_slice]
+                else:
+                    np.add.at(dst, dst_slice, src[src_slice])
         else:
             def step_copy():
                 dst[dst_slice] = src[src_slice]
```
hunse: In this case, if there are repeated indices in ``dst_slice``, then the behaviour is kind of undefined. I think we should have an assertion here to make sure that's not the case. I don't think a user should ever be able to trip that assertion, but it could save a developer some trouble.

hunse:
I'm going to go through and make some of the changes I suggested.

nengo/tests/test_connection.py
hunse: I tried this and it seems to work. I'll add a test for it.

AllenHW:
LGTM :shipit:

PR #1362, opened 2017-09-25
Fix bugs with Piecewise process output shape

drasmuss:
**Motivation and context:**
This fixes two related bugs with the Piecewise process:

1. `Piecewise({1: [1, 1, 1]})` would return `0.0` for times < 1 (rather than `[0, 0, 0]`).
2. `Piecewise({1: 1})` would return an array with shape `()` for times < 1 and shape `(1,)` for times >= 1, 
    because `ravel` upsizes its inputs to at least 1d.  This doesn't really produce any errors, but it's 
    probably good to be consistent.  I thought about adding some logic to check if the output shape 
    should be `()` or `(1,)`, but then figured that it was probably easier/more consistent to just always have 
    the output of the process be an array, instead of a scalar.

I wasn't sure whether to add a changelog entry or not.  Technically a bugfix, but it comes so closely on the heels of the feature release that it seems unnecessary (anyone looking at the changelog would probably be seeing both things for the first time).

**How has this been tested?**

Added a new test, see `test_shape_out`

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,8 @@ Release History
 - Added a ``Piecewise`` process, which replaces the now deprecated
   ``piecewise`` function.
   (`#1036 <https://github.com/nengo/nengo/issues/1036>`_,
-  `#1100 <https://github.com/nengo/nengo/pull/1100>`_)
+  `#1100 <https://github.com/nengo/nengo/pull/1100>`_,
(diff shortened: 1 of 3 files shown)


drasmuss:
Looks good, and that makes sense with the changelog entry.

Issue #1370, opened 2017-10-17
Misleading error thrown when invalid input returned from node function


Seanny123:
Running:

```
import nengo

with nengo.Network() as model:
    in_nd = nengo.Node(lambda t: "what")

with nengo.Simulator(model) as sim:
    sim.run(1)
```

Gives:
```
Building finished in 0:00:01.                                                   
Simulating finished in 0:00:01.                                                 
Traceback (most recent call last):
  File "invalid_node.py", line 8, in <module>
    sim.run(1)
  File "/home/saubin/git/nengo/nengo/simulator.py", line 309, in run
    self.run_steps(steps, progress_bar=progress_bar)
  File "/home/saubin/git/nengo/nengo/simulator.py", line 331, in run_steps
    self.step()
  File "/home/saubin/git/nengo/nengo/simulator.py", line 342, in step
    step_fn()
  File "/home/saubin/git/nengo/nengo/builder/operator.py", line 765, in step_simpyfunc
    if y is None or not np.all(np.isfinite(y)):
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

Which doesn't actually tell you:

1. which function is misbehaving
2. what the invalid type that was returned

I feel like this was discussed before and fixed, but then was broken as some point? I can't find the previous issue discussing this.

Issue #1232, opened 2016-12-12
Setting Oja learning rate too high gives an exception


jgosmann:
Setting the learning rate of the `Oja` rule to `0.1` raises following exception during the simulation:
```
---------------------------------------------------------------------------
FloatingPointError                        Traceback (most recent call last)
<ipython-input-8-169e0758f028> in <module>()
      1 with nengo.Simulator(model) as sim:
----> 2     sim.run(3.2)

/home/jgosmann/.virtualenvs/bleeding-edge/lib/python3.5/site-packages/nengo-2.1.1.dev0-py3.5.egg/nengo/simulator.py in run(self, time_in_seconds, progress_bar)
    268         logger.info("Running %s for %f seconds, or %d steps",
    269                     self.model.label, time_in_seconds, steps)
--> 270         self.run_steps(steps, progress_bar=progress_bar)
    271 
    272     def run_steps(self, steps, progress_bar=True):

/home/jgosmann/.virtualenvs/bleeding-edge/lib/python3.5/site-packages/nengo-2.1.1.dev0-py3.5.egg/nengo/simulator.py in run_steps(self, steps, progress_bar)
    288         with ProgressTracker(steps, progress_bar) as progress:
    289             for i in range(steps):
--> 290                 self.step()
    291                 progress.step()
    292 

/home/jgosmann/.virtualenvs/bleeding-edge/lib/python3.5/site-packages/nengo-2.1.1.dev0-py3.5.egg/nengo/simulator.py in step(self)
    299         try:
    300             for step_fn in self._steps:
--> 301                 step_fn()
    302         finally:
    303             np.seterr(**old_err)

/home/jgosmann/.virtualenvs/bleeding-edge/lib/python3.5/site-packages/nengo-2.1.1.dev0-py3.5.egg/nengo/builder/neurons.py in step_simneurons()
     66 
     67         def step_simneurons():
---> 68             self.neurons.step_math(dt, J, output, *states)
     69         return step_simneurons
     70 

/home/jgosmann/.virtualenvs/bleeding-edge/lib/python3.5/site-packages/nengo-2.1.1.dev0-py3.5.egg/nengo/neurons.py in step_math(self, dt, J, spiked, voltage, refractory_time)
    311         # since v(t) = v(0) + (J - v(0))*(1 - exp(-t/tau)) assuming
    312         # J is constant over the interval [t, t + dt)
--> 313         voltage -= (J - voltage) * np.expm1(-delta_t / self.tau_rc)
[...]

PR #1371, opened 2017-10-17
Add warning for invalid Node return types

Seanny123:
**Motivation and context:**
Solves #1370. Not sure if this needs a changelog entry or not.

**How has this been tested?**
Ran the code in #1370 and it returned a legible error. Also ran all tests. Didn't add a unit test, because `operators.py` isn't tested?

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,8 +22,10 @@ Release History
 2.6.1 (unreleased)
 ==================
 
+**Fixed**
 
-
+- Better error message for invalid return values in ``nengo.Node`` functions.
+  (`#1317 <https://github.com/nengo/nengo/pull/1317>`_)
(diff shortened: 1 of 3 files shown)


jgosmann:
This relates to the discussion we had about #1232 (just to cross-reference it).

[A try/except-block is pretty cheap](https://stackoverflow.com/questions/2522005/cost-of-exception-handlers-in-python), so I don't think performance should be an issue.

One suggestion though: The new try/except block is followed by another try/except block raising pretty much the same error (just with less information). Those two blocks could be merged into a single try/except.

jgosmann:
I think a changelog entry would be nice. I would see this as a bugfix.

PR #1375, opened 2017-10-25
Rich display progress bar

jgosmann:
**Motivation and context:**
`%load_ext nengo.ipynb` and the IPython progress bar currently break in in recent Jupyter notebook/IPython versions (potentially completely crashing the notebook frontend). Also, the implementation based on widgets has proven to be not very robust as the widget API has been evolving quickly.

This PR implements a new `IPython5ProgressBar` compatible with IPython 5 and newer. The HTML code is based on the old `IPython2ProgressBar`, but instead of using widgets, the IPython rich display system is used which allows updates to displays since IPython 5. As we do not require to capture any user interaction that is completely sufficient. It also allows us to display the HTML progress bar in notebooks without `%load_ext nengo.ipynb`.

The notebook extension `nengo.ipynb` is deprecated with this PR as well as the `IPython2ProgressBar` and loading it will issue a warning. At some point `%load_ext nengo.ipynb` should become a no-op or deleted (but that would break backwards compatibility).

none

**How has this been tested?**
Tested the following scenarios:
* Progress bar in Jupyter notebook
* Progress bar when executing with `ipython` on the terminal
* Progress bar when executing with `python` on the terminal

Still to test:
* [x] What happens in `ipython qtconsole`?

**Types of changes:**

**Checklist:**

**Still to do:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,15 @@ Release History
 2.6.1 (unreleased)
 ==================
 
+**Deprecated**
+
+- The ``nengo.ipynb`` IPython extension and the ``IPython2ProgressBar``
+  have been deprecated. ``IPython5ProgressBar`` can be used instead (supporting
+  IPython>=5) and will be automatically activated in IPython kernels (e.g.,
+  in Jupyter notebooks).
+  (`#1375 <https://github.com/nengo/nengo/pull/1375>`_,
+  `#1087 <https://github.com/nengo/nengo/issues/1087>`_)
+
 **Fixed**
 
 - Better error message for invalid return values in ``nengo.Node`` functions.
--- a/examples/advanced/functions_and_tuning_curves.ipynb
+++ b/examples/advanced/functions_and_tuning_curves.ipynb
@@ -33,7 +33,6 @@
       "\n",
       "import nengo\n",
       "\n",
-      "%load_ext nengo.ipynb\n",
       "%matplotlib inline"
      ],
      "language": "python",
--- a/examples/advanced/inhibitory_gating.ipynb
+++ b/examples/advanced/inhibitory_gating.ipynb
@@ -22,8 +22,7 @@
       "import matplotlib.pyplot as plt\n",
       "%matplotlib inline\n",
       "\n",
-      "import nengo\n",
-      "%load_ext nengo.ipynb"
+      "import nengo"
      ],
      "language": "python",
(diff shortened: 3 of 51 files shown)


jgosmann:
Does not work in the QtConsole (the qt-console requests the text/html conversion, but does not actually use it).

Seanny123:
Did the previous one work in the QT Console? Even if it did, I would
consider this PR to be such an important improvement that breaking QT
Console is acceptable.

jgosmann:
The QtConsole should have defaulted to the `TerminalProgressBar`.

jgosmann:
Ok, so the current state is now:
* `python` defaults to `TerminalProgressBar`
* `ipython` defaults to `IPython5ProgressBar` which will direct calls to `TerminalProgressBar`. This is detected by the fact that `_ipython_display_` will not be invoked in this case.
* All ipython kernels connected to via Jupyter (`jupyter console`, `jupyter qtconsole`, `jupyter notebook` etc) will use `IPython5ProgressBar`, but direct calls to `HtmlProgressBar`.
  * If the frontend supports HTML display (only `notebook` at the moment), a the HTML progress bar will be displayed.
  * All other frontends will display a message to set the progress bar to `TerminalProgressBar`.

For some reason the updating of the progress bar does not work for me in Jupyter notebook, but it seems that my Jupyter notebook installation might be broken.

jgosmann:
The choices in the comment are meant to reflect that most people (I assume) use either `python` to run scripts on the console or Jupyter notebook for interactive work. So these two cases work out of the box. But of course we could change it to the prior state: Always use the terminal progress bar and activate the HTML progress bar in the notebook with a special command (`%load_ext nengo.ipynb`).

Seanny123:
I think this is fine. Is there some other reason this is still tagged as WIP?

jgosmann:
Not really, just forgot to remove the tag. Might add some more documentation maybe.

jgosmann:
So #1380 is based on this PR and rewrites some of the code. Not sure whether to close this PR and move all discussion and review to the new PR or whether to get this one merged first and then rebase the other PR?

tbekolay:
Hmm... if it rewrites code I'm inclined to close this one and move to the new PR, unless you have strong feelings for doing this one first @jgosmann

jgosmann:
Sounds perfectly fine to me.

Issue #1017, opened 2016-04-20
Add config system and decoder cache API docs


tbekolay:
I didn't get to these in #959. I think we can leave them for 2.1.1 though.

Seanny123:
Somewhat related to my confusion in #914 .

Issue #1314, opened 2017-05-16
The build progress bar can take a long time to appear.


jgosmann:
No progress bar is shown, but the total time will be printed after completion.

jgosmann:
Hm, the simulation progress is shown ... maybe this is just because the initial build step takes long and the progress bar will not be printed before completion of this first step?

tbekolay:
Yeah this tends to happen with the build, I find that only specific types of models tend to get any progress shown. But actually this reminds me, it would be great to have the optimization process hooked into the builder progress if possible, as that will now take a significant portion of the build time.

jgosmann:
Part of the reason is probably that we only start showing the progress bar if the ETA exceeds one second. If the first build step takes very long it won't be shown for quite a while.

> it would be great to have the optimization process hooked into the builder progress if possible

Indeed. Would you prefer to have it as an additional stage (so that you get three progress bars for building, optimizing, and simulating) or should it be included in the build progress bar?

I think the difficulty is mostly that we don't know in advance how many passes will run. Also, the question is how fine grained should the updating of the progress bar be done?

tcstewar:
> Would you prefer to have it as an additional stage (so that you get three progress bars for building, optimizing, and simulating) or should it be included in the build progress bar?

Definitely part of the build progress bar.  The build bar should be everything done to set up the simulation, and the simulating bar should be everything done to run the simulation.

jgosmann:
It could be useful to know how long the optimizer takes, but I suppose one could activate logging to get that information.

tcstewar:
> It could be useful to know how long the optimizer takes, but I suppose one could activate logging to get that information.

It could be useful to know that, but it could also be useful to know the total amount of time taken solving for decoders, or generating A matrices, or any number of other sub-steps in the build process.  I don't think of the progress bar as a way to get fine-grained information like that -- it's more of a high-level "hey, stuff is going on; don't worry, your computer hasn't crashed; and we're about this proportion of the way through" indicator.  :)

arvoelke:
> know the total amount of time taken solving for decoders

FYI this helper might be handy to keep in the back pocket (disclaimer: not thoroughly tested):

```
def solver_time(sim, model):
    return sum(sim.data[conn].solver_info.get('time', 0)
               for conn in model.all_connections
               if sim.data[conn].solver_info is not None)
```

jgosmann:
[...]

PR #1380, opened 2017-11-13
Progress bar bug fixes and improvements

jgosmann:
**Motivation and context:**
This introduces a number of bug fixes improvements to the progress bars.

* Despite #1340, the build progress bar was not working correctly for nested networks.
* The updating frequency in Jupyter notebooks was lower than it should have been (due to the `AutoProgressBar` not reporting the capabilities of the delegate properly).
* Allows progress bar with an unknown number of steps.
* Implements special displays for unknown number of steps for the terminal and IPython progress bars (see below).
* Uses the same space for all build related progress bars and displays the total time for all sub steps in the end. (The ETA for single steps are just for that step.)

tty
nov-12-2017 19-10-41

Open questions:

* Remove/deprecate `AutoProgressBar`? The intention of it is to not display the progress bar for short running builds or simulations (of less than one second) to not needlessly clutter up the output. This might have made more sense when there was only a single progress bar (before the build progress bar was introduced). Maybe it makes more sense to always display the progress bars now?
* The build displays progress bars for the actual building and the optimizing as these are the longest running substeps, but there are a few more steps that a progress bar might sense for:
  * Initializing the cache (acquiring the index lock might take 10 seconds before a timeout, though with #1364 this should occur more rarely).
  * Constructing the operator dependency graph.
  * Topo-sorting the dependency graph
  * Initializing the signals

The downside of introducing progress bars for these steps is that a callback needs to introduced into the code (because updating of the progress bars is synchronous, we could make it asynchronous, but that would require to spawn threads which might not be a good idea in a library).

This PR is based on #1375 and rewrites some of that code again to do updates per JavaScript (to avoid messing with the CSS animations).

**How has this been tested?**
Manually on a bigger model. Also added a unit test to test some aspects of the `Simulator` progress bar (it fails for the buggy implementation of the build progress, but passes for the fixed version).

Might make sense to go through commit by commit, though for the `HtmlProgressBar` implementation it might save time to just look at the final implementation.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/ipynb.py
+++ b/nengo/ipynb.py
@@ -44,7 +44,16 @@ else:
 
 
 def load_ipython_extension(ipython):
-    if has_ipynb_widgets() and rc.get('progress', 'progress_bar') == 'auto':
+    if IPython.version_info[0] >= 5:
+        warnings.warn(
+            "Loading the nengo.ipynb notebook extension is no longer "
+            "required. Progress bars are automatically activated for IPython "
+            "version 5 and later.")
+    elif has_ipynb_widgets() and rc.get('progress', 'progress_bar') == 'auto':
+        warnings.warn(
+            "The nengo.ipynb notebook extension is deprecated. Please upgrade "
+            "to IPython version 5 or later.")
+
         IPythonProgressWidget.load_frontend(ipython)
         rc.set('progress', 'progress_bar', '.'.join((
             __name__, IPython2ProgressBar.__name__)))
@@ -135,6 +144,9 @@ class IPythonProgressWidget(DOMWidget):
     @classmethod
     def load_frontend(cls, ipython):
         """Loads the JavaScript front-end code required by then widget."""
+        warnings.warn(
+            "IPythonProgressWidget is deprecated. Please upgrade to "
+            "IPython version 5 or later.", DeprecationWarning)
         if notebook_version[0] < 4:
             ipython.run_cell_magic('javascript', '', cls.LEGACY_FRONTEND)
         elif ipywidgets.version_info[0] < 5:
@@ -155,6 +167,9 @@ class IPython2ProgressBar(ProgressBar):
     supports_fast_ipynb_updates = True
 
     def __init__(self, task):
+        warnings.warn(
+            "IPython2ProgressBar is deprecated. Please upgrade to IPython "
+            "version 5 or later.", DeprecationWarning)
         super(IPython2ProgressBar, self).__init__(task)
         self._escaped_task = escape(task)
         self._widget = IPythonProgressWidget()
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -159,11 +159,13 @@ class Simulator(object):
                 self.model.build(network,
                                  progress=pt.next_stage("Building", "Build"))
 
-        # Order the steps (they are made in `Simulator.reset`)
-        self.dg = operator_dependency_graph(self.model.operators)
+            # Order the steps (they are made in `Simulator.reset`)
+            self.dg = operator_dependency_graph(self.model.operators)
 
-        if optimize:
-            opmerge_optimize(self.model, self.dg)
+            if optimize:
+                with pt.next_stage(
+                        'Building (running optimizer)', 'Optimization'):
+                    opmerge_optimize(self.model, self.dg)
 
         self._step_order = [op for op in toposort(self.dg)
                             if hasattr(op, 'make_step')]
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -14,7 +14,7 @@ from nengo.cache import get_default_decoder_cache
 from nengo.exceptions import ReadonlyError, SimulatorClosed, ValidationError
 from nengo.utils.compat import range, ResourceWarning
 from nengo.utils.graphs import toposort
-from nengo.utils.progress import ProgressTracker
+from nengo.utils.progress import Progress, ProgressTracker
 from nengo.utils.simulator import operator_dependency_graph
 
 logger = logging.getLogger(__name__)
@@ -152,9 +152,12 @@ class Simulator(object):
         else:
             self.model = model
 
-        if network is not None:
-            # Build the network into the model
-            self.model.build(network, progress_bar=self.progress_bar)
+        pt = ProgressTracker(progress_bar, Progress("Building", "Build"))
+        with pt:
+            if network is not None:
+                # Build the network into the model
+                self.model.build(network,
+                                 progress=pt.next_stage("Building", "Build"))
 
         # Order the steps (they are made in `Simulator.reset`)
         self.dg = operator_dependency_graph(self.model.operators)
@@ -333,10 +336,12 @@ class Simulator(object):
         """
         if progress_bar is None:
             progress_bar = self.progress_bar
-        with ProgressTracker(steps, progress_bar, "Simulating") as progress:
+
+        with ProgressTracker(progress_bar, Progress(
+                "Simulating", "Simulation", steps)) as pt:
             for i in range(steps):
                 self.step()
-                progress.step()
+                pt.total_progress.step()
 
     def step(self):
         """Advance the simulator by 1 step (``dt`` seconds)."""
--- a/nengo/utils/progress.py
+++ b/nengo/utils/progress.py
@@ -51,8 +51,8 @@ class Progress(object):
     ----------
     steps : int
         Number of completed steps.
-    max_steps : int
-        The total number of calculation steps of the process.
+    max_steps : int, optional
+        The total number of calculation steps of the process, if known.
     start_time : float
         Time stamp of the time the process was started.
     end_time : float
@@ -72,8 +72,8 @@ class Progress(object):
 
     """
 
-    def __init__(self, max_steps):
-        if max_steps <= 0:
+    def __init__(self, max_steps=None):
+        if max_steps is not None and max_steps <= 0:
             raise ValidationError("must be at least 1 (got %d)"
                                   % (max_steps,), attr="max_steps")
 
@@ -91,6 +91,8 @@ class Progress(object):
         -------
         float
         """
+        if self.max_steps is None:
+            return 0.
         return min(1.0, self.n_steps / self.max_steps)
 
     def elapsed_seconds(self):
@@ -130,7 +132,7 @@ class Progress(object):
 
     def __exit__(self, exc_type, dummy_exc_value, dummy_traceback):
         self.success = exc_type is None
-        if self.success:
+        if self.success and self.max_steps is not None:
             self.n_steps = self.max_steps
         self.end_time = time.time()
         self.finished = True
@@ -188,6 +190,8 @@ class TerminalProgressBar(ProgressBar):
     def update(self, progress):
         if progress.finished:
             line = self._get_finished_line(progress)
+        elif progress.max_steps is None:
+            line = self._get_unknown_progress_line(progress)
         else:
             line = self._get_in_progress_line(progress)
         sys.stdout.write(line)
@@ -210,6 +214,29 @@ class TerminalProgressBar(ProgressBar):
                 progress_str[:percent_pos] + percent_str +
                 progress_str[percent_pos + len(percent_str):])
 
+        return '\r' + line.format(progress_str)
+
+    def _get_unknown_progress_line(self, progress):
+        """Generates a progress line with continuously moving marker.
+
+        This is to indicate processing while not knowing how far along we
+        progressed with the processing.
+        """
+        duration = progress.elapsed_seconds()
+        line = "[{{}}] duration: {duration}".format(
+            duration=timestamp2timedelta(duration))
+        text = " {}... ".format(progress.task)
+        width, _ = get_terminal_size()
+        marker = '>>>>'
+        progress_width = max(0, width - len(line) + 2)
+        index_width = progress_width + len(marker)
+        i = int(4. * duration) % (index_width + 1)
+        progress_str = (' ' * i) + marker + (' ' * (progress_width - i))
+        progress_str = progress_str[len(marker):-len(marker)]
+        text_pos = (len(progress_str) - len(text)) // 2
+        progress_str = (
+            progress_str[:text_pos] + text +
+            progress_str[text_pos + len(text):])
         return '\r' + line.format(progress_str)
 
     def _get_finished_line(self, progress):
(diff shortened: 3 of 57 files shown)


jgosmann:
Forgot to link #1314 with some related discussion.

drasmuss:
It would be nice to be able to pass in a full `ProgressTracker` instance to `run_steps`/`build`.  This would make it easier for other processes to hook into the progress system (e.g., a different backend that wants to add extra steps to the build process).

I'm not sure if this should be instead of or in addition to the option to pass in a `ProgressBar`/`ProgressUpdate` instance?

jgosmann:
Do you mean passing it to the `model.build` function or the `Simulator.__init__`?

> I'm not sure if this should be instead of or in addition to the option to pass in a ProgressBar/ProgressUpdate instance?

I think users should still be able to set the progress bar without construction the whole `ProgressTracker` themselves.

drasmuss:
In the main use case I'm thinking of it'd be the `model.build` function (a backend that has implemented a custom `Simulator` but is still using the `nengo` build process, like `nengo_dl`/`nengo_ocl`).

jgosmann:
The `model.build` is already expecting a `ProgressTracker` (in this PR). But I'm not sure if that fully covers your use case.

drasmuss:
Ah yeah you're right sorry! Let me try integrating this into `nengo_dl`, and I'll see if I hit any stumbling blocks.

drasmuss:
Pushed some suggested changes.

3f48fcc

Refactors things a bit to simplify MultiProgressTracker.

8e3e3b6

Spawns a thread to update progress bars with unknown duration (as Jan mentioned above).  Since unknown progress bars don't depend on the process they are tracking, it's kind of a pain to force that underlying process to add function calls that are really just triggering the progress bar display.  Or, in some cases (like `nengo_dl`), it may not be possible to insert progress bar update functions into the process being tracked.

d745b7a

Adds `start`/`stop` functions to ProgressTracker.  These are basically just more accessible versions of `__enter__`/`__exit__` that allow the ProgressTracker to be used without a context block if desired.  One added benefit is that `stop` can take an argument to change the name of the task when it finishes.  This lets you do things like display "Simulation finished in 0:00:00" instead of "Simulating finished in 0:00:00".

515d0ac / f2279c8

Bugfixes

nengo/utils/progress.py
```diff
@@ -645,35 +647,32 @@ class MultiProgressTracker(object):
...
+        super(MultiProgressTracker, self).__init__(1, progress_bar, task)
 
-    def activate_subtask(self, subtask):
-        self.subtask = subtask
-
-    def subprogress(self, max_steps):
+    def subprogress(self, max_steps, subtask):
         pt = ProgressTracker(
-            max_steps, self.progress_bar, self.subtask, close=False)
-        self.subtask_progress[self.subtask] = pt.progress
```
jgosmann: I put this in do provide a way to retrieve the time of substeps (which can be useful information sometimes) ... though the instance is not exposed by the simulator tso there is no way to get access to this anyways at the moment.

nengo/utils/progress.py
```diff
@@ -645,35 +647,32 @@ class MultiProgressTracker(object):
...
+            max_steps, self.progress_bar, subtask, close=False)
+
         return pt
 
+
+class NoopProgressTracker(ProgressTracker):
+    def __init__(self, *_):
+        pass
```
jgosmann: Should call the super constructor.

drasmuss: I don't think we want to call the super constructor in this case (it'd force us to construct Progress/ProgressBar/ProgressUpdater classes that we never need).

nengo/utils/progress.py
```diff
@@ -645,35 +647,32 @@ class MultiProgressTracker(object):
...
-        self.total_progress.__exit__(exc_type, exc_value, traceback)
-        self.progress_bar.update(self.total_progress)
-        self.progress_bar.close()
+        pass
+
+    def step(self):
+        pass
+
+    def subprogress(self, *_):
+        return NoopProgressTracker()
```
jgosmann: Why does this have a `subprogress` method?

drasmuss: So that you can use it as a replacement for MultiProgressTracker or ProgressTracker.

jgosmann:
I did not have `MultiProgressTracker` inherit from `ProgressTracker` because they are not supposed to be used in different ways. Technically you could use a `MultiProgressTracker` in place of a `ProgressTracker` but it does not really make sense. 

I used the `activate_subtask` method to have all the naming of the tasks in one place instead of distributing it across different unrelated files.

What is the rationale behind f2279c8a822f47addbbc5557ddca00219fc04111?

drasmuss:
> I did not have MultiProgressTracker inherit from ProgressTracker because they are not supposed to be used in different ways. Technically you could use a MultiProgressTracker in place of a ProgressTracker but it does not really make sense.

As far as I can tell, the inherited version has all the same functionality with less code.  Is there some specific behaviour of ProgressTracker that you want to disable in MultiProgressTracker?

> I used the activate_subtask method to have all the naming of the tasks in one place instead of distributing it across different unrelated files.

I think it's easier to have the name of the subtask at the same place where the subtask is created.  Otherwise you have to look across different files (as you point out) in order to figure out what the name of this subtask is, or where this subtask you just "activated" actually starts.

> What is the rationale behind f2279c8?

If you don't have this then whenever a subtask finishes it will display the "finished" line, even though `close=False`. E.g., try

``` python
with MultiProgressTracker(TerminalProgressBar(), "test") as pt:
    time.sleep(5.0)
    with pt.subprogress(None, "test2"):
        time.sleep(2.0)

    time.sleep(5.0)

    with pt.subprogress(None, "test3"):
        time.sleep(2.0)
```

jgosmann:
Regarding d745b7a338cd0bb68faa022a2c5f16f22be9a850: I would prefer to specify both task names at outset. That keeps related information together and allows `__exit__` to also make use of it.

jgosmann:
> As far as I can tell, the inherited version has all the same functionality with less code. Is there some specific behaviour of ProgressTracker that you want to disable in MultiProgressTracker?

The `__enter__` and `__exit__` methods are the same, but `MultiProgressTracker` did not have a `step` method because one should create sub-trackers to track the progress and not do it directly on the `MultiProgressTracker`. The difference in how those two classes are used justifies to me having some code duplication. 

> I think it's easier to have the name of the subtask at the same place where the subtask is created. Otherwise you have to look across different files (as you point out) in order to figure out what the name of this subtask is, or where this subtask you just "activated" actually starts.

Maybe the subtasks should be created in the Simulator then? That would also make it more consistent where those are created because at the moment one is created in the builder and one in the Simulator for the optimizer. However, in case of the builder we need to now `max_steps`.

> If you don't have this then whenever a subtask finishes it will display the "finished" line, even though close=False. E.g., try

Yes, I consider that as intended behaviour because the progress bar accurately displays what is stored in the model. The `close` is not about whether the task is finished, but about whether a new progress bar is supposed to be displayed on the next `update`.

jgosmann:
Also a thought on threads: If we go that way, I would probably take a look at whether we can make more extensive use of an updater thread to simplify the `ProgressUpdater` hierarchy.

jgosmann:
> If we go that way, I would probably take a look at whether we can make more extensive use of an updater thread to simplify the ProgressUpdater hierarchy.

On second thought: We might not want to throw out the `ProgressUpdater` hierarchy yet for backwards compatibility?

nengo/utils/progress.py
jgosmann: I think it is bad practice to derive from a class without calling the constructor (and this caused problems in the Nengo code base before). But I added a commit where this issue becomes irrelevant anyways.

nengo/utils/progress.py
jgosmann: Upon further thought it doesn't really make sense to introduce this without a way to access it.

jgosmann:
I pushed a commit that uses a always a thread for updating of the progress bars. This (and the fact that IPython/Jupyter Notebook does not store all output anymore since version 3) allows to deprecate the `ProgressUpdater` stuff (not sure if we want to delete it right now, technically breaks backward compatibility). I also merged `ProgresserTracker` and `MultiProgressTracker` into a single class. I hope that solves mostly the issues that you had with my previous implementation, but also issues that I had with yours. The new `ProgressTracker` has a `total_progress` attribute that can be used to update the progress if the are no subtasks. Otherwise, `subtask` allows to create progress instances that can be used to track subprogresses. Instead of a progress tracker, the `Progress` instance is no passed to the builder.

Furthermore, I changed `Progress` to have an `ongoing_name` and a `finished_name`. These names are all set in the simulator. We do not need to know the `max_steps` because that attribute can be later modified.

jgosmann:
> If you don't have this then whenever a subtask finishes it will display the "finished" line, even though close=False. E.g., try

I added another two commits to revert the `close` change (see 4f039fa256ce035abbb2cbd269502e66c9ce3bf8 for some more details) and make the progress bar show the “parent” task instead of the subtask during those periods.

Seanny123:
Should I wait until the commit history for this is cleaned up before reviewing or is that not really going to improve my ability to review?

jgosmann:
I don't want to squash all of it yet in case we wan to go back to an older implementation stage. But parts of it could maybe cleaned up ...

jgosmann:
I cleaned up all the fixup and squash commits. I would like to keep the rest of the history for now (and maybe squash some more before a merge to master).

jgosmann:
@arvoelke

nengo/utils/progress.py
```diff
@@ -212,12 +236,197 @@ def _get_in_progress_line(self, progress):
 
         return '\r' + line.format(progress_str)
 
+    def _get_unknown_progress_line(self, progress):
+        duration = progress.elapsed_seconds()
```
Seanny123: Maybe add a comment here that says it animates a marker continuously to show progress? It took me a while to understand what the code was doing without that context.

nengo/utils/progress.py
```diff
@@ -390,42 +609,47 @@ def update(self, progress):
...
+    def __init__(self, progress_bar, total_progress, update_interval=0.1):
+        self.progress_bar = to_progressbar(progress_bar)
+        self.total_progress = total_progress
+        self.update_interval = update_interval
+        self.update_thread = threading.Thread(target=self.update_loop)
+        self.update_thread.daemon = True
+        self._closing = False
+        self.sub_progress = None
+
+    def subtask(self, max_steps=None, ongoing_name='', finished_name=None):
```
Seanny123: Found this function name a little bit confusing. It's not just accessing the currently running subtask, it's creating a new subtask to track the progress of. Maybe rename it `next_subtask` or `inc_subtask` or `new_subtask` or `get_subtask`? I know naming things is hard, but I thought it might be worth mentioning.

nengo/utils/progress.py
```diff
@@ -168,16 +186,20 @@ def update(self, progress):
...
 
+    def close(self):
+        """Closes the progress bar.
+
+        Indicates that not further updates will be made.
+        """
+        pass
+
 
 class NoProgressBar(ProgressBar):
```
Seanny123: Given that there are a lot of Progress Bar implementations, would it be worth it to split them off into their own file for the sake of organization?

jgosmann: I think it is better to keep them in a single file.

nengo/ipynb.py
```diff
@@ -45,6 +45,12 @@
 
 
 def load_ipython_extension(ipython):
+    # Not a deprecation warning as this are hidden by default and we want to
+    # make sure the user sees this message.
+    warnings.warn(
+        "Loading the nengo.ipynb notebook extension will break with current "
+        "Jupyter notebook versions and is deprecated. All features provided "
+        "by the old extension are automatically activated for IPython>=5.")
```
tbekolay: Why are we keeping this extension around? For IPython versions 2-4? At the very least, we should ensure that this becomes a no-op if using IPython >= 5... but I would also be okay making it a no-op for all versions. IPython versions 2-4 having a terminal progress bar seems fine to me. If anyone really wants it back then can go back to previous Nengo versions.

nengo/simulator.py
```diff
@@ -151,15 +151,20 @@ def __init__(
         else:
             self.model = model
 
-        if network is not None:
-            # Build the network into the model
-            self.model.build(network, progress_bar=self.progress_bar)
+        with ProgressTracker(
```
tbekolay: This will break the progressbar for Nengo OCL. Just pointing it out to remind myself to fix it once this is merged.

nengo/utils/progress.py
```diff
@@ -301,6 +514,9 @@ class ProgressUpdater(object):
     """
 
     def __init__(self, progress_bar):
+        warnings.warn(DeprecationWarning(
+            "ProgressUpdater has been deprecated because progress bars "
+            "have been switched to asynchronous updating."))
```
tbekolay: I wouldn't consider `ProgressUpdater` to be part of a public API; can we remove this instead of deprecating?

jgosmann: [It is referenced in our documentation](https://www.nengo.ai/nengo/nengorc.html), but I'm ok with removing it. In any case, that part of the documentation (and the rc file template) should be updated.

nengo/ipynb.py
jgosmann: * There will be a number of existing notebooks that still load the extension. It seems like a bad idea to break them.
* No-op is fine with me if we completely remove the old IPython progress bar.
* Using the `TerminalProgressBar` without a special `ProgressUpdater` in IPython 2 can consume a lot of memory and might increase the notebook size to be large enough to crash your browser. (This is fixed in IPython 3.)

Seanny123:
Note: This partially resolves #1017

Seanny123:
@tbekolay is this ready to merge or do you have other concerns you'd like to address?

I'm excited about having this PR merged, because having a working progress bar is nice and this is blocking the other documentation PRs.

jgosmann:
+1
Note that the current implementation will not work with Jupyter lab because they disabled Javascript output. Due to the way CSS animations are used, we can't currently to without the JS. Though, we could do without it if Jupyter notebook were to change how they update the HTML, but not sure if that will happen. If that will not happen, the progress bar is probably best implemented as a MIME renderer. The core implementation would probably reuse most of the code in this PR, but it will require a bunch of boilerplate to transform it to a proper extension (and probably requires the user to issue a command once to activate the extension).

See jupyterlab/jupyterlab#3118 for more details on the discussion.

Note that this PR improves the current state at least for notebooks and by adding the optimizer progress bar also for the terminal progress bars. Jupyter Lab can be addressed in a different PR.

jgosmann:
Looks all fine to me.

PR #1386, opened 2017-11-17
Set sim seed from network if sim seed not given.

### LINKED ISSUE #980: How should node/process seeds interact?
Currently, we have a few interacting mechanisms for seeding. In most cases, it's pretty obvious what should happen, and that is what happens. Here's a brief explanation of what I think happens now:
### Builder seed hierarchy
1. If you set a seed for an object (specifically `Ensemble`, `Connection` and `Probe`) then that seed will be used for any random operation (which sometimes doesn't happen for `Probe`).
2. If you don't set a seed, the seed for the containing network is used.
3. If the containing network doesn't have a seed, it will be assigned one from its containing network.

This hierarchy ensures that the build process is deterministic if a top-level seed is set. Because seeds are assigned at the network level, changes within one subnetwork should not affect other subnetworks.
### Simulator seed hierarchy
1. If you set a seed for a `Process`, then that seed will be used.
2. If you don't set a seed, the seed for the Simulator is used.

The question that came up in #958 is how should this hierarchy change when a `Node` seed is set. Since nodes can contain processes, it makes sense to affect the seed of the Process, but since the seed of the process is managed by the simulator, it's not a straightforward change.

So, the first issue is whether we should keep two separate seeds. The main reason for the simulator seed was for random processes; if it's easier to manage things with one seed hierarchy then I think we should go that way.

Then the second issue is how the `Process` seed hierarchy should go. Processes can be in both `Ensemble.noise` and `Node` at the moment. Should it be `Process` > `Node`/`Ensemble` > `Network`?

jgosmann:
**Motivation and context:**
Issue #980 hasn't gotten any attention in a while, but apparently the last decision was to implement the behavior as in this PR, i.e. to set the simulator seed from the network seed if no simulator seed is given. Still seems like a good idea to me because today I was again surprised that my network with a process was behaving non-deterministic despite a network seed being set.

Fixes #980.

none

**How has this been tested?**
added a test

**Types of changes:**

- Bug fix (non-breaking change which fixes an issue)?
- New feature (non-breaking change which adds functionality)?

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,10 @@ Release History
 - Default values can no longer be set for
   ``Ensemble.n_neurons`` or ``Ensemble.dimensions``.
   (`#1372 <https://github.com/nengo/nengo/pull/1372>`__)
+- If the simulator seed is not specified, it will now be set
+  from the network seed if a network seed is specified.
+  (`#980 <https://github.com/nengo/nengo/issues/980>`__,
+  `#1386 <https://github.com/nengo/nengo/pull/1386>`__)
 
 **Fixed**
 
--- a/nengo/simulator.py
+++ b/nengo/simulator.py
@@ -89,6 +89,7 @@ class Simulator(object):
         The length of a simulator timestep, in seconds.
     seed : int, optional (Default: None)
         A seed for all stochastic operators used in this simulator.
+        Will be set to ``network.seed + 1`` if not given.
     model : Model, optional (Default: None)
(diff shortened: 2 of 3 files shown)


jgosmann:
Merry Christmas! 🎄

Can we get this PR merged? I once again lost a day trying to figure out why I get different results despite identical (network) seeds.

jgosmann:
Thanks! ❤️

Issue #1124, opened 2016-07-19
filt y0 not well-defined when synapse is unstable or has a passthrough


arvoelke:
If the `filt` method is passed a value of `y0`, it will flood both the input and output arrays with this value. To my understanding, this emulates the effect of providing `y0` as a constant input to the synapse, and letting it converge in the limit to its steady-state output of `y0`.

Relevant code below for reference:
```
    class General(Step):
        """An LTI step function for any given transfer function.

        Implements a discrete-time LTI system using the difference equation
        [1]_ for the given transfer function (num, den).

        References
        ----------
        .. [1] http://en.wikipedia.org/wiki/Digital_filter#Difference_equation
        """
        def __init__(self, num, den, output, y0=None):
            super(LinearFilter.General, self).__init__(num, den, output)
            self.x = collections.deque(maxlen=len(num))
            self.y = collections.deque(maxlen=len(den))
            if y0 is not None:
                self.output[...] = y0
                for _ in num:
                    self.x.appendleft(np.array(self.output))
                for _ in den:
                    self.y.appendleft(np.array(self.output))

        def __call__(self, t, signal):
            self.output[...] = 0

            self.x.appendleft(np.array(signal))
            for k, xk in enumerate(self.x):
                self.output += self.num[k] * xk
            for k, yk in enumerate(self.y):
                self.output -= self.den[k] * yk
            self.y.appendleft(np.array(self.output))

            return self.output
```

This is good for stuff like lowpass / alpha (as long as we have this understanding that it also initializes the second order based on a constant input). But there are problems for less common synapses.

Let `(A, B, C, D)` be the continuous-time state-space representation of the desired synapse. The above scheme is a consistent and well-defined way of initializing the synapse as long as:
 1. `eig(A).real < 0` (note that `eig(A) = poles(H)`)
 2. `D = 0`

The first condition says that the system is BIBO stable. Then a constant input of `y0 != 0` will cause the system to explode (instead of converging to `y0`). Note that an integrator does not satisfy this condition (it has an eigenvalue at 0). The "correct" thing to do for zero poles would be to flood the input vector with zeros (but keep the output vector full of `y0`). Then the initial state will be consistent with some possible history (e.g., a step to `y0` followed by zeros). However for strictly unstable systems (poles greater than zero) we shouldn't ever want this case (we should probably detect it and throw an error).

[...]

Issue #938, opened 2016-01-27
Off-by-one in LinearFilter.General step


arvoelke:
When simulating higher order synapses, the `General` step method is invoked, which seems to be off-by-one delay, possibly due to the ordering in which `y` depends on `output`. 

The following test shows that `scipy.signal.lfilter` agrees with the synapse implementation in the `Simple` step for a first-order synapse such as `Lowpass`, but not for a higher order differentiator.

```
import pylab
import numpy as np
from scipy.signal import lfilter, cont2discrete, tf2ss

import nengo

sys1 = nengo.Lowpass(0.001)
sys2 = nengo.LinearFilter([0.0001, 0, 0], [0.0001, 0.02, 1])

for sys in (sys1, sys2):

    with nengo.Network() as model:
        x = nengo.Node(output=np.cos)
        p = nengo.Probe(x, synapse=sys)

    sim = nengo.Simulator(model)
    sim.run(0.02)

    (num,), den, _ = cont2discrete((sys.num, sys.den), sim.dt)
    expected = lfilter(num, den, x.output(sim.trange()))

    pylab.figure()
    pylab.title("len(num)=%d, len(den)=%d" % (len(sys.num), len(sys.den)))
    pylab.plot(sim.trange(), sim.data[p])
    pylab.plot(sim.trange(), expected)
    pylab.show()
```

![download 15](https://cloud.githubusercontent.com/assets/5420057/12611683/12fec926-c4bc-11e5-80e1-4ec1a6d825ac.png)
![download 16](https://cloud.githubusercontent.com/assets/5420057/12611684/1304e720-c4bc-11e5-9509-7ff0a9d0bfee.png)

Currently, [my implementation in NengoLib](https://github.com/arvoelke/nengolib/blob/fcbf86c80142e9e83127f949de9f06939f9a6354/nengolib/signal/system.py#L167) avoids this problem by simulating the state-space directly. Moreover, by putting it in controllable form, the `A` matrix reduces to a dot product and a shift operation. This requires less state than the LCCD method, and also reduces total simulation time.

arvoelke:
Actually I think there's something slightly more serious/general going on, with how the discretization has been incorrectly formulated to be preemptive with the state at the cost of effectively putting a delay on the passthrough. This basically means that transfer functions that are not strictly proper are not going to be able to behave correctly. However, I need to be more sure of what this means should happen in the more general setting, so I'm closing this issue for now. Sorry for the spam.

[...]

Issue #316, opened 2014-04-09
Allow the user to set the initial state of a population


celiasmith:
It would be handy to be able to give a starting point (other than zero) by specifying the first population state vector.  It might even make sense to allow specifying the starting somatic voltages of the population somehow.

Seanny123:
Any ideas on how to implement this? I'm not really familiar with the simulator, but if someone pointed me in the right direction, I'd love to take a crack at it.

studywolf:
I think we've discussed this before, I thought we had it online but it might have been just in the office. If I remember correctly the problem here is setting up the presynaptic activity and outgoing connections of the population properly...is that right @drasmuss?

drasmuss:
It'd be relatively easy to figure out what the optimal activities are for a population, for a given target output (and given decoders).  However, if you just set the output activites to that value, it's immediately going to diverge from that, because the population isn't getting the inputs to support that activity.  So if you really wanted that to work robustly, you'd have to back it up to all the inputs to the population (and all the inputs to those populations, etc.).  And you would probably have to fiddle with the connection signals as well, so you don't get a bunch of transients induced by the synaptic filtering.

As a first step, you could just do the first part, and put the burden on the user to initialize everything properly.  That is, if the user is wanting to set the population output to 0.5, they probably also want to initialize the inputs to that population to support that.  That wouldn't solve the synaptic filtering transients, but maybe those won't be that bad?  Who knows!

hunse:
I think this is most useful in the case of things with recurrent connections, e.g. an integrator or an oscillator, since purely feedforward ensembles are quick to adjust their activities (i.e. little memory). For recurrent types of objects, the memory is in the synapses, so really the key would be initializing the synapses. So you're going to have to identify the recurrent connections on an ensemble that you want to initialize, which in general is tricky.

One way to constrain the problem would be to allow the user to set an initial value for a synapse. So if they wanted an integrator to have a specific initial value, they'd have to figure out what value the recurrent synapse would have, and assign that as the initial value. (This should be easy if you don't have a function on the recurrent connection, since then the target value for the recurrent population will be the same as the synapse value. If there is a function on the connection, then you have to apply that function to the synapse initial value, since the decoding happens first. We could make a helper function to do this, or even make it automatic.)

Then, if we want, we could add `initial_value` parameters to networks like Integrator and Oscillator, which will assign the initial value to the proper recurrent connection so the user doesn't have to worry about it.

jgosmann:
[...]

PR #1387, opened 2017-11-23
Pickleable simulator

hunse:
**Motivation and context:**

This allows a whole simulator to be pickled. My main motivation is to allow a simulator to be saved occasionally when doing a learning task.

Only the model, signal values, and probe values are pickled (as well as some attributes like whether the simulator has a progress bar or optimizer). The step functions can then be re-built from this (so no functions are pickled).

The better expm function (imported from Scipy) is not necessary for this PR, but I added it when I realized our current synapse building can be unstable e.g. for Lowpass or Alpha synapses with small tau.

This depends on/supersedes #1135.

**How has this been tested?**

I've added one test each for pickling models and simulators. I also slightly extended the ``expm`` test (with another test matrix) so that the old ``expm`` fails but the new one passes.

The commits should be reviewed individually. Each one is relatively self-contained and straightforward. The most complicated is adding signals to synapses, since now all Processes can return a dictionary of arrays that gets turned into signals for use in the ``make_step`` function.

**Types of changes:**

Currently this is breaking nengo_ocl, though it shouldn't be.

**Checklist:**

**Still to do:**

Why is this breaking nengo_ocl??


Seanny123:
I've marked this as WIP, since all the tests are failing, but let me know when you fix them.

jgosmann:
Haven't looked at anything in detail yet, but I'm somewhat excited about 43871ecb24eed80d551e0a78369686831302bdc2 because it is a step towards #316. 😃

hunse:
Working with nengo_ocl now. This is ready for review.

drasmuss:
Haven't done a thorough review, but one thought is that we might want to do some profiling of various different synapses/simulators after this change. I know in `nengo-dl` I played with making internal synapse variables part of the sim state, but it slowed things down a lot (because with the way things work in tensorflow, reading and writing to Signals is slower than reading and writing to temporary internal variables). If that's only a problem for `nengo-dl`, then we can just deal with it there (although we might want to actually look into doing that before merging this PR, just to make sure that it is solvable and won't permanently slow down `nengo-dl`). But it might affect other simulators too, in which case it'd be a more core concern to this PR.

hunse:
Here's a basic script I was using for profiling:
```python
import timeit

import numpy as np
import nengo

nt = 10000
n = 1000
x = np.random.uniform(-1, 1, size=(nt, n))

synapse = nengo.Alpha(0.01)

def test():
    return synapse.filt(x)

print(min(timeit.repeat(test, number=1, repeat=10)))
```
It showed a significant slowdown on this branch (about 0.5 s instead of 0.3). Not quite sure why that would be, since I think the main difference is that before the data stored by `LinearFilter` was copied and then added to a queue, and now it's copied into one big array. It could be that copying into the array is slower (because of how the data is arranged), or it could be some other issue.

It would be good to figure out some way to do this efficiently, though. Really, everything should be storing data in Signals. The signals/ops structure was set up to be a complete representation of the state and computation graph of the simulator, and it's been a bit of a hack that we haven't been using this for synapses.

As for backends, they're really free to do what they want under the hood. There's no reason that all simulators have to be pickleable. Even if we figure out a speedup here, it's not clear to me that the same thing would work in `nengo-dl` or any other backend. But my hope is to keep things clear enough here so that backends can determine the ops/signals involved in things like synapses, and make their decisions accordingly.

arvoelke:
> It showed a significant slowdown on this branch (about 0.5 s instead of 0.3).

I'm assuming this comment is out-of-date since you switched to a state-space implementation? I'm seeing a speed-up to 0.15s on this branch versus 0.20s on master now. This is consistent with the speed-ups I am getting in nengolib, which also uses the state-space representation.

arvoelke:
And you're right that this should be more accurate as well. That's because previous `cont2discrete` would convert back-and-forth from tf -> ss -> tf. Now it's just tf -> ss in CCF, and then a matrix `expm` to discretize with ZOH. Skips the inversion back to `tf`. Happy that core nengo is embracing this now. :)

drasmuss:
We could use the state space representation without the other changes in this PR though right? (i.e. it's possible that things would be faster if we just had the state space changes without the other ones)

docs/examples/advanced/processes.ipynb
```diff
@@ -447,7 +447,7 @@
    "outputs": [],
    "source": [
     "class SimpleOscillator(nengo.Process):\n",
-    "    def make_step(self, shape_in, shape_out, dt, rng):\n",
+    "    def make_step(self, shape_in, shape_out, dt, rng, state=None):\n",
```
arvoelke: Should the notebook explain the `state` parameter (since this is a notebook about how to make custom processes)?

CHANGES.rst
```diff
@@ -35,6 +35,10 @@ Release History
 - Added version tracking to documentation, so that documentation from old
   versions remains available.
   (`#1488 <https://github.com/nengo/nengo/pull/1488>`__)
+- The ``Simulator`` is now pickleable, allowing its state to be easily stored.
+  Also added the ``clear_probes`` function to clear ``Simulator`` probes
+  (can be used before pickling to reduce the pickle file size).
+  (`#1387 <https://github.com/nengo/nengo/pull/1387>`__)
```
arvoelke: Should add a note about the new `state` parameter, since this is a user-facing and backwards-incompatible change for anyone who makes custom processes?

nengo/_vendor/scipy/sparse/nonscipy_utils.py
```diff
@@ -0,0 +1,10 @@
+import math
+
+# simple custom implementation of n choose k
+def comb(n, k):
+    """Compute n choose k"""
+    if k > n or k < 0:
+        return 0
+    if k == 0:
+        return 1
+    return math.factorial(n) // math.factorial(n - k) // math.factorial(k)
```
arvoelke: Where did this implementation come from? For scipy when `exact=False` (the default) it uses this implementation: https://github.com/scipy/scipy/blob/v1.3.0/scipy/special/_comb.pyx#L7

nengo/_vendor/scipy/sparse/README.md
```diff
@@ -0,0 +1,6 @@
+linalg_expm
+-----------
+This code for the matrix exponential is taken from `scipy.sparse`.
+
+One function that could not be ported over (it is in C) is `scipy.special.comb`.
```
arvoelke: scipy has a Python implementation (https://github.com/scipy/scipy/blob/v1.3.0/scipy/special/_comb.pyx#L7) also see comment below.

nengo/base.py
```diff
@@ -330,6 +359,10 @@ def make_step(self, shape_in, shape_out, dt, rng):
             The simulation timestep.
         rng : `numpy.random.RandomState`
             A random number generator.
+        state : {string: `numpy.ndarray`}
```
arvoelke: Above in `allocate` the docstring says `initial_state : {string: .builder.Signal}` instead of `numpy.ndarray`. Is this intentional?

nengo/builder/processes.py
```diff
@@ -161,6 +178,14 @@ def build_synapse(model, synapse, sig_in, sig_out=None):
         sig_out = Signal(
             np.zeros(sig_in.shape), name="%s.%s" % (sig_in.name, synapse))
 
+    shape_in = sig_in.shape if sig_in is not None else (0,)
+    shape_out = sig_out.shape if sig_out is not None else (0,)
+    state_init = synapse.allocate(shape_in, shape_out, model.dt)
+    state = {}
+    for name, value in state_init.items():
+        state[name] = Signal(value)
+        model.sig[synapse][state_prefix + name] = state[name]
```
arvoelke: Is there a clean way to avoid copy-pasting this code (here and above)? This would also remove the need for a global `state_prefix` variable?

nengo/simulator.py
```diff
@@ -134,6 +134,7 @@ def __init__(
             dt=0.001, seed=None, model=None, progress_bar=True, optimize=True):
         self.closed = True  # Start closed in case constructor raises exception
         self.progress_bar = progress_bar
+        self.optimize = optimize
```
arvoelke: Update docstring attributes to include `optimize` (and also `progress_bar` which was missing) to be consistent with how other things like `nengo.[Ensemble|Connection|etc...]` list any attributes that were passed in through `__init__`.

nengo/simulator.py
```diff
@@ -194,6 +195,24 @@ def __enter__(self):
...
+
+    def __setstate__(self, state):
+        self.__init__(network=None, model=state['model'],
+                      dt=state['dt'], seed=state['seed'],
+                      progress_bar=state['progress_bar'],
+                      optimize=state['optimize'])
+        for key, value in state['signals'].items():
+            self.signals[key] = value
+        for key, value in state['probe_outputs'].items():
+            self._probe_outputs[key].extend(value)
```
arvoelke: Are there any situations where the value of `self.closed` can matter?

arvoelke: After looking at the main unit test, it's a little weird to me that I can't do:
```python
with sim:
    sim.run(0.001)
with sim:
    sim.run(0.001)
```
but I can if I pickle and then unpickle it in between. What was the motivation for disallowing reopening the simulator?

nengo/synapses.py
```diff
@@ -221,123 +197,123 @@ def evaluate(self, frequencies):
...
-        if y0 is not None:
-            output[:] = y0
-
-        return LinearFilter.NoDen(np.array([1.]), np.array([]), output)
+        A, B, C, D = self._get_ss(dt, method=method)
+        X = np.zeros((A.shape[0],) + shape_out, dtype=dtype)
+
+        if y0 is not None and len(X) > 0:
+            y0 = np.array(y0, copy=False, ndmin=1)
+            X[:] = np.linalg.solve(np.eye(len(A)) - A, B * y0)
```
arvoelke: Should `y0` be renamed to `u0` to indicate that it reflects a steady-state input and not an output? 

That is, this is solving:

`x = Ax + Bu` in order to find the stable-point `x` that corresponds to a constant input of `u`? The docstring can also be updated to reflect this.

But more importantly this has no solution if the system is exponentially unstable, which can happen in perfectly valid cases. For example, the integrator `1/s` is commonly used as a synapse by members of the lab (and users of nengolib) as an elegant way of performing perfect integration across some connection (often to compute the ideal output of an integrator, or as a dynamical primitive for some larger system). In this case we have `A = 0` before discretization and `A = 1` after discretization, hence:

`x = x + Bu` 

which has a solution of `x = 0` if and only if `u = 0`, otherwise it has no solution. This makes sense because an integrator never settles to a constant state given a non-zero input.

More detailed discussion is here for reference: https://github.com/nengo/nengo/issues/1124

nengo/_vendor/scipy/sparse/nonscipy_utils.py
hunse: I wrote it. I'm fine to use the other one.

nengo/synapses.py
```diff
@@ -221,123 +197,123 @@ def evaluate(self, frequencies):
...
+
+        return dict(X=X)
+
+    def make_step(self, shape_in, shape_out, dt, rng, state=None,
+                  method='zoh'):
+        """Returns a `.Step` instance that implements the linear filter."""
+        assert shape_in == shape_out
+        assert state is not None
+
+        A, B, C, D = self._get_ss(dt, method=method)
```
arvoelke: Is this being called twice in a row? One by the allocation and again by `make_step`?

nengo/simulator.py
hunse: On other backends, we may destroy the data when the simulator is closed.

nengo/synapses.py
```diff
@@ -221,123 +197,123 @@ def evaluate(self, frequencies):
...
+
+        @classmethod
+        def check(cls, A, B, C, D, X):
+            if A.shape[0] == 0:
+                return X.shape[0] == B.shape[0] == 0 and D.shape[0] == 1
+            else:
+                return (A.shape[0] == A.shape[1] == B.shape[0] == C.shape[1]
+                        == X.shape[0]
+                        and A.shape[0] == X.shape[0]
+                        and D.shape[0] == B.shape[1] == 1)
```
arvoelke: Might as well also check that `C.shape[0] == D.shape[1] == 1`?

nengo/synapses.py
```diff
@@ -221,123 +197,123 @@ def evaluate(self, frequencies):
...
+            raise NotImplementedError("Step object must implement __call__")
+
+        @classmethod
+        def check(cls, A, B, C, D, X):
+            if A.shape[0] == 0:
+                return X.shape[0] == B.shape[0] == 0 and D.shape[0] == 1
+            else:
+                return (A.shape[0] == A.shape[1] == B.shape[0] == C.shape[1]
+                        == X.shape[0]
+                        and A.shape[0] == X.shape[0]
```
arvoelke: This second check `A.shape[0] == X.shape[0]` is redundant because it is a logical subset of the first check.

nengo/base.py
hunse: No, they should both be ```{string: `numpy.ndarray`}```. The output of `allocate` is turned into `Signal`s in the builder, but is arrays.

nengo/synapses.py
hunse: I forgot to remove that when I rewrote it.

nengo/synapses.py
```diff
@@ -221,123 +197,123 @@ def evaluate(self, frequencies):
...
+            self.X *= -self.a
+            self.X += signal
+            return self.b * self.X[0]
 
-    class General(Step):
-        """An LTI step function for any given transfer function.
+        @classmethod
+        def check(cls, A, B, C, D, X):
+            return super(LinearFilter.OneX, cls).check(A, B, C, D, X) and (
+                len(A) == 1 and B.item() == 1 and (D == 0).all())
```
arvoelke: I think `B.item() == 1` should be changed? Discretization in general could make `B` something completely different than `1`.

nengo/synapses.py
hunse: No. `D` is not always 2D.

nengo/synapses.py
hunse: Yes. It didn't seem worth it to cache because it's quite cheap.

nengo/synapses.py
arvoelke: Did you mean to check `C` instead?

nengo/synapses.py
```diff
@@ -221,123 +197,123 @@ def evaluate(self, frequencies):
...
+        def __init__(self, A, B, C, D, X):
+            super().__init__(A, B, C, D, X)
+            self.a = A.item()
+            self.b = C.item()
 
         def __call__(self, t, signal):
-            self.output *= -self.a
-            self.output += self.b * signal
-            return self.output
+            self.X *= -self.a
```
arvoelke: Is this being unit-tested? According to the coverage report it isn't, but not sure how flaky that is. The reason I figured was because I think this should be `self.X *= self.a` without the minus sign. Since the linear system is of the form `x = Ax + Bu`.

arvoelke:
Are there any unit tests that are sensitive to single time-step delays in both `filt` and the use of each different synapse simulation method?

The reason I ask was because I had to do something quite ad-hoc in `nengolib` to make things consistent with `nengo` across both `filt` and normal simulation -- and you don't seem to be doing that here from what I can see. That was related to https://github.com/nengo/nengo/issues/938.

> We could use the state space representation without the other changes in this PR though right? (i.e. it's possible that things would be faster if we just had the state space changes without the other ones)

Yeah I'm thinking the slowdown might be due to the use of signals. So we're paying a cost for that, but then making up for it with the state-space representation. Haven't verified this though.

nengo/synapses.py
hunse: It seems to me that #1124 is the real issue we have to address here. But if renaming it to `u0` makes it clearer to you, I'm fine with that. It will be an API change, but I think not many people are using that parameter.

nengo/synapses.py
arvoelke: Can you give an example? I believe there was a time in an earlier version of scipy (<0.18.0) when it wasn't, but this was fixed after I reported this issue: https://github.com/scipy/scipy/issues/5760.

nengo/synapses.py
arvoelke: One potential issue I see later is if `method` is refactored or exposed in some way. Currently we're trusting that it's called with the exact same signature in each place, which opens up some surface area for things to go wrong. Is there a way to avoid this without caching?

nengo/simulator.py
arvoelke: Why do we disallow it for the reference backend then? This allows the reference simulator to be reopened by unpickling it. So It seems that for consistency either we should disallow pickling if the reference simulator is closed, or allow the reference simulator to be reopened.

hunse:
I don't see why using Signals should inherently make things slower. Just like other parts of the simulator, when we get down to the actual simulation, it's all in Numpy arrays. I think the first step is to do some better profiling of if there's a slowdown, and if so, where. I was finding it very difficult to do timing on my machine and get reproducible results.

As for delays, we had code in our previous `LinearFilter` that would remove a zero at the front of a discrete transfer function numerator to try to stop needless delays (e.g. if your numerator was `[0, 0.3, 0.4]`, it would become `[0.3, 0.4, 0]`). When switching to state space, I found that both of these numerators gave the same matrices from `tf2ss`, so I got rid of it (also, we now do the continuous->discrete conversion after we do `tf2ss`, for numerical stability reasons, so we can't do that manipulation anymore). If there are particular places where the behaviour changes, let me know. I don't think it's robustly tested. But I also don't think there's the same delay that we used to have to remove (from what I can tell by looking at the matrices for an Alpha filter, the input can affect the output the same timestep, which isn't true if you have a numerator like `[0, 0.3, 0.4]`).

arvoelke:
>  I don't see why using Signals should inherently make things slower. Just like other parts of the simulator, when we get down to the actual simulation, it's all in Numpy arrays. I think the first step is to do some better profiling of if there's a slowdown, and if so, where. I was finding it very difficult to do timing on my machine and get reproducible results.

nengolib is consistently faster in my tests. It's possible that it could be due to my conversion back to CCF in order to [exploit the off-diagonal structure](https://github.com/arvoelke/nengolib/blob/098cc3458eecc1449ebf615c409c4271658a7656/nengolib/signal/system.py#L178). I forgot I had been doing this. In general this is risky because it effectively does the same thing numerically as converting back to tf (https://github.com/arvoelke/nengolib/issues/168). I won't be doing this in every case -- maybe just the cases where I can detect no numerical loss in the conversion. At any rate it makes sense to investigate this separately by profiling various Nengo reference models.

> When switching to state space, I found that both of these numerators gave the same matrices from tf2ss, so I got rid of it

Can you give an example? It doesn't change the `A` matrix, because that is derived from the denominator. But it does change the other matrices, and can become a part of the passthrough matrix `D`. e.g.,

```python
>>> from scipy.signal import tf2ss
>>> tf2ss([0, 0.3, 0.4], [1, 0, 0])
(array([[-0., -0.],
       [ 1.,  0.]]), array([[1.],
       [0.]]), array([[0.3, 0.4]]), array([[0.]]))
>>> tf2ss([0.3, 0.4, 0], [1, 0, 0])
(array([[-0., -0.],
       [ 1.,  0.]]), array([[1.],
       [0.]]), array([[0.4, 0. ]]), array([[0.3]]))
```

I remember some of the weirdness basically came out of the case-handling that was required for when the numerator was full degree versus when it isn't. I have some [very thorough unit tests](https://github.com/arvoelke/nengolib/blob/master/nengolib/signal/tests/test_system.py#L319) here that are comparing things on the level of single time-step delays with Nengo in all of these cases -- none of these tests have changed since they were written, so it's still the same within Nengo.

I'm definitely okay with changing this because it was pretty hacky how I had to match it. I have a feeling the difference might just be in the `filt` method in this PR and not within the simulator.

nengo/synapses.py
hunse: You're right, it's not getting hit. That minus was left over from the transfer function implementation. Could be part of the reason why I was seeing a slowdown for Lowpass synapses.

I've added a commit now that I think fixes it.

nengo/simulator.py
hunse: Pickling can be used to save probe data, too. I think we should just store the value of `closed` as part of pickling, so if you pickle->unpickle a closed simulator, it's still closed.

hunse: I think the main argument for not allowing it on the reference backend is so that users don't get into the habit of being able to do that, since if they switch simulators they can't.

hunse:
I'm going to split out the state-space stuff and do it in a separate PR, then base this off that PR. That will make it easier to evaluate that by itself, and to do better speed testing of the benefits of state-space vs any potential slowdowns to the synapse state stuff.

hunse:
I redid my timing script now that this is based of #1535, and found that it's no slower than that branch for either `Lowpass` or `Alpha` synapses. I think the reason for the initial slowdown was that I had switched the transfer function implementation to reduce the number of states. Now that we no longer use transfer functions, it's a non-issue.

nengo/synapses.py
hunse: Yeah, this was related to the `__call__` for this class not getting hit. Fixed now.

nengo/synapses.py
hunse: This was on my laptop, and I can't remember what Scipy version I have on there. Could have been on an older version.

nengo/synapses.py
hunse: Yeah, I was wondering about having `method` as a LinearFilter class attribute.

nengo/_vendor/scipy/sparse/nonscipy_utils.py
hunse: Done in #1535

nengo/_vendor/scipy/sparse/README.md
hunse: Done in #1535

nengo/synapses.py
hunse: In #1535, `y0` now actually represents the steady-state output. We still have the issue of it not working for some synapses, but that's a larger issue described by #1124.

nengo/synapses.py
hunse: Done in #1535

hunse:
I think I've addressed all of @arvoelke's comments related to #1535 here. He can resolve them as appropriate, and further comments on that part of the code can go in #1535.

nengo/processes.py
```diff
@@ -98,14 +98,18 @@ def __repr__(self):
...
+    def make_step(self, shape_in, shape_out, dt, rng, state=None):
         assert shape_in == (0,)
         assert len(shape_out) == 1
 
         dist = self.dist
         scale = self.scale
         alpha = 1. / np.sqrt(dt)
-        filter_step = self.synapse.make_step(shape_out, shape_out, dt, None)
+        filter_step = self.synapse.make_step(
+            shape_out, shape_out, dt, rng=None, state=state)
```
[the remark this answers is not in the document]
hunse: Yeah, we had `None` before, too, basically because we assume synapses should not need it. But someone could implement some sort of probabilistic synapse that uses it. I'd make the change, but I'm not sure we need to test it (we have other processes that do use `rng`).

nengo/synapses.py
```diff
@@ -61,6 +60,9 @@ def __init__(self, default_size_in=1, default_size_out=None,
                          default_dt=default_dt,
                          seed=seed)
 
+    def allocate(self, shape_in, shape_out, dt, dtype=None, y0=None):
+        return {}  # pragma: no cover  Implement if the synapse has a state
```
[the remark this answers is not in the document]
hunse: I just wanted to make a clear API. If we want people to notice, I'd raise a `NotImplementedError` here instead.

nengo/synapses.py
```diff
@@ -462,9 +474,7 @@ def __init__(self, t, **kwargs):
         super().__init__(**kwargs)
         self.t = t
 
-    def make_step(self, shape_in, shape_out, dt, rng, y0=0, dtype=None):
-        """Returns a custom step function."""
-        assert shape_in == shape_out
+    def _get_n(self, dt, dtype=None):
```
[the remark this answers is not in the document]
hunse: `get_coefficients`? It basically gets the coefficients in a weird form specific to this implementation of the triangle filter.

nengo/tests/test_synapses.py
```diff
@@ -19,10 +19,11 @@
 butter_den = np.array([1., -3.18063855, 3.86119435, -2.11215536, 0.43826514])
 
 
-def run_synapse(Simulator, seed, synapse, dt=1e-3, runtime=1., n_neurons=None):
+def run_synapse(Simulator, seed, synapse, dt=1e-3, runtime=0.2, high=100,
+                n_neurons=None):
     model = nengo.Network(seed=seed)
     with model:
-        u = nengo.Node(output=WhiteSignal(runtime, high=10))
+        u = nengo.Node(output=WhiteSignal(runtime, high=high))
```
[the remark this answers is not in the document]
hunse: To get some higher frequency components in there and notice the effect of some of the filters more. Most of the filters we're testing with are some sort of lowpass filter, and with 10 the output signal often looked almost identical to the input. We can also use a shorter runtime and still get good coverage across a range of input values.

drasmuss:
There is some missing diff coverage here https://codecov.io/gh/nengo/nengo/compare/042fffe0b351b9ab60a9aa2d008229379cd53347...da0edfeb458141b5f3ed761d8d3a155e7dba0ef5/src/nengo/synapses.py?before=nengo/synapses.py#D1-500, which I don't know well enough to write a test for.  Could you add a test when you get the chance @hunse? (no rush)

hunse:
To test that, we just need a Triangle filter with `y0 != 0` (non-zero initialization). It could probably be as simple as having `y0 = 1`, filtering a signal of all ones, and making sure that the output is always one (i.e. doesn't start at 0 and go up to 1, which it would if `y0 = 0`). If you want to take a stab, go for it, otherwise I can do that next week.

tbekolay:
> * the time input to the function will be a scalar `np.array` rather than a pure float

The main difference (and why we did the pure float thing) is that the `np.array` version can be mutated by the process, while the pure float can't -- or, if it's mutated it won't affect the simulation. Maybe overly defensive, but I think probably better that a user can't modify the simulator time in a node.

docs/examples/advanced/processes.ipynb
tbekolay: Done, thanks for pointing this out :ok_hand:

docs/examples/advanced/processes.ipynb
```diff
@@ -447,15 +448,19 @@
    "outputs": [],
    "source": [
     "class SimpleOscillator(nengo.Process):\n",
-    "    def make_step(self, shape_in, shape_out, dt, rng):\n",
+    "    def make_state(self, shape_in, shape_out, dt, dtype=None):\n",
+    "        # return a dictionry mapping strings to their initial state\n",
```
drasmuss: dictionry -> dictionary

PR #1392, opened 2017-12-18
Allow PES.pre_tau to be 0 or None

drasmuss:
**Motivation and context:**
``PES.pre_tau`` was previously restricted to be > 0, but I don't think there's anything wrong with letting it be 0 or `None` if the user wants.

**How has this been tested?**
Added `pre_tau` parameter to the tests in `test_pes_ens_ens`

**How long should this take to review?**

- Quick (less than 40 lines changed or changes are straightforward)

**Types of changes:**
- New feature (non-breaking change which adds functionality)

**Checklist:**

- [x] I have read the **CONTRIBUTING.rst** document.
- [x] I have updated the documentation accordingly.
- [x] I have included a changelog entry.
- [x] I have added tests to cover my changes.
- [x] I have run the test suite locally and all tests passed.

PR #1381, opened 2017-11-13
Pytest warnings

jgosmann:
**Motivation and context:**
Starting with pytest 3.2 it displays warnings raised during tests. At the moment this produces a lot of noise when running the tests and requires a lot of scrolling to find the actual test failures; on Travis CI it might even cut off the log (though you can download a full log). Also, at least one of the warnings did highlight a relevant problem that required fixing for future compatibility.

Note that this bumps up the pytest requirement (for tests) to 3.2 (for `pytest.filterwarnings`).

This is currently work in progress. See to do list below for details. Creating this PR to make sure that the tests do not fail on Travis-CI (as we had warnings related problems in #1011 before, [though it seems to be fixed now](https://github.com/nengo/nengo/issues/1378#issuecomment-342571414)).

Closes #1378.

**Interactions with other PRs:**
For a particular Python 3.4 the latest available IPython/Jupyter notebook versions produce a bunch of warnings that are most likely related to the progress bar. These will probably be fixed with the merge of #1380.

**How has this been tested?**
existing tests

**How long should this take to review?**

- Average (neither quick nor lengthy)

**Types of changes:**

- Bug fix (non-breaking change which fixes an issue)

**Checklist:**

- [x] I have read the **CONTRIBUTING.rst** document.
- [n/a] I have updated the documentation accordingly.
- [n/a] I have included a changelog entry. [not user facing]
- [n/a] I have added tests to cover my changes.
- [x] I have run the test suite locally and all tests passed.

**Still to do:**

- [x] At the moment there is still a large number of warnings related to Sigmoid (and RectifiedLinear) neurons that will require some understanding of the test cases and what exactly causes the warnings for those neuron types. These might be warnings hinting at actual problems with the tests (at least for Sigmoid neurons).
- [x] There is a warning in one example produced on purpose. Not sure at the moment what the best way to filter it out is.
- [n/a] There is some `ResourceWarning` output that is shown immediatly when running the tests, but does not enter the warning summary. 😕 --> not our fault, but an issue with xdist: pytest-dev/pytest-xdist#247
- [x] Warnings when SciPy is not installed.
- [x] slow tests
- [x] logging tests
- [x] analytics + compare tests

PR #1394, opened 2017-12-23
Make conda install quiet as workaround

Seanny123:
**Motivation and context:**
As noted in #1392, TravisCI has started crashing during conda installs. After consulting [various issues](https://github.com/travis-ci/travis-ci/issues/8934) this seems to be caused by writing too much output. This is resolved by quietly installing `conda`.

**Note**: One of the recommended workarounds is setting `.travisci` to `filter_secret: false`, however this didn't work. ):

**Types of changes:**

**Checklist:**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
   - if [[ "$SCIPY" == "true" ]]; then
       export CONDA_DEPS="$CONDA_DEPS scipy";
     fi
-  - if [[ -n "$CONDA_DEPS" ]]; then eval conda install "$CONDA_DEPS"; fi
+  - if [[ -n "$CONDA_DEPS" ]]; then eval conda install --quiet "$CONDA_DEPS"; fi
   - if [[ -n "$PIP_DEPS" ]]; then eval pip install "$PIP_DEPS"; fi
 
 # Run the tests


Seanny123:
@tbekolay quieting conda does seem to work, however if there are too many warnings, the same error occurs. Consequently, if you could merge #1381 it would be greatly appreciated.

jgosmann:
`filter_secret: false` [seems to work too](https://travis-ci.org/nengo/nengo/builds/320474281), but seems like a potentially bad idea, even though it seems that the command lines itself still get filtered (I assume it is only the command output that won't be filtered anymore?). That flag also seems to be undocumented?

Seanny123:
> filter_secret: false seems to work too

It seemed to me it still wasn't working because it failed when pytest spewed out too many warnings?

jgosmann:
Yeah, it seems that that's what happend (only two build jobs failed and the one I looked at had a lot of warnings and didn't fail in the conda phase).

Seanny123:
I wasn't sure what making conda quiet would be sufficient and might have
some unattended side effects. However, if it's fine, then proceed!

tbekolay:
> might have some unattended side effects

Actually quite the opposite... `--quiet` only suppresses the output of the progress bars, not the list of packages installed (which I was worried about) so I kind of think the reduced noise in the build log is better.

Merging with one review as it doesn't touch code :+1:

PR #1395, opened 2017-12-29
Index lists with double references using negative indices

jgosmann:
**Motivation and context:**
There is one special case with sliced indexing that still does not correctly work: If a list of indices references the same index twice, but once with a positive integer and once with a negative integer, one of those will overwrite the other one when the slicing is done on an input. This PR fixes that by converting all indices to positive indices (as a canonical representation) before checking for repeated indices.

none

**How has this been tested?**
added a test

I'd recommend going commit by commit, but the PR is small enough that looking at the full diff should work too.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -55,6 +55,9 @@ Release History
 - Fixed validation checks that prevented the default
   from being set on certain parameters.
   (`#1372 <https://github.com/nengo/nengo/pull/1372>`__)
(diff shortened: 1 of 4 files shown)


Seanny123:
There's an issue with Numpy 1.8 on Python 2.7 that should be resolved before merging.

jgosmann:
I replaced the last commit, now the tests should pass.

PR #1398, opened 2018-01-05
Create a custom op for PES learning rule

drasmuss:
**Motivation and context:**

All the other learning rules act as `updates`, meaning that if you compute a parameter delta on step `t`  the weights change on `t+1`.  That's pretty standard for learning rule behaviour.

Our current PES implementation, on the other hand, actually applies the update on step `t`.  So you could, in theory, compute an error signal, update connection weights, and have that affect the behaviour of downstream ensembles on the same timestep.

One reason why I'd say this is bad is that, just conceptually, it is inconsistent with the behaviour of the other learning rules in Nengo, and I think the expected behaviour of learning rules in general.  Another reason is that it means you can get unexpected cycles in the graph if you don't have a synaptic delay on the error connection, e.g.:

``` python
import nengo

with nengo.Network() as net:
    a = nengo.Ensemble(10, 1)
    b = nengo.Node(size_in=1)
    c = nengo.Connection(a, b, synapse=None, learning_rule_type=nengo.PES())
    nengo.Connection(b, c.learning_rule, synapse=None)

    # this network won't build, because it has cycles in the op graph
```

This PR changes PES so that it acts as an `update`, like the other rules.  In order to do this we have to implement a custom PES operator (so that we can specify its output should be an `update`).  Technically we could get away without a custom op, and instead build a delay into PES by sticking a `SimProcess(..., mode="update")` into `build_pes`.  But that seemed like a weird/awkward solution to me, pretending that we have a Process in the PES rule just so that it will act as an `update`.

One potential downside of this change is related to this comment from the `build_pes` docstring: "Generic operators are used because they are more likely to be implemented on other backends like Nengo OCL."  Note that this is only relevant to backends that use the Nengo builder, which I think is just `nengo_ocl` and `nengo_dl` at the moment.  @hunse could comment for `nengo_ocl`, but for `nengo_dl` I'd actually find it easier if PES had its own operator.  That makes it easier for me to build an efficient implementation of that operator, because I have control over how it is implemented internally rather than being tied to the generic operators.  In addition, our current PES implementation already has some weird elements in it (e.g., it uses the `ElementwiseInc` operator to perform an outer product, which isn't the defined/expected behaviour for `ElementwiseInc`).  So backends already kind of need to be aware of the PES rule, even without the custom operator.

The only other downside (I think) is that the custom operator no longer exposes the `correction` signal for probing (since it is now internal to the operator).  Technically this is a breaking change, although I'd say a minor one.  Most people probably didn't even know that was probeable, I'd guess.  And the correction is just equal to the error signal (which is still probeable) times a constant (`learning_rate / n_neurons`), so the information is still pretty easy to retrieve.  However, if anyone is really worried about this, we could have the `SimPES` operator output that signal just for probing.

#1392 / #1095 touch build_pes, so will need to be updated
#1254 also implemented a custom PES operator, with a slightly different approach (along with a bunch of other changes)

**How has this been tested?**

Passes all existing tests, and I added a new test for the cycle case described above.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -95,6 +95,9 @@ Release History
 - The ``NeuronType.current`` and ``NeuronType.rates`` methods now document
   the supported shapes of parameters and return values.
   (`#1437 <https://github.com/nengo/nengo/pull/1437>`__)
+- PES learning updates are now applied on the next timestep rather than
+  the current one.
+  (`#1398 <https://github.com/nengo/nengo/pull/1398>`_)
 
 **Deprecated**
 
@@ -123,6 +126,8 @@ Release History
 - Removed the ``probe_all`` function. It can now be found in
   `nengo_extras <https://github.com/nengo/nengo_extras/>`__.
   (`#1187 <https://github.com/nengo/nengo/pull/1187>`_)
+- ``PES.correction`` is no longer probeable.
+  (`#1398 <https://github.com/nengo/nengo/pull/1398>`_)
 
 **Fixed**
 
--- a/docs/backend_api.rst
+++ b/docs/backend_api.rst
@@ -71,6 +71,8 @@ Operators
 
 .. autoclass:: nengo.builder.neurons.SimNeurons
 
+.. autoclass:: nengo.builder.learning_rules.SimPES
+
 .. autoclass:: nengo.builder.learning_rules.SimBCM
 
 .. autoclass:: nengo.builder.learning_rules.SimOja
--- a/docs/examples/usage/strings.ipynb
+++ b/docs/examples/usage/strings.ipynb
@@ -266,6 +266,8 @@
     "print(nengo.builder.operator.SimPyFunc(sig, lambda x: 0.0, True, sig))\n",
     "print(nengo.builder.operator.SimPyFunc(\n",
     "    sig, lambda x: 0.0, True, sig, tag=\"tag\"))\n",
+    "print(nengo.builder.learning_rules.SimPES(sig, sig, sig, sig, 0.1))\n",
+    "print(nengo.builder.learning_rules.SimPES(sig, sig, sig, sig, 0.1, tag=\"tag\"))\n",
     "print(nengo.builder.learning_rules.SimBCM(sig, sig, sig, sig, 0.1))\n",
     "print(nengo.builder.learning_rules.SimBCM(sig, sig, sig, sig, 0.1, tag=\"tag\"))\n",
     "print(nengo.builder.learning_rules.SimOja(sig, sig, sig, sig, 0.1, 1.0))\n",
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -1,12 +1,120 @@
 import numpy as np
 
 from nengo.builder import Builder, Operator, Signal
-from nengo.builder.operator import DotInc, ElementwiseInc, Copy, Reset
+from nengo.builder.operator import Copy, Reset
 from nengo.connection import LearningRule
-from nengo.ensemble import Ensemble, Neurons
+from nengo.ensemble import Ensemble
 from nengo.exceptions import BuildError
 from nengo.learning_rules import BCM, Oja, PES, Voja
 from nengo.node import Node
+
+
+class SimPES(Operator):
+    r"""Calculate connection weight change according to the PES rule.
+
+    Implements the PES learning rule of the form
+
+    .. math:: \Delta \omega_{ij} = \frac{\kappa}{n} e_j a_i
+
+    where
+
+    * :math:`\kappa` is a scalar learning rate,
+    * :math:`n` is the number of presynaptic neurons
+    * :math:`e_j` is the error for the jth output dimension, and
+    * :math:`a_i` is the activity of a presynaptic neuron.
+
+    Parameters
+    ----------
+    pre_filtered : Signal
+        The presynaptic activity, :math:`a_i`.
+    error : Signal
+        The error signal, :math:`e_j`.
+    delta : Signal
+        The synaptic weight change to be applied, :math:`\Delta \omega_{ij}`.
+    learning_rate : float
+        The scalar learning rate, :math:`\kappa`.
+    encoders : Signal, optional
+        If not None, multiply the error signal by these post-synaptic
+        encoders (in the case that we want to learn a neuron-to-neuron
+        weight matrix instead of decoder weights).
+    tag : str, optional (Default: None)
+        A label associated with the operator, for debugging purposes.
+
+    Attributes
+    ----------
+    pre_filtered : Signal
+        The presynaptic activity, :math:`a_i`.
+    error : Signal
+        The error signal, :math:`e_j`.
+    delta : Signal
+        The synaptic weight change to be applied, :math:`\Delta \omega_{ij}`.
+    learning_rate : float
+        The scalar learning rate, :math:`\kappa`.
+    encoders : Signal, optional
+        If not None, multiply the error signal by these post-synaptic
+        encoders (in the case that we want to learn a neuron-to-neuron
+        weight matrix instead of decoder weights).
+    tag : str, optional (Default: None)
+        A label associated with the operator, for debugging purposes.
+
+    Notes
+    -----
+    1. sets ``[]``
+    2. incs ``[]``
+    3. reads ``[pre_filtered, error, encoders]``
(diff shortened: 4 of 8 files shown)


Seanny123:
Note: this is failing for Python 2.7 and Numpy 1.8, as well as some `flake8` checks.

jgosmann:
> One potential downside of this change is related to this comment from the build_pes docstring:

One other downside: The optimizer can act on the standard operators, but special implementations are required for specific operators. Though, I assume that only few connections in a model actually use PES, so the impact is probably minor.

> it uses the ElementwiseInc operator to perform an outer product

I did not know that this was possible! I believe that would actually allow me to implement another custom learning rule to easily be implemented. But that should probably be better documented (if not even another operator specifically for outer products would be in order).

> Note: this is failing for Python 2.7 and Numpy 1.8, as well as some flake8 checks.

This is the second time in a short while that specifically the Numpy 1.8 tests where failing. Maybe want to think about dropping support for it?

hunse:
I had always assumed that ElementwiseInc was implemented the way it was precisely so that it *could* be used as an outer product. But I guess that's not clear in the docstring or by looking at the code necessarily.

I didn't realize that we were doing the update on the delta, rather than doing the update in ``build_learning_rule`` when ``delta`` is applied to the weights. To me, the latter makes more sense, because then a user/developer can implement a learning rule however they want, and don't have to worry about updates. But I guess we don't have any op that can do that operation as an update, so maybe that's why we did it the current way.

I like the idea of having learning rules done using basic operators more than writing custom ops for them. This makes things easier from an optimization and scheduling standpoint, since the constituent operators of something like PES can theoretically be done with those of BCM or other learning rules, or even other parts of the simulation. If anything, I would argue that our other learning rules should be done in terms of basic operators.

Practically, though, we typically only have a few learning rules per network, so I don't think it makes a huge difference if we're optimizing things perfectly ideally. I do like that this makes PES an update; I agree that that is an oversight. And I think in the short term it will work fine. In the long term, though, I'm for building things in terms of a few basic ops, and limiting the number of specialized ops.

drasmuss:
> I didn't realize that we were doing the update on the delta, rather than doing the update in build_learning_rule when delta is applied to the weights. To me, the latter makes more sense, because then a user/developer can implement a learning rule however they want, and don't have to worry about updates. But I guess we don't have any op that can do that operation as an update, so maybe that's why we did it the current way.

Yeah the problem is that `update` is always a `set` (as currently defined), so if you have multiple learning rules all targeting the same weights you couldn't have the `delta -> weights` part be an update.  You have to have each learning rule `update` their own individual signal (`delta`), and then combine that afterwards with an `inc`.  We'd have to create a new `update_inc` type, which might not be too complicated but would definitely be a more fundamental change.

> Practically, though, we typically only have a few learning rules per network, so I don't think it makes a huge difference if we're optimizing things perfectly ideally. I do like that this makes PES an update; I agree that that is an oversight. And I think in the short term it will work fine. In the long term, though, I'm for building things in terms of a few basic ops, and limiting the number of specialized ops.

Yeah I agree that having things in terms of basic operators is probably a good goal, as long as it gives us the desired behaviour.  Perhaps something to keep in mind for the builder refactoring, coming up with a system that'd let us move more of the existing `SimX` ops into generic operators?

It's also worth keeping in mind that there's nothing preventing a backend from defining a builder for learning rules in terms of basic operators instead (keeping in mind their own implementation of those basic operators, which might let them work with things a bit differently, e.g. whether or not they can use ElementwiseInc for outer products).

hunse:
> Yeah the problem is that `update` is always a `set` (as currently defined), so if you have multiple learning rules all targeting the same weights you couldn't have the `delta -> weights` part be an update.  You have to have each learning rule `update` their own individual signal (`delta`), and then combine that afterwards with an `inc`.  We'd have to create a new `update_inc` type, which might not be too complicated but would definitely be a more fundamental change.

Oh yes, we don't allow multiple updates per signal, that's the real problem. There's actually nothing that says whether an update can use the previous value in a signal as part of the update (as is the case if you're implementing) or not (as with overwriting). It does so happen that all our updates are overwrites, not increments, but there's nothing that says that has to be. So I don't think we'd need a new signal modification type. Ops that can update by incrementing or overwriting would have to distinguish between the two cases, of course, but that could just be with a flag that they store.

> It's also worth keeping in mind that there's nothing preventing a backend from defining a builder for learning rules in terms of basic operators instead

That is true, and is what I plan to do with PES in nengo_ocl if/when this is merged. But we are starting to add things to core Nengo (e.g. the optimizer) that could take advantage of having them as separate ops.

drasmuss:
I did some profiling before/after this change to see what the impact is.  Here is the test script I used:

``` python
import timeit

import nengo

n_neurons = 1000
d = 100
n_ensembles = 20

with nengo.Network() as net:
    for _ in range(n_ensembles):
        a = nengo.Ensemble(n_neurons, d)
        b = nengo.Node(size_in=d)
        c = nengo.Connection(a, b, learning_rule_type=nengo.PES())
        nengo.Connection(b, c.learning_rule)

with nengo.Simulator(net, optimize=True) as sim:
    t = 1e10
    for i in range(5):
        start = timeit.default_timer()
        sim.run(1.0)
        elapsed = timeit.default_timer() - start
        print("%d: %f" % (i, elapsed))
        t = min(elapsed, t)
    print("fastest time:", t)
```
Note that this is kind of a worst-case scenario for switching to a PES operator, because we have 20 independent paths which the optimizer should be able to combine (when using basic operators).  But the results still show the PES op significantly out performing the basic operators:

basic:
```
Build finished in 0:00:05.                                                      
Simulation finished in 0:00:11.                                                 
0: 10.375603
Simulation finished in 0:00:11.                                                 
1: 10.576693
Simulation finished in 0:00:11.                                                 
2: 10.884909
Simulation finished in 0:00:11.                                                 
3: 10.475991
Simulation finished in 0:00:12.                                                 
4: 11.080365
fastest time: 10.375602857898375
```

pes op:
```
Build finished in 0:00:05.                                                      
Simulation finished in 0:00:08.                                                 
0: 7.554621
Simulation finished in 0:00:08.                                                 
1: 7.857940
Simulation finished in 0:00:08.                                                 
2: 7.756308
Simulation finished in 0:00:08.                                                 
3: 7.655517
Simulation finished in 0:00:08.                                                 
4: 7.554778
fastest time: 7.55462109086078
```

I'm guessing this is because there is less reading and writing to intermediate Signals, but I haven't done any detailed profiling.

tbekolay:
There should also be fewer step functions to run, which I think contributes. In any case, faster is faster regardless of the reason 👌

hunse:
I think using `out=` on `np.outer` is also a significant part, but that's just a hunch.

I profiled on my machine. Before:
```
Build finished in 0:00:11.                                                                                           
Simulation finished in 0:00:36.                                                                                      
0: 35.323122
Simulation finished in 0:00:46.                                                                                      
1: 45.761717
Simulation finished in 0:00:40.                                                                                      
2: 39.770072
Simulation finished in 0:00:41.                                                                                      
3: 40.255951
Simulation finished in 0:00:38.                                                                                      
4: 37.894030
fastest time: 35.3231218320434
```
After:
```
Build finished in 0:00:11.                                                                                           
Simulation finished in 0:00:15.                                                                                      
0: 14.954900
Simulation finished in 0:00:13.                                                                                      
1: 12.843164
Simulation finished in 0:00:16.                                                                                      
2: 15.156235
Simulation finished in 0:00:15.                                                                                      
3: 14.767565
Simulation finished in 0:00:14.                                                                                      
4: 13.948803
fastest time: 12.843163657991681
```
So an even bigger relative improvement for me.

tbekolay:
The improvement for me is less dramatic, but still an improvement:

master:

```
Simulation finished in 0:00:19.                                                         0: 18.360750
Simulation finished in 0:00:19.                                                         1: 18.375270
Simulation finished in 0:00:20.                                                         2: 19.172345
Simulation finished in 0:00:20.                                                         3: 20.069227
Simulation finished in 0:00:19.                                                         4: 18.857892
fastest time: 18.360749627001496
```

pes_op:

```
Simulation finished in 0:00:18.                                                         0: 17.556179
Simulation finished in 0:00:18.                                                         1: 17.657640
Simulation finished in 0:00:19.                                                         2: 18.805179
Simulation finished in 0:00:18.                                                         3: 17.978860
Simulation finished in 0:00:18.                                                         4: 17.552026
fastest time: 17.552026199000466
```

PR #1406, opened 2018-01-19
Make sim.n_steps and sim.time scalars

drasmuss:
**Motivation and context:**

These values were being returned as arrays with shape `(1,)`, which means you can't do things like `for i in range(sim.n_steps)`.  This also doesn't match up with our documentation, which says these values are ints/floats.

**How has this been tested?**

Added a check in `test_simulator.py`

**Types of changes:**

**Checklist:**


jgosmann:
I commits changing how the testing for a scalar is done and adding a changelog entry.

Technically, this is a breaking change, but I assume those properties have not been used widely.

PR #1413, opened 2018-02-16
Misc. typos

luzpaz:
Found via `codespell -q 3`

PR #1424, opened 2018-04-02
Run codespell through TravisCI

tbekolay:
**Motivation and context:**
In #1413, @luzpaz was kind enough to fix several typos using the `codespell` tool. Rather than let typos keep piling up, this PR runs `codespell` through TravisCI to spot misspelling sooner rather than later.

Note that pylint also has some [spellchecking options](https://docs.pylint.org/en/1.8/technical_reference/features.html#spelling-checker), which I tried out, and did not work well because

- Tons of false positives as variable names are not whitelisted (keeping a whitelist for them would be very time consuming)
- Only checks comments and docstrings
- Requires an additional dependency, `pyenchant`

`codespell`, while requiring a `pip install codespell`, had very few false positives (which I fixed in this PR) and seems to check all text, including variable names and text in Jupyter notebooks. I believe it only looks for a set of common misspellings, but it seems to catch quite a bit, and certainly is better than having nothing.

**How has this been tested?**
Ran this locally with `codespell -q 3`; TravisCI currently testing it.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.ci/static.sh
+++ b/.ci/static.sh
@@ -12,9 +12,9 @@ function usage {
 }
 
 if [[ "$COMMAND" == "install" ]]; then
-    pip install flake8 pylint
+    pip install codespell flake8 pylint
 elif [[ "$COMMAND" == "run" ]]; then
-    flake8 -v nengo && pylint nengo
+    flake8 -v nengo && codespell -q 3 && pylint nengo
 else
     if [[ -z "$COMMAND" ]]; then
         echo "Command required"
--- a/nengo/spa/tests/test_pointer.py
+++ b/nengo/spa/tests/test_pointer.py
@@ -101,11 +101,11 @@ def test_convolution(rng):
     c = a.copy()
     c *= b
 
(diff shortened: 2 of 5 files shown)


luzpaz:
Very cool folks. Glad to help and glad to see you utilizing codespell in your workflow. Cheers!  
CC @larsoner

jgosmann:
Should codespell also run when invoking `tox -e static`?

tbekolay:
> Should codespell also run when invoking tox -e static?

Ah yes, I missed that as I no longer use `tox` to invoke tests on multiple virtualenvs (I use a tiny bash function instead).

Out of curiosity, do any Nengo devs  use `tox`? If so, I'll update `tox.ini`. If not, I'll remove it.

jgosmann:
I sometimes use it to run the static checkers locally, but if we had another simple commad for that it would be fine too (maybe even better as tox is slow due to installing everything in a new virtualenv).
Even rarer I use it to run things with Python 2.7 if that version specfically failed the Travis-CI tests.

tbekolay:
I've rebased this now that #1425 is merged, so this is ready for review again!

PR #1425, opened 2018-04-05
Simplify root directory

tbekolay:
**Motivation and context:**
In #1424, a change to `tox.ini` was missed because I don't use `tox.ini` anymore, and it seems no other developers do either, so it makes sense to remove it. There have been a few other things living in the root `nengo` directory that I've been wanting to remove for some time, so it makes sense to do it all in one PR.

#1424 is blocked by this; will rebase it once this is merged.

**How has this been tested?**
There are changes to `.travis.yml` and Appveyor so I'm relying on those tests to run.

**Types of changes:**

**Checklist:**

**Still to do:**

I'm pushing now to get the TravisCI tests running, and to start a discussion about some of the more controversial parts of this PR.

- [x] I want to do more rearranging of `.travis.yml` and `.appveyor.yml`; it would be nice if we didn't have to rewrite all of these for every project.


Seanny123:
TIL:
1. You can install dependencies with `'.[all]'` 
2. The difference between `'` and `"` is returning a char or a string in C++/Julia. I'm assuming this is the reason you're doing the switch?

tbekolay:
> You can install dependencies with '.[all]'

Only with this PR because of the changes to `setup.py` ;)

> The difference between ' and " is returning a char or a string in C++/Julia. I'm assuming this is the reason you're doing the switch?

No, there's no difference between ' and " in Python, except that they let you put " and ' in the string respectively (e.g. `'"hey"'` and `"'hey'"`). I just noticed it was inconsistent in a few places so I did a find/replace to make them all " (which I prefer, personally).

tbekolay:
OK this is all done and ready for review. Reviewers, do chime in with whether you're OK with removing the requirements files.

drasmuss:
What do you think about organizing the ci scripts by function? So instead of having `install-coverage.sh`, `script-coverage.sh`, and `post-coverage.sh`, we would just have a `coverage.sh` that we call like `coverage.sh install`, `coverage.sh script`, `coverage.sh post`?

tbekolay:
> So instead of having install-coverage.sh, script-coverage.sh, and post-coverage.sh, we would just have a coverage.sh that we call like coverage.sh install, coverage.sh script, coverage.sh post?

Sounds like a good idea! Will implement first thing next week :+1:

tbekolay:
Implemented your suggestion in 1f746b9699788849eedbd24a1a54dd3fe1 @drasmuss ... that what you had in mind?

PR #1430, opened 2018-04-19
Documentation and website updates

celiasmith:
**Motivation and context:**
Recent discussions in ABR and the main nengo website <https://github.com/nengo/nengo.github.io/issues/44> suggested some menu and wording alterations.

**How has this been tested?**
N/A

**Types of changes:**

**Checklist:**


docs/backend_api.rst
```diff
@@ -5,7 +5,7 @@ Reference simulator
 .. default-role:: obj
 
 Nengo is designed so that models created with the
-:doc:`Nengo modeling API <frontend_api>`
+:doc:`Nengo core API <core_api>`
```
drasmuss: I don't love the name "Nengo core", mainly because I don't think "core" really indicates what users should expect to find there.  E.g., is the backend less "core" than the frontend?  I also think we have been using "Nengo core" in other places to refer more generally to the frontend+default nengo backend.

Personally I thought "Nengo modeling API" was pretty clear, but of course that's all subjective :smile:.  If we want something different, perhaps we could just use the frontend/backend language directly (i.e. "Nengo frontend API")?

celiasmith: I picked 'nengo core' because it's what's in this picture: https://www.nengo.ai/projects.html and in that context is referring only to the frontend API, I think.

I'm definitely not wedded to it, though.  The frontend/backend would be better if we plan on having separate docs for both, for instance.  Mainly I wanted to avoid calling it 'frontend api' in once place (the url) and 'modelling api' in a different place (the menu link).

drasmuss: > I picked 'nengo core' because it's what's in this picture: https://www.nengo.ai/projects.html and in that context is referring only to the frontend API, I think.

Yeah that's true.  Although right below that, we have "Core frameworks" which includes a bunch of stuff not in what we'd call "Nengo core", and then inside "Core frameworks" we have "Nengo core" which includes the reference simulator.  So we might be overloading the word "core" there :wink:

> The frontend/backend would be better if we plan on having separate docs for both, for instance. 

We do have a backend_api page (describing the simulator api), although it's not labeled as such (https://www.nengo.ai/nengo/backend_api.html). So we could apply the frontend/backend terminology there if we want.

> Mainly I wanted to avoid calling it 'frontend api' in once place (the url) and 'modelling api' in a different place (the menu link).

I don't think it's a general expectation that menu headings match up with URLs.  E.g. if we look at https://www.tensorflow.org/programmers_guide/ (just the first thing I looked at since I had the tab open), a lot of the left-hand links direct to pages with URLs that don't match the menu heading.  Which isn't to say that we shouldn't have a more informative name, if we think "Nengo modeling API" isn't clear.  But I don't think we need to constrain ourselves to "what would be an informative name AND a good URL".  Or we could change the page to "modeling_api.html" if we like that name but want to have the names match.

celiasmith:
ok, these considerations seem to point to the frontend/backend api
distinction... are there any reasons not to go with that?

tbekolay:
> are there any reasons not to go with that?

Frontend sounds good, because we only have one Nengo frontend, the one in Nengo core. However, there are many backends, so calling the simulator docs the Nengo backend API could be a bit misleading... really, we do want generic Nengo backend documentation, and then also Nengo core backend documentation, so I think it's fine to switch to frontend/backend for now, then when we have generic backend docs put those there and move the core backend docs to a different file (or a different section in the same file)

README.rst
```diff
@@ -19,8 +19,7 @@
 Nengo: Large-scale brain modelling in Python
 ********************************************
 
-.. image:: http://c431376.r76.cf2.rackcdn.com/71388/fninf-07-00048-r2/image_m/fninf-07-00048-g001.jpg
-  :alt: An illustration of the three principles of the NEF
+Nengo is a Python library for building and simulating large-scale neural models. Nengo can create sophisticated spiking or non-spiking neural simulations with sensible defaults in a few lines of code. Yet, Nengo is highly extensible and flexible. You can define your own neuron types and learning rules, get input directly from hardware, build and run deep neural networks, drive robots, and even simulate your model on a completely different neural simulator or neuromorphic hardware.
```
tbekolay: I'll fix this in the merge, so not blocking, but even in RST docs we want to keep to around 80 characters per line in order to make future diffs easier to read (scanning a really long line for the change takes a while). Then also, since we're manually adding line breaks, I prefer to use [semantic line breaks](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) for even easier plaintext reading and diffing.

tbekolay: Also, I did kind of like that image as an eye catcher for the README... any objections to me adding it back in? I know the link is broken so I'll put it somewhere that we control.

docs/networks.rst
```diff
@@ -10,7 +10,8 @@ though usually not `.Probe` instances.)
 Like most abstractions, this helps with code-reuse and maintainability.
 You'll find the documentation for the various pre-built networks below.
 
-Building your own network can be a great way to encapsulate
+You may also want to pre-build your own networks.  Doing so can
```
tbekolay: The concept of "pre-building" a network feels strange to me, which makes me think that "pre-built" networks might be confusing too.

How about instead of "pre-built" we introduce the term "Reusable networks"; then, it makes more sense to tell users that they can also build reusable networks (rather the pre-built networks).

README.rst
celiasmith: nope no objection, i took it out because of the broken link.

docs/networks.rst
celiasmith: yeah, i wasn't a fan of 'pre-built' either (mostly when referring to ones users would make). '  'Reusable' is better.  I've been trying to think of something even better. Have come up with 'pre-built', 'template', 'component', but none of those sound any better, so 'reusable' is fine by me.

drasmuss:
Fixed an unrelated typo in the changelog I noticed when testing this.  Not sure why this is just popping up now, probably some sphinx change that made this an error instead of a silent failure.  But figured might as well stick it in here since we're touching a bunch of docs stuff with this PR :smile:

tbekolay:
I made a fixup commit (1525b5ed6c2209381968cce42) with the linebreaks, and a few other minor things (but not so minor that I can merge it unreviewed in good conscience).

One of the things that commit does is, similar to the PR for nengo.github.io, remove the `contrib_guide.rst` doc and put everything in `contributing.rst`, as the only things linked from there were that doc, the license (which I moved up to the main index) and nef_minimal (which I linked from `contributing.rst`).

But, I have to admit that `nef_minimal.rst` sticks out like a sore thumb... it doesn't really make sense anywhere, and it strange because it's mostly a Python file. Does anyone has any objections to me converting it to an example notebook and linking that notebook from `examples.rst` instead?

If this is contentious I'll merge this as is.

drasmuss:
Yeah those changes look good, as well as moving `nef_minimal` to an example notebook :+1:

tbekolay:
> Yeah those changes look good, as well as moving nef_minimal to an example notebook

OK, done in 6dd0e8e. Will merge once I build the docs locally and check them.

Issue #1431, opened 2018-04-20
Setting gains, biases, intercepts, and max_rates simultaneously.


pblouw:
It's currently possible to specify an inconsistent collection of these parameters when creating an ensemble, in which case the tuning curves are set by the gains and biases (since these are fundamental from the perspective of a neuron model). For example:

```
with nengo.Network() as net:
    ens = nengo.Ensemble(1, 1, encoders=[[1]], gain=[1], bias=[1], max_rates=[250], intercepts=[0.2])

with nengo.Simulator(net) as sim:
    eval_points, activities = tuning_curves(ens, sim)

plt.figure()
plt.plot(eval_points, activities)
plt.ylabel("Firing rate (Hz)")
plt.xlabel("Input Value")
```
This gives the following tuning curves, for which e.g. the intercept is clearly different from what is expected: 
![tuning_curve](https://user-images.githubusercontent.com/2708813/39073987-bf235d92-44bd-11e8-89d7-376cf79b2322.png)

Would it make sense to add a warning to explain what is happening here?

PR #1433, opened 2018-04-24
Warn if gain, bias override intercepts, max_rates

pblouw:
**Motivation and context:**

See issue [#1431](https://github.com/nengo/nengo/issues/1431). This PR adds a warning that occurs whenever an ensemble is built for which specified gains and biases override specified intercepts and/or max_rates. I'm happy to get feedback regarding, for example, whether it makes sense to either issue a warning elsewhere, or add something to the documentation that explains how gains, biases, max_rates, and intercepts are related, and then count on the user not to create ensembles of the sort under consideration here. 

None so far as I know.

**How has this been tested?**

I've added a test that checks whether a `UserWarning` is issued whenever an ensemble is built for which gains, biases and at least one of max_rates and intercepts are specified. 

The `get_gain_bias` function in `nengo/builder/ensemble.py` has the core change, so it's likely best to start there before moving on to looking at the corresponding test.

**Types of changes:**

**Checklist:**

There's no docstring for the `get_gain_bias` function, and it is likely not something that a user would ever call directly, so I haven't added any documentation. It also seems like a small enough change to avoid including a changelog entry, but I'm happy add this if it makes sense to. 

**Still to do:**

Nothing at this point.

### DIFF
--- a/nengo/builder/ensemble.py
+++ b/nengo/builder/ensemble.py
@@ -8,7 +8,7 @@ from nengo.builder import Builder, Signal
 from nengo.builder.operator import Copy, DotInc, Reset
 from nengo.dists import Distribution, get_samples
 from nengo.ensemble import Ensemble
-from nengo.exceptions import BuildError
+from nengo.exceptions import BuildError, NengoWarning
 from nengo.neurons import Direct
 from nengo.utils.builder import default_n_eval_points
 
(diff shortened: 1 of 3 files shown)


nengo/builder/ensemble.py
```diff
@@ -88,6 +88,13 @@ def get_gain_bias(ens, rng=np.random):
         bias = get_samples(ens.bias, ens.n_neurons, rng=rng)
         max_rates, intercepts = ens.neuron_type.max_rates_intercepts(
             gain, bias)
+        if (ens.max_rates is not Ensemble.max_rates.default or
+           ens.intercepts is not Ensemble.intercepts.default):
+            warnings.warn(
+                "Specifying the gains and biases for %s imposes a set of "
+                "maximum firing rates and intercepts. Further specifying "
+                "either max_rates or intercepts has no effect." % ens)
```
jgosmann: I would make this a `NengoWarning`.

pblouw: When initially thinking about this, I stuck with a `UserWarning` for consistency with the warning issued by `gen_eval_points`. In any event, I've added a fixup to use a `NengoWarning`, but it doesn't seem like there is a widespread usage of these throughout the codebase (i.e. most warnings are `UserWarning`s based on a quick search). Is there a rule of thumb for when one should preferred over the other? Or should most of the existing `UserWarning`s be switched over to `NengoWarning`s? (I'm happy to do this at some point if that is the case)

Issue #1419, opened 2018-03-09
Correction to NeuronType.current method docstring


arvoelke:
Currently the `NeuronType.current` method has the docstring:

```
        Parameters
        ----------
        x : (n_neurons,) array_like
            Vector-space input.
        gain : (n_neurons,) array_like
            Gains associated with each neuron.
        bias : (n_neurons,) array_like
            Bias current associated with each neuron.
```

But the `x` parameter does not need to have the shape `(n_neurons,)`. It can be shaped as `(m, 1)`. This is how it is used by `nengo.builder.ensemble.get_activities(...)`. Not sure if there are other ways it can also be shaped.

The docstring is also missing a `Returns` section. In the case where `x.shape == (m, 1)`, the shape of the returned array is `(m, n_neurons)`.

hunse:
Closed by #1437.

PR #1437, opened 2018-05-17
Reworking NeuronType.current and NeuronType.rates

tbekolay:
**Motivation and context:**
In #1419, @arvoelke pointed out that the docstring of the `current` method (and, by association, `rates`) did not reflect the way in which these methods are being used. When I looked into it more, I realized that these were not usages that I intended, hence the incorrect docstrings, but also that these usages were now the majority of use cases because they allow for more efficient code.

To explain, `current` is just a tiny helper function that does `gain * x + bias`. I originally intended for all three of these variables to be of shape `(n_neurons,)`, so it's all simple element-wise multiplication. However, at some point we started passing in `x` values with shapes `(n_samples, 1)`, which due to NumPy's broadcasting rules, ends up giving you a result that is of shape `(n_samples, n_neurons)`.

The use case I was envisioning, passing in `(n_neurons,)`, lets you provide different `x` values to each neuron. When passing in `(n_samples, 1)`, you provide the same input to all neuron, but you can provide multiple inputs at once. When passing in `(n_samples, n_neurons)`, you can provide multiple inputs and those inputs can be different for each neuron. Hopefully it's obvious that the more efficient thing to do is to just call `current` once with all of the samples (usually eval_points), so that tends to be the most common use case. Passing in `(n_samples, 1)` ends up being useful when generating response curves, since you want to see every neuron's response to the same input. Passing in `(n_neurons,)` is, in most cases, just an inefficient version of passing in `(n_samples, n_neurons)`.

Note that in all of these cases, `x` has already been projected onto the neuron's encoding vector and has been scaled by the radius, which has further confused things because

1. We don't have a separate name for the original input vector `x` and the `x` we're using in this case.
2. I called `x` the "input vector" in the previous docstring, which is sort of defensible but quite confusing.

Also note that while everything I said above was for `current`, the same is also true of `rates` because `rates` first calls `current` and then makes other variables the same shape as the value returned by `current`.

With all that context, it is hopefully easier to understand the changes that I made in this PR. I have essentially implemented two ways to clear up the confusion that I've explained above.

The first way (commit 81a6fdf692cbe40215aa89) tries to maintain backwards compatibility, but make things slightly more consistent, but having `current` (and consequently `rates`) always return a 2d array of shape `(n_samples, n_neurons)`, even if the cases with 1 sample -- which, as I have noted above, ends up being a relatively rare use case, even though it was the one in the docstring. It was relatively easy to fix any tests that were failing as a result of this change, though it resulted in some ugliness in parts.

The second way (commit 0c0d96510220ca5a) attempts to embrace the more common use cases, at the cost of backwards compatibility. This commit assumes that we pretty much always want to be passing in more than one sample, and makes the one-sample case the slightly awkward one (e.g., you have to pass in something of shape `(1, n_neurons)` if you really want one sample). This was also somewhat easy to fix, and I think the result is easier to deal with and easier to explain. However, it may be the case that other backends are relying on the current behavior.

So, if anyone has any thoughts on whether one of these two approaches is preferred, or if there's some obviously better third option I haven't thought of, please share it here. Also, if anyone knows of any usages of these methods that will be affected by these changes (@arvoelke ?) it would be good to know ahead of time to either defer this to a 3.0 release or at least knowingly break a few things.

**How has this been tested?**
The usual test suite.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -57,18 +57,35 @@ class NeuronType(FrozenObject):
     def current(self, x, gain, bias):
         """Compute current injected in each neuron given input, gain and bias.
 
+        Note that ``x`` is assumed to be already projected onto the encoders
+        associated with the neurons and normalized to radius 1, so the maximum
+        expected current for a neuron occurs when input for that neuron is 1.
+
         Parameters
         ----------
-        x : (n_neurons,) array_like
-            Vector-space input.
+        x : (n_samples,) or (n_samples, n_neurons) array_like
+            Scalar inputs for which to calculate current.
         gain : (n_neurons,) array_like
             Gains associated with each neuron.
         bias : (n_neurons,) array_like
             Bias current associated with each neuron.
+
+        Returns
+        -------
+        current : (n_samples, n_neurons)
+            Current to be injected in each neuron.
         """
         x = np.array(x, dtype=float, copy=False, ndmin=1)
         gain = np.array(gain, dtype=float, copy=False, ndmin=1)
         bias = np.array(bias, dtype=float, copy=False, ndmin=1)
+
+        if x.ndim == 1:
+            x = x[:, np.newaxis]
+        elif x.ndim >= 3 or x.shape[1] != gain.shape[0]:
+            raise ValidationError("Expected shape (%d, %d); got %s."
+                                  % (x.shape[0], gain.shape[0], x.shape),
+                                  attr="x", obj=self)
+
         return gain * x + bias
 
     def gain_bias(self, max_rates, intercepts):
@@ -102,9 +119,8 @@ class NeuronType(FrozenObject):
         max_rate = max_rates.max()
 
         # Start with dummy gain and bias so x == J in rate calculation
-        gain = np.ones(J_steps)
-        bias = np.zeros(J_steps)
-        rate = np.zeros(J_steps)
+        gain = np.ones(1)
+        bias = np.zeros(1)
 
         # Find range of J that will achieve max rates (assume monotonic)
         J_threshold = None
@@ -128,11 +144,12 @@ class NeuronType(FrozenObject):
                 raise RuntimeError("Could not find max current")
 
         J = np.linspace(J_threshold, J_max, J_steps)
-        rate = self.rates(J, gain, bias)
+        rate = self.rates(J, gain, bias).squeeze(axis=1)
 
         gain = np.zeros_like(max_rates)
         bias = np.zeros_like(max_rates)
         J_tops = np.interp(max_rates, rate, J)
+
         gain[:] = (J_threshold - J_tops) / (intercepts - 1)
         bias[:] = J_tops - gain
         return gain, bias
@@ -158,28 +175,31 @@ class NeuronType(FrozenObject):
             X-intercepts of neurons.
         """
 
-        max_rates = self.rates(np.ones_like(gain), gain, bias)
+        max_rates = self.rates(1, gain, bias).squeeze(axis=0)
 
         x_range = np.linspace(-1, 1, 101)
-        rates = np.asarray([self.rates(np.ones_like(gain) * x, gain, bias)
-                            for x in x_range])
+        rates = self.rates(x_range, gain, bias)
         last_zeros = np.maximum(np.argmax(rates > 0, axis=0) - 1, 0)
         intercepts = x_range[last_zeros]
 
         return max_rates, intercepts
 
     def rates(self, x, gain, bias):
-        """Compute firing rates (in Hz) for given vector input, ``x``.
+        """Compute firing rates (in Hz) for given input ``x``.
 
         This default implementation takes the naive approach of running the
         step function for a second. This should suffice for most rate-based
         neuron types; for spiking neurons it will likely fail (those models
         should override this function).
 
+        Note that ``x`` is assumed to be already projected onto the encoders
+        associated with the neurons and normalized to radius 1, so the maximum
+        expected rate for a neuron occurs when input for that neuron is 1.
+
         Parameters
         ----------
-        x : (n_neurons,) array_like
-            Vector-space input.
+        x : (n_samples,) or (n_samples, n_neurons) array_like
+            Scalar inputs for which to calculate rates.
         gain : (n_neurons,) array_like
             Gains associated with each neuron.
         bias : (n_neurons,) array_like
@@ -187,7 +207,7 @@ class NeuronType(FrozenObject):
 
         Returns
         -------
-        rates : (n_neurons,) ndarray
+        rates : (n_samples, n_neurons) ndarray
             The firing rates at each given value of ``x``.
         """
         J = self.current(x, gain, bias)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -92,6 +92,9 @@ Release History
   (`#1521 <https://github.com/nengo/nengo/pull/1521>`_)
 - Switched to nengo-bones templating system for TravisCI config/scripts.
   (`#1514 <https://github.com/nengo/nengo/pull/1514>`_)
+- The ``NeuronType.current`` and ``NeuronType.rates`` methods now document
+  the supported shapes of parameters and return values.
+  (`#1437 <https://github.com/nengo/nengo/pull/1437>`__)
 
 **Deprecated**
 
--- a/nengo/tests/test_neurons.py
+++ b/nengo/tests/test_neurons.py
@@ -4,7 +4,7 @@ import numpy as np
 import pytest
 
 import nengo
-from nengo.exceptions import BuildError, SimulationError
+from nengo.exceptions import BuildError, SimulationError, ValidationError
 from nengo.neurons import (
     AdaptiveLIF,
     AdaptiveLIFRate,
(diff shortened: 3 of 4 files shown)


arvoelke:
I've wrote and scrapped my reply a couple times now lol. Your commit descriptions make things very clear, but it is still very complicated trying to compare these three things while considering all of the possibilities.

I don't think anything in `nengolib` would be broken here by either version. I do have a couple unpublished notebooks and experiments that use `rates` in some way that might break, but it's really hard to tell. I wouldn't be surprised if others have accidentally relied upon these methods accepting/returning certain shapes, even just by virtue of how programming with NumPy goes some times (i.e., keep hacking the axes until it finally does what you wanted).  :P But this is more of a general observation than it is a specific concern about either of these two versions.

If I understand this right, backwards-compatibility here, for either version, is riding on the assumption that nobody has been calling it as `(n_neurons,)`, despite that being the main use-case advertised by the documentation. I've never called it that way. But, if someone had been calling it this way, I think there is potential for their code to break in very subtle (and possibly even silent) ways, with either version. This is concerning from the perspective of wanting users to trust our documentation.

For the first version, I think the only other difference is the case of `(1,)` which used to return `(n_neurons,)` but would now return `(1, n_neurons)`. If nobody has been calling it either way, I think the first version is the safest, and indeed should not have any other differences...?

Similarly, with the second version, weird things might happen when `n_neurons=1` and/or `n_samples=1` due to the squeezing. The new documentation currently only mentions the latter case.

I'm also not sure if we want to go down this rabbit hole, but the old version even accepts 3 or more dimensions. To maintain backwards-compatibility with that, I think both of your versions need to check against `x.shape[-1]` instead of `x.shape[1]`.

I don't think I have been relying on any of these edge-cases, but, again, it's hard to say for sure. After trying to do this in my head, I feel the need to enumerate out all of the possibilities and create a giant table (with a column for each of the three versions) to figure out what will happen in all cases just to be sure...

Speaking of which, I noticed that it also accepts a scalar, and you rely on this behaviour, but that isn't mentioned in the documentation.

arvoelke:
Just one last thing:

> The second way ... assumes that we pretty much always want to be passing in more than one sample, and makes the one-sample case the slightly awkward one (e.g., you have to pass in something of shape (1, n_neurons) if you really want one sample).

Is this true? It looks like you could pass in either a scalar (like `max_rates = self.rates(1, gain, bias)`) or something of shape `(1,)`.  I thought the awkward part was just the returned array being squeezed.

tbekolay:
> I've wrote and scrapped my reply a couple times now lol.

Haha the initial post took a few gos too ;)

> if someone had been calling it this way, I think there is potential for their code to break in very subtle (and possibly even silent) ways, with either version

That's true, there is also a slight variant of the first option that still does the squeeze, or doesn't change the behavior at all and just changes the documentation. The reason I didn't go for that was mainly because it made the docs pretty awkward to write, but if backward compatibility is important then we can go that route.

> Similarly, with the second version, weird things might happen when n_neurons=1 and/or n_samples=1 due to the squeezing. The new documentation currently only mentions the latter case.

Right, I did not consider the case with `n_neurons=1`. That dimension should probably never be squeezed away, so if we go with option 2, I'll add some tests for that case and not squeeze away the `n_neurons` dimension.

> I'm also not sure if we want to go down this rabbit hole, but the old version even accepts 3 or more dimensions. To maintain backwards-compatibility with that, I think both of your versions need to check against x.shape[-1] instead of x.shape[1].

I really hope people weren't doing this, but yes it is possible. I'd be in favor of explicitly disallowing it personally.

> Speaking of which, I noticed that it also accepts a scalar, and you rely on this behaviour, but that isn't mentioned in the documentation.

Right, I believe that it should be the case that passing in a scalar and a vector of shape `(1,)` (meaning the same one sample provided to all neurons) work the exact same, so I didn't mention it because my default assumption is that scalars and vectors of shape `(1,)` are the same thing unless otherwise stated. But if that's not most people's assumption I can be explicit about it... again, I was trying to make sure that the docs weren't super long and confusing.

> Is this true? It looks like you could pass in either a scalar (like max_rates = self.rates(1, gain, bias)) or something of shape (1,). I thought the awkward part was just the returned array being squeezed.

By "the one-sample case" I meant the case in which you have one sample per neuron, but a different input value for each neuron. Providing the same sample to every neuron is not awkward, as you say, but in the few cases in the Nengo core code that we were using the documented `(n_neurons,)` input, it has to be changed to `(1, n_neurons)` (even though the returned value is the same due to the squeezing -- the squeezing should make it easier to work with, not harder).

nengo/neurons.py
```diff
@@ -135,36 +148,39 @@ def max_rates_intercepts(self, gain, bias):
...
+        Note that ``x`` is assumed to be already projected onto the encoders
+        associated with the neurons and normalized to radius 1, so the maximum
+        expected rate for a neuron occurs when input for that neuron is 1.
+
         Parameters
         ----------
-        x : (n_neurons,) array_like
-            Vector-space input.
+        x : (n_samples, 1), (n_samples, n_neurons) array_like
+            Scalar inputs for which to calculate current.
```
arvoelke: current -> rates

nengo/neurons.py
```diff
@@ -171,8 +175,8 @@ def rates(self, x, gain, bias):
 
         Parameters
         ----------
-        x : (n_samples, 1), (n_samples, n_neurons) array_like
-            Scalar inputs for which to calculate current.
+        x : (n_samples,) array_like
```
arvoelke: Should this have been kept as `x : (n_samples, 1), (n_samples, n_neurons) array_like` (same as the first version)? Or am I misunderstanding why this is different in this second version?

arvoelke:
Just want to make a quick note here that, as of `nengolib==0.4.2` (https://github.com/arvoelke/nengolib/pull/137), I am now using the `rates` method in a way that relies on the current 2D input->output. I'm pretty sure this doesn't break with either of your proposed changes. If it does, then that is no big deal (I'll figure something out)!

tbekolay:
> I am now using the rates method in a way that relies on the current 2D input->output

2D inputs shouldn't be affected in either version :+1: Thanks for the note!

nengo/neurons.py
```diff
@@ -34,19 +34,37 @@ def _argreprs(self):
...
-        return gain * x + bias
+
+        if x.ndim == 1:
+            x = x[:, np.newaxis]
+        elif x.ndim > 1 and x.shape[1] != gain.shape[0]:
+            raise ValidationError("Expected shape (%d, %d); got (%d, %d)."
+                                  % ((x.shape[0], gain.shape[0]) + x.shape),
+                                  attr="x", obj=self)
+
+        return np.squeeze(gain * x + bias)
```
hunse: I don't like the `squeeze`. Often, I work explicitly with a certain number (e.g. 2) dimensions, and if `n_neurons` or `n_samples` happens to be 1, I don't want to lose those dimensions. This could potentially cause bugs if e.g. we have an ensemble with 1 neuron.

arvoelke: I agree that I'd prefer to avoid the squeeze. If you do a ctrl+F in the main PR thread you'll see a bunch of discussion on this and the various cases.

nengo/neurons.py
```diff
@@ -34,19 +34,37 @@ def _argreprs(self):
...
         x = np.array(x, dtype=float, copy=False, ndmin=1)
         gain = np.array(gain, dtype=float, copy=False, ndmin=1)
         bias = np.array(bias, dtype=float, copy=False, ndmin=1)
-        return gain * x + bias
+
+        if x.ndim == 1:
+            x = x[:, np.newaxis]
+        elif x.ndim > 1 and x.shape[1] != gain.shape[0]:
+            raise ValidationError("Expected shape (%d, %d); got (%d, %d)."
+                                  % ((x.shape[0], gain.shape[0]) + x.shape),
```
hunse: This string formatting will break if `x.ndim > 2`.

arvoelke: Do you think we should disallow that case? Previously it would "just work" according to numpy's usual broadcasting, but I can see an argument for limiting things to the typical use-cases.

hunse: Actually it would only work if we required `x.shape[-1]` to be equal to `gain.shape[0]`. Even though I really prefer more general code, it's probably better to disallow it. That way, we can keep the docstring as-is and have it properly documented.

arvoelke:
I reworked option 2 to get yet another option that does not `.squeeze()` the return value. Instead `current` and `rates` always return a 2D array of shape `(n_samples, n_neurons)`. I think this is much less error-prone / surprising because it doesn't do things differently for the special cases of `n_samples=1` or `n_neurons=1`. Squashed all of its history together onto master for readability, but it might be good to separate out the extra testing and some of the other misc/unrelated changes into another commit, if we're happy with this direction. See the commit in the `current-rates-nosqueeze` branch: https://github.com/nengo/nengo/commit/71f308f9cc220c24118ab1f7f029aa31f966650e.

nengo/neurons.py
arvoelke: This is done in option 3.

nengo/neurons.py
arvoelke: This is done in option 3.

tbekolay:
Option 3 in `current-rates-nosqueeze` looks great to me. A bit of history splitting makes sense too.

Issue #1421, opened 2018-03-23
Broken link in Nengo docs


Seanny123:
The link in `getting_started.rst` links to `user_api.rst`, which doesn't exist. Should be changed to `frontend_api.rst`. Ideally, these types of broken links should be caught automatically somehow? Maybe they will be caught once the new documentation build PRs are merged? Just wanted to make a note to check back on this later.

jgosmann:
Should be fixed?

PR #1438, opened 2018-05-17
Fix documentation links

tbekolay:
**Motivation and context:**
@Seanny123 found a broken link in #1421. A bit of grepping revealed a few other broken links. This PR fixes them by using actual Sphinx directives, which we should always use because they would have caught these issues. There is probably also a dead link checker Sphinx extension, but ... meh!

**How has this been tested?**
Built the docs, checked that the links work now.

**Types of changes:**

**Checklist:**

- [na] I have run the test suite locally and all tests passed.

### DIFF
--- a/.ci/docs.sh
+++ b/.ci/docs.sh
@@ -15,6 +15,8 @@ function usage {
 if [[ "$COMMAND" == "install" ]]; then
     conda install jupyter matplotlib scipy
     pip install "sphinx<1.7" nbsphinx numpydoc guzzle_sphinx_theme ghp-import
+elif [[ "$COMMAND" == "check" ]]; then
+    sphinx-build -b linkcheck -vW -D nbsphinx_execute=never docs docs/_build
 elif [[ "$COMMAND" == "run" ]]; then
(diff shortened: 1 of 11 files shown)


drasmuss:
Sphinx does indeed have a dead link checker, I use it in `nengo-dl` during the doc CI build.  Something like this

`sphinx-build -b linkcheck docs docs/_build -W -D nbsphinx_execute=never`

(`-W` makes it fail on warnings, and `-D nbsphinx_execute=never` means that it doesn't try to build the notebooks during the link checking).  I'm fine not having it in, but just thought I'd save some googling if we do want to add it to nengo as part of this PR :smile:

tbekolay:
I added the dead link checking and fixed all the dead or redirected links. One link (https://www.lfd.uci.edu/~gohlke/pythonlibs/) was unfortunately not dead, but could not be reached by `requests`, the library that Sphinx uses to check links. [Another project](https://github.com/SciTools/cartopy/issues/978) compromised by leaving in the link, but not clickable, which seemed an okay compromise, but I can switch to removing the link entirely if that's preferable.

PR #1440, opened 2018-05-23
Upgrade notebook format, fix style

tbekolay:
**Motivation and context:**
We have a fair bit of clunky code dedicated to dealing with old school IPython notebooks, and have been keeping our notebooks in the old nbformat 3 for compatibility reasons. I think we had talked about upgrading them for Nengo 3.0, but honestly I can't imagine anyone is still using those ancient versions of IPython anymore, so I feel like we can upgrade them to make our lives easier.

After doing the upgrade using nbconvert, I looked into how easy it would be to run the usual flake8 and pylint style checks on the example notebooks, and while the original `code-prettify` solution seemed difficult to get working, it ended up being relatively easy to use nbconvert and a few other shell tools to convert the notebook examples to a Python script and run the static checks on those scripts. The conversion process results in a few false positives, which for the most part I disabled (e.g., semicolons after some statements are things we want in the notebooks even though it's bad style in a Python script). Others I fixed, mostly long lines, which is a thing I was wanting to fix in our notebooks for a while anyway.

This is based on #1436, which fixes example testing. I'll be merging that shortly anyhow though.

**How has this been tested?**
I added tests to make sure that all examples are in notebook format 4, and don't contain some pieces of metadata that could be problematic.

You will want to look at this commit-by-commit, as the commit switching everything to format 4 is gigantic but all done automatically, so it doesn't need to be scrutinized.

**Types of changes:**

**Checklist:**


.ci/static.sh
```diff
@@ -12,9 +15,20 @@ function usage {
...
+    conda install jupyter
     pip install codespell flake8 pylint
 elif [[ "$COMMAND" == "run" ]]; then
-    flake8 -v nengo && codespell -q 3 && pylint nengo
+    jupyter-nbconvert \
+        --log-level WARN \
+        --to python \
+        --TemplateExporter.exclude_input_prompt=True \
+        -- **/*.ipynb
+    sed -i -e 's/# $/#/g' -e '/get_ipython()/d' -- docs/**/*.py
```
Seanny123: Would you mind adding a comment explaining what this line accomplishes for the shell noobs?

tbekolay: Can do :+1:

tbekolay: Did.

.ci/static.sh
```diff
@@ -18,11 +18,15 @@ if [[ "$COMMAND" == "install" ]]; then
...
 elif [[ "$COMMAND" == "run" ]]; then
+    # Convert notebooks to Python scripts
     jupyter-nbconvert \
         --log-level WARN \
         --to python \
         --TemplateExporter.exclude_input_prompt=True \
         -- **/*.ipynb
+    # Remove style issues introduced in the conversion:
+    #   s/# $/#/g replaces lines with just '# ' with '#'
+    #   /get_ipython()/d removes lines containing 'get_ipython()'
```
tbekolay: Doesn't warrant more comments in this file, but `get_ipython` raises flake8 errors because IPython injects it into the local namespace rather than importing it within the script, which honestly would make a lot more sense. In any case, `get_ipython` is only used for magic functions, which we wouldn't want to add to the script versions of these notebooks anyway.

.ci/static.sh
```diff
@@ -18,11 +18,15 @@ if [[ "$COMMAND" == "install" ]]; then
...
     pip install codespell flake8 pylint
 elif [[ "$COMMAND" == "run" ]]; then
+    # Convert notebooks to Python scripts
     jupyter-nbconvert \
         --log-level WARN \
         --to python \
         --TemplateExporter.exclude_input_prompt=True \
         -- **/*.ipynb
+    # Remove style issues introduced in the conversion:
+    #   s/# $/#/g replaces lines with just '# ' with '#'
```
tbekolay: Yeah for some reason when nbconvert converts markdown cells, it makes everything a comment, and when it makes empty lines into comments it does it by having the line contain `# ` rather than just `#`, and `# ` is a style issue because of trailing whitespace.

nengo/utils/ipython.py
```diff
@@ -55,21 +24,15 @@
     if IPython.version_info[0] <= 2:
```
Seanny123: Given we're only supporting versions 4 and greater with this PR, do we still need to try and meet all these cases?

nengo/tests/test_examples.py
```diff
@@ -73,6 +81,8 @@ def assert_noexceptions(nb_file, tmpdir):
 @pytest.mark.filterwarnings("ignore:Creating new attribute 'memory_location'")
 def test_fast_noexceptions(nb_file, tmpdir):
     """Ensure that no cells raise an exception."""
+    pytest.importorskip("IPython", minversion="1.0")
```
Seanny123: I'm not sure if this is relevant to this PR, but why is the `minversion="1.0"`. I think I'm confused about the relation between `Jupyter` and `IPython`. Please let me know if I shouldn't worry about this.

nengo/utils/ipython.py
tbekolay: Yeah we should probably go through and remove these old things ... just didn't want to risk breaking anything (and also lazy), but I think it's worth doing.

nengo/tests/test_examples.py
tbekolay: Right yeah we should probably bump these up, I copied the line from elsewhere in this file.

nengo/utils/ipython.py
Seanny123: Just to be clear, is it worth doing in this PR or in a different PR?

tbekolay: I'm happy to do it in this PR :man_shrugging:

Seanny123: Do you want me to make the fixup commit or are you already doing it?

tbekolay: I'm on it

tbekolay:
I fixed the things that @Seanny123 mentioned. Note that in some places there are still some things for earlier versions of IPython, but it's because notebook version 4 was introduced in IPython version 3, so it's still relevant (for the time being).

Also, since this touches all of the notebooks and will be a pain to rebase if notebooks change, it would be nice to have this reviewed before any other PR touching the notebooks.

docs/examples/dynamics/controlled_integrator2.ipynb
```diff
@@ -5,7 +5,20 @@
...
+    "However, the control signal is zero for integration,\n",
+    "less than one for low-pass filtering, and greater than 1 for saturation.\n",
+    "This behavior maps more directly to the differential equation\n",
+    "used to describe an integrator:\n",
+    "\n",
+    "$$\\dot{x} = \\mathrm{Ax}(t) + \\mathrm{Bu}(t)$$\n",
+    "\n",
+    "The control in this circuit is $A$ in that equation.\n",
+    "This is also the controlled integrator\n",
+    "described in the book \"How to build a brain.\""
```
jgosmann: I believe the puncutation should be outside of the quotes of the title (if the English rules don't differ from the German rules).

docs/examples/dynamics/lorenz_attractor.ipynb
```diff
@@ -6,22 +6,30 @@
...
+    "\\dot{x}_1 = - x_0 x_2' - x_1 \\\\\\\n",
     "\\dot{x}_2' = x_0 x_1 - \\beta (x_2' + \\rho) - \\rho\n",
     "$$\n",
     "\n",
-    "For more information, see http://compneuro.uwaterloo.ca/publications/eliasmith2005b.html \"Chris Eliasmith. A unified approach to building and controlling spiking attractor networks. Neural computation, 7(6):1276-1314, 2005.\""
+    "For more information, see\n",
+    "http://compneuro.uwaterloo.ca/publications/eliasmith2005b.html\n",
+    "\"Chris Eliasmith. A unified approach to building\n",
+    "and controlling spiking attractor networks.\n",
+    "Neural computation, 7(6):1276-1314, 2005.\""
```
jgosmann: Punctuation outside of quotation marks?

docs/examples/learning/learn_associations.ipynb
```diff
@@ -118,11 +157,19 @@
...
+    "the first half learns the set of associations,\n",
+    "and the second tests recall.\n",
+    "\n",
+    "The learning signal will be set to 0 \n",
+    "o allow learning during the first phase,\n",
+    "and -1 to inhibit learning during the second phase.\n",
+    "\n",
+    "The memory is confined to a single ensemble.\n",
+    "Roughly speaking, its encoders will hold the keys,\n",
+    "and its decoders will hold the values. "
```
jgosmann: Trailing whitespace

docs/examples/spa/associative_memory.ipynb
```diff
@@ -427,7 +524,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "def input_fun(t):\n",
+    "def input_start_fun(t):\n",
```
jgosmann: It's always fun when the input starts. :grinning: :wink:

.ci/static.sh
```diff
@@ -12,9 +15,24 @@ function usage {
...
+        --log-level WARN \
+        --to python \
+        --TemplateExporter.exclude_input_prompt=True \
+        -- **/*.ipynb
+    # Remove style issues introduced in the conversion:
+    #   s/# $/#/g replaces lines with just '# ' with '#'
+    #   /get_ipython()/d removes lines containing 'get_ipython()'
+    sed -i -e 's/# $/#/g' -e '/get_ipython()/d' -- docs/**/*.py
+    flake8 nengo || STATUS=1
+    flake8 --ignore=E703,W291,W391 docs || STATUS=1
```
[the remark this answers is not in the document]
hunse: I've made a commit that makes all the notebooks compliant. I've also updated `devscripts/clearoutputs.py` to remove trailing whitespace, so it should be relatively easy for developers to clean new notebooks. I haven't actually enabled the check yet, though, since I want to check with @tbekolay that there's not another reason not to do this.

.ci/static.sh
```diff
@@ -12,9 +15,24 @@ function usage {
...
+    pylint docs nengo || STATUS=1
+    rm docs/examples/**/*.py
+    codespell -q 3 --skip="./build,./docs/_build,*-checkpoint.ipynb"|| STATUS=1
```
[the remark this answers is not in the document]
hunse: It makes it slightly too long (80 chars). This gets fixed in #1488, so ignoring for now here.

docs/examples/dynamics/controlled_integrator2.ipynb
hunse: English often puts them inside, even when it doesn't make sense IMO. I don't think it's worth changing.

.ci/static.sh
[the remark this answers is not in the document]
tbekolay: Yeah the only reason to keep it was because of the comment above, that lines with `# ` instead of `#` are inserted, but I fixed that in the `sed` statement, so I'll reenable this check.

hunse:
LGTM. I squashed all the fixups, so the tests are re-running, but I did do a diff to ensure the squashed branch is identical to the unsquashed one that passed all the tests, so they should all pass again (not sure in situations like this if we care about re-running them or not).

PR #1443, opened 2018-05-24
Fix warning for deprecated pytest feature

tbekolay:
**Motivation and context:**
In a recently released version of pytest, we got a bunch of warnings for a [feature that will be removed in pytest 4](https://docs.pytest.org/en/latest/mark.html#updating-code). This PR updates our code to not use the deprecated feature.

**How has this been tested?**
Ran the tests locally, no longer got any warnings.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,7 +11,8 @@ install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   - conda info -a
-  - conda install numpy matplotlib jupyter pytest
+  - conda install numpy matplotlib jupyter
+  - pip install pytest>=3.6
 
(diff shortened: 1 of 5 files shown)


jgosmann:
I wonder whether we should rather use `iter_markers`. Multiple `parametrize` markers can be combined or can't they?

jgosmann:
Also, `get_closest_marker` does not exist in the Windows environments ... maybe they use an older version of pytest?

tbekolay:
Good call @jgosmann, it is indeed possible to chain multiple parametrizations, and this happens to be convenient when you want to test all combinations of some parameter values. See the recent squash commit for a test of this also.

Issue #1441, opened 2018-05-24
nengo.utils.testing.Mock objects iterate forever


tbekolay:
https://github.com/nengo/nengo-extras/pull/72 contained a tests which did this:

```python
fig = plt.Figure()
for ax in fig.axes:
   ...
```

This ended up iterating forever when not generating plots because the `nengo.utils.testing.Mock` object that is returned by `plt.Figure()` and `fig.axes` does not handle iteration, so I believe the default behavior ends up continuing to make `Mock` objects.

Hopefully we can fix this by implementing an `__iter__` method that raises `StopIteration` quickly.

PR #1444, opened 2018-05-25
Fix iterating over Mock objects

tbekolay:
**Motivation and context:**
As reported in #1441, previously if you tried to iterate over an object that was mocked out (e.g., something in `plt`) it would loop forever. This fixes it by returning a iterator to an empty list, so no iteration will occur.

Since this is a change only to testing infrastructure, I think it doesn't need a changelog entry as it's entirely an under-the-hood change.

**How has this been tested?**
Added a test, which will fail without the change to `Mock`.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/utils/tests/test_testing.py
+++ b/nengo/utils/tests/test_testing.py
@@ -72,3 +72,10 @@ def test_logger_norecord():
         logger.info("Testing that logger doesn't record")
     with pytest.raises(ValueError):
         logger_obj.get_filepath(ext='txt')
+
+
+def test_mock_iter(plt):
+    fig = plt.figure()
+    for i, ax in enumerate(fig.axes):
+        assert False, "Mock object iterating forever"
+    plt.saveas = None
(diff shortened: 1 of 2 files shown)


nengo/utils/tests/test_testing.py
```diff
@@ -72,3 +72,11 @@ def test_logger_norecord():
         logger.info("Testing that logger doesn't record")
     with pytest.raises(ValueError):
         logger_obj.get_filepath(ext='txt')
+
+
+def test_mock_iter(plt):
+    fig = plt.figure()
+    for i, ax in enumerate(fig.axes):
+        if i > 10:
```
jgosmann: This line seems somewhat arbitrary and isn't really required because this loop should not even execute once.

tbekolay: Feel free to push a fixup

tbekolay:
@jgosmann Can you give this a quick review? Might as well get it in the release.

PR #1449, opened 2018-06-12
Deactivate progress bar in cache benchmark.

### LINKED ISSUE #1447: Decoder cache benchmarks results regression
I'm not sure at which point this started, but I was going through the plots for a release and noticed that the cache benchmarks are showing that the decoder cache does not improve build times:

d

m

n

These plots definitely used to look different, in that they showed that the cache lowered build times significantly. I get similar (but not identical) results in both my Python 2 and Python 3 environments, so I am inclined to think that this is not due to some quirk of my environment, though it's possible.

jgosmann:
**Motivation and context:**
Using a progress bar starts a thread that sleeps for 0.1s for every
update. Joining the thread after completing the build thus takes up
to 0.1s too which swamps any build performance improvements for the
given test cases. By deactivating the progress bar we get the raw
build performance. Outside the tests added 0.1s are not noticeable
to the user and thus do not need to be changed.

Fixes #1447.

none

**How has this been tested?**
Looked at the plots of the decoder cache benchmark tests. Run them with `pytest --slow --plots -- nengo/tests/test_cache.py`.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/tests/test_cache.py
+++ b/nengo/tests/test_cache.py
@@ -382,7 +382,10 @@ with model:
     '''
 
     without_cache = {
-        'rc': 'rc.set("decoder_cache", "enabled", "False")',
+        'rc': '''
+rc.set("progress", "progress_bar", "none")
+rc.set("decoder_cache", "enabled", "False")
+''',
         'stmt': '''
 with nengo.Simulator(model):
     pass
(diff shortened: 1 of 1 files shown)


jgosmann:
Travis CI tests fail because of a URL that couldn't be found when building the docs and seems unrelated to this PR.

tbekolay:
> Travis CI tests fail because of a URL that couldn't be found when building the docs and seems unrelated to this PR.

That's fixed in #1448 if you want to give a quick look over!

PR #1453, opened 2018-06-29
Catch node output validation error edge cases

### LINKED ISSUE #1452: Nengo node unhelpful size error message
When you run the code:
```
import nengo

class TestNode(nengo.Node):
    def __init__(self):
        super().__init__(
            self.update,
            size_in=0,
            size_out=0)

    def update(self, t, x):
        return t / x

    nd = TestNode()

    sim.run(sim.dt)
```

I get the following confusing error:
```
  File "nd_size.py", line 17, in <module>
    sim.run(sim.dt)
  File "/home/saubin/git/nengo/nengo/simulator.py", line 321, in run
    self.run_steps(steps, progress_bar=progress_bar)
  File "/home/saubin/git/nengo/nengo/simulator.py", line 345, in run_steps
    self.step()
  File "/home/saubin/git/nengo/nengo/simulator.py", line 356, in step
    step_fn()
  File "/home/saubin/git/nengo/nengo/builder/operator.py", line 766, in step_simpyfunc
    y = fn(t.item(), *args) if t is not None else fn(*args)
TypeError: update() missing 1 required positional argument: 'x'
```

The problem does not occur if:

1. The `super` call is replaced with `super().__init__(self.update, size_in=0)
2. If the network is not simulated

Any ideas on how we could give a better error?

Seanny123:
**Motivation and context:**
Resolves #1452 

**How has this been tested?**
Added new tests!

**Types of changes:**

**Checklist:**


Seanny123:
Given this PR currently causes the error `.\nengo\node.py:34:5: C901 'OutputParam.coerce' is too complex (15)` what should I do to resolve it? Should I just use a `#noqa` comment or should I break the function into separate functions?

jgosmann:
I vote breaking the function up; it's getting quite long.

Seanny123:
This breaks for the case of `nengo.Node(1)` and `nengo.Node([1])`. Will fix later.

Seanny123:
When I made this code backwards compatible with Python 2, I broke the original test case, because `inspect.signature` knows how to identify whether the first argument in a function is `self`. However, `inspect.getfullargspec` does not. [Fixing this is non-trivial](https://stackoverflow.com/q/51178199/1079075), so I'm going to abandon this PR until I'm suddenly excited by the internals of Python or until Nengo abandons Python 2 (which I know is only going to happen in 2020).

Seanny123:
Thanks to the kind folks at StackOverflow, I figured out the problem and this is now ready for review. However, I'm afraid I should have broken up the code changes better. Suggestions for improving the commit organisation before review are welcome.

Seanny123:
Good catch on the other corner cases. I don't know why I missed the warning being excessive initially, but I agree with it's removal. Ping me once you fix the Python 2.7 error and I'll approve this PR.

Seanny123:
LGTM.

nengo/node.py
```diff
@@ -7,10 +8,21 @@
 from nengo.exceptions import ValidationError
 from nengo.params import Default, IntParam, Parameter
 from nengo.processes import Process
-from nengo.utils.compat import is_array_like
+from nengo.utils.compat import is_array_like, getfullargspec
 from nengo.utils.stdlib import checked_call
 
 
+class OutputFnArgsValidationError(ValidationError):
```
hunse: I'm all for OOP, but can we just have a function that does this? From what I can tell, it's just a helper for making the `msg`.

hunse: Otherwise it'll show up as `OutputFnArgsValidationError` in error traces, which is a little weird to me.

CHANGES.rst
```diff
@@ -84,6 +84,13 @@ Release History
   values preserved, causing issues with some node functions.
   (`#1474 <https://github.com/nengo/nengo/pull/1474>`__)
 
+**Fixed**
```
tbekolay: Should be only one Fixed seciton

nengo/node.py
```diff
@@ -7,7 +8,7 @@
 from nengo.exceptions import ValidationError
 from nengo.params import Default, IntParam, Parameter
 from nengo.processes import Process
-from nengo.utils.compat import is_array_like
+from nengo.utils.compat import is_array_like, getfullargspec
```
tbekolay: Should be sorted alphabetically

nengo/node.py
```diff
@@ -1,4 +1,5 @@
 import warnings
+import inspect
```
tbekolay: Should come before warnings, alphabetical

PR #1462, opened 2018-07-20
Activate pylint invalid-name warnings.

jgosmann:
**Motivation and context:**
Ensures some consistency. Regexes have been relaxed from the defaults to match current naming conventions in the code base.

based on #1278, related to other pylint PRs.

**How has this been tested?**

ran pylint

**Types of changes:**

**Checklist:**

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -17,12 +17,24 @@ enable = anomalous-backslash-in-string,
          cyclic-import,
          dangerous-default-value,
          function-redefined,
+         invalid-name,
          line-too-long,
          super-init-not-called,
          ungrouped-imports,
          unnecessary-lambda,
          unused-import,
          wrong-import-order,
+
+
+[BASIC]
+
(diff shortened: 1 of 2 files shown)


tbekolay:
I think I'm going to close this one for a few reasons:

1. I think the `invalid-name` warning would be useful as there are definitely some sections of code that I think could use better variables names. But matching the current conventions basically means we allow anything that isn't wildly out there, which we would likely catch on review without these warnings.
2. I find regexes harm readability wherever they are; in this case, it makes the `.pylintrc` file hard to read. What are the naming conventions being enforced here? We could add comments to explain each one, but then the comment and regex could become desynced. Using a [verbose regex](https://docs.python.org/3/library/re.html#re.X) works in a Python file, but I don't think we can do that in `.pylintrc` so I would only want to use non-default values for these regexes if there was a big benefit.
3. The only names actually changed here are ones in a test file that I think (not 100% sure) we copied from an open source project, so aren't really ones that we have to maintain anyhow.

jgosmann:
Fine with me.

PR #1465, opened 2018-08-30
Deprecate spa

jgosmann:
**Motivation and context:**

I think with the release of Nengo SPA 0.6 (which will happen soon) the project is mature enough to allow the deprecation of the `spa` module in core Nengo. [Nengo SPA is much more powerful](https://www.nengo.ai/nengo-spa/examples/intro_coming_from_legacy_spa.html#Why-switch-to-nengo_spa?) (and that list doesn't even mention the support for other binding operations like vector-derived transformation binding that will be introduced in 0.6).

A deprecation note is added to the documentation and a deprecation warning will be issued when using the `nengo.spa.SPA` class (which is a requirement to use most of `nengo.spa`'s functionality).

none

**How has this been tested?**

existing tests

**Types of changes:**

- Breaking change (fix or feature that causes existing functionality to change)??

**Checklist:**

**Still to do:**

- [ ] Adjust the version number in the deprecation notice in the documentation, depending on when this PR will be merged and released.


jgosmann:
I prefer to skip those tests by default instead of completely removing them.
Another question: should we exclude the spa module from the coverage report then?

tbekolay:
> I prefer to skip those tests by default instead of completely removing them.

Yeah that's true, why don't we mark them all as slow so that we still run them on each release.

> Another question: should we exclude the spa module from the coverage report then?

Sure, makes sense.

drasmuss:
Looks like just one line in `networks.circularconvolution` https://codecov.io/gh/nengo/nengo/compare/master...deprecate-spa/changes, so that seems alright.

drasmuss:
Went ahead and removed the spa examples while it was on my mind.  Note: it looks like this removes reference to SPA, but I'm anticipating that we'll soon be updating nengo docs to the new nengo-sphinx-theme, which uses the overall theme to link projects together (so we don't need to worry about cross-referencing different projects from within other projects' documentation).

jgosmann:
I agree with removing the examples, but I added a fixup commit to keep the reference to the new SPA project in case someone is still looking for SPA at that place. I'd like to keep that notice for at least one release.

PR #1466, opened 2018-08-31
Test deselection

### LINKED ISSUE #1082: Report total number of runable tests

jgosmann:
**Motivation and context:**

This shows them in the deselected count, but keeps them separate from
test expected to fail or being skipped. The slow tests are still being
skipped because to me these still belong into the tests of selected
tests that should run for a full pass of the test suite whereas the
other categories are more special cases (for example tests that only
plot things, but do not assert anything).

This commit also adds messages detailing what has been selected right
after the collection.

Supersedes #1350, closes #1082.

This does not count how many tests have been deselected in each category. To me this seems to be not worth the effort (because those number have to be stored awkwardly when the deselection is done to be able to print them when the report created).

I added a commit inverting the logic of `noexamples` (i.e. a value `True` meaning the no examples should be run) that makes more sense to me and I was initially confused by the existing logic. But that commit could be dropped from the PR.

Supersedes #1350

**How has this been tested?**

manually trying out different flags to pytest

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -290,7 +290,7 @@ def pytest_generate_tests(metafunc):
 
 
 def pytest_collection_modifyitems(session, config, items):
-    if not config.getvalue('noexamples'):
+    if config.getvalue('noexamples'):
         deselect_by_condition(
             lambda item: getattr(item.obj, 'example', None), items, config)
     if not config.getvalue('slow'):
@@ -346,7 +346,7 @@ def deselect_by_condition(condition, items, config):
 def pytest_report_collectionfinish(config, startdir, items):
     deselect_reasons = ["Nengo core tests collected"]
 
-    if not config.getvalue('noexamples'):
+    if config.getvalue('noexamples'):
         deselect_reasons.append(
             " example tests deselected (--noexamples passed)")
     if not config.getvalue('slow'):
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -289,6 +289,95 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize("nl_nodirect", nodirect)
 
 
+def pytest_collection_modifyitems(session, config, items):
+    if not config.getvalue('noexamples'):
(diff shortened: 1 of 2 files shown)


jgosmann:
I'm getting a bunch of warnings like these
```
nengo/utils/tests/test_ensemble.py::test_tuning_curves[SpikingRectifiedLinear-2]
  /Volumes/Home/blubb/Documents/projects/uni/nengo/nengo/conftest.py:140: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
  Please use node.get_closest_marker(name) or node.iter_markers(name).
  Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
    argnames = request.keywords['parametrize'].args[::2]
```
But I feel like these were already addressed somewhere?

tbekolay:
> But I feel like these were already addressed somewhere?

Yes, #1443

tbekolay:
From reading through the diff this is looking great; I'll test it out in more detail tomorrow but most likely it'll be good to go.

PR #1467, opened 2018-09-01
Fix warnings raised in test suite

tbekolay:
**Motivation and context:**
#1443 fixed most of the warnings recently raised in the test suite, but there were a few more related to recent NumPy and SciPy changes.

**How has this been tested?**
Ran the test suite locally and got 0 warnings.

**Types of changes:**

**Checklist:**

**Still to do:**
- [x] Is the `atol=0` value correct in the `scipy.sparse.linalg.cg` call?

### DIFF
--- a/nengo/utils/least_squares_solvers.py
+++ b/nengo/utils/least_squares_solvers.py
@@ -107,7 +107,7 @@ class ConjgradScipy(LeastSquaresSolver):
                 itns[i] += 1
 
             X[:, i], infos[i] = scipy.sparse.linalg.cg(
-                G, B[:, i], tol=self.tol, callback=callback)
+                G, B[:, i], tol=self.tol, callback=callback, atol=0)
 
         info = {'rmses': rmses(A, X, Y), 'iterations': itns, 'info': infos}
         return X if matrix_in else X.ravel(), info
--- a/nengo/utils/least_squares_solvers.py
+++ b/nengo/utils/least_squares_solvers.py
@@ -80,13 +80,28 @@ class Cholesky(LeastSquaresSolver):
 
 
 class ConjgradScipy(LeastSquaresSolver):
-    """Solve a least-squares system using Scipy's conjugate gradient."""
+    """Solve a least-squares system using Scipy's conjugate gradient.
+
+    Parameters
+    ----------
+    tol : float
+        Relative tolerance of the CG solver (see [1]_ for details).
+    atol : float
+        Absolute tolerance of the CG solver (see [1]_ for details).
+
+    References
+    ----------
+    .. [1] scipy.sparse.linalg.cg documentation, https://docs.scipy.org/
+        doc/scipy/reference/generated/scipy.sparse.linalg.cg.html
+    """
 
     tol = NumberParam('tol', low=0)
+    atol = NumberParam('atol', low=0)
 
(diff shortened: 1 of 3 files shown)


hunse:
So if we wanted to be the most like old Scipy, we'd set `atol = tol`, since this appears to be [what 0.18 did](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.sparse.linalg.cg.html). However, I think new Scipy split apart `tol` and `atol` for a reason, so I think we should do the same. I've added a commit that allows the user to set `atol` if they wish. I also set the default to non-zero, since I think we want it to come into play to help make sure the solver doesn't iterate forever in weird cases. I say weird because usually I think the relative tolerance `tol` is actually what's stopping the iterations. In support of this point, I played around with `atol` in the context of our small test (`test_solvers.test_scipy_solvers`) and changing `atol` to even as high as `tol` doesn't affect the number of iterations (though that system is very well regularized).

I also added a try-except so that we can still work with old Scipy (which doesn't have `atol`).

nengo/utils/least_squares_solvers.py
```diff
@@ -106,8 +121,16 @@ def __call__(self, A, Y, sigma, rng=None):
...
+            try:
+                X[:, i], infos[i] = scipy.sparse.linalg.cg(
+                    G, B[:, i], tol=self.tol, callback=callback,
+                    atol=self.atol)
+            except TypeError as e:
+                # no atol parameter in Scipy < 1.0
+                if 'atol' not in str(e):
+                    raise e
+                X[:, i], infos[i] = scipy.sparse.linalg.cg(
+                    G, B[:, i], tol=self.tol, callback=callback)
```
arvoelke: Should `.travis.yml` be modified to test both scipy < 1.0 and the current version? Since here we're explicitly supporting both versions, it would be good to make sure we don't break that in the future.

hunse: Testing everything with scipy < 1.0 seems like overkill just for this one case. I guess there should be a way to run just this test, so that might be worth it (not sure how much time just spooling up a new VM takes, though).

Frankly, I'm not too worried. I doubt that this would break, since I think it's pretty clear in the code what it's doing, so nobody should change it. But if it did break for scipy < 1.0, that's not a huge deal IMO. Most users should be able to upgrade their scipy pretty easily.

arvoelke: My thinking / intent is the check wouldn't be in place just for this code. The check would be for any other code that exists now, or in the future, that might differ across the two versions. If something else breaks across versions, but we ask that people upgrade scipy, then the above code ends up being unnecessary / unused. On the other hand, if we are trying to support scipy < 1.0, then we won't know for sure when new code is added if it's not tested. So I'm mainly looking for the the formal stance about what versions we're looking to support, and positing that for long-term consistency it makes sense to test both until support is dropped. Like you said, it should be easy enough to upgrade scipy if someone wants to use this solver, so we could just drop support now?

arvoelke: Might be worth mentioning that the `atol` is present in 1.1.0, but [_not_ 1.0.0](https://docs.scipy.org/doc/scipy-1.0.0/reference/generated/scipy.sparse.linalg.cg.html), so it's actually not a version 1.0.0 difference.

PR #1468, opened 2018-09-06
Streamlined neuron argreprs and added amplitude

hunse:
- Amplitude is now printed in the argreprs (`str` representation)
  for LIF and RectifiedLinear.
- Added an `_add_argrepr` helper function to make argreprs more
  concise for all neuron types (had been done only for Izhikevich
  before).

**Motivation and context:**

When printing the string representation for `LIF`, `LIFRate`, `RectifiedLinear`, and `SpikingRectifiedLinear`, the amplitude parameter was not included.

**How has this been tested?**

I encountered this when making plots with these neuron types in nengo_loihi, so I tested it there.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -24,13 +24,6 @@ class NeuronType(FrozenObject):
 
     probeable = ()
 
-    def __repr__(self):
-        return "%s(%s)" % (type(self).__name__, ", ".join(self._argreprs))
-
-    @property
-    def _argreprs(self):
-        return []
-
     def current(self, x, gain, bias):
         """Compute current injected in each neuron given input, gain and bias.
 
@@ -220,6 +213,7 @@ class Direct(NeuronType):
         """
         raise SimulationError("Direct mode neurons shouldn't be simulated.")
 
+
 # TODO: class BasisFunctions or Population or Express;
 #       uses non-neural basis functions to emulate neuron saturation,
 #       but still simulate very fast
@@ -240,6 +234,8 @@ class RectifiedLinear(NeuronType):
     """
 
     probeable = ('rates',)
+
+    amplitude = NumberParam('amplitude', low=0, low_open=True)
 
     def __init__(self, amplitude=1):
         super(RectifiedLinear, self).__init__()
@@ -315,10 +311,6 @@ class Sigmoid(NeuronType):
         super(Sigmoid, self).__init__()
         self.tau_ref = tau_ref
 
-    @property
-    def _argreprs(self):
-        return [] if self.tau_ref == 0.0025 else ["tau_ref=%s" % self.tau_ref]
-
     def gain_bias(self, max_rates, intercepts):
         """Analytically determine gain, bias."""
         max_rates = np.array(max_rates, dtype=float, copy=False, ndmin=1)
@@ -369,15 +361,6 @@ class LIFRate(NeuronType):
         self.tau_rc = tau_rc
         self.tau_ref = tau_ref
         self.amplitude = amplitude
-
-    @property
-    def _argreprs(self):
-        args = []
-        if self.tau_rc != 0.02:
-            args.append("tau_rc=%s" % self.tau_rc)
-        if self.tau_ref != 0.002:
-            args.append("tau_ref=%s" % self.tau_ref)
-        return args
 
     def gain_bias(self, max_rates, intercepts):
         """Analytically determine gain, bias."""
@@ -506,6 +489,9 @@ class AdaptiveLIFRate(LIFRate):
     tau_ref : float
         Absolute refractory period, in seconds. This is how long the
         membrane voltage is held at zero after a spike.
+    amplitude : float
+        Scaling factor on the neuron output. Corresponds to the relative
+        amplitude of the output spikes of the neuron.
 
     References
     ----------
@@ -523,15 +509,6 @@ class AdaptiveLIFRate(LIFRate):
         super(AdaptiveLIFRate, self).__init__(**lif_args)
         self.tau_n = tau_n
         self.inc_n = inc_n
-
-    @property
-    def _argreprs(self):
-        args = super(AdaptiveLIFRate, self)._argreprs
-        if self.tau_n != 1:
-            args.append("tau_n=%s" % self.tau_n)
-        if self.inc_n != 0.01:
-            args.append("inc_n=%s" % self.inc_n)
-        return args
 
     def step_math(self, dt, J, output, adaptation):
         """Implement the AdaptiveLIFRate nonlinearity."""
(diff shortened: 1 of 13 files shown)


tbekolay:
> By the way, what this the current convention / work flow for reviews? Am I to make certain changes myself?

Currently there's a fair bit of onus on the reviewer to make changes if they're not too big, but if they're big changes you can put it back on the PR author. It's resulted in a lack of reviewers though, so that will probably change in the near future.

tbekolay:
Current guidelines are at https://www.nengo.ai/reviewing.html and if we change the process we'll change that document.

hunse:
The biggest thing for me is to make it clear who is going to make the changes. If you feel comfortable doing them yourself, then assign the issue to yourself and go ahead. Otherwise, assign it to me, so that I know the ball's in my court.

arvoelke:
I've unit-tested the code now. In the process, I discovered the `repr()` for `nengo.LIF()` and `nengo.AdaptiveLIF()` were both incorrect, as was the order for `nengo.AdaptiveLIF()` and `nengo.AdaptiveLIFRate()`. Also both `nengo.AdaptiveLIF()` and `nengo.AdaptiveLIFRate()` were missing parameters in their docstrings. I've hopefully fixed it all.

arvoelke:
This looks good to me now, but someone should check my revisions too. I also recommend that the rest of the `docs/examples/usage/strings.ipynb` notebook be converted into unit tests and then removed.

nengo/neurons.py
```diff
@@ -463,6 +463,15 @@ def __init__(self, tau_rc=0.02, tau_ref=0.002, min_voltage=0, amplitude=1):
             tau_rc=tau_rc, tau_ref=tau_ref, amplitude=amplitude)
         self.min_voltage = min_voltage
 
+    @property
+    def _argreprs(self):
+        args = []
+        self._add_argrepr(args, 'tau_rc', 0.02)
+        self._add_argrepr(args, 'tau_ref', 0.002)
+        self._add_argrepr(args, 'min_voltage', 0)
```
hunse: These three lines are copied from the parent `LIFRate`, we should use their method instead.

nengo/neurons.py
```diff
@@ -545,9 +545,10 @@ def __init__(self, tau_n=1, inc_n=0.01, **lif_args):
 
     @property
     def _argreprs(self):
-        args = super(AdaptiveLIFRate, self)._argreprs
+        args = []
         self._add_argrepr(args, 'tau_n', 1)
         self._add_argrepr(args, 'inc_n', 0.01)
+        args.extend(super(AdaptiveLIFRate, self)._argreprs)
```
hunse: I don't understand why switching this order makes it consistent. What was inconsistent about it before? Other neuron types (e.g. LIF) have the base class parameters first, then the unique parameters.

nengo/tests/test_neurons.py
```diff
@@ -476,6 +476,95 @@ def test_frozen():
...
+    assert (repr(nengo.Izhikevich(coupling=0.3)) ==
+            "Izhikevich(coupling=0.3)")
+    assert (repr(nengo.Izhikevich(reset_voltage=-1)) ==
+            "Izhikevich(reset_voltage=-1)")
+    assert (repr(nengo.Izhikevich(reset_recovery=5)) ==
+            "Izhikevich(reset_recovery=5)")
+    assert (repr(nengo.Izhikevich(tau_recovery=0.1, coupling=0.3,
+                                  reset_voltage=-1, reset_recovery=5)) ==
+            "Izhikevich(tau_recovery=0.1, coupling=0.3, "
+            "reset_voltage=-1, reset_recovery=5)")
```
hunse: So the one tricky thing about this test is that it doesn't check for new unrepresented parameters, which is really what causes these issues in the first place. That said, the `strings` notebook would only help with this if people think to run the strings notebook. I think this test is definitely a good addition, but we might want to look at a more programmatic/introspective way of checking the attributes of our objects to make sure all `Parameters` get documented in the `repr` string somewhere (I'm happy to leave that as future work, though).

nengo/neurons.py
arvoelke: Everything is now in the same order as in the constructor. Can you give an example of a unit test that you think should give different results?

nengo/neurons.py
arvoelke: I had thought about this, but the third line's actually different, and it's lines [1, 2, 4] that are copied. That is, `min_voltage` comes in between `tau_ref` and `amplitude`. I wanted to keep the ordering consistent with their respective constructors.

nengo/neurons.py
hunse: Ahh, I just didn't understand what you were using as your metric for "consistent". Having them in the same order as the constructor makes sense.

hunse:
I've created a new PR that makes this test programmatic.

This LGTM.

hunse:
I'd like to bump this issue. It's causing my plots not to be named correctly when I use `pytest.mark.parametrize` with a neuron type with amplitude. I think it's ready for final review/merge now.

nengo/neurons.py
```diff
@@ -241,11 +245,19 @@ class RectifiedLinear(NeuronType):
 
     probeable = ('rates',)
 
+    amplitude = NumberParam('amplitude', low=0, low_open=True)
```
drasmuss: I would allow `amplitude=0`. There might be some legitimate, although rare, use cases (e.g., I want to disable the output of an ensemble for testing purposes).

nengo/neurons.py
```diff
@@ -241,11 +245,19 @@ class RectifiedLinear(NeuronType):
...
+
     def __init__(self, amplitude=1):
         super(RectifiedLinear, self).__init__()
 
         self.amplitude = amplitude
 
+    @property
+    def _argreprs(self):
+        args = []
+        self._add_argrepr(args, 'amplitude', 1)
```
drasmuss: Could we automate this (like we do in e.g. https://github.com/nengo/nengo/blob/master/nengo/params.py#L495), rather than having to remember to update this list whenever we add a parameter or change a default?

arvoelke: The tests that @hunse added in #1472 will notice if you forget to update this list after adding a parameter / changing a default. So I don't think forgetting to do this will be a problem.

nengo/neurons.py
arvoelke: Note that this addition was made to be consistent with `RectifiedLinear`.

arvoelke: More importantly, though, I believe (?) if you try this you will get a decoder solver error due to an all-zero activity matrix.

drasmuss: Ah yeah that's true.  Would still be nice to support, but not so nice that it's worth fiddling with all that 😄

hunse: `LIFRate` also has an amplitude parameter, so I just set one up to be the same.

hunse:
I made some comments on 3fcabae, but it got replaced by 6109245, so for reference that's where my comments are. It looks like they're all done in 6109245, so I think I'm good with this. Automatically doing argreprs passes the test introduced here and the new test is pretty robust I think, so I'm satisfied.

At some point, we might want to consider using the programmatic argreprs for things like Distributions and Synapses, but that's definitely future work (they're still manually making their own `__repr__`s at this point).

PR #1469, opened 2018-09-07
Fix the string representation of RectifiedLinear

tcstewar:
**Motivation and context:**

This just fixes the string representation of RectifiedLinear and SpikingRectifiedLinear neurons.  Without this fix, we got this behaviour:

```python
print(nengo.SpikingRectifiedLinear(amplitude=0.5))
>>> SpikingRectifiedLinear()
```

**How has this been tested?**

Just by doing the above example in Jupyter notebook

**Types of changes:**

**Checklist:**

This feels like too small a fix to add a test, and I'm not sure it even warrants a changelog entry.  If I'm wrong about that, let me know!

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -241,10 +241,19 @@ class RectifiedLinear(NeuronType):
 
     probeable = ('rates',)
 
+    amplitude = NumberParam('amplitude', low=0, low_open=True)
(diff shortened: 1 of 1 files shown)


drasmuss:
I think Eric did this in #1468?

PR #1472, opened 2018-09-14
Programmatically check `repr`

hunse:
**Motivation and context:**

This automates the testing of our `__repr__` functions for FrozenObjects, so that e.g. if new parameters are added we don't forget to add them to `__repr__`.

Based on #1468.

**How has this been tested?**

By printing the `repr` for each object in the tests, to ensure that they look as expected, and that we're checking all different combinations.

**Types of changes:**

- Testing improvements

**Checklist:**

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -538,19 +538,27 @@ class AdaptiveLIFRate(LIFRate):
     tau_n = NumberParam('tau_n', low=0, low_open=True)
     inc_n = NumberParam('inc_n', low=0)
 
-    def __init__(self, tau_n=1, inc_n=0.01, **lif_args):
-        super(AdaptiveLIFRate, self).__init__(**lif_args)
+    def __init__(
+        self,
+        tau_n=1,
+        inc_n=0.01,
+        tau_rc=0.02,
+        tau_ref=0.002,
+        amplitude=1,
+    ):
+        super(AdaptiveLIFRate, self).__init__(
+            tau_rc=tau_rc, tau_ref=tau_ref, amplitude=amplitude)
         self.tau_n = tau_n
         self.inc_n = inc_n
 
     def step_math(self, dt, J, output, adaptation):
         """Implement the AdaptiveLIFRate nonlinearity."""
         n = adaptation
-        LIFRate.step_math(self, dt, J - n, output)
+        super(AdaptiveLIFRate, self).step_math(dt, J - n, output)
         n += (dt / self.tau_n) * (self.inc_n * output - n)
 
 
-class AdaptiveLIF(AdaptiveLIFRate, LIF):
+class AdaptiveLIF(LIF):
     """Adaptive spiking version of the LIF neuron model.
 
     Works as the LIF model, except with adapation state ``n``, which is
@@ -590,10 +598,31 @@ class AdaptiveLIF(AdaptiveLIFRate, LIF):
 
     probeable = ('spikes', 'adaptation', 'voltage', 'refractory_time')
 
+    tau_n = NumberParam('tau_n', low=0, low_open=True)
+    inc_n = NumberParam('inc_n', low=0)
+
+    def __init__(
+        self,
+        tau_n=1,
+        inc_n=0.01,
+        tau_rc=0.02,
+        tau_ref=0.002,
+        min_voltage=0,
+        amplitude=1,
+    ):
+        super(AdaptiveLIF, self).__init__(
+            tau_rc=tau_rc,
+            tau_ref=tau_ref,
+            min_voltage=min_voltage,
+            amplitude=amplitude,
+        )
+        self.tau_n = tau_n
+        self.inc_n = inc_n
+
     def step_math(self, dt, J, output, voltage, ref, adaptation):
         """Implement the AdaptiveLIF nonlinearity."""
         n = adaptation
-        LIF.step_math(self, dt, J - n, output, voltage, ref)
+        super(AdaptiveLIF, self).step_math(dt, J - n, output, voltage, ref)
         n += (dt / self.tau_n) * (self.inc_n * output - n)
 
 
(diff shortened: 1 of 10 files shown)


tbekolay:
I have some concerns with the complexity-benefit tradeoff for this PR. We add seven functions (some of which are IMO pretty complex) to test files in order to ensure that `__repr__` methods are correct. Unlike other types of repr testing, this will find unanticipated issues, so for example if a new parameter is added this will find if it has not been added to the repr. The PR fixes several reprs that had missing parameters.

The seven functions appear to do two main things.

1. Generate instances of `FrozenObject`s with random parameters. Since we use `Parameter` subclasses for all these arguments, it's relatively easy to do, except for the [cases in which it's not easy](https://github.com/nengo/nengo/pull/1472/files#diff-43b918f6535d46b6b7c43a44fcfbb62bR40).
2. Check the reprs of those `FrozenObject`s by generating what the repr should look like and testing against that.

My main concern with the complexity-benefit tradeoff here is that the added complexity is in the tests and not in the main codebase. If you look at [one of the tests that this PR replaces](https://github.com/nengo/nengo/pull/1472/files#diff-273f7dd17cf9940b99c916464def1fabL478), you see some dumb, obvious code that does exactly what it says it does. This is ideal for test cases, because if one of these conditions fail, the issue will be obvious, even if the fix itself is not. In the new tests, however, the [actual assertion](https://github.com/nengo/nengo/pull/1472/files#diff-43b918f6535d46b6b7c43a44fcfbb62bR101) is two levels deep in the call stack, and is basically meaningless without dropping into a debugger where the assertion fails. There is a fair bit of work involved figuring out what the actual issue is before we get to fixing that issue.

I have two ideas for how to move the complexity into the main codebase and keep the tests simple.

1. Use the code that is generating the candidate reprs to generate actual reprs in frozen objects.

   If we've gone through the effort to figure out how to check if a random object's repr is correct, then it seems to follow that we could instead use that same logic to generate a random object's repr. Currently we are doing this manually for individual objects (hence the changes to `__repr__` methods in this PR) but we could instead do this in `FrozenObject.__repr__`. Looking at `FrozenObject.__repr__` right now, we do seem to do this, but if we are having to override it for some objects then there might be issues with that implementation that we should figure out.

   In this case, we would not care so much about making instances of `FrozenObject`s with random parameters, and instead make some worst-case smoke tests of various object `reprs` (i.e., we wouldn't care about number 1 above).

2. Add the ability to generate instances of `FrozenObject`s with random parameters to the main codebase. Test reprs with a simpler approach (i.e., `eval(repr(obj)) == obj`).

   This option makes sense if the previous option overlooks some of the specifics of what's actually going on. Perhaps it is not actually the case that we're automatically generating correct `__repr__`s. In that case, the main complexity lies in randomly generating instances of `FrozenObject`s. The main question is whether moving that to the main codebase is useful. Most of my thoughts about use cases involve testing and providing practical limitations on parameters for users, but you could imagine that some people could use it for hyperparameter optimization (though only really for random searches).

   In this case, we would essentially be ignoring number 2 above by relying on something like `eval(repr(obj)) == obj`.

There are of course other options, like removing the complexity and manually writing the test cases that this PR found through the automated parts, but that would ignore any of the potential benefits. If anyone can see other good options, please post them here for consideration.

nengo/neurons.py
```diff
@@ -545,24 +545,20 @@ def __init__(
         tau_rc=0.02,
         tau_ref=0.002,
         amplitude=1,
-        **kwargs
     ):
         super(AdaptiveLIFRate, self).__init__(
-            tau_rc=tau_rc,
-            tau_ref=tau_ref,
-            amplitude=amplitude,
-            **kwargs)
```
hunse: The `LIF` constructor was getting called, and `min_voltage` was getting properly set. I tested that, and so did `test_neurons.test_argreprs`. I'm not quite sure how it was happening, but this `**kwargs` was needed for it.

Anyway, the new way seems cleaner, so I'm fine with it.

tbekolay: You're right, `super` looks through the MRO and call both `__init__` methods, so `AdaptiveLIF` was working correctly because `AdaptiveLIFRate` accepted and ignored `min_voltage` thanks to the `**kwargs` (this [SO thread](https://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance) helped clarify). I had removed the `**kwargs` before making the other changes so didn't realize I had broken it.

hunse:
Changes look good to me.

I'm not totally sold on the `eval` checking for reprs. This definitely wouldn't work in all cases; for example, if you've got a `Sample` distribution with a large array as its argument, the `repr` won't print the whole array (I'll print something like `array([1, 2, 3, ..., 9997, 9998, 9999])`, though you actually need a pretty large array to get it to do this). But I think for our test cases, it's fine.

Currently, Transforms make their own `_argreprs`. It would be nice if they didn't have to, but that doesn't seem trivial. I think we should make an issue, or we can look at that again when we make more Transforms.

tbekolay:
> I'm not totally sold on the `eval` checking for reprs

As I said in the commit message, there are definitely tradeoffs, and it might be a good idea for us to specifically check a few specific reprs in addition to doing the eval checking. Maybe not in this PR though in the interest of getting it done today.

For the specific case of large NumPy arrays, one thing we could do is set the threshold very high with [np.set_printoptions](https://docs.scipy.org/doc/numpy/reference/generated/numpy.set_printoptions.html#numpy.set_printoptions). However, that does change some global NumPy state, and the default threshold is set to 1000, which is pretty high. I think it's fine for now, but if this does come up, there is a resolution in the form of `np.set_printoptions` (or implementing a specific test for those cases; doing the eval testing doesn't mean you can't do specific repr checking in addition).

> I think we should make an issue, or we can look at that again when we make more Transforms.

Agreed, I made #1518 to track that.

PR #1474, opened 2018-09-18
Fix signal offset when reshaping

hunse:
**Motivation and context:**

When reshaping a signal, the offset was not preserved. This means that the view was not on the correct part of the signal.

This fixes nengo/nengo_ocl#160.

**How has this been tested?**

I've written a unit test, but it currently fails for unclear reasons. (Because somehow reshaping a numpy array with a base does not preserve the base?)

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] get unit test working


jgosmann:
[I think the problem is that `reshape` is not guaranteed to create a view, but might return a copy](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.reshape.html) (in which case the `Signal` reshape logic does not work).

hunse:
Ok, fixed this up. We now throw an error (just the standard Numpy error; do we want a Nengo error?) if someone tries to reshape a non-contiguous array that does not work with that shape. I've added to the signal reshape test to check this, as well as check that non-contiguous arrays with offsets that can be reshaped are done so properly (the original issue that this PR is addressing).

nengo/builder/signal.py
```diff
@@ -238,9 +238,15 @@ def reshape(self, *shape):
         Any number of integers can be passed to this method,
         describing the desired shape of the returned signal.
         """
-        return Signal(self._initial_value.reshape(*shape),
+        if len(shape) == 1:
+            shape = shape[0]  # in case a tuple is passed in
```
drasmuss: Why is it a problem if the shape is set to `(x,)` instead of `x`?

nengo/builder/tests/test_signal.py
```diff
@@ -123,6 +124,39 @@ def test_signal_reshape():
...
+    s0slice = slice(0, 3), slice(None, None, 2)
+    s0shape = 2, 3
+    s0 = s[s0slice].reshape(*s0shape)
+    assert s.offset == 0
+    assert np.array_equal(s0.initial_value, value[s0slice].reshape(*s0shape))
+
+    s1slice = slice(1, None), slice(None, None, 2)
+    s1shape = 2, 4
+    s1 = s[s1slice].reshape(s1shape)
+    assert s1.offset > 0
```
drasmuss: Don't we know exactly what the offset should be here?

nengo/tests/test_builder.py
```diff
@@ -99,158 +99,6 @@ def create(make_extra, seed):
         assert same1seeds[same1obj] == same2seeds[same2obj]
 
 
-def test_signal():
```
drasmuss: I swear I removed these in some other PR, but I don't remember where 🤷‍♂️ .  No harm in doing it twice though.

drasmuss:
> We now throw an error (just the standard Numpy error; do we want a Nengo error?)

I'd vote for a custom nengo error, so that we can explain more clearly what is going wrong.

nengo/builder/signal.py
hunse: It's not. The problem is because we're using the `*` for the arguments, if you call e.g. `reshape((1, 1))` you'll get `shape = ((1, 1),)` (wrapped inside an extra tuple), which is a problem.

nengo/builder/tests/test_signal.py
hunse: Yeah, you could figure it out. I just wrote it generally in case someone changes the test. The main thing I wanted to check is that we're testing one case with zero offset and one with non-zero offset, so that's why I put the asserts in. The point of this test is not to check the offset (there's a separate test for that), but there's also no reason we can't check the exact offset here.

hunse:
The Numpy error is actually not bad:
```
    def reshape(self, *shape):
        """Return a view on this signal with a different shape.
    
            Note that ``reshape`` cannot change the overall size of the signal.
            See `numpy.reshape` for more details.
    
            Any number of integers can be passed to this method,
            describing the desired shape of the returned signal.
            """
        if len(shape) == 1:
            shape = shape[0]  # in case a tuple is passed in
        initial_value = self.initial_value.view()
>       initial_value.shape = shape  # raises AttributeError if cannot
E       AttributeError: incompatible shape for a non-contiguous array
```
To get our own, we'd just have to trap this and then raise our own, which makes the code a bit messier. But you're right, we would be able to add a little bit more information. What exactly would we want to add that would help somebody debug this?

nengo/builder/signal.py
drasmuss: Ah, got ya.  We don't have to do this (it's not introduced by this PR), but I'd prefer just passing the shape as a tuple (i.e. `def reshape(self, shape)`), for consistency with `np.reshape`.  Not critical, but in case others felt the same way might as well do it now.

drasmuss:
Basically the unique aspect is that we require our reshape method to always return a view, not a copy.  To me that error message suggests that the problem is the shape, so the user should figure out what a compatible shape would be.  But in reality the problem is the copying.  Something like `raise BuildError("Reshaping <signal> would require the array to be copied (because it is not contiguous), which is not supported")`.

nengo/builder/signal.py
hunse: `np.reshape` works either way, which is why we did it like this.

hunse: Before I realized this simple trick would allow it to work both ways, I was considering dropping support for passing them separately. I'd be fine with that. This is a function that's mostly used internally, anyway, so dropping support would have little effect on user experience (though it's possible some backends are relying on it).

drasmuss: I'm fine with leaving it in if other people find it useful 🤷‍♂️

nengo/builder/signal.py
```diff
@@ -241,8 +241,14 @@ def reshape(self, *shape):
...
-        initial_value.shape = shape  # raises AttributeError if cannot
-                                     # reshape without copying.
+        try:
+            # this raises AttributeError if cannot reshape without copying
+            initial_value.shape = shape
+        except AttributeError:
+            raise SignalError(
+                "Reshaping %s to %s would require the array to be copied "
+                "(because it is not contiguous), which is not supported" % (
+                    self, shape))
```
hunse: This is slightly misleading, since some non-contiguous arrays can be reshaped, depending on the shape and the way that they are non-contiguous. What if we said: "due to the way it is not contiguous" instead of "because ...". Is that any clearer? Or would most people just read that as a more-poorly-worded version of the current statement?

drasmuss: I would read the "because" as applying to "require the array to be copied".  I.e., the reason it is being copied is because it is not contiguous, which is true.  I don't think it implies that all non-contiguous arrays would be copied.  But I also don't think it's that important, so feel free to change it to whatever wording makes sense to you.

PR #1475, opened 2018-09-18
Subclass config

hunse:
**Motivation and context:**

This allows you to subclass a Nengo object (e.g. Ensemble) and still have the config/default system work.

**How has this been tested?**

I've written a unit test.

**Types of changes:**

**Checklist:**

**TODO**

- [x] There's some dead code at the bottom of the unit test, which illustrates a case where I'm not sure what the expected behaviour should be. We need to figure this out.

### DIFF
--- a/nengo/config.py
+++ b/nengo/config.py
@@ -417,10 +417,10 @@ class Config:
                               "keyword with an unconfigurable parameter.")
 
         for config in reversed(Config.context):
-
-            # If a default has been set for this config, return it
-            if nengo_cls in config.params and config[nengo_cls] in desc:
-                return getattr(config[nengo_cls], param)
+            for cls in nengo_cls.__mro__:
+                # If a default has been set for this config, return it
+                if cls in config.params and config[cls] in desc:
+                    return getattr(config[cls], param)
 
         # Otherwise, return the param default
         return desc.default
--- a/nengo/tests/test_config.py
+++ b/nengo/tests/test_config.py
@@ -2,8 +2,9 @@ import pytest
 
 import nengo
 import nengo.synapses
+from nengo.config import SupportDefaultsMixin
 from nengo.exceptions import ConfigError, ReadonlyError
-from nengo.params import Parameter
+from nengo.params import Default, Parameter
 from nengo.utils.testing import ThreadedAssertion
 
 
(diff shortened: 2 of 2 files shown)


tbekolay:
While the change itself seems OK, I feel pretty strongly about not encouraging subclassing any Nengo objects, except `nengo.Network`. If there's a legitimate use case for subclassing a Nengo object, it would be great to discuss other ways it could be done.

hunse:
I'm using subclassing right now to make an Ensemble with convolutional encoders. I'll admit it is a temporary hack (while we wait for the NEP to be approved/implemented), so you might be right, I don't know if it is something that we need long term. But I am finding subclassing useful to get things working in the short term.

tcstewar:
> I feel pretty strongly about not encouraging subclassing any Nengo objects, except `nengo.Network`.

Do you also feel that way about `nengo.Node`?  I subclass it fairly frequently in situations where it's overkill to do a full `nengo.Process`....

tbekolay:
> Do you also feel that way about `nengo.Node`?

Yes, 100%. `nengo.Process` is another thing that we do explicitly want people to subclass, so if subclassing it feels difficult then we should make it easier.

tcstewar:
> Yes, 100%. `nengo.Process` is another thing that we do explicitly want people to subclass, so if subclassing it feels difficult then we should make it easier.

Fair enough.  I'll take another try at forcing myself to use `nengo.Process` instead and see if I can come up with ideas for where I find it to be overkill and what I can do about that.  I've done that more in the past (when we last had this conversation, I think), but have slipped back into old habits and can't quite clearly state where that difference is for me.  :)  Thanks for the reminder!

hunse:
I think this needs more discussion. Do we want to go ahead with this? My opinion is that whether we want people to subclass or not, this should work. If we really want it not to work, then we would need to be much more explicit about not allowing subclassing. Having it silently not work as expected (as is currently the case) is not a good option IMO.

hunse:
Rebased.

One thing that needs discussion is the commented-out code at the bottom of the test. The question is: If we have a general config (i.e. for a parent class) in the current context, but a more specific config (i.e. for the class in question) in a higher context, which do we prefer? The current behaviour is that we prefer the current context (the general config for the parent class, in this case). I'm fine with this (doing it the other way would be more work to implement).

PR #1477, opened 2018-09-20
Add configurable allclose fixture for tests

drasmuss:
**Motivation and context:**

Other backends often want to use Nengo's unit tests.  However, different platforms may expect different levels of accuracy (e.g., custom neuromorphic hardware with discretized weights).  The idea here is to make it easy for backends to modify the tolerances on Nengo's unit tests, through the pytest config system.

**How has this been tested?**

All current unit tests continue to pass, and added some new tests to check the new pytest config options.

I'd recommend going through commit-by-commit.  The diff looks pretty big, but a lot of it is just renaming variables.  I've tried to separate the functional changes from the refactoring in the commit history.

**Types of changes:**

Note: technically some breaking changes in here, but only with respect to the test suite.

**Checklist:**

### DIFF
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -1,9 +1,10 @@
+from fnmatch import fnmatch
 import hashlib
 import inspect
 import importlib
 import os
 import re
-from fnmatch import fnmatch
+import shlex
 import warnings
 
 import matplotlib
@@ -437,18 +438,19 @@ def pytest_runtest_setup(item):  # noqa: C901
     if not hasattr(item, 'obj'):
         return  # Occurs for doctests, possibly other weird tests
 
-    test_uses_sim = 'Simulator' in item.fixturenames
-    test_uses_refsim = 'RefSimulator' in item.fixturenames
-    tests_frontend = not (test_uses_sim or test_uses_refsim)
+    item_name = get_item_name(item)
 
-    if not tests_frontend:
-        item_name = get_item_name(item)
+    # join all the lines and then split (preserving quoted strings)
+    unsupported = shlex.split(
+        " ".join(item.config.getini("nengo_test_unsupported")))
+    # group pairs (representing test name/reason)
+    unsupported = [unsupported[i:i + 2] for i in range(0, len(unsupported), 2)]
 
-        for test, reason in TestConfig.Simulator.unsupported:
-            # We add a '*' before test to eliminate the surprise of needing
-            # a '*' before the name of a test function.
-            if fnmatch(item_name, '*' + test):
-                pytest.xfail(reason)
+    for test, reason in unsupported:
+        # We add a '*' before test to eliminate the surprise of needing
+        # a '*' before the name of a test function.
+        if fnmatch(item_name, "*" + test):
+            pytest.xfail(reason)
 
 
 def pytest_terminal_summary(terminalreporter):
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -287,7 +287,8 @@ def allclose(request):
 
     If a test contains multiple ``allclose`` checks then multiple entries
     with the same testname can be added to the config, and will be applied
-    in the given order.
+    in the given order. If there are fewer config entries than tests, the
+    last entry will be applied to remaining tests with the same testname.
     """
 
     call_count = [0]  # need to access the closure this way for 2.7 compat
--- a/nengo/conftest.py
+++ b/nengo/conftest.py
@@ -269,6 +269,57 @@ def seed(request):
     return function_seed(request.function, mod=TestConfig.test_seed)
 
 
+@pytest.fixture
+def allclose(request):
+    """A functional replacement for np.allclose.
+
+    This allows backends to specify different tolerances by adding
+
+    .. code-block:: ini
+
+        nengo_test_tolerances =
+            path/to/test/file:testname atol=x rtol=y
+            path/to/other/file:testname2 atol=z  # this is a comment
+            ...
+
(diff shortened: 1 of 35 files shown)


nengo/conftest.py
```diff
@@ -269,6 +269,57 @@ def seed(request):
...
+                if entry.startswith("#"):
+                    break
+                k, v = entry.split("=")
+                kwargs[k] = float(v)
+            override.append(kwargs)
+
+    def _allclose(a, b, **kwargs):
+        # set tolerances
+        if override:
+            kwargs.update(override[min(call_count[0], len(override) - 1)])
```
hunse: So if `call_count` is outside the number of overrides, then we just keep applying the last one? Maybe we should state this in the docstring.

nengo/tests/test_pytest.py
```diff
@@ -6,3 +6,10 @@ def test_seed_fixture(seed):
     """The seed should be the same on all machines"""
     i = (seed - TestConfig.test_seed) % npext.maxint
     assert i == 1832276344
+
+
+def test_allclose(allclose):
```
hunse: A note that overrides are being set in `setup.cfg` would be good

nengo/tests/test_solvers.py
```diff
@@ -363,7 +362,7 @@ def input_function(t):
 
     for i, Solver in enumerate(Solvers):
         for j, n_neurons in enumerate(neurons):
-            plt.subplot(len(neurons), len(Solvers), len(Solvers)*j + i + 1)
```
hunse: PEP8 actually says that this should be done the way it originally was here: https://www.python.org/dev/peps/pep-0008/?#other-recommendations

Also, I've noticed that there are a number of stylistic changes in this commit. In my opinion, if they're not violating our static checks (`flake8` and `pylint`), then they should be part of a separate commit. Here, it feels like they make it more difficult to check what is actually being changed.

nengo/utils/numpy.py
```diff
@@ -138,7 +138,7 @@ def norm(x, axis=None, keepdims=False):
         If True, the reduced axes are left in the result. See `np.sum` in
         newer versions of Numpy (>= 1.7).
     """
-    return np.sqrt(np.sum(x**2, axis=axis, keepdims=keepdims))
```
hunse: Why not cast it up here, too?

nengo/utils/numpy.py
```diff
@@ -161,7 +161,9 @@ def rms(x, axis=None, keepdims=False):
         If True, the reduced axes are left in the result. See `np.sum` in
         newer versions of Numpy (>= 1.7).
     """
-    return np.sqrt(np.mean(x**2, axis=axis, keepdims=keepdims))
+
+    x = np.asarray(x)
+    return np.sqrt(np.mean(x ** 2, axis=axis, keepdims=keepdims))
 
 
 def rmse(x, y, axis=None, keepdims=False):
```
hunse: Unrelated, but I really hate that I ever made this function. It's so easy just to do `rms(x - y)`, there's no reason to even have it. Could we deprecate it?

nengo/tests/test_solvers.py
drasmuss: I'm fine reverting those or making them a separate commit, doesn't seem super important to me either way.

nengo/utils/numpy.py
drasmuss: We could (there are probably other places we could do the same as well).  But also the only reason I was touching these functions was because I was getting type errors in the `rmse` part of the `allclose` fixture, which is no longer part of this PR.  So we could just revert this commit as well.

nengo/conftest.py
hunse: Also, just for me to understand, with parametrized tests, there's no way of explicitly targeting a particular parameter setting? The way to do that would be to have multiple entries for the test in the order that the parameters are used?

One thing that's a little confusing about this is when a test is parametrized multiple times. For example
```
pytest.mark.parametrize('a', [0, 1])
pytest.mark.parametrize('b', [2, 3])
def test_my_stuff(a, b):
    print((a, b))
```
I assume this will print "(0, 2) (0, 3) (1, 2) (1, 3)" but I'm not sure.

I'm wondering if there's a better way to do this. I think the `request` does include information about the parameters, so maybe there's a way to check this?

hunse:
I've added some commits to do the things I requested. I still have an open question about how to deal with `parametrize` tests.

hunse:
I also added a commit to deprecate `rmse`. Maybe it's not fair to tag it on to this PR, but it just seems like such a small change.

nengo/conftest.py
drasmuss: Yes you can explicitly target parameter settings with e.g.
```
nengo_test_unsupported = file/path/testname[parameter_value]
```
I added that to the tests/documentation so it's more obvious.  I also added a thing to automatically escape special characters, so that people don't have to remember to do that.

drasmuss:
> I think it's actually a very nice thing to have. We don't have to do it now, but I would push to take what @tbekolay has done in nengo-loihi (with the automatic reporting of RMSE in the test results) and bring that into this PR. Since we're adding the allclose fixture now, it seems like it's worth spending a little extra time to get the features we want.

We do need the tolerances feature soon for Loihi, and we're both pretty busy, so my inclination would be to get that sorted out first and worry about RMSE features later.

hunse:
Fair enough

tbekolay:
Done with the `pytest-allclose` fixture instead in #1563.

PR #1481, opened 2018-10-29
Add support for convolutional transformations

drasmuss:
**Motivation and context:**
It would be good to be able to represent transforms other than dense 2D weight matrices in a consistent way across the nengo ecosystem.  This focuses specifically on convolutional connections, although with the assumption that we will want to implement more of these (e.g. sparse weight matrices).

See https://github.com/nengo/enhancement-proposals/pull/14 for more detailed motivation/discussion.

Some discussion points:
- Do we want to add an example notebook demonstrating this new syntax?
- I added the ``Transforms`` superclass based on the assumption that we will want more than one of these kinds of things.  But right now we only have one, so perhaps this is premature.  We could just have everything refer directly to ``Convolution`` instead of ``Transform`` (but would then need to go and reintroduce some generality if we add other types of transforms).

**How has this been tested?**

Added new tests in `test_transform.py` and `builder/tests/test_transform.py`.

I'd start out by looking at the content in `transform.py`, as that is the main user-facing change.  Then take a look at the builder changes to see how that gets implemented.  The rest of the changes have to do with documentation/validation/testing.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/_vendor/npconv2d/conv2d.py
+++ b/nengo/_vendor/npconv2d/conv2d.py
@@ -1,0 +1,132 @@
+"""
+Vendorized from https://github.com/renmengye/np-conv2d
+
+MIT License
+
+Copyright (c) 2017 Mengye Ren
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+"""
+from __future__ import division
+
+import numpy as np
+from nengo.utils.numpy import array_offset
+
+
+def calc_pad(pad, in_siz, out_siz, stride, ksize):
+    """Calculate padding width.
+    Args:
+        pad: padding method, "SAME", "VALID", or manually speicified.
+        ksize: kernel size [I, J].
+    Returns:
+        pad_: Actual padding width.
+    """
+    if pad == 'SAME':
+        return (out_siz - 1) * stride + ksize - in_siz
+    elif pad == 'VALID':
+        return 0
+    else:
+        return pad
+
+
+def calc_size(h, kh, pad, sh):
+    """Calculate output image size on one dimension.
+    Args:
+        h: input image size.
+        kh: kernel size.
+        pad: padding strategy.
+        sh: stride.
+    Returns:
+        s: output size.
+    """
+
+    if pad == 'VALID':
+        return np.ceil((h - kh + 1) / sh)
+    elif pad == 'SAME':
+        return np.ceil(h / sh)
+    else:
+        return int(np.ceil((h - kh + pad + 1) / sh))
+
+
+def extract_sliding_windows(x, ksize, pad, stride, floor_first=True):
+    """Converts a tensor to sliding windows.
+    Args:
+        x: [N, H, W, C]
+        k: [KH, KW]
+        pad: [PH, PW]
+        stride: [SH, SW]
+    Returns:
+        y: [N, (H-KH+PH+1)/SH, (W-KW+PW+1)/SW, KH * KW, C]
+    """
+    n = x.shape[0]
+    h = x.shape[1]
+    w = x.shape[2]
+    c = x.shape[3]
+    kh = ksize[0]
+    kw = ksize[1]
+    sh = stride[0]
+    sw = stride[1]
+
+    h2 = int(calc_size(h, kh, pad, sh))
+    w2 = int(calc_size(w, kw, pad, sw))
+    ph = int(calc_pad(pad, h, h2, sh, kh))
+    pw = int(calc_pad(pad, w, w2, sw, kw))
+
+    ph0 = int(np.floor(ph / 2))
+    ph1 = int(np.ceil(ph / 2))
+    pw0 = int(np.floor(pw / 2))
+    pw1 = int(np.ceil(pw / 2))
+
+    if floor_first:
+        pph = (ph0, ph1)
+        ppw = (pw0, pw1)
+    else:
+        pph = (ph1, ph0)
+        ppw = (pw1, pw0)
+    x = np.pad(
+        x, ((0, 0), pph, ppw, (0, 0)),
+        mode='constant',
+        constant_values=(0.0,))
+
+    x_sn, x_sh, x_sw, x_sc = x.strides
+    y_strides = (x_sn, sh*x_sh, sw*x_sw, x_sh, x_sw, x_sc)
+    y = np.ndarray((n, h2, w2, kh, kw, c), dtype=x.dtype, buffer=x.data,
+                   offset=array_offset(x), strides=y_strides)
+    return y
+
+
+def conv2d(x, w, pad='SAME', stride=(1, 1)):
+    """2D convolution (technically speaking, correlation).
+    Args:
+        x: [N, H, W, C]
+        w: [I, J, C, K]
+        pad: [PH, PW]
+        stride: [SH, SW]
+    Returns:
+        y: [N, H', W', K]
+    """
+    ksize = w.shape[:2]
+    x = extract_sliding_windows(x, ksize, pad, stride)
+    ws = w.shape
+    w = w.reshape([ws[0] * ws[1] * ws[2], ws[3]])
+    xs = x.shape
+    x = x.reshape([xs[0] * xs[1] * xs[2], -1])
+    y = x.dot(w)
+    y = y.reshape([xs[0], xs[1], xs[2], -1])
+    return y
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -5,10 +5,9 @@ import numpy as np
 from nengo.builder import Builder, Signal
 from nengo.builder.ensemble import gen_eval_points, get_activities
 from nengo.builder.node import SimPyFunc
-from nengo.builder.operator import Copy, DotInc, ElementwiseInc, Reset
-from nengo.builder.transforms import ConvInc
+from nengo.builder.operator import Copy, ElementwiseInc
 from nengo.connection import Connection
-from nengo.transforms import Convolution, get_transform
+from nengo.transforms import Convolution
 from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError, ObsoleteError
 from nengo.neurons import Direct
@@ -100,7 +99,7 @@ def build_linear_system(model, conn, rng):
     return eval_points, activities, targets
 
 
-def build_decoders(model, conn, rng, transform):
+def build_decoders(model, conn, rng):
     encoders = model.params[conn.pre_obj].encoders
     gain = model.params[conn.pre_obj].gain
     bias = model.params[conn.pre_obj].bias
@@ -109,20 +108,12 @@ def build_decoders(model, conn, rng, transform):
     targets = get_targets(conn, eval_points)
 
     x = np.dot(eval_points, encoders.T / conn.pre_obj.radius)
-    E = None
-    if conn.solver.weights:
-        E = model.params[conn.post_obj].scaled_encoders.T[conn.post_slice]
-        # include transform in solved weights
-        targets = multiply(targets, transform.T)
-
     wrapped_solver = (model.decoder_cache.wrap_solver(solve_for_decoders)
                       if model.seeded[conn] else solve_for_decoders)
     decoders, solver_info = wrapped_solver(
-        conn, gain, bias, x, targets, rng=rng, E=E)
+        conn, gain, bias, x, targets, rng=rng)
 
-    weights = (decoders.T if conn.solver.weights else
-               multiply(transform, decoders.T))
-    return eval_points, weights, solver_info
+    return eval_points, decoders.T, solver_info
 
 
 def solve_for_decoders(conn, gain, bias, x, targets, rng, E=None):
@@ -161,19 +152,17 @@ def slice_signal(model, signal, sl):
 
 
 @Builder.register(Solver)
-def build_solver(model, solver, conn, rng, transform):
-    return build_decoders(model, conn, rng, transform)
+def build_solver(model, solver, conn, rng):
+    return build_decoders(model, conn, rng)
 
 
 @Builder.register(NoSolver)
-def build_no_solver(model, solver, conn, rng, transform):
+def build_no_solver(model, solver, conn, rng):
     activities = np.zeros((1, conn.pre_obj.n_neurons))
     targets = np.zeros((1, conn.size_mid))
-    E = np.zeros((1, conn.post_obj.n_neurons)) if solver.weights else None
     # No need to invoke the cache for NoSolver
-    decoders, solver_info = conn.solver(activities, targets, rng=rng, E=E)
-    weights = (decoders.T if conn.solver.weights else
-               multiply(transform, decoders.T))
+    decoders, solver_info = conn.solver(activities, targets, rng=rng)
+    weights = decoders.T
     return None, weights, solver_info
 
 
@@ -229,15 +218,11 @@ def build_connection(model, conn):
     model.sig[conn]['in'] = get_prepost_signal(is_pre=True)
     model.sig[conn]['out'] = get_prepost_signal(is_pre=False)
 
-    weights = None
+    decoders = None
+    encoders = None
     eval_points = None
     solver_info = None
-    signal_size = conn.size_out
     post_slice = conn.post_slice
-
-    # Sample transform if given a distribution
-    transform = get_transform(
-        conn.transform, (conn.size_out, conn.size_mid), rng=rng)
 
     # Figure out the signal going across this connection
     in_signal = model.sig[conn]['in']
@@ -245,7 +230,6 @@ def build_connection(model, conn):
             or (isinstance(conn.pre_obj, Ensemble)
                 and isinstance(conn.pre_obj.neuron_type, Direct))):
         # Node or Decoded connection in directmode
-        weights = transform
         sliced_in = slice_signal(model, in_signal, conn.pre_slice)
         if conn.function is None:
             in_signal = sliced_in
@@ -255,46 +239,34 @@ def build_connection(model, conn):
             in_signal = Signal(np.zeros(conn.size_mid), name='%s.func' % conn)
             model.add_op(SimPyFunc(in_signal, conn.function, None, sliced_in))
     elif isinstance(conn.pre_obj, Ensemble):  # Normal decoded connection
-        if isinstance(conn.transform, Convolution):
-            raise NotImplementedError(
-                "Convolutional transforms on decoded connections are "
-                "not supported")
-
-        eval_points, weights, solver_info = model.build(
-            conn.solver, conn, rng, transform)
+        eval_points, decoders, solver_info = model.build(
+            conn.solver, conn, rng)
         if conn.solver.weights:
             model.sig[conn]['out'] = model.sig[conn.post_obj.neurons]['in']
-            signal_size = conn.post_obj.neurons.size_in
-            post_slice = None  # don't apply slice later
+
+            if isinstance(conn.post_obj, Ensemble):
+                encoders = model.params[conn.post_obj].scaled_encoders.T
+                encoders = encoders[conn.post_slice]
+
+                # post slice already applied to encoders, don't apply later
+                post_slice = None
     else:
-        weights = transform
         in_signal = slice_signal(model, in_signal, conn.pre_slice)
 
-    # Add operator for applying weights
-    model.sig[conn]['weights'] = Signal(
-        weights, name="%s.weights" % conn, readonly=True)
-    signal = Signal(np.zeros(signal_size), name="%s.weighted" % conn)
-    model.add_op(Reset(signal))
+    # Build transform
+    weighted, weights = model.build(conn.transform,
+                                    in_signal,
+                                    decoders=decoders,
+                                    encoders=encoders,
+                                    rng=rng)
+    model.sig[conn]["weights"] = weights
 
-    if isinstance(conn.transform, Convolution):
-        model.add_op(ConvInc(model.sig[conn]["weights"],
-                             in_signal,
-                             signal,
-                             conn.transform,
-                             tag="%s.weights_convinc" % conn))
-    else:
-        op = ElementwiseInc if weights.ndim < 2 else DotInc
-        model.add_op(op(model.sig[conn]['weights'],
-                        in_signal,
-                        signal,
-                        tag="%s.weights_elementwiseinc" % conn))
-
-    # Add operator for filtering
+    # Build synapse
     if conn.synapse is not None:
-        signal = model.build(conn.synapse, signal)
+        weighted = model.build(conn.synapse, weighted)
 
     # Store the weighted-filtered output in case we want to probe it
-    model.sig[conn]['weighted'] = signal
+    model.sig[conn]['weighted'] = weighted
 
     if isinstance(conn.post_obj, Neurons):
         # Apply neuron gains (we don't need to do this if we're connecting to
@@ -302,16 +274,17 @@ def build_connection(model, conn):
         gains = Signal(model.params[conn.post_obj.ensemble].gain[post_slice],
                        name="%s.gains" % conn)
         model.add_op(ElementwiseInc(
-            gains, signal, model.sig[conn]['out'][post_slice],
+            gains, weighted, model.sig[conn]['out'][post_slice],
             tag="%s.gains_elementwiseinc" % conn))
     else:
         # Copy to the proper slice
         model.add_op(Copy(
-            signal, model.sig[conn]['out'], dst_slice=post_slice,
+            weighted, model.sig[conn]['out'], dst_slice=post_slice,
             inc=True, tag="%s" % conn))
 
     # Build learning rules
     if conn.learning_rule is not None:
+        # TODO: provide a general way for transforms to expose learnable params
         if isinstance(conn.transform, Convolution):
             raise NotImplementedError(
                 "Learning on convolutional connections is not supported")
@@ -334,5 +307,5 @@ def build_connection(model, conn):
 
     model.params[conn] = BuiltConnection(eval_points=eval_points,
                                          solver_info=solver_info,
-                                         transform=transform,
-                                         weights=weights)
+                                         transform=conn.transform,
+                                         weights=weights.initial_value)
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -6,8 +6,9 @@ from nengo.builder import Builder, Signal
 from nengo.builder.ensemble import gen_eval_points, get_activities
 from nengo.builder.node import SimPyFunc
 from nengo.builder.operator import Copy, DotInc, ElementwiseInc, Reset
+from nengo.builder.transforms import ConvInc
 from nengo.connection import Connection
-from nengo.dists import get_samples
+from nengo.transforms import Convolution, get_transform
 from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError, ObsoleteError
 from nengo.neurons import Direct
@@ -235,8 +236,8 @@ def build_connection(model, conn):
     post_slice = conn.post_slice
 
     # Sample transform if given a distribution
-    transform = get_samples(
-        conn.transform, conn.size_out, d=conn.size_mid, rng=rng)
+    transform = get_transform(
+        conn.transform, (conn.size_out, conn.size_mid), rng=rng)
 
     # Figure out the signal going across this connection
     in_signal = model.sig[conn]['in']
@@ -254,6 +255,11 @@ def build_connection(model, conn):
             in_signal = Signal(np.zeros(conn.size_mid), name='%s.func' % conn)
             model.add_op(SimPyFunc(in_signal, conn.function, None, sliced_in))
     elif isinstance(conn.pre_obj, Ensemble):  # Normal decoded connection
+        if isinstance(conn.transform, Convolution):
+            raise NotImplementedError(
+                "Convolutional transforms on decoded connections are "
+                "not supported")
+
         eval_points, weights, solver_info = model.build(
             conn.solver, conn, rng, transform)
         if conn.solver.weights:
@@ -269,11 +275,19 @@ def build_connection(model, conn):
         weights, name="%s.weights" % conn, readonly=True)
     signal = Signal(np.zeros(signal_size), name="%s.weighted" % conn)
     model.add_op(Reset(signal))
-    op = ElementwiseInc if weights.ndim < 2 else DotInc
-    model.add_op(op(model.sig[conn]['weights'],
-                    in_signal,
-                    signal,
-                    tag="%s.weights_elementwiseinc" % conn))
+
+    if isinstance(conn.transform, Convolution):
+        model.add_op(ConvInc(model.sig[conn]["weights"],
+                             in_signal,
+                             signal,
+                             conn.transform,
+                             tag="%s.weights_convinc" % conn))
+    else:
+        op = ElementwiseInc if weights.ndim < 2 else DotInc
+        model.add_op(op(model.sig[conn]['weights'],
+                        in_signal,
+                        signal,
+                        tag="%s.weights_elementwiseinc" % conn))
 
     # Add operator for filtering
     if conn.synapse is not None:
@@ -298,6 +312,10 @@ def build_connection(model, conn):
 
     # Build learning rules
     if conn.learning_rule is not None:
+        if isinstance(conn.transform, Convolution):
+            raise NotImplementedError(
+                "Learning on convolutional connections is not supported")
+
         rule = conn.learning_rule
         rule = [rule] if not is_iterable(rule) else rule
         targets = []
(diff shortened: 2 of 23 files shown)


nengo/_vendor/npconv2d/conv2d.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    ksize = w.shape[:2]
+    x = extract_sliding_windows(x, ksize, pad, stride)
+    ws = w.shape
+    w = w.reshape([ws[0] * ws[1] * ws[2], ws[3]])
+    xs = x.shape
+    x = x.reshape([xs[0] * xs[1] * xs[2], -1])
+    y = x.dot(w)
+    y = y.reshape([xs[0], xs[1], xs[2], -1])
+    return y
```
hunse: I'm not convinced this is the best way to do convolution. Did you test it for speed and memory use? It looks to me like `extract_sliding_windows` will create a version of `x` that is potentially quite large. Its shape is `(n, h2, w2, kh, kw, c)`, where h2 and w2 are the output size (fine), but kh and kw are the size of the kernel. If you've got e.g. a 9x9 kernel, then this intermediary object is going to use 81 times as much memory as the output object (not accounting for any difference in input and output channels). The advantage to this method is that you can line everything up to run it through a single `dot` call, which may save time. I'll try to do some profiling on this to see how much potential time saving there is.

nengo/builder/connection.py
```diff
@@ -269,11 +275,16 @@ def get_prepost_signal(is_pre):
         weights, name="%s.weights" % conn, readonly=True)
     signal = Signal(np.zeros(signal_size), name="%s.weighted" % conn)
     model.add_op(Reset(signal))
-    op = ElementwiseInc if weights.ndim < 2 else DotInc
-    model.add_op(op(model.sig[conn]['weights'],
-                    in_signal,
-                    signal,
-                    tag="%s.weights_elementwiseinc" % conn))
+
+    if isinstance(conn.transform, Convolution):
```
hunse: The problem with this is if somebody wants to add their own transform, they'd have to list it here, which is not really possible. Would it be possible to have a `build` function for the transform, to make it pluggable?

nengo/builder/connection.py
```diff
@@ -298,6 +309,10 @@ def get_prepost_signal(is_pre):
 
     # Build learning rules
     if conn.learning_rule is not None:
+        if isinstance(conn.transform, Convolution):
```
hunse: I'm less worried about this, but ultimately, it might be nice to have a way that users can define if their transform supports learning, and if so specify how the learning happens. I don't think this is high priority, though, so I'm happy to leave that as future work.

nengo/connection.py
```diff
@@ -168,18 +169,28 @@ def check_function_can_be_applied(self, conn, function_info):
         size_mid = conn.size_in if size is None else size
         transform = conn.transform
 
-        if isinstance(transform, np.ndarray):
-            if transform.ndim < 2 and size_mid != conn.size_out:
-                raise ValidationError(
-                    "function output size is incorrect; should return a "
-                    "vector of size %d" % conn.size_out, attr=self.name,
-                    obj=conn)
+        if isinstance(transform, (np.ndarray, Transform)):
```
hunse: I think having full weights also be a transform (as discussed at the meeting) will make the following sections a lot cleaner. I'm going to hold off looking at them too closely until that's done.

nengo/transforms.py
```diff
@@ -0,0 +1,233 @@
...
+    @property
+    def size_in(self):
+        return self.input_shape.size
+
+    @property
+    def size_out(self):
+        return self.output_shape.size
+
+
+class ConvShape(object):
```
hunse: This could potentially be used in the future for non-convolutional things. For example, you could have local filters which are like convolution but different at different image locations. These still deal in channels. I'm wondering if we should call it `ChannelShape` or something more generic.

nengo/builder/tests/test_transforms.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.mark.parametrize("channels_last", (True, False))
+@pytest.mark.parametrize("stride0", (1, 2))
+@pytest.mark.parametrize("stride1", (1, 2))
+@pytest.mark.parametrize("kernel0", (4, 5))
+@pytest.mark.parametrize("kernel1", (4, 5))
+@pytest.mark.parametrize("shape0", (16, 17))
+@pytest.mark.parametrize("shape1", (16, 17))
+@pytest.mark.parametrize("padding", ("same", "valid"))
```
hunse: Not sure if we're worried about time for the nengo tests, but this consists of 256 tests and adds 25 seconds to the test suite. It's good that it's thorough, but I'm wondering if there's a way to make it a little shorter at least for the typical tests. We could still keep more extended tests that we run manually if someone's making changes to the conv code.

drasmuss: Note that this is already automatically skipped in the typical tests on TravisCI, as it is skipped if scipy isn't installed.  So it's only an issue for the one run we do with scipy, or for people testing locally.  But we also have a pytest marker for slow tests, if we wanted to add an additional opt-in layer.

nengo/_vendor/npconv2d/conv2d.py
drasmuss: It does definitely save time, but I haven't done any memory profiling to see how it compares to other options in that regard.

hunse: I added a commit that uses a view instead of copying in `extract_sliding_windows`, so the memory should no longer be an issue. And as per your profiling, the speed is much faster.

drasmuss:
Pushed some changes to make transform building pluggable, and to treat full weight matrices as the new ``Dense`` transform object.

There are two functional changes as a result of this.
``` python
my_conn = nengo.Connection(a, b, transform=1)
my_conn.transform  # this will be a nengo.Dense object now
my_conn.transform.init  # the original value passed to the transform
```
This is the same way `synapses` work (e.g. `nengo.Connection(a, b, synapse=1)` will result in `my_conn.synapse == Lowpass(1)`, so shouldn't be too surprising.

The other change is to `NoSolver(..., weights=True)`.  The way that the solver process usually works is that the solver provides decoders, and then those are automatically combined with the transform and encoders (in the `weights=True`) case.  The new transform build logic slightly modifies this, as the responsibility for building the full matrix now lies in the transform builder rather than the solver. 
 In most cases this doesn't make any difference, except for the specific case of using NoSolver with weights=True.  Previously, in this case the user would pass the full weight matrix (not just the decoders), and then encoders were not automatically applied (unlike with all the other solvers).  The transform was automatically applied, but it wouldn't really work for anything except scalar transforms because the dimensions wouldn't line up (this was technically a bug in the previous setup, but we never noticed because I suspect no one really uses NoSolver with weights=True).  

I modified this so that with NoSolver the user always specifies just the decoders, and then if `weights=True` then the transform/encoders are applied on top of that (as with all the other solvers).  Note that the previous NoSolver behaviour is equivalent to `nengo.Connection(a.neurons, b.neurons, transform=weights)`, so we haven't lost functionality.  And this doesn't affect ``NoSolver(..., weights=False)``, which is the common use case.  So, a minorly-breaking change, but seems like an overall improvement I hope.

Semi-related to the above, currently all our solvers have an option to produce full weight matrices (but you have to pass the transform and encoders into them so that they can do the combination).  With the new setup where all that is encapsulated in the transform build process, we don't actually use that functionality of the solvers any more.  I left it in because people might be using the solvers in other ways in their own code, and still relying on that functionality.  But we could consider taking that code out if we want to clean things up a bit.

nengo/utils/builder.py
```diff
@@ -27,7 +27,7 @@ def full_transform(conn, slice_pre=True, slice_post=True, allow_scalars=True):
         not using slicing, since these work fine in the reference builder.
         If false, these scalars will be turned into scaled identity matrices.
     """
-    transform = conn.transform
+    transform = conn.transform.init
```
hunse: My guess is that this only works for Dense transforms? Maybe we should check/specify that?

drasmuss: Done. As a sidenote, I did look at moving this into the `Dense` class like we talked about, but I saw that this function is only used in other `utils` functions.  So it seemed weird to move a relatively large chunk of code from utils into the core in order to support utils.  We could do it though if we find more common uses for `full_transform` .

nengo/builder/tests/test_transforms.py
drasmuss: I also reduced this to 64 conditions now.

nengo/builder/connection.py
drasmuss: I put a TODO marking this as something we should do in the future

nengo/builder/tests/test_transforms.py
hunse: Yeah, I might make them opt-in. It would be nice if one of the CI tests ran them, though. Would it be possible to make the Scipy test-run opt in to slow tests as well? I know there's been at least one time in the past when a slow test has been broken in master and we didn't notice because they don't always get run. Or would adding those to the Scipy run put us at risk of timing out? @tbekolay

nengo/transforms.py
```diff
@@ -0,0 +1,348 @@
...
+
+    coerce_defaults = False
+
+    def __init__(self, name, default, optional=False, readonly=False):
+        super(TransformParam, self).__init__(
+            name, default, optional, readonly)
+
+    def coerce(self, conn, transform):
+        if not isinstance(transform, Transform):
+            transform = Dense((conn.size_out, conn.size_mid), transform)
```
hunse: Rather than having this be tied to a Connection (in which case, it really seems like something that should live in `connection.py`), could we have a `shape` param where we pass `('size_out', 'size_mid')` like with `NdArrayParam.shape` for Connections? Or maybe we'd want them as separate `size_in` and `size_out` params. Then, this could potentially be used other places.

One potential place it could immediately be used is for `Ensemble.encoders`. It would be nice (for me at least) to be able to have convolutional encoders. This might be a bit of work, so I'm not saying it should happen in this PR necessarily. And maybe the suggestion I made about generalizing TransformParam shouldn't be implemented until we're using it in another place so that we really have a general test for it. But long term, that's where I see all of this going.

Maybe I'll take a quick look at that now and see how much work it would be.

nengo/builder/connection.py
```diff
@@ -254,50 +239,59 @@ def get_prepost_signal(is_pre):
             in_signal = Signal(np.zeros(conn.size_mid), name='%s.func' % conn)
             model.add_op(SimPyFunc(in_signal, conn.function, None, sliced_in))
     elif isinstance(conn.pre_obj, Ensemble):  # Normal decoded connection
-        eval_points, weights, solver_info = model.build(
-            conn.solver, conn, rng, transform)
+        if isinstance(conn.transform, Convolution):
```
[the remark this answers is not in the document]
drasmuss: Woops I actually just forgot to take that out, it isn't needed anymore (it is checked in the transform builder).

drasmuss:
Meant to ask this earlier: we're failing the `numpy==1.10` check on Travis-CI because I use `np.moveaxis`, which doesn't exist until 1.11.  I could implement a workaround for the 1.10 case, but before I do that I thought I'd check if we just want to bump the minimum numpy version (I figure we'll be gradually dropping old support at some point, so when we run into something like this is a good reminder to think about if now is the time).

hunse:
You could also replace `moveaxis` with `np.transpose`, though looking at your use-case it might be a little more cumbersome. And `rollaxis` is still around (and we still use it in Nengo), but `moveaxis` is preferred and again less cumbersome.

I'm happy to drop 1.10 support at some point, but I'd want to do a little research before we do. For example, what is the current version in the repo for the oldest LTS version of Ubuntu? That might be a good metric to decide when to drop support.

drasmuss:
It looks like 14.04 ships with 1.8 (so we're not supporting that anyway), and then 16.04 ships with 1.11 (https://packages.ubuntu.com/search?keywords=python-numpy).

tbekolay:
> 14.04 ships with 1.8 (so we're not supporting that anyway)

I think [we are](https://github.com/nengo/nengo/blob/master/setup.py#L31), we just stopped testing with it in TravisCI because the conda package was having some issue. Though, I feel like we have older releases that support older numpys (for people in dire situations where they have to use old packages) so I'm :+1: on upping the requirement, especially since we have lots of documentation on how to install numpy in all sorts of situations.

hunse:
Ok, I've added two commits. cc60464 is tiny and just switches the order of multiplication to make it more readable IMO.

37c6d72 moves TransformParam, which is specific to Connections, to `connections.py`. It also makes a few little changes to the exceptions (mostly just in wording, and I think there was one exception where `size_out` was used when it should've been `size_mid`). I also added a more general `TransformParam` that can theoretically be used down the road in other places. Somewhat ironically, I'm not actually using it for ConnectionTransformParam, since that had so much unique logic that it seemed difficult to actually separate any of the general parts out in a meaningful way. Since this hasn't been tested, I can see an argument for taking it out. I'd be fine with that, but I think it'd be a good starting block if we want to have encoder transforms down the road, so it'd be nice to have a place to put it. Maybe I could just copy the code into an issue for making encoder transforms.

Finally, I did look into encoder transforms, but they're going to be a little bit of work. We'll have to add an `apply` function to transforms (needed for figuring out how the transform acts on the eval points), as well as some way to normalize the transform if desired so that we get unit output for unit input. So I'm not going to do that now.

nengo/transforms.py
```diff
@@ -46,54 +47,37 @@ class TransformParam(Parameter):
 
     coerce_defaults = False
 
-    def __init__(self, name, default, optional=False, readonly=False):
```
drasmuss: I would take this out, since it isn't used anywhere.

nengo/transforms.py
```diff
@@ -135,11 +119,13 @@ class Dense(Transform):
...
         super(Dense, self).__init__()
 
-        self.shape = shape
-
         if is_array_like(init):
             init = np.asarray(init)
 
+            # replace any unspecified shape values
+            shape = tuple(s if s != '*' else init.shape[k] if k < init.ndim
+                          else 1 for k, s in enumerate(shape))
```
drasmuss: I'm OK with having this in, but since it's not actually being used in the code we should add a test (so we know it's working).

drasmuss:
Ah yeah you're right, I was reading the conditions wrong.  Fixups all look good to me then!

nengo/transforms.py
```diff
@@ -0,0 +1,290 @@
...
+    strides : tuple of int, optional
+        Stride of the convolution (1 element for a 1D convolution, 2 for
+        a 2D convolution, etc.).
+    padding : ``"same"`` or ``"valid"``, optional
+        Padding method for input signal.  "Valid" means no padding, and
+        convolution will only be applied to the fully-overlapping areas of the
+        input signal (meaning the output will be smaller).  "Same" means that
+        the input signal is zero-padded so that the output is the same shape
+        as the input.
+    channels_last : bool, optional
```
tbekolay: Is there some historical reason why you would want to give this flexibility? Like is it common in one type of model to have channels last and in other models first? My inclination would be too choose one and force it (maybe auto-reshape if it's possible) but if there are two competing conventions then I'm fine to leave it.

nengo/transforms.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def sample(self, rng=np.random):
+        if isinstance(self.init, Distribution):
+            return self.init.sample(*self.shape, rng=rng)
+
+        return self.init
+
+    @property
+    def init_shape(self):
+        """The shape of the initial value."""
```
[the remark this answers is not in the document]
drasmuss: It would be different due to the broadcasting we allow (e.g. passing `transform=1.0` instead of the full weight matrix).  So in that case `init_shape` would be `()`, but `shape` would be `(size_out, size_in)`.

nengo/transforms.py
drasmuss: It's a pretty common parameter for convolution (e.g. https://www.tensorflow.org/api_docs/python/tf/nn/convolution, https://keras.io/layers/convolutional/).  And there are competing conventions (e.g. the default in Theano is channels_first, default in TensorFlow is channels_last).

There are also concrete reasons why users might be forced to use one or the other (e.g., if you're using TensorFlow without a GPU you have to use channels_last, but if you're using nengo-loihi you pretty much have to use channels_first).  So we can't just tell users "always use channels_last".

We could have the user always specify the model with, e.g., channels_last, and then the backend could move dimensions around before each conv op to channels_first or channels_last according to their preference.  But then we're forcing backends that could efficiently support both styles to only support one.  And there's a small but non-trivial amount of overhead in moving those dimensions around.

Another option would be to remove the `channels_last` parameter from `Convolution`, and force users to specify a `ChannelShape` for `input_shape` if they want to specify `channels_last`.  However, in the current design I've tried to obscure `ChannelShape` from the user, so they don't really need to know it exists.  So we'd be adding some complexity in learning `ChannelShape`, in exchange for removing complexity from `Convolution`.

Overall my inclination would be to leave the parameter as-is, since it has some concrete benefits and should hopefully be familiar to CNN users.

nengo/builder/transforms.py
```diff
@@ -0,0 +1,127 @@
...
+        if not self.conv.channels_last:
+            X = np.moveaxis(X, 0, -1)
+            Y = np.moveaxis(Y, 0, -1)
+
+        if self.conv.dimensions == 1:
+            # add extra dimension to make it a 2D convolution
+            X = X[:, None, :]
+            W = W[:, None, :, :]
+            Y = Y[:, None, :]
+            stride = stride + (1,)
```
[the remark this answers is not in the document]
drasmuss: It does affect what is happening in `make_step` though, as it changes whether we're doing a `(N, 1)` convolution or a `(1, N)` convolution.  I did some quick benchmarking and it seemed like the latter was slightly faster, so I switched it around.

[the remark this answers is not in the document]
hunse: Cool, thanks!

tbekolay:
When running the test locally, I'm getting some shape off-by-one errors in `builder/tests/test_transforms.py`. Here's one of them:

```
_________________________ test_convinc_2d[valid-5-5-2-2-False] _________________________

channels_last = False, stride0 = 2, stride1 = 2, kernel0 = 5, kernel1 = 5
padding = 'valid', rng = <mtrand.RandomState object at 0x7f36059e2410>

    @pytest.mark.parametrize("channels_last", (True, False))
    @pytest.mark.parametrize("stride0", (1, 2))
    @pytest.mark.parametrize("stride1", (1, 2))
    @pytest.mark.parametrize("kernel0", (4, 5))
    @pytest.mark.parametrize("kernel1", (4, 5))
    @pytest.mark.parametrize("padding", ("same", "valid"))
    def test_convinc_2d(channels_last, stride0, stride1, kernel0, kernel1,
                        padding, rng):
        correlate2d = pytest.importorskip("scipy.signal").correlate2d
    
        shape0 = 16
        shape1 = 17
        in_channels = 32
        out_channels = 64
        x_shape = (shape0, shape1, in_channels) if channels_last else (
            in_channels, shape0, shape1)
        x = Signal(rng.randn(*x_shape))
        w = Signal(rng.randn(kernel0, kernel1, in_channels, out_channels))
    
        conv = Convolution(out_channels, x_shape, kernel_size=(kernel0, kernel1),
                           strides=(stride0, stride1), padding=padding,
                           channels_last=channels_last)
    
        y = Signal(np.zeros(conv.output_shape.shape))
    
        signals = {sig: np.array(sig.initial_value) for sig in (x, w, y)}
        step = ConvInc(w, x, y, conv).make_step(signals, None, None)
    
>       step()

nengo/builder/tests/test_transforms.py:37: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def step_conv():
>       Y[...] += conv2d.conv2d(X, W, pad=pad, stride=stride)[0]
E       ValueError: operands could not be broadcast together with shapes (6,7,64) (6,6,64) (6,7,64)

nengo/builder/transforms.py:125: ValueError
```

In all, 42 test fail with one of these and 22 pass. This only happens to me in Python 2.7, with NumPy 1.15.1 and Scipy 1.1.0; my Python 3 environment (with NumPy 1.14.2 and Scipy 1.0.0) passes all the tests. Since the test depends on Scipy, TravisCI isn't testing it on Python 2.7, so I don't think this is specific to my environment, I think it's a Python 2 issue (possibly in Numpy or Scipy).

This issue seems to happen before and after cleaning up the history, so I'm going to push my cleaned up history anyhow.

tbekolay:
In case it helps with debugging, here's the list of which pass and which fail:

```
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-4-1-1-True] FAILED [  1%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-4-1-1-False] FAILED [  3%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-4-1-2-True] FAILED [  4%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-4-1-2-False] FAILED [  6%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-4-2-1-True] FAILED [  7%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-4-2-1-False] FAILED [  9%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-4-2-2-True] FAILED [ 10%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-4-2-2-False] FAILED [ 12%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-5-1-1-True] FAILED [ 14%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-5-1-1-False] FAILED [ 15%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-5-1-2-True] FAILED [ 17%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-5-1-2-False] FAILED [ 18%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-5-2-1-True] FAILED [ 20%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-5-2-1-False] FAILED [ 21%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-5-2-2-True] FAILED [ 23%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-4-5-2-2-False] FAILED [ 25%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-4-1-1-True] FAILED [ 26%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-4-1-1-False] FAILED [ 28%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-4-1-2-True] PASSED [ 29%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-4-1-2-False] PASSED [ 31%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-4-2-1-True] FAILED [ 32%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-4-2-1-False] FAILED [ 34%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-4-2-2-True] FAILED [ 35%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-4-2-2-False] FAILED [ 37%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-5-1-1-True] PASSED [ 39%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-5-1-1-False] PASSED [ 40%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-5-1-2-True] FAILED [ 42%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-5-1-2-False] FAILED [ 43%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-5-2-1-True] FAILED [ 45%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-5-2-1-False] FAILED [ 46%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-5-2-2-True] FAILED [ 48%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[same-5-5-2-2-False] FAILED [ 50%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-4-1-1-True] PASSED [ 51%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-4-1-1-False] PASSED [ 53%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-4-1-2-True] FAILED [ 54%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-4-1-2-False] FAILED [ 56%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-4-2-1-True] PASSED [ 57%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-4-2-1-False] PASSED [ 59%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-4-2-2-True] FAILED [ 60%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-4-2-2-False] FAILED [ 62%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-5-1-1-True] PASSED [ 64%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-5-1-1-False] PASSED [ 65%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-5-1-2-True] PASSED [ 67%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-5-1-2-False] PASSED [ 68%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-5-2-1-True] PASSED [ 70%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-5-2-1-False] PASSED [ 71%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-5-2-2-True] PASSED [ 73%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-4-5-2-2-False] PASSED [ 75%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-4-1-1-True] PASSED [ 76%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-4-1-1-False] PASSED [ 78%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-4-1-2-True] FAILED [ 79%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-4-1-2-False] FAILED [ 81%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-4-2-1-True] FAILED [ 82%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-4-2-1-False] FAILED [ 84%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-4-2-2-True] FAILED [ 85%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-4-2-2-False] FAILED [ 87%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-5-1-1-True] PASSED [ 89%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-5-1-1-False] PASSED [ 90%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-5-1-2-True] PASSED [ 92%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-5-1-2-False] PASSED [ 93%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-5-2-1-True] FAILED [ 95%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-5-2-1-False] FAILED [ 96%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-5-2-2-True] FAILED [ 98%]
nengo/builder/tests/test_transforms.py::test_convinc_2d[valid-5-5-2-2-False] FAILED [100%]
```

hunse:
Fixed the Python2 errors. We were relying on Python3 division.

tbekolay:
> Fixed the Python2 errors. We were relying on Python3 division.

Thanks! Working now.

nengo/transforms.py
```diff
@@ -42,8 +44,7 @@ class Transform(FrozenObject):
     """A base class for connection transforms."""
 
     def sample(self, rng=np.random):
-        """
-        Returns concrete weights needed to implement the specific transform.
+        """Returns concrete weights to implement the specified transform.
```
drasmuss: I was actually using some automated parser at some point that didn't parse the docstrings properly if they didn't start on the next line (which is why I started doing it that way).  I don't remember what that parser was in order to demonstrate though 🤷‍♂️ .

nengo/transforms.py
```diff
@@ -172,9 +174,11 @@ def sample(self, rng=np.random):
             kernel = []
```
drasmuss: Can remove this line with the list comprehension

nengo/transforms.py
tbekolay: Hm, interesting. There's a lot of docstrings we'll have to change if that's the case so I say we keep it consistent for now and if we find a tool that doesn't parse it correctly we can switch.

nengo/transforms.py
tbekolay: Yep, I thought I had removed it but you know how rebasing be

tbekolay:
I switched to the newline for each argument convention after hearing some PyCon talk that advocated for it, and since changing I've liked it a lot, because it

1. results in smaller diffs (usually)
2. reduces decisions (i.e. no "which arg do I linebreak after?" choice)
3. makes it easier to see how many args a callable takes, and makes all the kwargs obvious as well
4. is more similar to the conventions that are typical in JS/Typescript, which I have sort of grown to appreciate

One thing that I've been meaning to look into is an automated Python code formatter, [black](https://github.com/ambv/black). I forget which project I saw it used in, but it's got a bunch of stars and would mean that we can stop thinking about how to make these kinds of decisions and just trust the tool.

All this for separate PRs though. I can remove the `kernel = []` line in the merge if you want, just lmk @drasmuss

drasmuss:
I just find it kind of hard to tell which variables go with what in blocks like this

https://github.com/nengo/nengo/pull/1481/commits/df09b81f534688495da2dfbb579e2a8b22b225cb#diff-b417881310f8e7370c44d7f0094e60c6R55

https://github.com/nengo/nengo/pull/1481/commits/df09b81f534688495da2dfbb579e2a8b22b225cb#diff-b417881310f8e7370c44d7f0094e60c6R61

I have to do a lot of mental bracket parsing, as opposed to things being organized by white space.  But it could just be because that's not a style I'm used to looking at 🤷‍♂️ .

Anyway, all looks good to me, fixing the `[]` in the merge sounds good.

hunse:
In that first diff Dan just posted, shouldn't the line break be after `x0,` anyway? That's the first argument. Rather than having the line break at the start of the list.

In general, I think we need to make a decision on how we're doing style stuff. In my opinion, the two possible routes are:
1. Have a general style guideline (like PEP8), and whatever obeys those rules is okay. If somebody else writes code that you don't like the look of, but is still fine according to PEP8, don't touch it, since it's there choice as the author of the code how to do it. This is most of an issue when the code is already in `master` and the only reason you're touching it is style, because this results in more difficult rebases because more lines are touched. If the code is not in `master` (such as here), or if you're making other changes to those lines and want to "fix" a few style things, then I think it's less of a problem to make a change. However, I still think it's important to have some respect for other people's style choices, at the very least to avoid flip-flopping where one person likes it one way, another person changes it, the first person changes it back in the future, etc.
2. Have a rigid style guideline that has rules for everything what we can apply blindly, and allows no developer freedom. This is more along the lines of what Trevor was suggesting, and would work even better if we have a tool to help enforce it like he was suggesting. Initially, there'll be a mess of changes as we fix everything to conform with guidelines, but once we do, then it's more straightforward to maintain. There also might be arguments about the best way to do things initially, but once we've decided that, then it's set in stone and there's no more discussion.
3. The third of two options is to start with 1 and slowly move towards 2, introducing new rules as situations come up. When a new rule is introduced, we could either apply it across the whole repository, or only apply it as code that does not conform is revisited for other reasons.

Of course, we'll never take all the developer choice out of developing, nor would we want to. Something like variable names is a good example: there's a ton of choice there. I think that's actually a place where we do very well balancing respect but also allowing for improvement: normally, you won't go and change somebody else's variable names, but sometimes a situation comes up where things aren't clear and you can provide a new more descriptive name.

Anyway, I'm not sure which way I'm leaning on this. I'm bringing it up, though, because I feel like right now we're in a kind of middle ground, and there have been some cases where it's annoyed me. If there are multiple acceptable ways of doing something, then I'd like others to respect my choice. That said, some of the "variable name" situation does apply: sometimes, you're in a situation to make something significantly and relatively unambiguously more readable by making e.g. a linebreak change. The trick is, there are sometimes different opinions on what is more readable, and what might seem unambiguously better to one person is not to another.

tbekolay:
> I just find it kind of hard to tell which variables go with what in blocks like this

I think those two blocks in particular are hard to format no matter what way you go ... in both versions I have to stare at it for a while. I find my version easier to parse, but yeah it's most likely just because it's what I'm used to now. It would be good to keep these examples in mind if we do want to make more rules about style.

> I think we need to make a decision on how we're doing style stuff

I don't disagree, but I'm also not sure how high a priority it is. Finding out about the `black` tool, though, piqued my interest because it could save us time in the long run not having to make style decisions. To a certain extent I stopped worrying about this once we changed the dev process to put more onus on the maintainer (to do all merges and whatnot) because the codebase is small enough that I would consider myself the owner of most of it (in the sense of providing support for it and maintaining it moving forward, not in the sense of authorship), and so as the owner I feel like I am free to change style because I, generally, will be interacting with that code the most going forward. Obviously there are a bunch of assumptions in there that aren't shared by other people, so definitely open to discussion on that (in a separate issue / PR, methinks).

> I still think it's important to have some respect for other people's style choices

Again I don't disagree, and hopefully I haven't come off as disrespectful when making style choices. I do think a lot about it and am not making changes arbitrarily or just to make sure I touch everything. I think as a project scales up and gets more stable you should expect that styles across files is going to get more consistent. And yeah, in case it's not clear, I'm only talking about style in this repo specifically here, I feel like many of the overarching decisions in terms of style and whatnot are made by the maintainer. So I would expect that in NengoDL I would generally only see Dan's line break conventions. I would try to follow that if I made a PR on NengoDL, but if I slip and Dan changes it, that feels fine to me.

> there have been some cases where it's annoyed me

Please let me know when that happens, as if you don't let me know then I will keep assuming that everyone's content with the current state of things. I do always try to leave at least a half day or so for people to review significant style changes before I do merges, though I understand that it's usually not the kind of thing that is important enough to slow down progress, but important enough that it's annoying. I feel like Dan saying "I preferred my way but go ahead" is good, it means that we'll think about this more in the next little while but it doesn't block this PR.

> what might seem unambiguously better to one person is not to another

Yep for sure ... there's not really any way around that, so I was going on the assumption that the primary maintainer's taste trumps the rest. That'll be different on different repos (with different maintainers). Part of why I favor this is that actually writing down all of our preferences for things (code style, git commit message formatting, etc etc) is pretty time consuming, whereas enforcing those things in merges is something I'm already doing and means I don't have to burden everyone with my neuroses. Maybe it also provides incentives to people to start maintaining projects so they can wield their iron fist :smirk:

---

Anyhow this is truly a digression, so I made #1491 to continue the discussion. For now, I'll fix that `kernel` list, squash the history down, and merge.

PR #1482, opened 2018-10-31
Numba LIF model reduces simulation time by 30% or more 

arvoelke:
**Motivation and context:**
The neuron model is one of the inner-most loops within the reference backend. Speeding this code up with JIT compilation would offer a significant reduction in simulation times to virtually all Nengo models using the default simulator. 

To this end, `nengo.neurons.NumbaLIF` works as a drop-in replacement for `nengo.LIF`. This new neuron model uses Numba's `nopython` just-in-time compilation feature to execute optimized machine code (instead of using the Python interpreter). See details: http://numba.pydata.org/numba-doc/latest/user/performance-tips.html.

**How has this been tested?**
 - `pip install numba`
 - `py.test nengo/tests/test_neurons.py::test_numbalif_benchmark --plots=plots --slow`
 - `xdg-open plots/test_neurons.test_numbalif_benchmark.pdf`

profile_numbalif

This is a work in progress. How should this be exposed to the user? If `numba` is already installed, should a factory substitute this in for `LIF`? Should the config system use this as the default if available?

**Types of changes:**

**Checklist:**

**Still to do:**
 - [x] Test for correctness / consistency
 - [ ] Test depends on `seaborn` and `pandas`
 - [ ] Extend to other models in Nengo, including synapse models
 - [x] Rerun existing Nengo test suite against `NumbaLIF`
 - [x] Refactor code to eliminate duplication of `LIF.step_math` code.
 - [x] ~~Move `custom_clip` to a utility file for numba-related functions?~~ Extended Numba to support `np.clip`

### DIFF
--- a/nengo/tests/test_neurons.py
+++ b/nengo/tests/test_neurons.py
@@ -565,3 +565,29 @@ def test_numbalif_benchmark(plt):
     plt.figure()
     sns.lineplot(data=DataFrame(data),
                  x="# Neurons", y="Time (ms)", hue="Neuron Type")
+
+
+@pytest.mark.parametrize("dt", [0.0005, 0.001, 0.002])
+def test_numbalif_correctness(Simulator, seed, dt):
+    # Test spike-trains for numerical precision, using a
+    # sine wave covering one period across the simulation time
+    n_neurons = 500
+    sim_t = 2.0
+    with nengo.Network() as model:
+        u = nengo.Node(output=lambda t: np.sin(2*np.pi*t/sim_t))
+
+        x_numba = nengo.Ensemble(n_neurons, 1, seed=seed,
+                                 neuron_type=NumbaLIF())
+        x_lif = nengo.Ensemble(n_neurons, 1, seed=seed,
+                               neuron_type=nengo.LIF())
+
+        nengo.Connection(u, x_numba, synapse=None)
+        nengo.Connection(u, x_lif, synapse=None)
+
+        p_numba = nengo.Probe(x_numba.neurons, 'spikes', synapse=None)
+        p_lif = nengo.Probe(x_lif.neurons, 'spikes', synapse=None)
+
+    with Simulator(model, dt=dt) as sim:
+        sim.run(sim_t)
+
+    assert np.allclose(sim.data[p_numba], sim.data[p_lif])
--- a/nengo/tests/test_neurons.py
+++ b/nengo/tests/test_neurons.py
@@ -1,9 +1,12 @@
+import timeit
+from collections import defaultdict
+
 import numpy as np
 import pytest
 
 import nengo
 from nengo.exceptions import BuildError, SimulationError
-from nengo.neurons import Direct, NeuronTypeParam
+from nengo.neurons import Direct, NeuronTypeParam, NumbaLIF
 from nengo.processes import WhiteSignal
 from nengo.solvers import LstsqL2nz
 from nengo.utils.ensemble import tuning_curves
@@ -528,3 +531,37 @@ def test_gain_bias(rng, nl_nodirect, generic):
 
     assert np.allclose(max_rates, max_rates0, atol=tolerance)
     assert np.allclose(intercepts, intercepts0, atol=tolerance)
+
+
+def _benchmark(neuron_type, n, d=1, sim_time=0.5, n_trials=10):
+    """Estimates number of milliseconds per simulation time-step."""
+
+    with nengo.Network() as model:
+        nengo.Ensemble(n_neurons=n, dimensions=d, neuron_type=neuron_type)
+
+    with nengo.Simulator(model, progress_bar=False) as sim:
+        sim.step()  # invokes the JIT compilation (cached from here on)
+
+        def to_time():  # timed portion of code
+            sim.run(sim_time, progress_bar=False)
+
+        n_steps = float(sim_time) / sim.dt
+        return 1e3 * timeit.timeit(to_time, number=n_trials) / float(n_trials) / n_steps
+
+
+@pytest.mark.slow
+@pytest.mark.noassertions
+def test_numbalif_benchmark(plt):
+    import seaborn as sns
+    from pandas import DataFrame
+
+    data = defaultdict(list)
+    for neuron_type in (NumbaLIF(), nengo.LIF()):
+        for n in np.linspace(1, 10000, 11, dtype=int):
+            data['Neuron Type'].append(type(neuron_type).__name__)
+            data['# Neurons'].append(n)
+            data['Time (ms)'].append(_benchmark(neuron_type, n))
+
+    plt.figure()
+    sns.lineplot(data=DataFrame(data),
+                 x="# Neurons", y="Time (ms)", hue="Neuron Type")
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -4,7 +4,9 @@ import logging
 import warnings
 
 import numpy as np
+from numba import njit
 
+import nengo.utils.numba  # triggers np.clip support
 from nengo.exceptions import SimulationError, ValidationError
 from nengo.params import Parameter, NumberParam, FrozenObject
 from nengo.utils.compat import range
@@ -424,6 +426,37 @@ class LIFRate(NeuronType):
         # (nan > 0 -> error), and not pass x < -1 to log1p
 
 
+def _lif_step_math(dt, J, spiked, voltage, refractory_time,
+                   tau_rc, tau_ref, min_voltage, amplitude):
+    # reduce all refractory times by dt
+    refractory_time -= dt
+
+    # compute effective dt for each neuron, based on remaining time.
+    # note that refractory times that have completed midway into this
+    # timestep will be given a partial timestep, and moreover these will
+    # be subtracted to zero at the next timestep (or reset by a spike)
+    delta_t = np.clip(dt - refractory_time, 0, dt)
+
+    # update voltage using discretized lowpass filter
+    # since v(t) = v(0) + (J - v(0))*(1 - exp(-t/tau)) assuming
+    # J is constant over the interval [t, t + dt)
+    voltage -= (J - voltage) * np.expm1(-delta_t / tau_rc)
+
+    # determine which neurons spiked (set them to 1/dt, else 0)
+    spiked_mask = voltage > 1
+    spiked[:] = spiked_mask * (amplitude / dt)
+
+    # set v(0) = 1 and solve for t to compute the spike time
+    t_spike = dt + tau_rc * np.log1p(
+        -(voltage[spiked_mask] - 1) / (J[spiked_mask] - 1))
+
+    # set spiked voltages to zero, refractory times to tau_ref, and
+    # rectify negative voltages to a floor of min_voltage
+    voltage[voltage < min_voltage] = min_voltage
+    voltage[spiked_mask] = 0
+    refractory_time[spiked_mask] = tau_ref + t_spike
+
+
 class LIF(LIFRate):
     """Spiking version of the leaky integrate-and-fire (LIF) neuron model.
 
@@ -447,39 +480,39 @@ class LIF(LIFRate):
 
     min_voltage = NumberParam('min_voltage', high=0)
 
+    _static_step_math = staticmethod(_lif_step_math)
+
     def __init__(self, tau_rc=0.02, tau_ref=0.002, min_voltage=0, amplitude=1):
         super(LIF, self).__init__(
             tau_rc=tau_rc, tau_ref=tau_ref, amplitude=amplitude)
         self.min_voltage = min_voltage
 
     def step_math(self, dt, J, spiked, voltage, refractory_time):
-        # reduce all refractory times by dt
-        refractory_time -= dt
+        return self._static_step_math(
+            dt, J, spiked, voltage, refractory_time,
+            self.tau_rc, self.tau_ref, self.min_voltage, self.amplitude)
 
-        # compute effective dt for each neuron, based on remaining time.
-        # note that refractory times that have completed midway into this
-        # timestep will be given a partial timestep, and moreover these will
-        # be subtracted to zero at the next timestep (or reset by a spike)
-        delta_t = (dt - refractory_time).clip(0, dt)
 
-        # update voltage using discretized lowpass filter
-        # since v(t) = v(0) + (J - v(0))*(1 - exp(-t/tau)) assuming
-        # J is constant over the interval [t, t + dt)
-        voltage -= (J - voltage) * np.expm1(-delta_t / self.tau_rc)
+class NumbaLIF(LIF):
+    """Numba-compiled version of the LIF model.
 
-        # determine which neurons spiked (set them to 1/dt, else 0)
-        spiked_mask = voltage > 1
-        spiked[:] = spiked_mask * (self.amplitude / dt)
+    Parameters
+    ----------
+    tau_rc : float
+        Membrane RC time constant, in seconds. Affects how quickly the membrane
+        voltage decays to zero in the absence of input (larger = slower decay).
+    tau_ref : float
+        Absolute refractory period, in seconds. This is how long the
+        membrane voltage is held at zero after a spike.
+    min_voltage : float
+        Minimum value for the membrane voltage. If ``-np.inf``, the voltage
+        is never clipped.
+    amplitude : float
+        Scaling factor on the neuron output. Corresponds to the relative
+        amplitude of the output spikes of the neuron.
+    """
 
-        # set v(0) = 1 and solve for t to compute the spike time
-        t_spike = dt + self.tau_rc * np.log1p(
-            -(voltage[spiked_mask] - 1) / (J[spiked_mask] - 1))
-
-        # set spiked voltages to zero, refractory times to tau_ref, and
-        # rectify negative voltages to a floor of min_voltage
-        voltage[voltage < self.min_voltage] = self.min_voltage
-        voltage[spiked_mask] = 0
-        refractory_time[spiked_mask] = self.tau_ref + t_spike
+    _static_step_math = staticmethod(njit(_lif_step_math))
 
 
 class AdaptiveLIFRate(LIFRate):
--- a/nengo/utils/__init__.py
+++ b/nengo/utils/__init__.py
@@ -12,6 +12,7 @@ from . import magic
 from . import nco
 from . import network
 from . import neurons
+from . import numba
 from . import numpy
 from . import paths
 from . import probe
--- a/nengo/utils/numba.py
+++ b/nengo/utils/numba.py
@@ -1,0 +1,27 @@
+"""
+Extra functions to extend the capabilities of Numba.
+"""
+
+from __future__ import absolute_import
+
+import numpy as np
+
+from numba.extending import overload
+
+
+@overload(np.clip)
+def np_clip(a, a_min, a_max):
+    """Numba-implementation of np.clip."""
+    # Does not support `out` argument, optional arguments, nor a.clip
+    # https://github.com/numba/numba/pull/3468
+    def np_clip_impl(a, a_min, a_max):
+        out = np.empty_like(a)
+        for index, val in np.ndenumerate(a):
+            if val < a_min:
+                out[index] = a_min
+            elif val > a_max:
+                out[index] = a_max
+            else:
+                out[index] = val
+        return out
+    return np_clip_impl


arvoelke:
FYI, I tried speeding up the `Cholesky` sub-solver using Numba, in three different ways, none of which worked:
 - Apply `@njit` to:
    ```python
    L = np.linalg.cholesky(G)
    L = np.linalg.inv(L.T)
    X = np.dot(L, np.dot(L.T, b))
    ```
    This ended up being slower than `scipy`.

 - Apply `@njit` to:
    ```python
    factor = scipy.linalg.cho_factor(G, overwrite_a=True)
    X = scipy.linalg.cho_solve(factor, b)
    ```
    This gave an unsupported method error.

 - Apply `@jit` to the previous code.
    This gave no noticeable improvement for large systems.

Benchmarking snippet for future reference:
```python
import nengo
from nengo.utils.least_squares_solvers import Cholesky, NumbaCholesky
from nengo.solvers import LstsqL2

import numpy as np

N = 3000
M = 3000
D = 256

rng = np.random.RandomState(seed=0)
A = rng.randn(M, N)
Y = rng.randn(M, D)

for subsolver in (Cholesky(), NumbaCholesky()):
    solver = LstsqL2(solver=subsolver)
    d, info = solver(A, Y)
    print(type(subsolver).__name__, info['time'])
```

arvoelke:
Also, I tried to deduplicate the code between `LIF` and `NumbaLIF`, by allowing the custom `clip` method to be provided as an additional argument to the static `_lif_step_math` method. However, it seems that passing a function pointer to a numba-compiled function has a significant associated cost. Specifically, it halved the `NumbaLIF` improvement at 10,000 neurons, without changing the time for `LIF`.

nengo/tests/test_neurons.py
```diff
@@ -528,3 +531,36 @@ def test_gain_bias(rng, nl_nodirect, generic):
...
+
+        def to_time():  # timed portion of code
+            sim.run(sim_time, progress_bar=False)
+
+        n_steps = float(sim_time) / sim.dt
+        return 1e3 * timeit.timeit(to_time, number=n_trials) / float(n_trials) / n_steps
+
+
+@pytest.mark.slow
+def test_numbalif_benchmark(plt):
```
jgosmann: Should be marked with @pytest.mark.noassertions
```suggestion
@pytest.mark.noassertions
def test_numbalif_benchmark(plt):
```

jgosmann:
> How should this be exposed to the user? If `numba` is already installed, should a factory substitute this in for `LIF`?

Is it possible to substitute the `numba` implentation if the `numba` module can be imported and otherwise use the default implementation?

arvoelke:
> Is it possible to substitute the numba implentation if the numba module can be imported and otherwise use the default implementation?

Should be doable. But wondering if there is any reason to allow the standard numpy implementation to still be accessible, even when `numba` is installed? And how should this work for inheritance (e.g., `AdaptiveLIF` calls `LIF.step_math`)? I also want to figure out what everyone would be happy with before applying this to the other neuron models.

tbekolay:
Very cool! And would be even cooler if we can apply it to more aspects of the sim!

> want to figure out what everyone would be happy with

Do you have a few possible ways forward in mind that we can vote on? I can think of a few possibilities but figured that you have more of a sense of what's possible having implemented it.

arvoelke:
With some help, I figured out the correct way to extend Numba to support the missing `np.clip` implementation. I pushed this as a PR to Numba: https://github.com/numba/numba/pull/3468 and included the relevant (self-contained) portion in this PR to avoid depending on some future Numba release. With this utility file, it basically becomes a one-line change to make the LIF model Numba-compatible.

Note that `numba` depends on:
 - `llvmlite` (pip install)
 - `numpy>=1.9`

I also tried changing `LIF` entirely over to the Numba implementation, and re-ran the test suite locally. All tests passed except for `nengo/tests/test_neurons.py::test_lif_builtin`, because this uses a 2D-array for the current `J`. See #1419 and #1437 for relevant documentation fixes / changes that might affect this test. This helps give some assurances that the implementation is up to spec.

Then I added another test to this PR that compares the spike-trains of `LIF` against `NumbaLIF` for numerical precision. This new test passes on my machine, which gives more supporting evidence that the implementation is correct.

arvoelke:
> Do you have a few possible ways forward in mind that we can vote on? I can think of a few possibilities but figured that you have more of a sense of what's possible having implemented it.

Here's my current thinking:

1. Move this to `nengo_extras`. This would mean the user must make the effort to install `numba`, `nengo_extras`, and then switch instances of `nengo.LIF()` over to `nengo_extras.NumbaLIF()` either on a per-ensemble basis or at the config level.

2. Move this to `nengolib`. This would be the same as option 1, except I could provide the support in terms of documentation / testing / maintenance, and I can also make it the default neuron model (if `numba` is installed) via [`nengolib.Network()`](https://github.com/arvoelke/nengolib/blob/master/nengolib/network.py). 

3. Keep this in `nengo` as a separate neuron model, as the PR currently stands. This would be similar to option 1, but could improve user-discovery and reduce the amount of work needed to install (one fewer package).

4. Replace `nengo.LIF()` with `nengo.NumbaLIF()` and force users to install `numba` by adding it as a pip dependency. If they can't install it, then they can't use the reference simulator. I don't expect people to go for this option, but including it, because, hey, options. Although this does have the benefit of reducing the surface area for maintenance / testing / debugging / etc. It also forces discovery of the improvement.

5. Replace `nengo.LIF()` with `nengo.NumbaLIF()` if `numba` is already installed, and otherwise fall back to the standard implementation. Assuming the implementation is okay, users may experience a free and automatic speed-up without having to do much. However, they would need to be made aware of the suggestion to install `numba`. We may also want to expose some flag that reveals whether or not the switch was made.

6. Replace `nengo.LIF()` with `nengo.NumbaLIF()` if `numba` is already installed _and_ some flag is set to True in the config. This flag could default to either value, e.g., perhaps defaulting to False for the first couple of releases, and then switched over to True if there prove to be no serious issues. This has the benefits of option 5, while allowing a user to turn it off for whatever reason, even if `numba` is still installed.

tcstewar:
I would lean towards 1 or 2 right now (no strong preference between them), but with an eye to eventually go to option 5.  I could also see going straight to 5.  I'm not too worried about the problem of letting the users know that they could get a speedup if they install numba -- we currently have exactly that same situation with letting people know that if they install an optimized numpy, then things will go much faster.  We should have a "how to make your models go faster" document somewhere and include both of those two things as suggestions.

jgosmann:
1. + 2. For any serious large scale models, one would want the most performant install. Each additional step to achieve this is an annoyance to me.

2. How would this work together with `nengo_spa.Network`?

4. I assume this might require a working C compiler? This would be quite a departure from not having to worry about such things once Numpy is running.

5. would be the option I favour. We could consider giving a warning if Numba isn't installed (we do the same with Scipy in some cases which gives better accuracy in the Basal Ganglia and gives additional speed ups with the optimizer). The optimized Numpy mentioned by @tcstewar is a slightly different case because afaik we cannot detect whether the installed Numpy is optimized.

6. My first reaction is to better not introduce a new config option. But that might be only because I'm currently dealing with a system at work where pretty much everything is configurable and it turns out to be pretty annoying.

arvoelke:
> * 2 How would this work together with `nengo_spa.Network`?

After a quick look, they should be compatible with each other, in the sense that you could subclass a network from both of them. All my network does is change a couple of defaults. Maybe a better solution is needed if people want to do this though.

> * 4 I assume this might require a working C compiler? This would be quite a departure from not having to worry about such things once Numpy is running.

My understanding is you actually don't! If you use `pip` or `conda` then you get a pre-built binary where `llvmlite` has been statically linked to the required subset of the LLVM compiler. Only if you want to build the dev branch of `llvmlite` do you have to worry about having a compiler, AFAIK. `llvmlite` is the only dependency other than `numpy>=1.9`.

jgosmann:
With this information 4 seems viable to me too give thorough prior testing on as much different systems as we can reasonably do.

hunse:
I'm surprised too. Getting `numba` installed seems much easier than I expected.

That said, I'm still wary of introducing another required dependency. While a lot of people will want the speed, for others (perhaps interested in trying some tutorials or running some small models to get started) speed is not so important. The other thing to remember is that while the plot at the top shows a 25-30% speedup in the neuron model, this is in the neuron model only. For many (most?) models, the majority of the computation is spent on multiplying weights, so the overall speedup might be more like 10%.

I think my long term ideal is option 6. I like having the choice to turn it off, if only for debugging and development. I don't think adding an extra config option is a big burden.

Short-term, I'm not sure if we need to use 1 or 2 as a stepping stone or not. It would be good to test this out on some other models and with other backends. My worry with 1 or 2 is that we'll put it there but none of us will really test it. I think I'd rather go with option 6 but have it default to disabled as Aaron suggested. All of us can then set our config files to use it (we should perhaps also have a print statement just to remind/ensure that it's being used). Once it's been in use on our end for a bit, then we can default it to enabled.

arvoelke:
> The other thing to remember is that while the plot at the top shows a 25-30% speedup in the neuron model, this is in the neuron model only. For many (most?) models, the majority of the computation is spent on multiplying weights, so the overall speedup might be more like 10%.

You are right that I haven't done enough here to explore the impact outside of a model that contains one ensemble and no connections. Below, I've tried something a bit more thorough by running the complete suite of [nengo-benchmarks](https://github.com/nengo/nengo-benchmarks). Each benchmark, except for "Parsing" (requires investigation), and "MNIST" (raises an error), performed 7-48% better in simulation time averaged across 10 trials (although there is considerable variability). Lorenz had the best improvement at 48%.

The mean across all benchmarks is 16% -- pretty consistent with your guess of 10%. I'm optimistic that adding JIT support to other important computations such as dot-products and elementwise-products will make further improvements, consistent with the intuition that the majority of computation is spent on multiplying weights.

benchmark_individual

benchmark_combined

```python
from collections import defaultdict

import nengo
import nengo_benchmarks

import numpy as np
import seaborn as sns
from pandas import DataFrame

benchmarks = [
    nengo_benchmarks.CommunicationChannel,
    nengo_benchmarks.ConvolutionCleanup,
    nengo_benchmarks.CircularConvolution,
    nengo_benchmarks.InhibitionTrial,
    nengo_benchmarks.LearningSpeedup,
    nengo_benchmarks.Lorenz,
    nengo_benchmarks.MatrixMultiply,
    nengo_benchmarks.SemanticMemory,
    nengo_benchmarks.SemanticMemoryWithRecall,
    # nengo_benchmarks.MNIST,
    nengo_benchmarks.Oscillator,
    nengo_benchmarks.Parsing,
    nengo_benchmarks.SPASequence,
    nengo_benchmarks.SPASequenceRouted,
]

data = defaultdict(list)

kwargs = {
    'verbose': False,
}

for benchmark in benchmarks:
    for trial in range(10):
        result_ref = benchmark().run(**kwargs)
        result_numba = benchmark().run(neuron_type=nengo.neurons.NumbaLIF(), **kwargs)

        name = benchmark.__name__
        print(name, trial)

        data['Benchmark'].append(name)
        data['Trial'].append(trial)
        # speed is inversely proportional to time, so this computes numba's improvement
        data['Improvement'].append(result_numba['speed'] / result_ref['speed'])

df = DataFrame(data)

plt.figure(figsize=(16, 6))
plt.hlines(1, -1, len(benchmarks), linestyle='--')
sns.boxplot(data=df, x='Benchmark', y='Improvement')
plt.xticks(rotation=50)
plt.show()

avg = np.mean(df['Improvement'])

plt.figure()
sns.kdeplot(df['Improvement'], shade=True)
plt.vlines([avg], 0, 1, linestyle='--')
plt.text(avg, 1.1, "%.2f" % avg, horizontalalignment='center')
plt.xlabel("Improvement")
plt.ylabel("Density")
plt.show()

for benchmark in df["Benchmark"].unique():
    print(benchmark, np.mean(df[df['Benchmark'] == benchmark]['Improvement']))
```

```
...
('CommunicationChannel', 1.0711993000693962)
('ConvolutionCleanup', 1.083407656191113)
('CircularConvolution', 1.1739388343539425)
('InhibitionTrial', 1.2655979857474466)
('LearningSpeedup', 1.1582004179285226)
('Lorenz', 1.4812901392236884)
('MatrixMultiply', 1.15535191082224)
('SemanticMemory', 1.171235946916829)
('SemanticMemoryWithRecall', 1.1197507715888813)
('Oscillator', 1.265969636887784)
('Parsing', 0.8497689418473513)
('SPASequence', 1.1256255668578001)
('SPASequenceRouted', 1.1679054635787818)
```

hunse:
Dot products are already done by a highly-optimized BLAS library though, assuming you've set up your Numpy correctly. The main way I can see saving there is doing the things that I think the optimizer already does, namely grouping the operations together so that you need fewer calls. Elementwise multiply might be slightly easier for the JIT compiler to optimize, but again it's essentially one Numpy call per Operator, so the overhead isn't huge. I think what makes neuron models such a good target is that there's a number of very simple Numpy calls within one neuron model (e.g. logs, inverses, multiplies, indexing) making for a lot of overhead that the JIT compiler can optimize out. Anyway, I'm not saying it's not worth trying the other operations, I'm just managing expectations.

It is interesting that some models (like Lorenz) have such big improvements. That's an even bigger improvement than you saw with a raw population of neurons (I assume that's what's in `test_numbalif_benchmark`), which is surprising.

xchoo:
Commenting from the view of building large models (i.e., spaun), with the `nengo_ocl` backend, simulation time actually is not an issue compared to the time it takes to actually construct the model. That being the case, I'd favour the solution that prefers a slimmer nengo install (i.e., options 1 & 2)

jgosmann:
> Commenting from the view of building large models (i.e., spaun), with the `nengo_ocl` backend, simulation time actually is not an issue compared to the time it takes to actually construct the model.

It can be in issue if your forced to use the reference backend, e.g. because of learning rules not implemented in Nengo OCL.

arvoelke:
> afaik we cannot detect whether the installed Numpy is optimized

For what it's worth, you can check whether numpy is using OpenBLAS via `np.__config__.show()` or `np.__config__.openblas_info` (https://github.com/numpy/numpy/issues/3912).

> Dot products are already done by a highly-optimized BLAS library though, assuming you've set up your Numpy correctly

In my experiment below, I found that Numba is 70% faster at dot-products than [Numpy with OpenBLAS](https://hunseblog.wordpress.com/2014/09/15/installing-numpy-and-openblas). `np.__config__.openblas_info` returns information on `openblas`, while  `np.__config__.show()` does not show any info on atlas, and so I think I do have it properly installed and configured. I remember this being relatively difficult to do, and think it would be much easier for an average person to `pip install numba`. And I could only find the instructions linked from the original `nengo==2.0.0` pypi release readme: https://pypi.org/project/nengo/2.0.0/. I could not find instructions elsewhere in the Nengo documentation.

```python
import time
from numba import njit

def np_dot(A, B):
    return np.dot(A, B)

@njit
def numba_dot(A, B):
    return np.dot(A, B)

trials = 500
n = 1000
d = 10
m = 5000

A = np.random.randn(n, d)
B = np.random.randn(d, m)

print(np.__config__.openblas_info)
for func in (np_dot, numba_dot):
    t = time.time()
    for i in range(trials):
        func(A, B)
    td = time.time() - t
    print(func.__name__, td)
```

```
{'define_macros': [('HAVE_CBLAS', None)],
 'language': 'c',
 'libraries': ['openblas', 'openblas'],
 'library_dirs': ['/usr/local/lib']}
('np_dot', 14.827863931655884)
('numba_dot', 8.773243188858032)
```

If we can optimize away any loops within the reference simulator there should be more gains. Although I am running into some challenges elsewhere.

> It is interesting that some models (like Lorenz) have such big improvements. That's an even bigger improvement than you saw with a raw population of neurons (I assume that's what's in test_numbalif_benchmark), which is surprising.

For the plot at the very top, 30% was the worst-case improvement. For `n=2000` neurons, which is how many the Lorenz benchmark uses, the improvement is closer to 50%. This also makes sense considering the Lorenz benchmark is a single ensemble with one connection and a probe.

> Commenting from the view of building & running large models (i.e., spaun), with the nengo_ocl backend, simulation time actually is not an issue compared to the time it takes to actually construct the model (30s simulation vs 45min build time). That being the case, I'd favour the solution that prefers a slimmer nengo install (i.e., options 1 & 2).

From this perspective, supposing Numba can be successfully applied to core parts of the builder, such as sampling, collection of tuning curves, solving for decoders, etc., then options 1 and 2 would in fact prevent the builder from having access to Numba and these potential improvements in build time. Would this change your vote? If build time is the big win for Spaun, I can try to prove out some improvements on the builder side before taking this to a vote?

drasmuss:
> And I could only find the instructions linked from the original nengo==2.0.0 pypi release readme: https://pypi.org/project/nengo/2.0.0/. I could not find instructions elsewhere in the Nengo documentation.

The instructions are here, just for anyone else looking for this information in the future :smile:
https://www.nengo.ai/nengo/getting_started.html#installing-numpy

arvoelke:
> The instructions are here, just for anyone else looking for this information in the future 
> https://www.nengo.ai/nengo/getting_started.html#installing-numpy

Ah, thank you. Specifically where it says "If speed is an issue and you know your way around a terminal, installing NumPy from source is flexible and performant. See the detailed instructions [here]." I was looking for keywords such as "BLAS", "ATLAS", "LAPACK", or Fortran.

hunse:
I tried that numba dot script on my machine, and got the following:
```
{'define_macros': [('HAVE_CBLAS', None)], 'library_dirs': ['/usr/local/lib'], 'libraries': ['openblas', 'openblas'], 'language': 'c'}
np_dot 16.1305673122406
numba_dot 19.067010164260864
```
I think that Numpy with some sort of BLAS is much easier to install now than in the past. I've got the OpenBLAS binaries from `apt-get` installed, and if I do `pip install numpy` in a fresh virtualenv it detects them and uses them.

I'm actually finding the opposite for `numba`. When I try to pip install it in a fresh virtualenv, it's not finding the pre-build wheels for some reason and is trying to build them, and fails because it can't find `llvm-config`. So I wasn't actually successful in comparing them in a fresh virtualenv, which is what I set out to do.

arvoelke:
There could certainly be some differences from machine to machine. I encourage others to also run both the benchmark test in this branch (see first post) and the script above (https://github.com/nengo/nengo/pull/1482#issuecomment-438761188). Here's details of my architecture given by the `lscpu` command on Ubuntu 16.04 LTS:
```
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                8
On-line CPU(s) list:   0-7
Thread(s) per core:    2
Core(s) per socket:    4
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 60
Model name:            Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz
Stepping:              3
CPU MHz:               3265.195
CPU max MHz:           3900.0000
CPU min MHz:           800.0000
BogoMIPS:              6784.94
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              8192K
NUMA node0 CPU(s):     0-7
Flags:                 fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm epb invpcid_single ssbd ibrs ibpb stibp kaiser tpr_shadow vnmi flexpriority ept vpid fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid xsaveopt dtherm ida arat pln pts flush_l1d
```

I tried running the script from a fresh environment, via:
```
conda create --name py36 python=3.6
conda activate py36
pip install numpy scipy numba
python -c "<copy-script-here>"
```
and got the same results as before. One thing I discovered though, is that without the `scipy` install this produces the error `scipy 0.16+ is required for linear algebra`, which curiously only seems to be documented in a "Note" here: http://numba.pydata.org/numba-doc/latest/reference/numpysupported.html#linear-algebra. This is only true for the above script which uses `np.dot` (not for the `njit`'d code in this branch, which uses only standard ufuncs and array indexing).

xchoo:
If you do `conda install numpy` instead of `pip install numpy` conda will also install prebuilt mkl packages that should improve the run time of numpy code.

arvoelke:
> If you do `conda install numpy` instead of `pip install numpy` conda will also install prebuilt mkl packages that should improve the run time of numpy code.

Wow! Using `mkl` instead of `openblas` gives me way better results for both `numba` and `numpy`, while providing an even larger relative Numba speedup -- the dot-product script is over 3.2x faster than `numpy`!

`mkl`:
```
np_dot 9.89888334274292
numba_dot 3.046924352645874
```
compared to before, using `openblas`:
```
np_dot 14.827863931655884
numba_dot 8.773243188858032
```

`np.__config__.show()`
```
mkl_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
blas_mkl_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
blas_opt_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
lapack_mkl_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
lapack_opt_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
```

arvoelke:
My workstation also has a 3.2x speedup with Numba on the dot script with `mkl`.

`lscpu`
```
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              12
On-line CPU(s) list: 0-11
Thread(s) per core:  2
Core(s) per socket:  6
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               63
Model name:          Intel(R) Xeon(R) CPU E5-1650 v3 @ 3.50GHz
Stepping:            2
CPU MHz:             1199.489
CPU max MHz:         3800.0000
CPU min MHz:         1200.0000
BogoMIPS:            6996.69
Virtualization:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            15360K
NUMA node0 CPU(s):   0-11
Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm cpuid_fault epb invpcid_single pti intel_ppin ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid cqm xsaveopt cqm_llc cqm_occup_llc dtherm ida arat pln pts flush_l1d
```

```
mkl_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
blas_mkl_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
blas_opt_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
lapack_mkl_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
lapack_opt_info:
    libraries = ['mkl_rt', 'pthread']
    library_dirs = ['/home/arvoelke/anaconda3/envs/temp/lib']
    define_macros = [('SCIPY_MKL_H', None), ('HAVE_CBLAS', None)]
    include_dirs = ['/home/arvoelke/anaconda3/envs/temp/include']
np_dot 7.823083162307739
numba_dot 2.4039652347564697
```

drasmuss:
After some internal discussion we decided to go with option 1 (moving this to `nengo_extras`), for now. See https://github.com/nengo/nengo-extras/pull/86. That doesn't preclude moving this back to nengo core in the future if we find ourselves using it a lot. nengo-extras just seemed like a more lightweight first step for numba integration, before we start adding numba logic (even if optional) into the core codebase.

PR #1488, opened 2018-11-08
Travis updates

drasmuss:
**Motivation and context:**

This adds some new quality of life stuff to our Travis builds.  

- c955c80 makes sure that we aren't accidentally installing a different version of numpy than we thought.  E.g., previously if we had set $NUMPY=1.3, but our internal nengo requirement was `numpy==1.10`, then we'd install `numpy==1.10` overtop of 1.3, giving us the mistaken impression that we were successfully testing 1.3.
- 8f79e5e changes the build environment to `minimal`.  Not that important, but means a slightly smaller docker image when spinning up a new job.
- 79996d9 activates TravisCI's caching functionality (so that we can reuse our python environment rather than building it from scratch this time).  However, testing this out shows that it doesn't offer a significant improvement to the build time (it did in NengoDL, but NengoDL has more dependencies to install).  And in the case of cache misses, it increases the build time.  I'll play around with it a little bit, but probably we don't end up using it.
- c5a6049 sets up automatic PyPI deployment for tags, with some error checking (so making new releases requires less manual effort).  It will also build test releases for branches named `release-candidate-x.y.z`; you can see an example here https://test.pypi.org/project/nengo/, generated by this build https://travis-ci.org/nengo/nengo/builds/452386601

**How has this been tested?**

Travis builds continue to run successfully.  Tested 79996d9 by rebuilding that commit several times, observing the build time.  No significant improvement, as noted.  Tested c5a6049 by doing the test release at https://test.pypi.org/project/nengo/.

Each commit is independent, so take them one at a time.

**Types of changes:**

**Checklist:**


drasmuss:
Improved the caching, so it does seem to provide a reliable ~10min improvement in total build time now (e.g., see https://travis-ci.org/nengo/nengo/builds/452486928).

.travis.yml
```diff
@@ -12,11 +15,13 @@ env:
...
 
-matrix:
+jobs:
   include:
-    - env: PYTHON="2.7" STATIC="true"
+    - stage: test
+      env: PYTHON="2.7" STATIC="true"
     - env: PYTHON="2.7" COVERAGE="true"
     - env: PYTHON="2.7" NUMPY="1.10"
+      cache: false  # disable the cache for one build to make sure that works
```
Seanny123: So, we're just going to leave this one build always without caching as a smoke-test?

drasmuss: Yeah that's the idea, just to make sure that there isn't some unanticipated issue that the caching is masking somehow.

hunse:
I took out the caching for now, and brought in most of the `nengo-loihi` changes.

I've preserved the version of this branch with caching in `travis-updates-cache`, but it doesn't have all the `nengo-loihi` changes in it.

hunse:
Just realized there's already a commit for the docs changes in cf363f254, so I'm going to pull that over here.

hunse:
This is re-doing some of 19b7143. Ideally, we'd rebase this onto that somehow to maintain the original history. One solution could be merging #1440 first, so that this is then based off that.

hunse:
Rebased this onto #1440, so that should be merged first.

.ci/static.sh
```diff
@@ -1,45 +1,51 @@
...
-    echo "  install  Install dependencies for running static checks"
-    echo "  run      Run static checks"
-    exit 1
-}
-
 if [[ "$COMMAND" == "install" ]]; then
-    conda install jupyter
-    pip install codespell flake8 pylint
-elif [[ "$COMMAND" == "run" ]]; then
+    # pip installs a more recent entrypoints version than conda
```
drasmuss: We should check if this is still necessary, or if conda has updated their `entrypoints` version now.

.travis.yml
```diff
@@ -9,29 +9,30 @@ cache:
...
-    - env: PYTHON="2.7" NUMPY="1.14"
-    - env: PYTHON="3.4" NUMPY="1.11"  # <=1.12 not packaged for py34 in conda
-    - env: PYTHON="3.5"
-    - env: PYTHON="3.6" COVERAGE="true" SCIPY="true"
-    - env: PYTHON="3.6" DOCS="true"
+    - env: MODE="test" PYTHON="2.7" NUMPY="1.12"
+    - env: MODE="test" PYTHON="2.7" NUMPY="1.13"
+    - env: MODE="test" PYTHON="2.7" NUMPY="1.14"
+    # <=1.12 not packaged for py34 in conda
+    - env: MODE="test" PYTHON="3.4" NUMPY="1.11"
```
drasmuss: This isn't new in this PR, but just noticed it.  Isn't our build here (using `numpy==1.11`) contradicting the comment?

tbekolay: Yeah I think this is supposed to say >=, my b

.gitignore
```diff
@@ -25,3 +25,52 @@ htmlcov
...
+docs/examples/spa/spa_sequence.py
+docs/examples/spa/spa_sequence_routed.py
+docs/examples/usage/config.py
+docs/examples/usage/delay_node.py
+docs/examples/usage/exceptions.py
+docs/examples/usage/network_design.py
+docs/examples/usage/network_design_advanced.py
+docs/examples/usage/rectified_linear.py
+docs/examples/usage/strings.py
+docs/examples/usage/tuning_curves.py
```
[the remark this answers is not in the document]
tbekolay: There shouldn't be any that we need to keep I don't think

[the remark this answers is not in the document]
hunse: I had wanted to keep it open in case we get any `.py` examples, but that's a good point that we do that `rm` anyway, so we should just ignore them all as long as we have that line in there.

.travis.yml
```diff
@@ -1,62 +1,71 @@
...
+        - provider: pypi
+          user: tbekolay
+          password: $PYPI_TOKEN
+          distributions: "sdist bdist_wheel"
+          on:
+            all_branches: true
+            condition: $TRAVIS_TAG =~ ^v[0-9]*
 
 before_install:
+  - chmod +x .ci/deploy.sh  # so we can keep it -x on local machines
```
tbekolay: Why do we want to keep this `-x` on local machines?

hunse: So that somebody doesn't call it accidentally, I assume. This is the way it was before; I just added the comment. I had actually switched it, then I thought that this was probably done intentionally, so I added the comment instead.

tbekolay: I think it was initially because marking things as executable is hard (impossible?) on Windows, but I think we should just mark it executable.

.ci/static.sh
hunse: I did a quick check, and it appears they're both 0.3 now.

drasmuss: Cool, I'd just take this out then to keep things simple

PR #1489, opened 2018-11-12
Switch documentation to nengo_sphinx_theme

drasmuss:
**Motivation and context:**

Switches the documentation over to our new theme (https://github.com/nengo/nengo-sphinx-theme)

**How has this been tested?**

Built the docs locally, looked over them, looks good :+1:.

**Types of changes:**

**Checklist:**

- [n/a] I have run the test suite locally and all tests passed.


drasmuss:
I added a commit that adds version tracking to the documentation.  Every time a tagged release is built, that documentation will be permanently added to the docs so that people can go back and refer to the documentation for previous versions.  The published docs themselves are updated every time a new commit is pushed to master, same as before.  I also have it set up so that docs are built and pushed to `gh-pages-test` whenever a non-master branch is built.  We could get rid of that if we want, as it does add to the regular build time.  But I've found it handy in NengoDL for those times when you want to make sure the documentation will build/display properly within the overall doc versioning framework (since it isn't as easy to build a full version of the docs locally anymore).  An example of that `gh-pages-test` output, for this branch, can be seen at https://github.com/nengo/nengo/commit/4534b8ed7fbcf1fa714df758dbffd96dda843ec0 if you want to see what the versioned docs look like.

drasmuss:
Added a commit to apply the patch from https://github.com/nengo/nengo-bones/pull/3, and added a favicon/logo.  At the moment I'm just using the generic Nengo logos.  We might want to think about creating separate assets for "Nengo core", distinct from the general "Nengo ecosystem" assets.

hunse:
It appears that the favicon logo file was not actually added (it's not in `docs/_static`, and TravisCI can't find it when building the docs).

PR #1492, opened 2018-11-15
Add ConvInc operator to backend docs

drasmuss:
**Motivation and context:**

Should have done this in #1481, I updated the frontend docs but not the backend docs.

**How has this been tested?**

Built the docs locally, looks good.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/transforms.py
+++ b/nengo/builder/transforms.py
@@ -66,6 +66,41 @@ def build_convolution(model, transform, sig_in,
 
 
 class ConvInc(Operator):
+    """Apply convolutional weights to input signal.
+
+    Parameters
+    ----------
+    W : Signal
+        The convolutional weights (a.k.a. the kernel).
+    X : Signal
+        The input signal.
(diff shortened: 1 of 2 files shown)


nengo/builder/transforms.py
```diff
@@ -66,6 +66,41 @@ def build_convolution(model, transform, sig_in,
 
 
 class ConvInc(Operator):
+    """Apply convolutional weights to input signal.
+
+    Parameters
+    ----------
+    W : Signal
+        The convolutional weights.
```
hunse: Maybe we should put "(aka. kernel)" after "weights", just to make it clear we're talking about the same thing (since "kernel" is the most common term in ML and computer vision). Same down below in "Attributes".

Also, this is just a general comment, but I've never been sure if/when we need both "Parameters" and "Attributes" sections, or when it's fine to just have one. They often have a bunch of overlap, and the redundancy can make them more of a pain to keep up to date.

PR #1494, opened 2018-12-05
Copy small arrays in array_hash

### LINKED ISSUE #1122: array_hash fails on reversed arrays
Alternative bug title: "Cannot use `PresentInput` process with a reversed array".

```
do_bug = True      # set to True to use array that will bug on array_hash
fix_bug = False    # set to True to fix the array so that it can be hashed
check_bug = False  # set to True to invoke the bugged code directly

if do_bug:
    inputs = np.asarray([0, 1, 2])[::-1]
else:
    inputs = np.asarray([0, 1, 2])

if fix_bug:
    inputs = np.array(inputs)

if check_bug:
    nengo.utils.numpy.array_hash(inputs)

    nengo.Node(output=PresentInput(inputs, 1.0))

    sim.run(1.0)
```

```
c:\users\aaron\ctn\nengo\nengo\utils\numpy.pyc in array_hash(a, n)
     72         v = a.view()
     73         v.setflags(write=False)
---> 74         return hash(v.data if PY2 else v.data.tobytes())
     75     else:
     76         # pick random elements to hash

AttributeError: cannot get single-segment buffer for discontiguous array
```

Submitted on behalf of @celiasmith.

arvoelke:
**Motivation and context:**
Closes #1122 issue with numpy<1.13 and Python 2.7.

**How has this been tested?**
Tested both cases in the code (small array and big array), and ran locally on the conda environment that had this issue:

```
cd nengo
conda create --name=py27 python=2.7 numpy=1.10 scipy=0.17.0
conda activate py27
python setup.py -q install
```

By reading issue #1122 and then proceeding to the fix (array copy) and tests.

**Types of changes:**

**Checklist:**
- (N/A) I have updated the documentation accordingly.

### DIFF
--- a/nengo/utils/tests/test_numpy.py
+++ b/nengo/utils/tests/test_numpy.py
@@ -25,7 +25,7 @@ def test_meshgrid_nd():
     assert np.allclose(expected, actual)
 
 
-@pytest.mark.parametrize("shape", [(5,), (10, 5), (500, 2)])
+@pytest.mark.parametrize("shape", [(5,), (10, 5), (200,), (500, 2)])
 def test_array_hash(shape):
     n = 100  # explicitly pass in array_hash default
     flat = np.arange(np.prod(shape))
--- a/nengo/utils/tests/test_numpy.py
+++ b/nengo/utils/tests/test_numpy.py
@@ -1,8 +1,10 @@
 from __future__ import absolute_import
 
+import pytest
+
 import numpy as np
 
-from nengo.utils.numpy import meshgrid_nd
+from nengo.utils.numpy import meshgrid_nd, array_hash
 
 
 def test_meshgrid_nd():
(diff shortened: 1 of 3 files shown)


nengo/utils/tests/test_numpy.py
```diff
@@ -21,3 +23,23 @@ def test_meshgrid_nd():
                   [[23, 42], [23, 42], [23, 42]]])]
     actual = meshgrid_nd(a, b, c)
     assert np.allclose(expected, actual)
+
+
+@pytest.mark.parametrize("shape", [(50,), (500, 2)])
```
hunse: Just thinking but maybe the size of the first array should be 5 instead of 50. At some point, we may decrease the size of arrays that get copied (really, we should do an analysis and find at what point is generating the random numbers for larger array hashing more costly than copying). Anyway, having a smaller number would help ensure that even if we decrease the default, this still works. The other option is to set the `n` parameter on `array_hash` to explicitly set the point at which we randomly hash, to make sure that we test both cases.

arvoelke: I've pushed an update that explicitly sets the `n` parameter in the tests. In either case though, I trust that if the default for `array_hash` is ever changed, that the unit tests will be looked at and changed accordingly as a part of that review.

arvoelke:
Closing as we no longer support Python 2.7.

PR #1497, opened 2019-01-07
Fixes for pytest 4.1

drasmuss:
**Motivation and context:**

The 4.1.0 release of pytest broke some of our testing infrastructure, this fixes it.

Note that https://github.com/nengo/nengo/commit/35b213383a4e6385f186e5e5a48067a036936d2b is kind of a workaround fix.  Something in more recent pytest versions causes PermissionErrors when renaming the cache file (non-deterministically).  But my thinking is that if pytest can trigger that somehow, then users might be able to as well (and we already knew that this part of the code was error-prone, based on the comments in the code).  So rather than trying to avoid the problem on the pytest side, I just handled this additional PermissionError case.

**How has this been tested?**

Ran all the unit tests (with `--plots --logs --analytics --slow`) on pytest 4.1.0, everything passed.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -7,6 +7,7 @@ import os
 import shutil
 import struct
 from subprocess import CalledProcessError
+import sys
 from uuid import uuid1
 import warnings
 
@@ -30,6 +31,10 @@ from nengo.utils.least_squares_solvers import (
 from nengo.utils.lock import FileLock
 
 logger = logging.getLogger(__name__)
+
+if sys.version_info < (3, 3, 0):
+    # there was no PermissionError before 3.3
+    PermissionError = OSError
 
 
 def get_fragment_size(path):
@@ -399,9 +404,8 @@ class WriteableCacheIndex(CacheIndex):
             pickle.dump(self._index, f, pickle.HIGHEST_PROTOCOL)
         try:
             replace(self.index_path + '.part', self.index_path)
-        except CalledProcessError:
-            # May happen with Python 2.7 on Windows where replace is
-            # implemented by a callout to the `move` command. It may fail when
+        except (CalledProcessError, PermissionError):
+            # It may fail when
             # another program like a virus scanner is accessing the file to be
             # moved. There is not a lot we could do about this. See
             # <https://github.com/nengo/nengo/issues/1200> for more info.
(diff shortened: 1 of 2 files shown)


nengo/cache.py
```diff
@@ -411,10 +415,9 @@ def _write_index(self):
                 "software. It is safe to ignore this warning. But if you see "
                 "it a lot, you might want to consider doing one of the "
                 "following for the best Nengo performance:\n"
-                "1. Try using Python 3 instead of Python 2.\n"
-                "2. Configure your anti-virus to ignore the Nengo cache "
+                "1. Configure your anti-virus to ignore the Nengo cache "
                 "folder ('{cache_dir}').\n"
-                "3. Disable the cache.\n"
+                "2. Disable the cache.\n"
```
hunse: Could this happen if the user simply doesn't have permission to write to the cache dir? Should we add that as one of the steps to check, before disabling the cache?

nengo/conftest.py
```diff
@@ -292,15 +292,15 @@ def mark_nl(nl):
 def pytest_collection_modifyitems(session, config, items):
     if config.getvalue('noexamples'):
         deselect_by_condition(
-            lambda item: getattr(item.obj, 'example', None), items, config)
+            lambda item: item.get_closest_marker("example"), items, config)
```
hunse: I checked, and this appears in pytest 3.6.0, which is the earliest version we support anyway :+1: I ran the tests with 3.6.0, and they appear to work, with slow tests being skipped (I didn't test for other markers).

nengo/cache.py
drasmuss: I don't think we really have a good grasp on why this happens or how to fix it.  For example in this case it was failing non-deterministically, so it's not a simple permission error.  Until we know we have a fix that works, I'm inclined to just leave it as the general "ignore this warning, or disable the cache".  I even thought about taking out the anti-virus part (since we don't really know if that's causing the issue either), but left it in as at least we have one well-documented case where that seemed to fix it.

hunse: If you have the cache in a folder that you explicitly do not have permissions for, do you get a permission error consistently? If that's the case, then I would suggest checking permissions.

As for the anti-virus part, I'm not sure who put that in or how well supported that is. I had assumed it was more related to the CalledProcessError, but who knows.

hunse: I did a quick test, and it looks like (at least on my system) if you don't have permissions for the cache dir, you get an error in `DecoderCache.__enter__`, and the warning there gets displayed. It's not terribly helpful, because we group all errors together and display the same generic message for all of them, so maybe we want to look at that in the future.

But for this, I'm fine with it as-is, because I'm convinced enough now that it won't come up for easily predictable or fixable reasons on the user's side.

jgosmann: > Could this happen if the user simply doesn't have permission to write to the cache dir?

I'd have to check, but I believe that we already write to a temporary file in the cache directory and the error only occurs trying to swap out the actual file with the temporary one.

PR #1476, opened 2018-09-18
Do not overwrite existing seeds when building

hunse:
**Motivation and context:**

This allows someone to pre-set seeds for some or all objects. I'm using this in `nengo_loihi` so that I can make sure the `nengo` models that we use to simulate the parts of the model that are on the host, have the same seeds.

**Interactions with other PRs:**

This is supporting nengo/nengo-loihi#70.

**How has this been tested?**

This is tested with `nengo_loihi`.

**How long should this take to review?**

- Quick (less than 40 lines changed or changes are straightforward)

**Types of changes:**

- New feature (non-breaking change which adds functionality)

**Checklist:**

- [ ] I have read the **CONTRIBUTING.rst** document.
- [ ] I have updated the documentation accordingly.
- [ ] I have included a changelog entry.
- [ ] I have added tests to cover my changes.
- [ ] I have run the test suite locally and all tests passed.

**Still to do:**

- [ ] Discuss whether we want to have this and the `nengo_loihi` PR both done as currently proposed, or whether we want to bring the `nengo_loihi` function for pre-seeding networks over here. Computing seeds before build-time has been something that we've talked about for a while, and it would make a lot of backend stuff easier.

PR #1501, opened 2019-02-01
Seed backends

hunse:
**Motivation and context:**

This makes seeding more accessible to backends. It separates out a seeding function for networks that backends can use. It respects existing values in the `model.seeds` and `model.seeded` dictionaries so that backends can set these for some (or all) objects and they won't be overwritten.

This replaces #1476. 

**How has this been tested?**

There is a test making sure that values don't get overwritten. Otherwise, our current seeding tests cover this, and it's tested for backends in nengo/nengo-loihi#70.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/network.py
+++ b/nengo/builder/network.py
@@ -49,21 +49,13 @@ def build_network(model, network, progress=None):
     -----
     Sets ``model.params[network]`` to ``None``.
     """
-    def get_seed(obj, rng):
-        # Generate a seed no matter what, so that setting a seed or not on
-        # one object doesn't affect the seeds of other objects.
-        seed = rng.randint(npext.maxint)
-        return (seed if not hasattr(obj, 'seed') or obj.seed is None
-                else obj.seed)
-
     if model.toplevel is None:
         model.toplevel = network
-        model.seeds[network] = get_seed(network, np.random)
-        model.seeded[network] = getattr(network, 'seed', None) is not None
-        max_steps = len(network.all_objects) + 1  # +1 for top level network
+        seed_network(network, seeds=model.seeds, seeded=model.seeded)
 
         if progress is not None:
-            progress.max_steps = max_steps
+            # number of sub-objects, plus 1 to accont for this network
+            progress.max_steps = len(network.all_objects) + 1
 
             def build_callback(obj):
                 if isinstance(obj, tuple(network.objects)):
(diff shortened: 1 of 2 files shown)


hunse:
So the static test here is failing because of an over-indented line in `params.py`. I'm confused, because a) my local flake8 doesn't catch this, and b) this line appears to have been like this since d2acfcd7. How is this just coming up now??

The other tests appear to just have errored mysteriously. They all pass locally for me, so I believe this is just a random thing.

hunse:
When I upgrade my flake8 it catches this, so must've been a flake8 bug that it was missing it.

tbekolay:
> The other tests appear to just have errored mysteriously. They all pass locally for me, so I believe this is just a random thing.

Definitely mysterious... but unfortunately I don't think it's random because [it's been happening for the past 2+ weeks](https://travis-ci.org/nengo/nengo/builds) and it [happens on Appveyor too](https://ci.appveyor.com/project/nengo/nengo/builds/22062967). I just pushed an empty commit to #1488 and you can see the [previously passing build](https://travis-ci.org/nengo/nengo/builds/454513077?utm_source=github_status&utm_medium=notification) versus the [currently erroring build](https://travis-ci.org/nengo/nengo/builds/488799842?utm_source=github_status&utm_medium=notification): previously speedy builds are now taking so long that they don't finish and end up timing out in the middle of running tests. I would think that this would indicate that something changed in the TravisCI images such that the way we're doing things is no longer performant, but since the same thing is happening on Appveyor that seems unlikely. The likely suspect is therefore Conda, but then we would have experienced a similar regression in terms of TravisCI build speed for Nengo Loihi, which I do not believe we experienced.

Looking through the logs in more detail, many of the logs seem to halt around the cache tests. I'm going to try skipping those tests to see if we can get through the test suite, as it could be a specific cache-related regression :man_shrugging:

tbekolay:
Hm, so looking through the log [without cache tests](https://travis-ci.org/nengo/nengo/builds/488813762), it is getting farther, but I'm noticing that we are now running the tests marked `slow` even though we've asked for them not to be run, so now I'm thinking that the issue is actually due to newer versions of pytest affecting how we're marking and skipping slow tests. We were stalling out at the cache tests likely because the cache benchmarks take a long time to run in general, and I believe are some of the worst case scenarios for the virtual machines that TravisCI / Appveyor use, so this is starting to make more sense.

This might be fixed in #1497, so I will take a look at that PR now.

hunse:
Fixups LGTM.

PR #1504, opened 2019-02-08
Skip unsupported tests instead of xfail

hunse:
When a simulator does not support a test, skip it instead of
xfailing. This does not actually change how the test is run,
because the way we were using xfail it would not actually
execute the test, so skip will do the same. Now tests will
be marked as skipped in the log, to differentiate them from
tests that are marked xfail and run (which could potentially xpass).

**How has this been tested?**

When nengo_loihi is working with nengo master again, I will test it there.

**Types of changes:**

**Checklist:**

- n/a I have updated the documentation accordingly.
- n/a I have included a changelog entry.
- n/a I have added tests to cover my changes.

**Still to do:**

- [x] test with nengo_loihi

### DIFF
--- a/nengo/tests/test_pytest.py
+++ b/nengo/tests/test_pytest.py
@@ -1,8 +1,37 @@
 import nengo.utils.numpy as npext
 from nengo.conftest import TestConfig
 
+pytest_plugins = ["pytester"]
+
 
 def test_seed_fixture(seed):
     """The seed should be the same on all machines"""
     i = (seed - TestConfig.test_seed) % npext.maxint
     assert i == 1832276344
+
+
+def test_unsupported(testdir):
+    testdir.makeconftest(
+        """
+        import nengo.conftest
+
+        class MockSimulator(object):
+            unsupported = [('*', 'mock simulator')]
+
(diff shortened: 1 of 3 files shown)


nengo/tests/test_pytest.py
```diff
@@ -1,8 +1,35 @@
...
+    ).parseoutcomes()
+    assert outcomes["skipped"] > 350
+    assert outcomes["deselected"] > 700
+    assert "passed" not in outcomes
+    assert "failed" not in outcomes
+
+    outcomes = testdir.runpytest(
+        "-p", "nengo.tests.options", "--pyargs", "nengo", "--unsupported",
+    ).parseoutcomes()
+    assert "passed" not in outcomes
```
hunse: Should we be checking that `outcomes['xpassed'] + outsomes['xfailed'] > 350` or something like that here?

tbekolay: Huh... so yesterday this part of the test wasn't working correctly for me due to some weirdness with Matplotlib, but I just ran it now and it seems to be working correctly, so yeah we should.

hunse: Looks good!

Issue #1487, opened 2018-11-07
Revisit Poisson spiking LIF model


arvoelke:
Related:
 - #579
 - #1225

I think we should revisit this old issue / PR of including Poisson spiking models within Nengo.

There is a nice mathematical argument for why the Poisson spiking version of the LIF model is superior to its regular spiking LIF counterpart. I have some math in my upcoming thesis that shows our LIF model is actually systematically biased to incorrectly represent high-frequency signals, even after washing out the initial transient (#1415). And especially as the spike rates are lowered relative to the frequency of the signal. There will always be some global error due to a systematic shift from some ideal state that I've quantified. Intuitively, this arises from the neurons not being "uniformly ready to spike" at any given moment in time. No amount of scaling `n_neurons` can correct for this bias (see plot at bottom of post for experimental validation, noting that the performance plateaus for the `LIF` model)!

However, Poisson spiking can! The following `PoissonLIF()` is actually optimal in the sense of expectation, as the expected filtered error theoretically scales to `0` as `n_neurons -> infinity`, regardless of the input signal frequency. The [memoryless](https://en.wikipedia.org/wiki/Memorylessness) feature of the exponential distribution (the time between Poisson events) is what gives us this property. Even more strongly, this guarantees that there is no approximation being made when switching between filtered `LIFRate()` and filtered `PoissonLIF()`, apart from variability in the PSC (which again scales to `0`). But this comes at the cost of increased variability for lower numbers of neurons and constant inputs, compared to `LIF()`. I have a currently-unpublished proof of this in my upcoming thesis and in discussions with @celiasmith.

Some drawbacks:
 - Time-to-first-spike is now probabilistic (e.g., consider rank-order encoding, or WTA mutual inhibition).
 - No notion of resetting the voltage to zero via inhibition.
 - Can violate its own refractory period (with small probability).
 - Higher variability for constant inputs.

Some advantages:
 - Better scaling guarantees (see two figures below).
 - Potentially faster.
 - Less memory.
 - No initial transient at the start of the simulation (see #1415).
 - No latency for discontinuous jumps in input (related to previous point, and higher frequencies in general).

Here is some experimental validation, with elements of the code borrowed from @tbekolay and @tcstewar:

```python
class PoissonLIF(LIFRate):
    """Poisson-spiking leaky integrate-and-fire (LIF) neuron model.

    Parameters
    ----------
    tau_rc : float
        Membrane RC time constant, in seconds. Affects how quickly the membrane
[...]

PR #1505, opened 2019-02-08
Add Poisson neuron type and corresponding test

HugoChateauLaurent:
**Motivation and context:**
Adding a Poisson generator is useful to get a realistic dummy spiking behaviour.
https://github.com/nengo/nengo/issues/579

**How has this been tested?**
I used the neuron type within ensembles. 
A test has been added in the test_neurons.py file to check the firing rate.

**Types of changes:**

**Checklist:**

I am not sure about the changelog entry. What exactly should I add, and where?

### DIFF
--- a/nengo/neurons.py
+++ b/nengo/neurons.py
@@ -679,6 +679,40 @@ class Izhikevich(NeuronType):
         recovery[spiked > 0] = recovery[spiked > 0] + self.reset_recovery
 
 
+class Poisson(NeuronType):
+    """Poisson spiking generator.
+
+    Emits a spike with a probability of ``p(s|x) = x*dt`` where ``x`` is the expected
+    firing rate and ``dt`` is the simulation timestep.
+
+    Parameters
+    ----------
+    amplitude : float, optional (Default: 1)
+        Scaling factor on the neuron output. Corresponds to the relative
+        amplitude of the output of the neuron.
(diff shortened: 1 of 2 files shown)


arvoelke:
Thank you for putting this together. It's encouraging to see these feature proposals and useful to know what is needed!

I think it's important to be clear that, from what I can see, this implements a Poisson version of the `nengo.SpikingRectifiedLinear()` model. And so it replaces a rectified linear unit with Poisson spikes at the same expected frequency as `max(0, J)`, where `J` is the encoded current -- but not the LIF's tuning curve with respect to the same current. More information can be found in my comment here: https://github.com/nengo/nengo/issues/1487#issuecomment-461976723

arvoelke:
By the way, an alternative to the approach you've taken would be to extend the code from #1487 to customize how the `rates` vector is obtained. Currently it comes from the inheritance `LIFRate.step_math(self, dt=1, J=J, output=rates)`, but in theory it can be whatever you want (any statically defined tuning curve, including ReLU). Another benefit of this implementation is that it correctly handles multiple spikes per time-step by use of Knuth's algorithm for Poisson sampling via repeated exponential sampling. I'd very much welcome such an extension, that is able to support Poisson versions of many different models, in [nengolib](https://github.com/arvoelke/nengolib/). :)

PR #1506, opened 2019-02-14
Weight solvers

hunse:
**Motivation and context:**

This gets weight solvers working properly with the new transforms.

**How has this been tested?**

I added a test to make sure that a non-factorizable weight solver (specifically NNLS) works as expected and produces non-negative weights. This should act as a smoke test for bigger problems with weight solvers, since this test will fail if weight solvers are being done in a factorized way (i.e. split up into decoders, transform, encoders).

**How long should this take to review?**

- Average (neither quick nor lengthy)

**Types of changes:**

- Bug fix (non-breaking change which fixes an issue)

**Checklist:**

- [x] I have read the **CONTRIBUTING.rst** document.
- [x] I have updated the documentation accordingly.
- [ ] I have included a changelog entry.
- [x] I have added tests to cover my changes.
- [x] I have run the test suite locally and all tests passed.

Issue #161, opened 2013-09-25
EventProbe


tbekolay:
Probing spikes right now is very expensive as the output current of every neuron is recorded on each timestep. Additionally, this forces the user to write a few lines of code to convert that representation to a list of spike times, which is easier to plot. It would be nice if it were possible to only record the spike times.

More generally, it's often very convenient to record the time that certain events occur, whether those events are spikes of the time at which a simulated lever is pressed (i.e., the first timestep that a particular signal goes from `< 1` to `1`). This capability is present in most experimental neuroscience software (e.g., it's part of [`python-neo`](http://pythonhosted.org/neo/core.html), which reads many neuroscientific data formats).

tbekolay:
Closing this in favor of #447. It's now _The_ spike-time / even probe issue. Godspeed #447!

PR #1507, opened 2019-02-15
Fix behaviour of non-compositional weight solvers

drasmuss:
**Motivation and context:**

The pluggable transform builder approach assumes that the Transform build function is responsible for combining decoders, transform, and encoders to generate a full weight matrix.  However, certain solvers must operate on the full weight matrix, they cannot operate in the decoder space and then be linearly combined with transform/encoders.  So the new Transform API broke those solvers; this fixes the behaviour by adding a special case for these "non-compositional" solvers.

I also added a second commit that removes the encoders argument from all the solvers (since it isn't being used anymore).

Alternate implementation of https://github.com/nengo/nengo/pull/1506

**How has this been tested?**

Pulled in a slightly modified version of the test from https://github.com/nengo/nengo/pull/1506.

The second commit is independent of the first (i.e., we could apply the non-compositional fix, but leave the encoder argument in as well).  So they can be treated relatively independently.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -7,7 +7,7 @@ from nengo.builder.ensemble import gen_eval_points, get_activities
 from nengo.builder.node import SimPyFunc
 from nengo.builder.operator import Copy, ElementwiseInc
 from nengo.connection import Connection
-from nengo.transforms import Convolution
+from nengo.transforms import Convolution, Dense
 from nengo.ensemble import Ensemble, Neurons
 from nengo.exceptions import BuildError, ObsoleteError
 from nengo.neurons import Direct
@@ -106,6 +106,19 @@ def build_decoders(model, conn, rng):
 
     eval_points = get_eval_points(model, conn, rng)
     targets = get_targets(conn, eval_points)
+
+    if conn.solver.weights and not conn.solver.compositional:
+        # solver is solving for the whole weight matrix, so apply
+        # transform/encoders to targets
+        if not isinstance(conn.transform, Dense):
+            raise BuildError(
+                "Non-compositional solvers only work with Dense transforms")
+        transform = conn.transform.sample(rng=rng)
+        targets = np.dot(targets, transform.T)
+        # weight solvers only allowed on ensemble->ensemble connections
+        assert isinstance(conn.post_obj, Ensemble)
+        post_enc = model.params[conn.post_obj].scaled_encoders
+        targets = np.dot(targets, post_enc.T[conn.post_slice])
 
     x = np.dot(eval_points, encoders.T / conn.pre_obj.radius)
     wrapped_solver = (model.decoder_cache.wrap_solver(solve_for_decoders)
@@ -244,21 +257,33 @@ def build_connection(model, conn):
         if conn.solver.weights:
             model.sig[conn]['out'] = model.sig[conn.post_obj.neurons]['in']
 
-            if isinstance(conn.post_obj, Ensemble):
-                encoders = model.params[conn.post_obj].scaled_encoders.T
-                encoders = encoders[conn.post_slice]
+            # weight solvers only allowed on ensemble->ensemble connections
+            assert isinstance(conn.post_obj, Ensemble)
 
-                # post slice already applied to encoders, don't apply later
-                post_slice = None
+            encoders = model.params[conn.post_obj].scaled_encoders.T
+            encoders = encoders[conn.post_slice]
+
+            # post slice already applied to encoders (either here or in
+            # `build_decoders`), so don't apply later
+            post_slice = None
     else:
         in_signal = slice_signal(model, in_signal, conn.pre_slice)
 
     # Build transform
-    weighted, weights = model.build(conn.transform,
-                                    in_signal,
-                                    decoders=decoders,
-                                    encoders=encoders,
-                                    rng=rng)
+    if conn.solver.weights and not conn.solver.compositional:
+        # special case for non-compositional weight solvers, where
+        # the solver is solving for the full weight matrix. so we don't
+        # need to combine decoders/transform/encoders.
+        weighted, weights = model.build(Dense(decoders.shape, init=decoders),
+                                        in_signal,
+                                        rng=rng)
+    else:
+        weighted, weights = model.build(conn.transform,
+                                        in_signal,
+                                        decoders=decoders,
+                                        encoders=encoders,
+                                        rng=rng)
+
     model.sig[conn]["weights"] = weights
 
     # Build synapse
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -129,7 +129,7 @@ def build_decoders(model, conn, rng):
     return eval_points, decoders.T, solver_info
 
 
-def solve_for_decoders(conn, gain, bias, x, targets, rng, E=None):
+def solve_for_decoders(conn, gain, bias, x, targets, rng):
     activities = conn.pre_obj.neuron_type.rates(x, gain, bias)
     if np.count_nonzero(activities) == 0:
         raise BuildError(
@@ -137,7 +137,7 @@ def solve_for_decoders(conn, gain, bias, x, targets, rng, E=None):
             "This is because no evaluation points fall in the firing "
             "ranges of any neurons." % (conn, conn.pre_obj))
 
-    decoders, solver_info = conn.solver(activities, targets, rng=rng, E=E)
+    decoders, solver_info = conn.solver(activities, targets, rng=rng)
     return decoders, solver_info
 
 
--- a/nengo/tests/test_solvers.py
+++ b/nengo/tests/test_solvers.py
@@ -8,7 +8,7 @@ import pytest
 
 import nengo
 from nengo.dists import UniformHypersphere
-from nengo.exceptions import ValidationError
+from nengo.exceptions import BuildError, ValidationError
 from nengo.utils.compat import iteritems, range
 from nengo.utils.numpy import rms, norm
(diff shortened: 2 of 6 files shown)


hunse:
I'd prefer the terms "factorable" and "non-factorable", since those are what are typically used when talking about matrices, like we are here. But compositionality is a more general concept, which people might be more familiar with in at least some context, so maybe it would hold more wide-spread meaning. It just seems slightly less precise to me.

drasmuss:
Yeah I was thinking that there might be a case where the solver was not factorable, in the matrix sense, but was compositional (i.e. the transform builder has a way to combine the solver output with transform/encoders that isn't linear matrix multiplication).  But with our current Transforms and solvers, compositionality is equivalent to factorability, so maybe it is better to be more specific 🤷‍♂️ .

nengo/builder/connection.py
```diff
@@ -107,6 +107,16 @@ def build_decoders(model, conn, rng):
     eval_points = get_eval_points(model, conn, rng)
     targets = get_targets(conn, eval_points)
 
+    if conn.solver.weights and not conn.solver.compositional:
+        # solver is solving for the whole weight matrix, so apply
+        # transform/encoders to targets
+        if not isinstance(conn.transform, Dense):
+            raise BuildError(
+                "Non-compositional solvers only work with Dense transforms")
+        targets = np.dot(targets, conn.transform.init.T)
```
hunse: The transform could be a distribution. Furthermore, to ensure that the sampled transform ends up being the same whether we use a we use a weight solver or not, I think we should sample it in `build_connection` so that it gets sampled the same either way.

nengo/builder/connection.py
```diff
@@ -254,11 +264,23 @@ def get_prepost_signal(is_pre):
...
-    weighted, weights = model.build(conn.transform,
-                                    in_signal,
-                                    decoders=decoders,
-                                    encoders=encoders,
-                                    rng=rng)
+    if conn.solver.weights and not conn.solver.compositional:
+        # special case for non-compositional weight solvers, where
+        # the solver is solving for the full weight matrix, which already
+        # includes the transform
+        weights = Signal(decoders, name="%s.weights" % conn)
```
hunse: Would it be cleaner to make a new `Dense` transform here and then call the transform builder on that?

drasmuss:
I talked about the NoSolver change in the original PR here https://github.com/nengo/nengo/pull/1481#issuecomment-435150936.  I do think this is still a change for the better.  Basically the question is, in the new language, should NoSolver be compositional or non-compositional?  I think it should be compositional, as that is how almost all the solvers work, so it seems like more expected behaviour.  And, as you point out, if users want to set the weights non-compositionally, they can use a neuron-to-neuron connection.

But I definitely did forget to update the docstring for NoSolver when I made that change, and will do that assuming you're OK with the above reasoning.

hunse:
As I said, I find it hard to really weigh in because I've never used `NoSolver`. But given what I know, I actually think that having it non-compositional (as it originally was) makes more sense.

Even though the majority of solvers are compositional, the whole point of `weights=True` on solvers is really to support non-compositional solvers. If you have a compositional solver, there's very few use cases for using `weights=True`. Basically all it does is slow your simulation down, because rather than doing decoders and encoders as two relatively small matrix multiplications, you have to do all the weights at once as one huge matrix multiplication. Not to mention you have to store all those weights. (When it comes to something like Loihi, there are more reasons to use `weights=True` with a compositional solver, because it eliminates the interneurons.) So the real reason that we have `weights=True` in my opinion is to support non-compositional solvers (I think that that's what motivated the introduction of that feature).

Having NoSolver be non-compositional covers the two main cases as I see it: wanting to specify decoders (like when we have NEF-style connections), and wanting to specify weights (as people not familiar with the NEF would be used to). The case of where you want to specify weights but have them multiplied by the encoders seems less common to me (because it's computationally equivalent to just having decoders then encoders, except perhaps on neuromorphic platforms).

That said, this behaviour of specifying your own decoders and having them multiplied by the encoders to get weights is unique. We don't have another way to do that. Whereas specifying decoders can be done with Ensemble->Neuron connections, and weights with Neuron->Neuron connections.

So really, I can't decide without a better idea of who is actually using NoSolver, and for what purpose. The way it originally was seems more intuitive to me, and is how I would expect it to be implemented if I didn't know anything more about it. I can see arguments for the new way, though.

Finally, how tricky would it be if NoSolver took `compositional` as an argument? Then you could specify it either way depending on what you want.

nengo/builder/connection.py
drasmuss: Changed it so this is properly handled if transform is a Distribution.  I didn't move the transform sampling into `build_connection`, because I think we want the Transform build functions to be able to control that sampling (since they are the "authority" on the transform, and might need to customize that behaviour).  But actually the way `build_connection` works out, `rng` is only either passed to the Transform builder or this function, so the random state will be the same in both cases.

nengo/builder/connection.py
drasmuss: Done, definitely way cleaner

drasmuss:
I think there are three use cases:

1. You want to specify the decoders, and have a factored connection (use `NoSolver(weights=False)`)
2. You want to specify the decoders, and have a full weight matrix (use `NoSolver(weights=True)`)
3. You want to specify the full weight matrix (use `transform=x`).

If we change NoSolver to be non-compositional, then you can't do 2 anymore (and now there are two different ways of doing 3).

Also practically speaking, I don't think anyone actually uses `NoSolver(weights=True)`.  As mentioned in that older comment, this was actually broken ever since it was first created, and no one noticed.

tcstewar:
> I think there are three use cases:

The only other use case I can think of is "getting around implementation-specific details of a particular backend".  For example, `nengo_spinnaker` never supported `a.neurons->b` style Connections, but you could fake that by using NoSolver.   I'm not sure that's an important thing to consider in this discussion about what the correct behaviour should be, but I just thought I'd mention this for completeness.

hunse:
@tcstewar: Did you ever use `NoSolver` with `weights=False`?

tcstewar:
> @tcstewar: Did you ever use `NoSolver` with `weights=False`?

I don't think so, since `nengo_spinnaker` never supported anything that bypassed encoders, and that was the main place I was using initially using NoSolver.

hunse:
I've added a test that runs a non-compositional solver, and fixed a couple bugs that that revealed. I think this is ready to go now.

nengo/tests/test_solvers.py
```diff
@@ -576,7 +576,41 @@ def test_nosolver_validation():
     NoSolver(values=[[1], [1]])
 
 
-def test_non_compositional_solver(Simulator):
+@pytest.mark.parametrize('solver', [
+    LstsqDrop(weights=True),
+    # LstsqL1(weights=True),
+    # NnlsL2(weights=True),
```
drasmuss: I'm guessing these are meant to be un-commented?

tbekolay: I uncommented them and the `NnlsL2` test fails... should we be concerned about this?

tbekolay: Also, these tests are pretty slow on my machine, is it possible to lower the simulation run time to something more like 0.1?

hunse: No, we should probably get rid of them. `LstsqL1` works, it's just slow. I couldn't figure a way to mark it slow without marking the whole test as slow. `NnlsL2` "works", but because it forces non-negative weights the communication channel doesn't work as well, so it fails the test the way it's written now.

drasmuss: So do we have any tests that LstsqL1 is working correctly?  I notice that you changed it to non-compositional, which I assume means that it was incorrect previously, but no tests were failing.

hunse: I think we have tests for all the solvers, but most of them are unit tests, not integration tests. So they wouldn't be affected by `compositional` being wrong.

It would be nice to have integration tests, but it's hard to have ones that really work. For example, with LstsqL1, you'd have to check that the sparsity of the final weights is as desired, to test whether `compositional` is set correctly. And it's not clear what the output sparsity should be given a particular regularization value. With LstsqDrop and Nnls, it's a little easier, since you can check for the desired sparsity or that the weights are non-negative, respectively, though as I said for Nnls, the weights don't work as well because they're non-negative.

One option would be to have a common test helper function that creates this network, and then have separate tests for each of the three compositional solvers that can check both the output signal and the weights specifically for that solver. Then we can mark LstsqL1 as slow without affecting the others. Let me know if you want me to do this.

arvoelke:
Since `E=E` was removed, is there a recommended way to write solvers in a way that is compatible across multiple versions of Nengo? I have several solvers that wrap existing solvers, and just pipe the `E=E` through. Replacing this with `*args, **kwargs` does not work for `<=2.6.0` because of #1358. Previously my codebase supported nengo 2.1.0-2.8.0 and this seems like a weird reason to limit support to `>=2.6.1`.

arvoelke:
Figured out a hacky work-around in #161. Summarized below, as a way of circumventing #1358. The trick is to make sure the inspection finds at least one default argument. I'm not sure this 100% safe, but all my tests pass. ¯\\\_(ツ)\_/¯

```python
class MockSolver(nengo.solvers.LstsqL2):

    def __call__(self, A, Y, __hack__=None, **kwargs):
        assert __hack__ is None
        # __hack__ is necessary prior to nengo PR #1359 (<2.6.1)
        # and following nengo PR #1507 (>2.8.0)

        ...
        return super(MockSolver, self).__call__(A, Y, **kwargs)
```

PR #1514, opened 2019-03-22
Switch to nengo-bones templating system

drasmuss:
**Motivation and context:**

Switches our TravisCI setup over to the new templating system from nengo-bones (so that these can be managed from a central location across the nengo ecosystem).

There was a weird pip installation issue with Python 3.4, but since official 3.4 support ends this month I didn't worry too much about it, and just dropped that build.

Depends on https://github.com/nengo/nengo-bones/pull/5

**How has this been tested?**

All our TravisCI builds continue to pass after the change, and I manually inspected the logs to check that things are running as expected.

**Types of changes:**

**Checklist:**


hunse:
What is the `pip` installation issue? I realize Python 3.4 won't continue to be updated/supported, but it would be nice if we could not break things. Many older OSes still ship with 3.4 (I'm still using 3.4). If the pip error is really beyond our control, though, then I guess I'm okay to break things. It would just mean we'd force a lot of people to upgrade their Python *now*.

drasmuss:
You can see the issue here if you want to spend some time on it https://travis-ci.org/nengo/nengo/jobs/509901277.  Note that we don't explicitly disallow python 3.4, we're just saying that we don't officially support it (because we're not testing against it).  So it isn't forcing anyone to upgrade; someone installing their environment in a different way that works with python 3.4 should still be fine.

arvoelke:
It seems that files like `simulator.py` are still not a part of the coverage. Is this considered outside the scope of this PR?

drasmuss:
Yeah in this PR all I was trying to do was make sure that everything was working the same as before.  Those files are actually being covered, if you look at the log output (e.g. here https://travis-ci.org/nengo/nengo/jobs/513096421#L2155).  So it's just something weird with the Codecov reports.  I'll make a card to look into that though, thanks for the reminder.

arvoelke:
Pushed a commit to fix up the README for dropping Python 3.4 support and cleaned up the history.

.nengobones.yml
```diff
@@ -0,0 +1,59 @@
...
+    pip_install:
+      - scikit-learn
+    coverage: true
+  - template: docs
+    conda_install:
+      - $NUMPY_VERSION
+      - scipy
+      - matplotlib
+
+travis_yml:
```
tbekolay: Does order matter in this file? I ask because it would be nice to move this before `ci_scripts` because it defines `NUMPY_VERSION`, which is used in `ci_scripts`.

arvoelke: In the interest of providing a rapid response I will say: I think the order doesn't matter.

PR #1502, opened 2019-02-06
Update README nef-summary image

arvoelke:
**Motivation and context:**
Broken image on the landing page README: https://github.com/nengo/nengo -- an unintentional consequence of a design change made on Dec 4, 2018: https://github.com/nengo/design/pull/7

**How has this been tested?**
Loaded new image via GitHub's preview

**How long should this take to review?**
- Quick (less than 40 lines changed or changes are straightforward)

**Types of changes:**
- Bug fix (non-breaking change which fixes an issue)

**Checklist:**
- [x] I have read the **CONTRIBUTING.rst** document.
- [N/A] I have updated the documentation accordingly.
- [N/A] I have included a changelog entry.
- [N/A] I have added tests to cover my changes.
- [N/A] I have run the test suite locally and all tests passed.

PR #1520, opened 2019-04-04
Drop Python 2 support

tbekolay:
**Motivation and context:**
The next release will be Nengo 3.0, and official Python 2.7 support will be ending soon, so let's bite the bullet and drop Python 2.7 support for Nengo. This is not done in a backwards compatible way here, but the `nengo.utils.compat` modules is still present and deprecated so that packages that depend on Nengo have time to drop Python 2.7 support between the 3.0 release and the next minor release.

**How has this been tested?**
Ran the test suite. TravisCI configuration was also changed.

**Types of changes:**

**Checklist:**

**Still to do:**
- [x] Update documentation
- [x] Update the changelog
- [x] Decide what to do with the `is_*` function in `nengo.utils.compat`

Currently this PR does not do anything with the `nengo.utils.compat.is_string` etc functions even though they exist in the `compat` module and are largely there due to Python 2/3 differences. However, they also serve to simplify some checks (especially those that could be checking both normal objects and NumPy arrays) so I am not sure whether we should outright remove them, perhaps move them? I will make a commit removing our usage of them to see what the diff looks like.


tbekolay:
The most recent commit in this PR removes uses of the `is_integer` and `is_string` functions, which looks pretty good and simple. Other functions, however, like `is_array_like` and `is_iterable` are harder to remove without adding lines.

Also the first Travis build failed due to some cached environments being Python 2.7 even though they should now be 3.7, so I'm hoping that removing the caches fixes them.

jgosmann:
Not sure in which `__init__` method we have cyclic imports, but I think that pylint is correct to consider them cyclic in any case. As long as module A imports module B anywhere in its code, A has a dependency on B. (Same if B imports A in which case in gets cyclic.) In general, I think that cyclic imports should be avoided. If they are really necessary I would rather disable the warning for that one file instead for the whole project.

CHANGES.rst
```diff
@@ -38,6 +38,9 @@ Release History
 
 **Changed**
 
+- Python 2 is no longer supported. The oldest supported Python version is 3.4.
```
arvoelke: This will need to be fixed up when 5ce285f4deb81437b1dc90f7ccaecb0b934c65ac is in master (#1514).

README.rst
```diff
@@ -58,7 +58,8 @@ If you'd like to install Nengo from source,
 please read the `developer installation instructions
 <https://www.nengo.ai/nengo/contributing.html#developer-installation>`_.
 
-Nengo is tested to work on Python 2.7 and 3.4+.
+Nengo is tested to work on Python 3.4 and above.
```
arvoelke: Note to self, this will conflict with change in #1514.

.pylintrc
```diff
@@ -41,6 +41,7 @@ enable = anomalous-backslash-in-string,
          unreachable,
          unused-import,
          useless-else-on-loop,
+         useless-object-inheritance,
```
arvoelke: Note these changes will need to be moved over to `setup.cfg` following #1514.

docs/examples/advanced/nef_algorithm.ipynb
```diff
@@ -156,7 +156,7 @@
     "    voltage and v=1 is the firing threshold.\n",
     "    \"\"\"\n",
     "    spikes = []\n",
-    "    for i in range(len(v)):\n",
+    "    for i, _ in enumerate(v):\n",
```
arvoelke: Why was this change made? I see at least one other place in the code where it's still `range()`, but not sure why there but not here.

.travis.yml
```diff
@@ -8,30 +8,29 @@ cache:
...
+    - env: MODE="test" NUMPY="numpy=1.12"
+    - env: MODE="test" NUMPY="numpy=1.13"
+    - env: MODE="test" NUMPY="numpy=1.14"
+    - env: MODE="test" NUMPY="numpy=1.15"
+    - env: MODE="test" PYTHON="3.4"
     - env: MODE="test" PYTHON="3.5"
     - stage: deploy
       if: branch =~ ^release-candidate-* OR tag =~ ^v[0-9]*
-      env: MODE="deploy" PYTHON="3.6"
+      env: MODE="deploy"
```
arvoelke: Note to self: this will all change when rebased onto `template-ci` (#1514).

nengo/utils/compat.py
```diff
@@ -12,6 +11,9 @@
 
 # If something's changed from Python 2 to 3, we handle that here
 if PY2:
+    warnings.warn("Python 2 compatibility has been dropped from Nengo as of "
+                  "version 3.0.0; nengo.utils.compat will be removed in the "
+                  "next minor release.")
```
arvoelke: Do we have a process in place to ensure that things like this are in fact done before the release?

docs/examples/advanced/nef_algorithm.ipynb
tbekolay: This tripped the `consider-using-enumerate` pylint rule. This happens here and not in other places because `i` is being used to index into `v`. I do not actually use the value from enumerate in this case because `v[i]` appears on the left-hand side of an equation, so using that value would have no effect. It also appears on the right hand side of some lines, so I could use it there, but it feels confusing to have the same quantity using two different identifiers in the same local block. Similarly, it's much easier/cleaner to make this change than it is to add a `pylint disable` comment.

nengo/utils/compat.py
tbekolay: Not really, we could add something to the [release guide](https://www.nengo.ai/releasing.html) to check for deprecated things and remove them if appropriate? Would that be enough process?

arvoelke: Hmm it would be kind of awkward to get to the stage of doing the release and finding out that a PR needs to happen. I'm thinking more along the lines of triggering the TODO item to appear during the appropriate release cycle so that it's not forgotten about. I'll also make a note for myself to add this to `CHANGES.rst`.

tbekolay: It's pretty common for things to come up in that first stage of the release, so I wouldn't worry about that too much. For example, we rarely run the slow tests, but we run them when releasing and that's required separate PRs a few times and it hasn't been a big issue.

setup.py
```diff
@@ -88,5 +88,6 @@ def read(*filenames, **kwargs):
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Topic :: Scientific/Engineering :: Artificial Intelligence",
+        "Framework :: Nengo",
```
tbekolay: This should be in alphabetical order, as the other classifiers are, and honestly I would leave this for a separate PR because it's not really related to the point of this one.

CHANGES.rst
```diff
@@ -25,22 +25,22 @@ Release History
 **Added**
 
 - Added progress bar support for Jupyter Lab >=0.32.
-  (`#1428 <https://github.com/nengo/nengo/pull/1428>`__,
-  `#1087 <https://github.com/nengo/nengo/issues/1087>`__)
+  (`#1428 <https://github.com/nengo/nengo/pull/1428>`_,
+  `#1087 <https://github.com/nengo/nengo/issues/1087>`_)
```
tbekolay: So this will cause ReST parsing issues whenever we have a hyperlink that is used multiple times. The `_` gives you a link that is associated with the text, while `__` gives an "anonymous" link (see [here](http://docutils.sourceforge.net/docs/user/rst/quickref.html#inline-markup)). If you use a link with `_` twice with different text you get a parse error.

So, if we wanted to be fully consistent we should use `__` everywhere, which we've sort of started to do since we ran into that problem in the past. However, I really think we should be touching these old lines because they don't cause any parse errors, and it messed up the `git blame` if you want to look at the changelog and see when something was added.

nengo/utils/compat.py
arvoelke: Resolved by additional commit to CHANGES.rst and this issue: https://github.com/nengo/nengo.github.io/issues/71

CHANGES.rst
arvoelke: Fair enough (resolved).

setup.py
arvoelke: I wanted to be efficient and also avoid the possibility of a one line change taking over a month to get through (e.g., https://github.com/nengo/nengo/pull/1502) so I just did it, especially since the PyPi team requested that it be "of immediate use". Accordingly I plan to do a `nengolib` release with this asap. But sure I've reverted the commit and will make a separate PR.

nengo/__init__.py
```diff
@@ -10,9 +12,30 @@
...
+Nengo requires at least Python 3.4.
+
+The fact that this version was installed on your system probably means that you
+are using an older version of pip; you should consider upgrading with
+
+ $ pip install pip setuptools --upgrade
+
+There are two options for getting Nengo working:
+
+- Upgrade to Python >= 3.4
```
arvoelke: Note this message and check `(3, 4)` will also need to be updated to 3.5 following #1514.

nengo/__init__.py
```diff
@@ -40,5 +63,5 @@
 except AttributeError:
     pass
 
-__copyright__ = "2013-2017, Applied Brain Research"
+__copyright__ = "2013-2019, Applied Brain Research"
```
arvoelke: `LICENSE.rst` should also be updated to `2019`. Using the need to make #1522 as precedent, I split these changes off into #1523 and added a fixup commit here to revert this change.

.travis.yml
```diff
@@ -8,30 +8,29 @@ cache:
...
+    - PYTHON="3.6"
 
 jobs:
   include:
     - stage: basic
-      env: MODE="static" PYTHON="2.7"
-    - env: MODE="test" PYTHON="3.6" COVERAGE="true" SCIPY="true"
-    - env: MODE="test" PYTHON="2.7" COVERAGE="true"
+      env: MODE="static"
+    - env: MODE="test" COVERAGE="true" SCIPY="true"
```
arvoelke: The fact we're no longer collecting coverage on 2.7 seems to be changing things quite a bit. Should I be going through the codecov report in detail and fixing up all the missing coverage?

nengo/utils/testing.py
```diff
@@ -436,7 +434,7 @@ def __init__(self, n_threads):
         for t in threads:
             t.join()
             if not t.assertion_result:
-                reraise(*t.exc_info)
+                raise self.exc_info[0].with_traceback(*self.exc_info[1:])
```
arvoelke: Is this not a bug? I think this should be:
```python
    raise self.exc_info[1].with_traceback(self.exc_info[2])
```
to be consistent with `compat.py`:
```python
    def reraise(tp, value, tb):
        raise value.with_traceback(tb)
```
and https://docs.python.org/3/library/exceptions.html#BaseException.with_traceback.

setup.py
```diff
@@ -82,10 +83,10 @@ def read(*filenames, **kwargs):
         "Intended Audience :: Science/Research",
         "License :: Free for non-commercial use",
         "Operating System :: OS Independent",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3.4",
         "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
```
arvoelke: I guess before #1514 gets merged in we should decide if we are making 3.7 support official? Over there I was assuming not, but we are here.

arvoelke: Note there is a `Programming Language :: Python :: 3 :: Only` classifier. Should we include it? https://pypi.org/pypi?%3Aaction=list_classifiers

nengo/utils/tests/test_stdlib.py
```diff
@@ -95,12 +99,12 @@ def test_groupby(hashable, force_list, rng):
     # call groupby
     keygroups = groupby(pairs, lambda p: p[0], force_list=force_list)
 
-    keys2 = sorted(map(lambda x: x[0], keygroups))
+    keys2 = sorted([x[0] for x in keygroups])
```
arvoelke: FFR, you can even leave out the `[]` as in: `sorted(x[0] for x in keygroups)`.

setup.py
```diff
@@ -72,6 +72,7 @@ def read(*filenames, **kwargs):
         "tests": tests_require,
     },
     tests_require=tests_require,
+    python_requires=">=3.4",
```
arvoelke: Note to self: this will also change to 3.5 following #1514.

setup.py
arvoelke: After some offline discussion, my mistake was that I had been defining "support" as testing against a clean/reproducible environment. But saying that "we support Python 3.7" does not mean that we are necessarily testing against a clean/reproducible Python 3.7 environment. Rather, the consensus was that the testing within the release process on a developer's py3.7 environment is sufficient and that official "support" means we will simply try to fix any bugs that emerge on py3.7 environments.

arvoelke: Went ahead and added this as a fixup to the first commit. Note that `"3 " < "3."` lexicographically and so it goes at the top.

arvoelke:
I went through the codecov report which was quite difficult to read because it doesn't directly show the difference in lines that used to be covered in [master](https://codecov.io/gh/nengo/nengo/tree/master/nengo) but are no longer covered here. The things I found were:
 - `utils/magic.py::ObjectProxyMethods.__module__` is no longer called anywhere.
 - `utils/nco.py` no longer hits `raise CacheIOError("Not a Nengo cache object file.")`.

It's not obvious why this is the case. But since these are both `utils` I think it's fine to ignore this for now, and deal with it when the `utils` are moved or when there is an effort to increase the overall coverage.

I did however add a commit to prune the branching in `stdlib.py` that was there to handle older Python versions. I also removed the need to cover `utils/compat.py` but left the tests there because they only take a couple milliseconds to run on my workstation. Lastly I added `coverage=true` to the `python=3.5; scipy=false` build in order to ensure we're covering both `scipy` branches throughout the codebase.

I'll create another branch to rebase this onto `template-ci` and clean up the history.

.travis.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     - env: MODE="test" NUMPY="numpy=1.14"
     - env: MODE="test" NUMPY="numpy=1.15"
     - env: MODE="test" PYTHON="3.4"
-    - env: MODE="test" PYTHON="3.5"
+    - env: MODE="test" COVERAGE="true" PYTHON="3.5"
```
drasmuss: I might add this build to the "Basic" stage, just so that we have both coverage builds happening there.  Otherwise I can anticipate some confusion if one of these builds runs but the other doesn't, and all of a sudden it looks like the coverage has changed if you just glance at the status checks.

arvoelke: If the static checks fail but the first basic codecov build passes, would codecov still report its results? I had thought if any part of the job failed (even non-codecov related builds) then codecov would auto-fail.

drasmuss: No we have it set up so that codecov always reports its status checks, regardless of whether the build passed or failed.

arvoelke: Done

setup.py
arvoelke: Done

nengo/utils/tests/test_stdlib.py
arvoelke: Done

nengo/utils/testing.py
arvoelke: Done

nengo/__init__.py
arvoelke: Done

.travis.yml
arvoelke: Done

.pylintrc
arvoelke: Done

README.rst
arvoelke: Done

CHANGES.rst
arvoelke: Done

setup.cfg
```diff
@@ -55,6 +52,7 @@ filterwarnings =
 ignore = _vendor, compat.py
```
arvoelke: Reminder to selves: we can remove "compat.py" from here once it's removed.

PR #1532, opened 2019-05-24
Implement SparseTransforms

hunse:
To allow more efficient connections when weights are sparse.

**How has this been tested?**

Added a test of the transform in four different situations (with and without Scipy, with and without a distribution initializer).

Start by looking at the `Sparse` transform class itself in `transforms.py`. Then maybe look at the test, and finally all the builder details.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] It probably needs some general cleanup/more documentation.

### DIFF
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -599,12 +599,19 @@ class DotInc(Operator):
         A = signals[self.A]
         Y = signals[self.Y]
 
-        def step_dotinc():
-            inc = A.dot(X)
-            if self.reshape:
-                inc = inc.reshape(Y.shape)
-            Y[...] += inc
-        return step_dotinc
+        if self.reshape:
+
+            def step_dotinc_reshape():
+                Y[...] += A.dot(X).reshape(Y.shape)
+
+            return step_dotinc_reshape
+
+        else:
+
+            def step_dotinc():
+                Y[...] += A.dot(X)
+
+            return step_dotinc
 
 
 class SparseDotInc(DotInc):
--- a/nengo/builder/operator.py
+++ b/nengo/builder/operator.py
@@ -520,7 +520,7 @@ def reshape_dot(A, X, Y, tag=None):
                          % (tag, A.shape, X.shape, Y.shape))
 
     # Reshape to handle case when np.dot(A, X) and Y are both scalars
-    return (np.dot(A, X)).size == Y.size == 1
+    return A.dot(X).size == Y.size == 1
 
 
 class DotInc(Operator):
@@ -600,11 +600,22 @@ class DotInc(Operator):
         Y = signals[self.Y]
 
         def step_dotinc():
-            inc = np.dot(A, X)
+            inc = A.dot(X)
             if self.reshape:
-                inc = np.asarray(inc).reshape(Y.shape)
+                inc = inc.reshape(Y.shape)
             Y[...] += inc
         return step_dotinc
+
+
+class SparseDotInc(DotInc):
+    """Like `.DotInc` but ``A`` is a sparse matrix."""
+
+    def __init__(self, A, X, Y, tag=None):
+        if not A.sparse:
+            raise BuildError("%s: A must be a sparse Signal")
+
+        # Disallow reshaping
+        super().__init__(A, X, Y, reshape=False, tag=tag)
 
 
 class BsrDotInc(DotInc):
--- a/nengo/builder/signal.py
+++ b/nengo/builder/signal.py
@@ -3,7 +3,12 @@ from io import StringIO
 import numpy as np
 
 from nengo.exceptions import SignalError
+from nengo.transforms import SparseMatrix
 import nengo.utils.numpy as npext
+
+
+def is_sparse(obj):
+    return isinstance(obj, SparseMatrix) or npext.is_spmatrix(obj)
 
 
 class Signal:
@@ -51,12 +56,26 @@ class Signal:
             assert name
         self._name = name
 
-        initial_value = np.asarray(initial_value)
-        if initial_value.ndim > 0 and base is None:
-            self._initial_value = np.ascontiguousarray(initial_value).view()
+        self._initial_value = initial_value
+        if self.sparse:
+            assert initial_value.ndim == 2
+            assert offset == 0
+            assert base is None
+            if npext.is_spmatrix(initial_value):
+                self._initial_value.data.setflags(write=False)
+
         else:
-            self._initial_value = initial_value.view()
-        self._initial_value.setflags(write=False)
+            # To ensure we do not modify data passed into the signal,
+            # we make a view of the data and mark it as not writeable.
+            # Consumers (like SignalDict) are responsible for making copies
+            # that can be modified, or using the readonly view appropriately.
+            readonly_view = np.asarray(self._initial_value)
+            if readonly_view.ndim > 0 and base is None:
+                readonly_view = np.ascontiguousarray(readonly_view)
+            # Ensure we have a view and aren't modifying the original's flags
+            readonly_view = readonly_view.view()
+            readonly_view.setflags(write=False)
+            self._initial_value = readonly_view
 
         if base is not None:
             assert isinstance(base, Signal) and not base.is_view
@@ -69,22 +88,35 @@ class Signal:
 
     def __getstate__(self):
         state = dict(self.__dict__)
-        v = self._initial_value
-        state['_initial_value'] = (
-            v.shape, v.base, npext.array_offset(v), v.strides)
+
+        if not self.sparse:
+            # For normal arrays, the initial value could be a view on another
+            # signal's data. To make sure we do not make a copy of the data,
+            # we store the underlying metadata in a tuple, which pickle can
+            # inspect to see that v.base is the same in different signals
+            # and avoid serializing it multiple times.
+            v = self._initial_value
+            state['_initial_value'] = (
+                v.shape, v.base, npext.array_offset(v), v.strides)
+
         return state
 
     def __setstate__(self, state):
         for k, v in state.items():
             setattr(self, k, v)
 
-        shape, base, offset, strides = self._initial_value
-        self._initial_value = np.ndarray(
-            shape, buffer=base, offset=offset, strides=strides)
-        self._initial_value.setflags(write=False)
+        if not self.sparse:
+            shape, base, offset, strides = self._initial_value
+            self._initial_value = np.ndarray(
+                shape, buffer=base, offset=offset, strides=strides)
+            self._initial_value.setflags(write=False)
 
     def __getitem__(self, item):
         """Index or slice into array"""
+
+        if self.sparse:
+            raise SignalError("Attempting to create a view of a sparse Signal")
+
         if item is Ellipsis or (
                 isinstance(item, slice) and item == slice(None)):
             return self
@@ -106,11 +138,11 @@ class Signal:
                       base=self.base, offset=offset)
 
     def __repr__(self):
-        return "Signal(%s, shape=%s)" % (self._name, self.shape)
+        return "Signal(name=%s, shape=%s)" % (self._name, self.shape)
 
     @property
     def base(self):
-        """(Signal or None) The base signal, if this signal is a view.
+        """(Signal) The base signal, if this signal is a view.
 
         Linking the two signals with the ``base`` argument is necessary
         to ensure that their live data is also linked.
@@ -130,7 +162,8 @@ class Signal:
     @property
     def elemstrides(self):
         """(int) Strides of data in elements."""
-        return tuple(s // self.itemsize for s in self.strides)
+        return (None if self.sparse
+                else tuple(s // self.itemsize for s in self.strides))
 
     @property
     def initial_value(self):
@@ -153,7 +186,7 @@ class Signal:
     @property
     def itemsize(self):
         """(int) Size of an array element in bytes."""
-        return self.initial_value.itemsize
+        return self.dtype.itemsize
 
     @property
     def name(self):
@@ -167,8 +200,7 @@ class Signal:
     @property
     def nbytes(self):
         """(int) Number of bytes consumed by the signal."""
-        # Equivalent to self.itemsize * self.size
-        return self.initial_value.nbytes
+        return self.itemsize * self.size
 
     @property
     def ndim(self):
@@ -210,7 +242,12 @@ class Signal:
     @property
     def strides(self):
         """(tuple) Strides of data in bytes."""
-        return self.initial_value.strides
+        return None if self.sparse else self.initial_value.strides
+
+    @property
+    def sparse(self):
+        """(bool) Whether the signal is sparse."""
+        return is_sparse(self.initial_value)
 
     def may_share_memory(self, other):
         """Determine if two signals might overlap in memory.
@@ -223,7 +260,8 @@ class Signal:
         other : Signal
             The other signal we are investigating.
         """
-        return np.may_share_memory(self.initial_value, other.initial_value)
+        return (self.is_view or other.is_view) and np.may_share_memory(
+            self.initial_value, other.initial_value)
 
     def reshape(self, *shape):
         """Return a view on this signal with a different shape.
@@ -234,6 +272,10 @@ class Signal:
         Any number of integers can be passed to this method,
         describing the desired shape of the returned signal.
         """
+
+        if self.sparse:
+            raise SignalError("Attempting to create a view of a sparse Signal")
+
         if len(shape) == 1:
             shape = shape[0]  # in case a tuple is passed in
         initial_value = self.initial_value.view()
@@ -291,25 +333,29 @@ class SignalDict(dict):
         return sio.getvalue()
 
     def init(self, signal):
-        """Set up a permanent mapping from signal -> ndarray."""
+        """Set up a permanent mapping from signal -> data."""
         if signal in self:
             raise SignalError("Cannot add signal twice")
 
-        x = signal.initial_value
+        assert isinstance(signal, Signal)
+        data = signal.initial_value
+        if isinstance(data, SparseMatrix):
+            data = data.allocate()
+
         if signal.is_view:
             if signal.base not in self:
                 self.init(signal.base)
 
             # get a view onto the base data
             view = np.ndarray(
-                shape=x.shape, strides=x.strides, offset=signal.offset,
-                dtype=x.dtype, buffer=self[signal.base].data)
-            assert np.array_equal(view, x)
+                shape=data.shape, strides=data.strides, offset=signal.offset,
+                dtype=data.dtype, buffer=self[signal.base].data)
+            assert np.array_equal(view, data)
             view.setflags(write=not signal.readonly)
             dict.__setitem__(self, signal, view)
         else:
-            x = x.view() if signal.readonly else x.copy()
-            dict.__setitem__(self, signal, x)
+            data = data if signal.readonly else data.copy()
+            dict.__setitem__(self, signal, data)
 
     def reset(self, signal):
         """Reset ndarray to the base value of the signal that maps to it"""
--- a/nengo/builder/transforms.py
+++ b/nengo/builder/transforms.py
@@ -49,9 +49,11 @@ def build_convolution(model, transform, sig_in,
     if decoders is not None:
         raise BuildError("Applying a convolution transform to a decoded "
                          "connection is not supported")
-    if encoders is not None:
-        raise BuildError(
-            "Applying encoders to a convolution transform is not supported")
+
+    # shouldn't be possible for encoders to be non-None, since that only
+    # occurs for a connection solver with weights=True, and those can only
+    # be applied to decoded connections (which are disallowed above)
+    assert encoders is None
 
     weights = transform.sample(rng=rng)
     weight_sig = Signal(weights, name="%s.weights" % transform, readonly=True)
--- a/nengo/builder/transforms.py
+++ b/nengo/builder/transforms.py
@@ -1,9 +1,9 @@
 import numpy as np
 
-from nengo.builder import Operator, Builder, Signal
-from nengo.builder.operator import Reset, ElementwiseInc, DotInc
+from nengo.builder import Builder, Operator, Signal
+from nengo.builder.operator import DotInc, ElementwiseInc, Reset, SparseDotInc
 from nengo.exceptions import BuildError
-from nengo.transforms import Dense, Convolution
+from nengo.transforms import Convolution, Dense, Sparse
 from nengo._vendor.npconv2d import conv2d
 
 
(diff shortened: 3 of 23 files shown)


nengo/builder/signal.py
```diff
@@ -259,6 +260,92 @@ def row(self):
         return self.reshape((1, self.size))
 
 
+class SparseSignal:
```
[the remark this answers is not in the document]
hunse: I agree that that would be nice. Just be aware that it's a reasonably big undertaking. We make assumptions in a few places about how signals are done internally, and I worry that backends are making more assumptions.

The reason I added the `SparseDotInc` operator was so that this could explicitly break on backends (though frankly, for NengoOCL, it might try to allocate data for the SparseSignals first and break in a weird way because of that, so I don't think that's the best way to do it). It would be nice if whatever way we do it with Signals, it could be obvious to backends what type of signal it is, and they can immediately error if they don't support that. Originally, I was imagining an abstract `Signal` base class that then gets implemented as e.g. `DenseSignal`, `SparseSignal`. But there are definitely some differences between the APIs (e.g. the dense version has `offset`, `strides`, etc., which are not applicable to the sparse one, though we could just return dummy values). 

I had hoped that it would be possible to do this in a way that would break NengoOCL gracefully if somebody uses an unsupported SparseTransform (i.e. we'd get the nice "That operator is not supported" message that's currently implemented for unsupported ops), but I'm getting less and less sure if that's possible. So maybe we just do things in whatever way makes the most sense here, and then backends have to catch up. It would be nice, though, to have an easier way to check what type of signal something is than having to do e.g. `isinstance(sig.initial_value, scipy.sparse.spmatrix)` or something like that.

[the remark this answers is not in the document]
drasmuss: Yeah, definitely agreed on all of that. I think in this case I just won't have a good sense of how feasible this is until I start trying it and see what breaks. But if trying to remove `SparseTransform` seems like it'll make things more complicated rather than less so, then I'm fine with just rolling back to this.

nengo/builder/signal.py
```diff
@@ -299,25 +386,31 @@ def __str__(self):
...
-
-            # get a view onto the base data
-            view = np.ndarray(
-                shape=x.shape, strides=x.strides, offset=signal.offset,
-                dtype=x.dtype, buffer=self[signal.base].data)
-            assert np.array_equal(view, x)
-            view.setflags(write=not signal.readonly)
-            dict.__setitem__(self, signal, view)
-        else:
-            x = x.view() if signal.readonly else x.copy()
```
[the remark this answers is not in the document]
hunse: Yes, I think this is true for now. It's actually not clear to me if Scipy sparse matrices can be views (it seems to me that it should be possible to share the same data and get a view, but I haven't figured out how to do it), so for now I've just disallowed it with `SparseSignal`. At some point we might want to move some of this logic for getting views into `Signal` itself, so that it can be done differently depending on the type of signal. But for now, I think it's fine to assume all signals that can have views are based on Numpy arrays.

nengo/builder/signal.py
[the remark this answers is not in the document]
jgosmann: > We make assumptions in a few places about how signals are done internally, and I worry that backends are making more assumptions.

One of the major places where this is the case, is probably the optimizer.

drasmuss:
Added a couple commits:

https://github.com/nengo/nengo/pull/1532/commits/5f3960602c411d5fdc8289eeafcf2b49c10652c5 removes SparseSignal, and instead has Signal support either `spmatrix` or `ndarray` initial values.  I have a question in there about how this would work in `__getstate__` or `__setstate__`. Those weren't implemented in the original `SparseSignal`, but could we support them?

https://github.com/nengo/nengo/pull/1532/commits/687373efee93634c2b4d1ee0e398be9926f51e42 further modifies this with a new idea that occurred to me afterwards.  In this setup we don't require scipy at all to work with sparse transforms.  Scipy only shows up inside the `SparseDotInc` operator (or, optionally, when creating a `SparseTransform`).  This way backends that have different sparse matrix implementations (e.g. NengoDL) can still use `SparseTransform` without requiring the user have `scipy` installed as an intermediate representation.

https://github.com/nengo/nengo/pull/1532/commits/b768ec5ea126db0aad09887464e552d26c94c3a1 is just because after the above changes pylint was complaining about too many public methods in `Signal`, and I noticed that `Signal.row` and `Signal.column` were only being used in one place in the codebase so I took them out.

nengo/builder/signal.py
```diff
@@ -69,13 +77,15 @@ def __init__(self, initial_value,
         self._readonly = bool(readonly)
 
     def __getstate__(self):
+        # TODO: support sparse type?
```
hunse: Shouldn't be too hard to do this. Because we don't have to worry about views, we should just be able to return the Scipy sparse matrix itself (which I assume is pickleable).

nengo/builder/signal.py
```diff
@@ -211,7 +227,19 @@ def size(self):
...
         """(tuple) Strides of data in bytes."""
-        return self.initial_value.strides
+        return self.initial_value.strides if self.viewable else None
+
+    @property
+    def type(self):
+        """(str) Type of the Signal ("dense" or "sparse")."""
+
+        return (
+            "sparse" if isinstance(self._initial_value, scipy_sparse.spmatrix)
```
hunse: `scipy_sparse` could be `None` here, I think, so we have to account for that.

drasmuss: This part is removed in the next commit

nengo/builder/transforms.py
```diff
@@ -68,18 +65,9 @@ def build_sparse(model, transform, sig_in,
...
-        assert isinstance(weights, np.ndarray)
-        OperatorType = DotInc
-        warnings.warn("Sparse Transform requires Scipy for sparse "
-                      "implementation. Falling back on dense matrix.")
-
     weight_sig = Signal(weights, name="%s.weights" % transform,
-                        readonly=True)
-    model.add_op(OperatorType(weight_sig, sig_in, weighted,
+                        readonly=True, type="sparse")
+    model.add_op(SparseDotInc(weight_sig, sig_in, weighted,
```
hunse: So my worry with this is that weights may not actually be sparse (i.e. if Scipy is not installed). Then, this is confusing to backends, because we've got a Signal that says it's sparse and an operator that says it's a SparseDotInc, when neither of them are actually sparse.

This commit is supposed to make things easier for backends, but at least for NengoOCL (what I'm familiar with) it will make it harder. Before, if you didn't have Scipy installed, you'd get a dense Signal and a regular DotInc, which nengo-ocl supports and so SparseTransform would work without Scipy. However, with this commit, things will be expressed in sparse terms that nengo-ocl is not familiar with, and it will break.

The old code here was only making a sparse Signal and SparseDotInc if Scipy is installed, so I think it should have worked fine on backends.

drasmuss: > So my worry with this is that weights may not actually be sparse (i.e. if Scipy is not installed). Then, this is confusing to backends, because we've got a Signal that says it's sparse and an operator that says it's a SparseDotInc, when neither of them are actually sparse.

I don't think it'd be confusing to backends, since the Signal explicitly tells the backend that they should treat this as a sparse signal (and same with the op).  The only tricky part is that in the case that the `initial_value` is not represented by a `scipy.sparse` format (e.g., because `scipy` is not installed), then the backend will have to manually convert the `initial_value` into the backend-specific sparse format.  But that seems like something that backends will probably have to do anyway.  So the only downside I can think of is that it might be slightly more difficult to convert a `numpy` array into the backend's sparse format than to convert a `scipy.sparse` array into that format, but I can't imagine there would be too much of a difference.

> This commit is supposed to make things easier for backends, but at least for NengoOCL (what I'm familiar with) it will make it harder. Before, if you didn't have Scipy installed, you'd get a dense Signal and a regular DotInc, which nengo-ocl supports and so SparseTransform would work without Scipy. However, with this commit, things will be expressed in sparse terms that nengo-ocl is not familiar with, and it will break.

If I'm understanding things correctly, the old behaviour only worked smoothly with NengoOCL if we assume that the user doesn't have scipy installed.  If the user does have scipy installed, then we would have the same problem ("things will be expressed in sparse terms that nengo-ocl is not familiar with, and it will break").  That doesn't seem like behaviour we'd want to rely on, so it seems like in either case NengoOCL will have to add logic to explicitly handle SparseDotInc (even if that behaviour is just to convert it to the equivalent dense DotInc).

With that in mind, this seems like an improvement for NengoOCL.  If you have support in NengoOCL for sparse signals and SparseDotInc, I presume that would be independent of `scipy`.  So in the old behaviour, if the user didn't have `scipy` installed, then their `SparseTransforms` would be converted to `DenseTransforms`, even though they could be running using NengoOCL's sparse support.  In the new behaviour, if the user doesn't have scipy installed then the same data is available to NengoOCL (the dense equivalent of the sparse transform).  So if you wanted, you could just implement that with a regular DotInc, same as before.  But you could also notice that this is a sparse op (even though the initial value isn't sparse), and take advantage of that to use the sparse support in NengoOCL.

> The old code here was only making a sparse Signal and SparseDotInc if Scipy is installed, so I think it should have worked fine on backends.

It works, but it requires scipy to be installed to work.  Unless backends are actually using `scipy` to do the implementation, it seems awkward to force `scipy` to be installed just to activate that functionality.

hunse: Those are some fair points. I had been treating the fall-back to Numpy as just to allow simple examples to not break for users, not as something with more practical applications.

In the long run, I do worry about representing sparse signals with dense matrices, for more than just size concerns. One issue comes up when you get to learning, where you may have a sparse matrix where you want to adjust the values with a learning rule, while maintaining the sparsity structure. In that case, the Numpy matrix won't necessarily define the sparsity structure, if e.g. you start with zeros for some elements but allow those elements to change.

In the short term, my biggest problem with this is just that it makes it less predictable what the data format in sparse Signals will be. It's obviously not hard for backends to check and deal with this accordingly, it's just another thing they have to be aware of.

Also, I just realised it makes the API a little weird. As things are, a user could not pass in a Numpy array with zeros and have it be treated sparsely by a backend (`Sparse.init` has to be the same length as `Sparse.inds` if it's a Numpy array). We could change this so that if you pass `init` without `inds`, it can be a Scipy sparse matrix or a Numpy matrix.

Personally, I don't have a problem with requiring Scipy to do anything practical with Sparse transforms, and I would slightly lean to that side just to simplify things a bit for backends (they don't have to interpret a Numpy array as a sparse matrix). But I'm also fine with these changes if that's the direction that others are leaning. Maybe we should do a straw poll.

tbekolay: Are there examples anywhere of models that would make use of sparse transforms? I feel like right now there are lots of hypothetical situations being discussed and if those were made into full examples (which we probably want for our documentation anyway) it would make the decision pretty easy.

drasmuss: > Also, I just realised it makes the API a little weird. As things are, a user could not pass in a Numpy array with zeros and have it be treated sparsely by a backend (Sparse.init has to be the same length as Sparse.inds if it's a Numpy array). We could change this so that if you pass init without inds, it can be a Scipy sparse matrix or a Numpy matrix.

I think we want users to always be specifying things in a sparse format if they want to use a SparseTransform (either by using `scipy.sparse`, or by specifying data/`inds`).  

> One issue comes up when you get to learning, where you may have a sparse matrix where you want to adjust the values with a learning rule, while maintaining the sparsity structure. In that case, the Numpy matrix won't necessarily define the sparsity structure, if e.g. you start with zeros for some elements but allow those elements to change.

Yeah that is true, the numpy fallback assumes that the sparsity structure won't change (so you can detect it just by searching for non-zero elements). But we don't support learning on `SparseTransforms` at the moment (and I'm not sure we would any time soon), so I'd say that's something we can worry about if/when the time comes.

> Are there examples anywhere of models that would make use of sparse transforms? I feel like right now there are lots of hypothetical situations being discussed and if those were made into full examples (which we probably want for our documentation anyway) it would make the decision pretty easy.

Really none of this makes a difference from the user's perspective, so I don't think the difference would show up in any examples.

The difference we're focusing on in this commit is how data is represented in-between the time the user specifies a ` SparseTransform` and when it is delivered to the backend.  Basically, whether we want to have a `scipy` bottleneck in-between the user and the backend (where the data is represented internally as `scipy` data structures during the build process).  The advantage of that is that we get a sparse matrix representation for free, which naturally stores the data we need to represent.  The disadvantage is that the user needs to have `scipy` installed, even if they're not using `scipy` to specify the `SparseTransform` and the backend isn't using `scipy` to implement it.

hunse: https://github.com/nengo/nengo-extras/blob/master/docs/examples/mnist_single_layer.ipynb

I'm not sure if it clarifies things, though. A lot of what we're talking about here is on the backend side, not the frontend API.

hunse: > The disadvantage is that the user needs to have scipy installed, even if they're not using scipy to specify the SparseTransform and the backend isn't using scipy to implement it.

I had thought originally about essentially defining our own sparse format for the sparse Signals, like the data/inds format we accept for creating a sparse transform. Then I realised that this would be re-inventing the wheel, since `scipy.sparse` already does this, but in a more flexible and efficient manner. So I opted to go with Scipy's sparse matrix as our internal representation. But you're right, that it does mean we're tied to it even if neither the user nor the backend are using it.

hunse: Just had a brief chat with Dan and Trevor. Dan is going to try implementing the data/inds format in `Signal`. However, if the user passes in a Scipy sparse matrix, we'd preserve that. So the data in a sparse `Signal` could either be a Scipy sparse matrix or data/inds. Then the operator (i.e. `SparseDotInc`) implementation can do what it wants, like pumping data/inds straight to whatever sparse format the backend uses for simulation. Thus, we don't require Scipy to do sparse stuff.

nengo/builder/signal.py
```diff
@@ -51,12 +51,19 @@ def __init__(self, initial_value,
             assert name
         self._name = name
 
-        initial_value = np.asarray(initial_value)
-        if initial_value.ndim > 0 and base is None:
-            self._initial_value = np.ascontiguousarray(initial_value).view()
+        self._initial_value = initial_value
+        if self.type == "dense":
```
jgosmann: How can `self.type` be already set here?

drasmuss: `self.type` is a property

jgosmann:
I think the current implementation requires some changes to the optimizer to prevent sparse signal from being improperly merged with other sparse signals or dense signals.

drasmuss:
Do you have a test that would illustrate the problem?  I thought that since there is no merge logic defined for `SparseDotInc`, the default would be not to merge them (and therefore we wouldn't have to worry about merging sparse signals).

jgosmann:
That's true for `SparseDotInc`, bot what ensures that only `SparseDotInc` uses sparse signals?

drasmuss:
> That's true for SparseDotInc, bot what ensures that only SparseDotInc uses sparse signals?

All kinds of other things would break if someone tried to use a sparse signal with an op that didn't support sparse signals, do I don't think we have to be too worried about that.  But I added an explicit check (and a test) in any case.

arvoelke:
What challenges would be involved with supporting learning on sparse transforms? e.g., [if one would like](https://forum.nengo.ai/t/recurrent-bcm-connection-problems/817/) only a sparse subset of the weights to be updated by the learning rule.

What would be involved with getting Nengo's solver to solve for a sparse decoding matrix? Should it work if I were to create a solver that uses L0-regularization and cast the returned weights to `Sparse`? Or is the primary use-case currently when manually specifying the `transform` (is there an example for when one should use this to improve the efficiency of their model)?

hunse:
For sparse learning, we would need learning ops that accommodate sparse signals. It would really be up to them how to do this. Doing them in a non-efficient way should be not too difficult.

Solvers don't return transforms, they just return arrays. So we'd have to update the connection builder to know what to do if a solver returns a Scipy sparse matrix. There are actually some "sparse" solvers (e.g. that set small weights to zero), but they currently work with dense matrices. You need > 50% sparsity (or more) for a sparse matrix to even begin to become efficient memory-wise (and probably much more than that for computational efficiency), so I would not have these solvers always return sparse matrices even if we support that.

The main use-case right now is user-provided weights on neuron-neuron connections (or node-neuron connections I think should work too). If you want to create an ensemble-ensemble connection with sparse weights, you'd need to solve for those weights yourself (e.g. manually call a Solver with eval points, encoders, etc).

nengo/builder/operator.py
```diff
@@ -601,12 +601,19 @@ def make_step(self, signals, dt, rng):
...
-        def step_dotinc():
-            inc = A.dot(X)
-            if self.reshape:
-                inc = inc.reshape(Y.shape)
-            Y[...] += inc
-        return step_dotinc
+        if self.reshape:
+
+            def step_dotinc_reshape():
+                Y[...] += A.dot(X).reshape(Y.shape)
```
hunse: I think it actually still makes a temporary variable for the reshaped matrix under the hood. In either case (old code or new code), that's just a matrix view, no data is copied.

nengo/transforms.py
```diff
@@ -210,21 +250,12 @@ def sample(self, rng=np.random):
...
 
-    @property
-    def ndim(self):
-        return len(self.shape)
-
-    @property
-    def size(self):
-        return self.indices.shape[0]
+        self._dense = np.zeros(self.shape, dtype=self.dtype)
+        self._dense[self.indices[:, 0], self.indices[:, 1]] = self.data
```
drasmuss: Should probably mark this as `readonly` as well, since we're caching it.

nengo/builder/operator.py
tbekolay: Yeah sorry I wasn't claiming that there's any change in anything other than the fact that the locals dictionary gets a new entry added to it, which is a very negligible but non-zero cost.

nengo/transforms.py
tbekolay: Done

nengo/params.py
```diff
@@ -374,17 +374,18 @@ class NdarrayParam(Parameter):
 
     If the passed value is an ndarray, a view onto that array is stored.
     If the passed value is not an ndarray, it will be cast to an ndarray
-    of float64s and stored.
+    of ``dtype`` and stored.
```
tbekolay: Note to self: adding `dtype` arg should be split into separate commit

nengo/builder/operator.py
```diff
@@ -615,24 +615,12 @@ def step_dotinc():
...
 class SparseDotInc(DotInc):
-    """Like `.DotInc` but ``A`` is a sparse matrix.
-
-    This class only overrides ``make_step``. See `.DotInc` for other details.
-    """
+    """Like `.DotInc` but ``A`` is a sparse matrix."""
 
     def __init__(self, *args, **kwargs):
+        # Disallow reshaping
         super().__init__(*args, reshape=False, **kwargs)
```
hunse: Should we explicitly check that `A` is sparse here?

tbekolay: Sure, no reason not to.

nengo/builder/signal.py
```diff
@@ -69,22 +85,30 @@ def __init__(self, initial_value,
 
     def __getstate__(self):
         state = dict(self.__dict__)
-        v = self._initial_value
-        state['_initial_value'] = (
-            v.shape, v.base, npext.array_offset(v), v.strides)
+
+        if not self.sparse:
```
hunse: Maybe make a note that we don't need this for sparse because we don't do views of sparse. If we do add sparse views at some point, this will have to change.

hunse: Or we could do the note the other way, saying that we DO do this for normal arrays because they could be views.

tbekolay: It's not clear to me that sparse views are possible... and I would assume if we're making a big change to how sparse signals are handled, we're going to do a search for `self.sparse` here and make changes as appropriate. So I'm not sure it warrants a specific comment right now.

tbekolay: The other way round makes more sense, I'll add that comment

nengo/builder/signal.py
```diff
@@ -210,7 +234,12 @@ def size(self):
     @property
     def strides(self):
         """(tuple) Strides of data in bytes."""
-        return self.initial_value.strides
+        return None if self.sparse else self.initial_value.strides
+
+    @property
+    def sparse(self):
+        """(bool) Whether the signal is sparse."""
+        return not npext.is_array_like(self.initial_value)
```
hunse: For sparse, `initial_value` has to be `spmatrix` or `SparseMatrix`, right? I'd prefer to check for those, a) in case `is_array_like` changes to include e.g. `spmatrix`, and b) so it's clearer reading the code what constitutes a sparse matrix.

nengo/builder/signal.py
```diff
@@ -291,25 +325,29 @@ def __str__(self):
...
-            assert np.array_equal(view, x)
+                shape=data.shape, strides=data.strides, offset=signal.offset,
+                dtype=data.dtype, buffer=self[signal.base].data)
+            assert np.array_equal(view, data)
             view.setflags(write=not signal.readonly)
             dict.__setitem__(self, signal, view)
         else:
-            x = x.view() if signal.readonly else x.copy()
-            dict.__setitem__(self, signal, x)
+            data = data if signal.readonly else data.copy()
```
hunse: So there is a slight difference now that we don't return a view on a readonly array (in the dense case). I don't think this is a big deal, because I don't think we ever change the `shape` of that, or set flags, or things like that. Just wanted to point it out in case anyone is aware of if/how it might affect backends.

tbekolay: It was always a view of a view anyway because the base `_initial_value` is always already a readonly view. I tried to make that more clear in `Signal.__init__` through new variable names and comments.

A view of a readonly view is still readonly, there's no practical difference that I can tell.

tbekolay: I guess the thing you can't prevent is modifying the flags of the thing that's returned, and so if you modify the flags you get from this then you modify it for all consumers of it, is that the distinction you're pointing out? It's true but if an op is doing something like that to a numpy array that is supposed to be readonly they should just not mark it as readonly.

nengo/utils/numpy.py
```diff
@@ -108,6 +111,33 @@ def array_hash(a, n=100):
         return hash(v.data.tobytes())
 
 
+def array_hash(a, n=100):
+    """Simple fast array hash function.
+
+    For arrays with size larger than ``n``, pick ``n`` elements at random
+    to hash. This strategy should work well for dense matrices, but for
+    sparse ones it is more likely to give hash collisions.
```
hunse: This comment needs updating. The thing about "sparse ones" was referring to numpy arrays with a lot of zeros. It doesn't apply to scipy sparse matrices.

nengo/builder/signal.py
tbekolay: Makes sense, changed!

nengo/builder/signal.py
hunse: As I mentioned in the slack channel, I was able to make two sparse matrices that share data (i.e. a sparse view of sorts). However, I think the method by which you create it would be different for each different sparse matrix subtype (i.e. CSR, COO, etc.), and we don't have an imminent need for it, so not worth pursuing now.

I just wanted to make it clear that this is happening because of views, since it would be easy to search for `self.sparse`, but completely ignore this code if you thought that it was something specific to sparse matrices, not to views. (This is why I wanted to keep the `viewable` property, to make this clearer, but a comment here about why we do this for dense will serve the same purpose.)

nengo/builder/signal.py
hunse: Yeah, that would be the distinction. Even if a view is readonly (i.e. has the `write=False` flag), you can still change the shape of that view (by modifying `.shape`), or modify flags, which would change it for all consumers as you say. But I agree, ops shouldn't be doing these things to readonly arrays.

tbekolay:
I think I've addressed all comments in recent fixups. Raising the build error also required writing a new test to get 100% diff coverage.

hunse:
LGTM

Issue #1527, opened 2019-05-01
Particular small synaptic taus cause FloatingPointError in LIF


arvoelke:
```python
import nengo

with nengo.Network() as model:
    nengo.Connection(nengo.Node(1), nengo.Ensemble(10, 1), synapse=1e-4)
    
with nengo.Simulator(model) as sim:
    sim.run(1.0)
```

```
~/CTN/nengo/nengo/builder/neurons.py in step_simneurons()
     75 
     76         def step_simneurons():
---> 77             self.neurons.step_math(dt, J, output, *states)
     78         return step_simneurons
     79 

~/CTN/nengo/nengo/neurons.py in step_math(self, dt, J, spiked, voltage, refractory_time)
    466         # since v(t) = v(0) + (J - v(0))*(1 - exp(-t/tau)) assuming
    467         # J is constant over the interval [t, t + dt)
--> 468         voltage -= (J - voltage) * np.expm1(-delta_t / self.tau_rc)
    469 
    470         # determine which neurons spiked (set them to 1/dt, else 0)

FloatingPointError: invalid value encountered in multiply
```

hunse:
Fixed in #1387 (126ff2cf41919d2c4c81391c7a872bcd3e204e0c). This is because we're still using my `expm` function that just uses a Taylor series, and is unstable for small `tau`. That commit switches to using Scipy's (vendorized), which uses Pade approximants and is more stable. It still fails for very small `tau`, but during build instead of simulation (we could make a better exception there).

PR #1535, opened 2019-05-29
State space synapses

hunse:
**Motivation and context:**

This makes synapses faster and more accurate, and simplifies the synapse code. The first commit (using Scipy's expm) makes filter discretization more accurate, fixing issues like #1527.

The second commit (using state-space for synapses) makes their simulation faster (and they should be more accurate, too, because Scipy's `cont2discrete` function works in state-space internally, so this avoids converting back to transfer function form).

This PR was separated out from #1387.

**How has this been tested?**

I used the following script for speed profiling. Though the results were slightly variable from one run to the next, runs on the old branch took around 0.9 s, whereas runs on the new branch took around 0.66 s. I also tried with the Lowpass synapse, where speed remained the same (as expected, because the simulation code there is the same).

```python
import timeit

import numpy as np
import nengo

nt = 50000
n = 1000
x = np.random.uniform(-1, 1, size=(nt, n))

synapse = nengo.Alpha(0.01)

def test():
    return synapse.filt(x)

print(min(timeit.repeat(test, number=1, repeat=5)))
```

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -118,6 +118,9 @@ Release History
   which speed up build times slightly. To re-enable this testing,
   enable the ``fail_fast`` RC setting.
   (`#1532 <https://github.com/nengo/nengo/pull/1532>`__)
+- ``LinearFilter`` now uses state space representations internally,
+  which is faster and potentially more accurate.
+  (`#1535 <https://github.com/nengo/nengo/pull/1535>`__)
 
 **Deprecated**
 
@@ -167,6 +170,9 @@ Release History
 - Fixed the progress bar in recent Jupyter Lab versions.
   (`#1499 <https://github.com/nengo/nengo/issues/1499>`_,
   `#1500 <https://github.com/nengo/nengo/pull/1500>`_)
+- Some higher-order ``LinearFilter`` synapses had unnecessary delays
+  that have now been removed.
+  (`#1535 <https://github.com/nengo/nengo/pull/1535>`__)
 
 2.8.0 (June 9, 2018)
 ====================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -121,6 +121,10 @@ Release History
 - ``LinearFilter`` now uses state space representations internally,
   which is faster and potentially more accurate.
   (`#1535 <https://github.com/nengo/nengo/pull/1535>`__)
+- The default value of ``y0`` in ``Synapse.filt`` is now 0 instead of
+  the initial value of the input signal. This allows unstable filters
+  (e.g., integrators) to be used with ``filt``.
+  (`#1535 <https://github.com/nengo/nengo/pull/1535>`__)
 
 **Deprecated**
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -125,6 +125,11 @@ Release History
   the initial value of the input signal. This allows unstable filters
   (e.g., integrators) to be used with ``filt``.
   (`#1535 <https://github.com/nengo/nengo/pull/1535>`__)
+- ``LinearFilter`` now accepts the discretization method as an argument,
+  rather than having it specified in ``make_step``.
+  (`#1535 <https://github.com/nengo/nengo/pull/1535>`__)
+- The ``synapse_kwargs`` argument to ``FilteredNoise`` has been removed.
+  (`#1535 <https://github.com/nengo/nengo/pull/1535>`__)
 
 **Deprecated**
 
--- a/nengo/synapses.py
+++ b/nengo/synapses.py
@@ -61,7 +61,7 @@ class Synapse(Process):
                          default_dt=default_dt,
                          seed=seed)
 
-    def filt(self, x, dt=None, axis=0, y0=None, copy=True, filtfilt=False):
+    def filt(self, x, dt=None, axis=0, y0=0, copy=True, filtfilt=False):
         """Filter ``x`` with this synapse model.
 
         Parameters
@@ -74,8 +74,8 @@ class Synapse(Process):
         axis : int, optional
             The axis along which to filter.
         y0 : array_like, optional
-            The starting state of the filter output. If None, the initial
-            value of the input signal along the axis filtered will be used.
+            The starting state of the filter output. Must be zero for
+            unstable linear systems.
         copy : bool, optional
             Whether to copy the input data, or simply work in-place.
         filtfilt : bool, optional
@@ -87,9 +87,6 @@ class Synapse(Process):
         dt = self.default_dt if dt is None else dt
         filtered = np.array(x, copy=copy, dtype=rc.float_dtype)
         filt_view = np.rollaxis(filtered, axis=axis)  # rolled view on filtered
-
-        if y0 is None:
-            y0 = filt_view[0]
 
         shape_in = shape_out = as_shape(filt_view[0].shape, min_dim=1)
         step = self.make_step(
@@ -113,7 +110,7 @@ class Synapse(Process):
         """
         return self.filt(x, filtfilt=True, **kwargs)
 
-    def make_step(self, shape_in, shape_out, dt, rng, y0=None, dtype=None):
+    def make_step(self, shape_in, shape_out, dt, rng, y0=0, dtype=None):
         """Create function that advances the synapse forward one time step.
 
         At a minimum, Synapse subclasses must implement this method.
@@ -131,7 +128,7 @@ class Synapse(Process):
         rng : `numpy.random.RandomState`
             Random number generator.
         y0 : array_like, optional
-            The starting state of the filter output. If None, each dimension
+            The starting state of the filter output. If zero, each dimension
             of the state will start at zero.
         dtype : `numpy.dtype`, optional
             Type of data used by the synapse model. This is important for
@@ -233,7 +230,7 @@ class LinearFilter(Synapse):
         return A, B, C, D
 
     def make_step(self, shape_in, shape_out, dt, rng,
-                  y0=None, dtype=None, method='zoh'):
+                  y0=0, dtype=None, method='zoh'):
         """Returns a `.Step` instance that implements the linear filter."""
         assert shape_in == shape_out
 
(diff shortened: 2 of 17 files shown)


arvoelke:
General comments regarding #1124:
 - This resolves my concerns in #1387 regarding `y0` by changing the default from `y0=None` to `y0=0` and then ensuring there is no error for unstable systems such as the integrator `1/s`.
 - This is a backwards-incompatible change for people who relied on the `filt` behaviour for `y0=None` (which used to initialize the state based on the first value in the input array). 
 - I don't think any tests needed to be changed as a result of this PR, which indicates a lack of coverage. We should add a test that is sensitive to this removal of the `y0=None` behaviour.
 - We can defer the rest of this issue for a future PR.

Regarding #938 I still have a number of concerns, primirily with documenting/testing the associated changes in single time-step delays with passthroughs. For example, I've found the following regression which doesn't appear to be covered by our unit tests:

```python
LinearFilter([-1, 0, 1e6], [1, 2, 1]).filt([1000., 0., 0.], y0=0)
```

On version 2.8.0 this produces: 

```
[-1000.          501.6652923  1499.6640453]
```

But on this branch it produces:

```
[-498.3347077  1499.6640453  2495.66729729]
```

This is related to the fact that previously we were eliminating a single time-step delay if and only if there was no passthrough. But here, we are swapping the order of the equations to do things in a different order, which doesn't do the same thing in every case (e.g., if the system initially had a passthrough). If this new behaviour is what we want, then it should be documented and tested for future regression.

There are also two new bugs introduced by the changes in this PR, that were not covered by our unit-testing. Both of these used to work (on master), and should be fixed and tested here.

```python
nengo.Lowpass(0.5).filt([1, 0, 0])
```

```
~/CTN/nengo/nengo/synapses.py in filt(self, x, dt, axis, y0, copy, filtfilt)
     93 
     94         for i, signal_in in enumerate(filt_view):
---> 95             filt_view[i] = step(i * dt, signal_in)
     96 
     97         if filtfilt:  # Flip the filt_view and filter again

~/CTN/nengo/nengo/synapses.py in __call__(self, t, signal)
    319 
    320         def __call__(self, t, signal):
--> 321             self.X *= self.a
    322             self.X += self.b * signal
    323             return self.X[0]

TypeError: Cannot cast ufunc multiply output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

```python
nengo.LinearFilter([1], [1]).filt([0.])
```

```
~/CTN/nengo/nengo/synapses.py in filt(self, x, dt, axis, y0, copy, filtfilt)
     94         shape_in = shape_out = as_shape(filt_view[0].shape, min_dim=1)
     95         step = self.make_step(
---> 96             shape_in, shape_out, dt, None, y0=y0, dtype=filtered.dtype)
     97 
     98         for i, signal_in in enumerate(filt_view):

~/CTN/nengo/nengo/synapses.py in make_step(self, shape_in, shape_out, dt, rng, y0, dtype, method)
    234         num, den = self.num, self.den
    235         if self.analog:
--> 236             num, den, _ = cont2discrete((num, den), dt, method=method)
    237             num = num.flatten()
    238 

~/CTN/nengo/nengo/utils/filter_design.py in cont2discrete(sys, dt, method, alpha)
    448     if len(sys) == 2:
    449         sysd = cont2discrete(tf2ss(sys[0], sys[1]), dt, method=method,
--> 450                              alpha=alpha)
    451         return ss2tf(sysd[0], sysd[1], sysd[2], sysd[3]) + (dt,)
    452     elif len(sys) == 3:

~/CTN/nengo/nengo/utils/filter_design.py in cont2discrete(sys, dt, method, alpha)
    493 
    494         # Need to stack zeros under the a and b matrices
--> 495         em_lower = np.hstack((np.zeros((b.shape[1], a.shape[0])),
    496                               np.zeros((b.shape[1], b.shape[1]))))
    497 

IndexError: tuple index out of range
```

arvoelke:
Here's another difference in behaviour in this PR versus nengo, involving chains of synapses that don't contain a passthrough:

```python
import numpy as np
import nengo
from nengo.utils.filter_design import cont2discrete
from numpy import poly1d

tau = 0.005
s = nengo.Lowpass(tau)
(num,), den, _ = cont2discrete((s.num, s.den), dt=0.001, method='zoh')
ss = nengo.LinearFilter(poly1d(num)**2, poly1d(den)**2, analog=False)

    a = nengo.Node(1)
    b = nengo.Node(size_in=1)
    nengo.Connection(a, b, synapse=s)
    p1 = nengo.Probe(b, synapse=s)
    p2 = nengo.Probe(a, synapse=ss)

    sim.run_steps(5)

print(np.allclose(sim.data[p1], sim.data[p2]))
print(sim.data[p1])
print(sim.data[p2])
```

On this PR we get the correct behaviour:

```
True
[[0.        ]
 [0.        ]
 [0.03285854]
 [0.08666313]
 [0.15274035]]
[[0.        ]
 [0.        ]
 [0.03285854]
 [0.08666313]
 [0.15274035]]
```

but on master we get:

```
False
[[0.        ]
 [0.        ]
 [0.03285854]
 [0.08666313]
 [0.15274035]]
[[0.        ]
 [0.03285854]
 [0.08666313]
 [0.15274035]
 [0.22487294]]
```

So that's a definite plus with this PR. But another thing that should be both tested and documented. The reason this happens is summarized in https://github.com/nengo/nengo/issues/938#issuecomment-497954987 as the difference between switching around the order of the equations (this PR) versus removal of a delay in each transfer function (master).

nengo/synapses.py
```diff
@@ -221,123 +223,130 @@ def evaluate(self, frequencies):
...
-        """An LTI step function for any given transfer function.
+        @classmethod
+        def check(cls, A, B, C, D, X):
+            return super(LinearFilter.OneX, cls).check(A, B, C, D, X) and (
+                len(A) == 1 and (D == 0).all())
 
-        Implements a discrete-time LTI system using the difference equation
-        [1]_ for the given transfer function (num, den).
+    class NoD(Step):
+        """An LTI step function for any given state-space system."""
```
arvoelke: Incorrect docstring (copied from `General`).

nengo/synapses.py
```diff
@@ -221,123 +223,130 @@ def evaluate(self, frequencies):
...
-                raise ValidationError("'den' must be length 1 (got %d)"
-                                      % len(den), attr='den', obj=self)
-
-            super().__init__(num, den, output)
-            self.b = num[0]
-            self.a = den[0]
-            if y0 is not None:
-                self.output[...] = y0
+    class OneX(Step):
+        """Step function for transfer functions with one state element."""
```
arvoelke: Also no passthrough (note you can have a one-dimensional state with a passthrough, but due to the `check`s below that will fall into the `General` case).

nengo/synapses.py
```diff
@@ -221,123 +223,130 @@ def evaluate(self, frequencies):
...
-                                      % len(den), attr='den', obj=self)
-            super().__init__(num, den, output)
-            self.b = num[0]
+        @classmethod
+        def check(cls, A, B, C, D, X):
+            if A.shape[0] == 0:
+                return X.shape[0] == B.shape[0] == 0 and D.shape[0] == 1
+            else:
+                return (A.shape[0] == A.shape[1] == B.shape[0] == C.shape[1]
+                        and D.shape[0] == B.shape[1] == 1
```
arvoelke: `C.shape[0] == 1` as well?

nengo/synapses.py
```diff
@@ -221,123 +223,130 @@ def evaluate(self, frequencies):
...
+        elif LinearFilter.OneX.check(A, B, C, D, X):
+            # OneX combines B and C into one scaling value `b`
+            b = B.item() * C.item()
+            X[:] = (b / (1 - A.item())) * y0
+        else:
+            # Solve for u0 (input) given y0 (output), then X given u0
+            assert B.ndim == 1 or B.ndim == 2 and B.shape[1] == 1
+            y0 = np.array(y0, copy=False, ndmin=2)
+            IAB = np.linalg.solve(np.eye(len(A)) - A, B)
+            Q = C.dot(IAB) + D  # multiplier from input to output
```
arvoelke: FYI, this is the DC gain. See discrete-time equation here https://www.mathworks.com/help/control/ref/dcgain.html and discussion in #1124, which matches the equation here.

nengo/synapses.py
```diff
@@ -221,123 +223,130 @@ def evaluate(self, frequencies):
...
+            assert Q.size == 1 and np.abs(Q.item()) > 1e-8
+            u0 = y0 / Q.item()
```
arvoelke: `Q` can be `0` in perfectly valid (stable) systems. For example,

```python
>>> nengo.LinearFilter([1, -1], [1, 0], analog=False).filt([1.], y0=1)
  File "/home/arvoelke/CTN/nengo/nengo/synapses.py", line 92, in filt
    shape_in, shape_out, dt, None, y0=y0, dtype=filtered.dtype)
  File "/home/arvoelke/CTN/nengo/nengo/synapses.py", line 260, in make_step
    assert Q.size == 1 and np.abs(Q.item()) > 1e-8
AssertionError
```

Note this has the discrete-time state-space matrices `(A, B, C, D) = (0, 1, -1, 1)`. So it is delaying the input by one time-step, and then subtracting it from the passthrough. For _any_ steady-state input, this produces a steady-state output of zero (hence zero DC-gain).

hunse: Added a ValidationError.

nengo/synapses.py
hunse: Noted in the code.

nengo/synapses.py
hunse: Done

nengo/synapses.py
hunse: Fixed

nengo/synapses.py
hunse: Fixed

hunse:
Ok, I think I've addressed the main issues here.

For parts of it, though, I'm not sure what tests/documentation would be helpful. For example, `LinearFilter([-1, 0, 1e6], [1, 2, 1]).filt([1000., 0., 0.], y0=0)` now produces the same output as master. But I don't think it's useful to test that that particular command has a particular output. What is the underlying thing being tested?

arvoelke:
Looks great thank you!

> But I don't think it's useful to test that that particular command has a particular output. What is the underlying thing being tested?

Maybe could test this synapse:
```python
du = nengo.LinearFilter([1, -1], [1, 0], analog=False)
```
to ensure that the output is a differentiated (and delayed by one time-step) version of the input (i.e., `np.diff` with a delay). I believe this should tap into the `General` stepper in the same way as the more convoluted example I gave above. But I think this example makes it semantically clear what is being tested?

hunse:
Ok, I've added such a test.

arvoelke:
> Ok, I've added such a test.

Great. And I noticed you don't need to compensate for the extra delay when you're using `filt` because that's only present in the `Simulator` (i.e., one time-step difference between `filt` and `Simulator`).

I don't think the example I gave in https://github.com/nengo/nengo/pull/1535#issuecomment-498067962 was documented or tested, which is an important change in this PR in my opinion. So I added such a test and changelog entry. I've also verified that this test fails prior to this PR. There's a subtle discussion point here that I hadn't actually mentioned yet pertaining to #938 in the case of `D = 0` (no passthrough). The old transfer function implementation can actually remove more than one delay. It will potentially remove _all_ of the delays in the transfer function, by implicitly treating the numerator `[1]` as `[1, 0, 0, ..., 0]` instead of `[0, 0, ..., 0, 1]`. This is doing the right thing now. I hadn't spotted it before because my `nengolib` implementation corrected this issue and my unit-testing comparing to `nengo` didn't include any examples where `len(den) - len(num) >= 2`. I can talk about this in greater length if desired, but hopefully the test and changelog entry suffice.

Since this maintains the previous behaviour of correctly implementing the requested transfer function (with an added delay in the case of `D != 0`), while also giving the correct behaviour for the example in https://github.com/nengo/nengo/pull/1535#issuecomment-498067962, I think this is a definite improvement that LGTM.

I'd like to have #1536 considered at a later time in order to remove the extra delay in the most general case, while keeping everything else the same, and potentially allowing the graph optimizer to merge the synaptic operations together.

CHANGES.rst
```diff
@@ -112,6 +112,10 @@ Release History
   which speed up build times slightly. To re-enable this testing,
   enable the ``fail_fast`` RC setting.
   (`#1532 <https://github.com/nengo/nengo/pull/1532>`__)
+- Higher-order synapses no longer remove unnecessary time-step delays when
```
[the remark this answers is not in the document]
arvoelke: Yes it is. For an example see `test_combined_delay` which failed prior to this PR. The explanation is in https://github.com/nengo/nengo/pull/1535#issuecomment-499200176. To be more accurate could say "Some higher-order synapses" rather than just "Higher-order synapses". Specifically, the ones that have `len(den) - len(num) >= 2` after discretization.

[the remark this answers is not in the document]
tbekolay: Okay cool, thanks!

tbekolay:
Okies, I reworded the changelog entries and added some new ones to try to be complete. With that, this PR is almost good to go, just curious about the other inline comment I added in regard to the difference between `scipy.linalg.expm` and `nengo._vendor.scipy.expm`.

nengo/utils/numpy.py
```diff
@@ -7,6 +7,12 @@
 import numpy as np
 
 from ..exceptions import ValidationError
+try:
+    from scipy.linalg import expm  # noqa: F401 pylint: disable=unused-import
```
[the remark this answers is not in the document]
hunse: I think my intention was in case Scipy improves their expm function in any way, we use the best. But I think that's unlikely, and probably worth the simplicity of always using the vendorized one.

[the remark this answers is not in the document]
tbekolay: Gotcha, will change to using the vendorized one :ok_hand:

Issue #1539, opened 2019-06-07
Some solvers compute only a single RMSE


arvoelke:
**Describe the bug**

Every solver returns a `solver_info['rmses']` (note: plural) which is expected to be the RMSE for each target (the number of post neurons for `weights=True`, or the number of dimensions for `weights=False`). However, some solvers (notably the most common ones) only return a single overall RMSE, and this can differ depending on whether `weights=True` or `weights=False`.

**To reproduce**

```python
import nengo
import numpy as np

def trial(solver, seed=0, n_pre=50, n_post=100, dim_post=1):

    with nengo.Network(seed=seed) as model:
        x = nengo.Ensemble(n_pre, 1)
        y = nengo.Ensemble(n_post, dim_post)
        conn = nengo.Connection(x, y[0], solver=solver)

    with nengo.Simulator(model, progress_bar=None) as sim:
        pass
    
    return(sim.data[conn].solver_info['rmses'].shape)

for cls in (nengo.solvers.Lstsq,
            nengo.solvers.LstsqNoise,
            nengo.solvers.LstsqMultNoise,
            nengo.solvers.LstsqL2,
            nengo.solvers.LstsqL2nz,
            nengo.solvers.LstsqL1,
            nengo.solvers.LstsqDrop,
            nengo.solvers.Nnls,
            nengo.solvers.NnlsL2,
            nengo.solvers.NnlsL2nz):
    print(cls.__name__, "W", trial(cls(weights=True)))
    print(cls.__name__, "D", trial(cls(weights=False), dim_post=2))
```

```
Lstsq W (1,)
Lstsq D (1,)
LstsqNoise W (1,)
LstsqNoise D (1,)
LstsqMultNoise W (1,)
LstsqMultNoise D (1,)
LstsqL2 W (1,)
LstsqL2 D (1,)
LstsqL2nz W (1,)
LstsqL2nz D (1,)
LstsqL1 W (100,)
LstsqL1 D (1,)
LstsqDrop W (100,)
LstsqDrop D (1,)
Nnls W (100,)
Nnls D (1,)
NnlsL2 W (100,)
NnlsL2 D (1,)
NnlsL2nz W (100,)
NnlsL2nz D (1,)
```

**Expected behavior**

Two ways this could be made consistent. 

1. The `rmses` are always across flattened arrays, in which case all of the shapes should be `(1,)`, and `rmses` (plural) should be renamed to `rmse` (singular).

[...]

PR #1540, opened 2019-06-09
Solver documentation fixup

arvoelke:
**Motivation and context:**
Fixes changelog and improves documentation of base solver class (confused me while I was looking into https://github.com/nengo/nengo/issues/1539).

None

**How has this been tested?**
N/A

**Types of changes:**

**Checklist:**

**Todo:**
 - [x] I think the solver docstring can be improved. In particular, `non-compositional` solvers don't necessarily work with neuron-to-neuron weight matrices if `weights=False`? This docstring makes the case sound different: https://github.com/nengo/nengo/blob/18c13fb417748630c9f07d329031de59b7d8353a/nengo/solvers.py#L21-L26

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -102,8 +102,8 @@ Release History
   when solving for weights.
   (`#1507 <https://github.com/nengo/nengo/pull/1507>`__)
 - Solvers no longer take encoders as an argument. Instead, encoders will
-  be applied to the targets before the solve function for compositional solvers
-  and applied by the Transform builder for non-compositional solvers.
+  be applied to the targets before the solve function for non-compositional
+  solvers and applied by the Transform builder for compositional solvers.
   (`#1507 <https://github.com/nengo/nengo/pull/1507>`__)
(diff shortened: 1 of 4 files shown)


arvoelke:
Pushed another commit which I think would have avoided #1539 being raised as an issue by me.

CHANGES.rst
```diff
@@ -82,8 +82,8 @@ Release History
   when solving for weights.
   (`#1507 <https://github.com/nengo/nengo/pull/1507>`__)
 - Solvers no longer take encoders as an argument. Instead, encoders will
-  be applied to the targets before the solve function for compositional solvers
-  and applied by the Transform builder for non-compositional solvers.
+  be applied to the targets before the solve function for non-compositional
+  solvers and applied by the Transform builder for compositional solvers.
```
arvoelke: Note: This still only applies to the case of `weights=True`. For the case of `weights=False` there should not be a difference. This is now clarified in the `Solver` documentation.

Issue #1547, opened 2019-06-27
Cannot change `n_neurons` post ensemble creation


xchoo:
**Describe the bug**

Trying to change the `n_neurons` attribute of an ensemble (post creation) causes an error message to be thrown.

**To reproduce**

A minimal code snippet to reproduce the behavior:

```python
import nengo

nengo.log(level="debug")

with nengo.Network() as net:
    ens = nengo.Ensemble(50, 1)
    ens.n_neurons = 100
```

**Expected behavior**

Previous versions of Nengo would not throw an error when doing this.

**Error message and logs**

The code throws the following error message:
```
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "/mnt/d/Users/xchoo/GitHub/nengo/nengo/base.py", line 107, in __setattr__
    super().__setattr__(name, val)
  File "/mnt/d/Users/xchoo/GitHub/nengo/nengo/config.py", line 455, in __setattr__
    raise exc_info[1].with_traceback(None)
nengo.exceptions.ReadonlyError: Ensemble.n_neurons: n_neurons is read-only and cannot be changed
```

**Versions**

 - OS: Windows 10, Ubuntu
 - Python: Anaconda Python 3.X
 - Nengo: latest master
 - Other: [e.g. NumPy v1.16.1 with MKL]

**Additional context**

Ran across this bug trying to do `n_neuron` parameter sweeps.

jgosmann:
This is due to fb8f5e80bf7430f503f511200ddc77c5c8ef9ebe/#1372 and I'm wondering whether this might have been on purpose. But in the discussion nothing hints at this, so probably this is indeed a bug.

tcstewar:
I also ran into this during the summer school...  I had a vague memory of discussing ages ago whether or not `n_neurons` should be configurable, but I'd definitely like it to be!  This does mean, of course, that it's possible to build broken models (the transform on a Connection doing to an Ensemble.neurons may now be the wrong size), but I do think that's worth it....

drasmuss:
After some more discussion, we think this is the desired behaviour (for now). Allowing n_neurons to change post-initialization can introduce all kinds of errors, and the errors tend to be obscure and hard to diagnose/fix because they don't occur until the build stage. 

[...]

PR #1548, opened 2019-06-28
Set n_neurons/dimensions to not read only

drasmuss:
**Motivation and context:**

Fixes https://github.com/nengo/nengo/issues/1547

**How has this been tested?**

Added a quick test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -182,6 +182,9 @@ Release History
 - Some higher-order ``LinearFilter`` synapses had unnecessary delays
   that have now been removed.
   (`#1535 <https://github.com/nengo/nengo/pull/1535>`__)
+- ``Ensemble.n_neurons/dimensions`` can now be changed after Ensemble
+  creation again.
+  (`#1548 <https://github.com/nengo/nengo/pull/1548>`__)
 
 2.8.0 (June 9, 2018)
 ====================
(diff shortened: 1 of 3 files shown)


jgosmann:
Doing something like
```python
with nengo.Network() as  model:
    ens = nengo.Ensemble(2, 1, max_rates=[10, 20, 30])
```
one immediately gets a useful error:
```
ValidationError: Ensemble.max_rates: shape[0] should be 2 (got 3)
```
However, if trying to do
```

with nengo.Network() as  model:
    ens = nengo.Ensemble(3, 1, max_rates=[10, 20, 30])
    ens.n_neurons = 2
```
no error is raised until one tries to build the model and the error message (and stack trace) are rather unhelpful:
```
ValueError: operands could not be broadcast together with shapes (3,) (2,)
```

* Could this be the reason for disallowing changing `n_neurons` and `dimensions`?
* Would it be possible to revalidate parameters? (Though, if you wanted to change `n_neurons` and `max_rates`, you would need a atomic operation.)

drasmuss:
After some more discussion, we're going to close this and leave the behaviour as-is for now (i.e. `n_neurons`/`dimensions` will stay read-only). 

There are just all kinds of ways that changing those values post-initialization can break other parts of the model (like the example pointed out above), and properly validating those changes to catch errors and provide informative error messages would require some significant changes to how parameter validation currently works.  Which is not to say we couldn't do that, if a compelling use-case arises. But so far there doesn't seem to be a strong need for it (e.g. these parameters have been read-only for almost 2 years without causing any issues), so we don't have any plans to undertake that work at the moment.

PR #1550, opened 2019-06-28
Add SpikingRectifiedLinear to cache whitelist

### LINKED ISSUE #1549: SpikingRectifiedLinear does not cache decoders
**Describe the bug**

`SpikingRectifiedLinear` does not cache decoders.

**To reproduce**

A minimal code snippet to reproduce the behavior:

```python
import nengo
from nengo.cache import Fingerprint

print(Fingerprint.supports(nengo.neurons.SpikingRectifiedLinear()))
print(Fingerprint.supports(nengo.neurons.RectifiedLinear()))
```

**Expected behavior**

```
True
True
```

**Error message and logs**

```
False
True
```

**Versions**

 - OS: N/A
 - Python: N/A
 - Nengo: latest master
 - Other: N/A

**Additional context**

May need a better process for keeping these things in sync when new models / etc are added. Have not gone through to check for others that could be missing from the whitelist.

arvoelke:
Previously ensembles with this neuron type would not have its decoders cached.

**Motivation and context:**
Closes #1549.

**How has this been tested?**
Added a unit test for this isolated incident as a short-term solution.

**Types of changes:**

**Checklist:**
- [N/A] I have updated the documentation accordingly.

### DIFF
--- a/nengo/cache.py
+++ b/nengo/cache.py
@@ -29,6 +29,7 @@ from nengo.neurons import (
     LIFRate,
     RectifiedLinear,
     Sigmoid,
+    SpikingRectifiedLinear,
 )
 from nengo.rc import rc
 from nengo.solvers import (
@@ -172,10 +173,11 @@ class Fingerprint:
         LIFRate,
         RectifiedLinear,
         Sigmoid,
+        SpikingRectifiedLinear,
     )
 
     WHITELIST = set(
-        (bool, float, complex, bytes, list, tuple, np.ndarray, int, str)
+        (type(None), bool, float, complex, bytes, list, tuple, np.ndarray, int, str)
         + SOLVERS
         + LSTSQ_METHODS
         + NEURON_TYPES
--- a/nengo/tests/test_cache.py
+++ b/nengo/tests/test_cache.py
@@ -1,6 +1,7 @@
 import errno
 import multiprocessing
 import os
+import sys
 import timeit
 from subprocess import CalledProcessError
 
@@ -17,7 +18,36 @@ from nengo.cache import (
     WriteableCacheIndex,
 )
 from nengo.exceptions import CacheIOWarning, FingerprintError
+import nengo.neurons
 from nengo.solvers import LstsqL2
+import nengo.utils.least_squares_solvers
+from nengo.utils.testing import Mock
+
+
+def isstrictsubclass(x, cls):
+    return isinstance(x, type) and issubclass(x, cls) and x is not cls
+
+
+def list_objects(module):
+    return [getattr(module, key) for key in dir(module) if not key.startswith("_")]
+
+
+neuron_types = [
+    obj
+    for obj in list_objects(nengo.neurons)
+    if isstrictsubclass(obj, nengo.neurons.NeuronType)
+]
+solver_types = [
+    obj
+    for obj in list_objects(nengo.solvers)
+    if isstrictsubclass(obj, nengo.solvers.Solver)
+    and obj is not nengo.solvers.NoSolver  # NoSolver is not cacheable
+]
+lstsq_solver_types = [
+    obj
+    for obj in list_objects(nengo.utils.least_squares_solvers)
+    if isstrictsubclass(obj, nengo.utils.least_squares_solvers.LeastSquaresSolver)
+]
 
 
 class SolverMock:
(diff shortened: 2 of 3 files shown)


nengo/tests/test_cache.py
```diff
@@ -299,6 +299,13 @@ def test_unsupported_fingerprinting(obj):
         Fingerprint(obj)
 
 
+@pytest.mark.parametrize("obj", [
+    nengo.SpikingRectifiedLinear(),  # issue 1549
+])
+def test_supported_fingerprinting(obj):
+    assert Fingerprint.supports(obj)
```
jgosmann: Not sure, how much value this tests add, but I guess there might be some as `supports` is doing a little bit more than testing list inclusion.

But shouldn't this test then be done for all whitelisted types?

arvoelke: This was to satisfy the required guideline _"I have added tests to cover my changes."_ I can add some extra testing, but I was expecting someone might first have an idea to ensure this stays in sync with new models/etc.

arvoelke: One option could be to create a whitelist decorator and then prepend that to anything we want in the whitelist. That way it is localized alongside the object being whitelisted, rather than hidden within `cache.py`.

hunse: We could have this test test all `NeuronTypes` in `neurons.py`. Then if there's one that isn't in the whitelist, we should catch it right away.

jgosmann: > This was to satisfy the required guideline _"I have added tests to cover my changes."_

I would think of this as a guideline, not a strict rule. Tests should add some actual benefit and not just satisfy this guideline.



> One option could be to create a whitelist decorator and then prepend that to anything we want in the whitelist.

That would make `neurons.py` depend on `cache.py` which might be a circular reference that I would like to try to avoid.

nengo/cache.py
```diff
@@ -105,7 +106,9 @@ def check_seq(tpl):
 
 def check_attrs(obj):
     attrs = [getattr(obj, x) for x in dir(obj) if not x.startswith("_")]
-    return all(Fingerprint.supports(x) for x in attrs if not callable(x))
+    return all(
+        Fingerprint.supports(x) for x in attrs if not callable(x) and x is not None
```
arvoelke: When might it be `None`?

arvoelke:
@hunse very nice, I like this solution. For reference I printed out the list that is auto-generated to parameterize the test, and it looks like so:
```
<class 'nengo.neurons.AdaptiveLIF'>
<class 'nengo.neurons.AdaptiveLIFRate'>
<class 'nengo.neurons.Direct'>
<class 'nengo.neurons.Izhikevich'>
<class 'nengo.neurons.LIF'>
<class 'nengo.neurons.LIFRate'>
<class 'nengo.neurons.RectifiedLinear'>
<class 'nengo.neurons.Sigmoid'>
<class 'nengo.neurons.SpikingRectifiedLinear'>
<class 'nengo.solvers.Lstsq'>
<class 'nengo.solvers.LstsqDrop'>
<class 'nengo.solvers.LstsqL1'>
<class 'nengo.solvers.LstsqL2'>
<class 'nengo.solvers.LstsqL2nz'>
<class 'nengo.solvers.LstsqMultNoise'>
<class 'nengo.solvers.LstsqNoise'>
<class 'nengo.solvers.Nnls'>
<class 'nengo.solvers.NnlsL2'>
<class 'nengo.solvers.NnlsL2nz'>
<class 'nengo.utils.least_squares_solvers.BlockConjgrad'>
<class 'nengo.utils.least_squares_solvers.Cholesky'>
<class 'nengo.utils.least_squares_solvers.Conjgrad'>
<class 'nengo.utils.least_squares_solvers.ConjgradScipy'>
<class 'nengo.utils.least_squares_solvers.LSMRScipy'>
<class 'nengo.utils.least_squares_solvers.RandomizedSVD'>
<class 'nengo.utils.least_squares_solvers.SVD'>
```
Briefly eye-balling this, it looks correct? Your changes LGTM.

nengo/cache.py
jgosmann: Whenever an attribute on an object is set to `None`?

hunse: Yeah, this came up with I think the `Cholesky` solver and maybe a few others that have a `transpose` parameter which by default is `None` (meaning the solver can choose which way to do the computation internally). `Fingerprint` seems to have no problem hashing these. I think it makes more sense to add `NoneType` to the whitelist, though, so I went and did that instead.

Also, this raised the point that the test only tests whether `Fingerprint.supports` is true for the type. I added a second part to the test to make sure it actually hashes without a problem (just in case something gets on the whitelist that isn't hashable for some reason).

PR #1555, opened 2019-07-27
Apply Black to source (and otherwise update to Bones 0.4.1)

tbekolay:
**Motivation and context:**
Nengo Bones had a release recently, so we'll need to update for any other PR to pass tests. As part of the release, we are autoformatting the source with Black, which is done in a separate commit so that it can ignored when looking through blames.

**Types of changes:**

**Checklist:**


hunse:
Can we keep Black out of the `_vendor` folder? I'd prefer to leave those untouched so it's easy to diff in the future.

tbekolay:
> Can we keep Black out of the `_vendor` folder? I'd prefer to leave those untouched so it's easy to diff in the future.

Yes definitely, forgot about the `_vendor` folder. Will update.

PR #1557, opened 2019-07-31
Docstyle fixups

hunse:
**Motivation and context:**

While `pydocstyle` doesn't appear to be a useful tool to apply regularly with the static checks, it did make some suggestions that help make our docstrings slightly more uniform.

**How has this been tested?**

Documentation changes only.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/solvers.py
+++ b/nengo/solvers.py
@@ -1,4 +1,10 @@
-"""Functions concerned with solving for decoders or full weight matrices."""
+"""Classes concerned with solving for decoders or full weight matrices.
+
+.. inheritance-diagram:: nengo.solvers
+   :parts: 1
+   :top-classes: nengo.solvers.Solver, nengo.solvers.SolverParam
+
+"""
 
 import logging
 import time
@@ -12,12 +18,11 @@ from nengo.utils.least_squares_solvers import (
     rmses,
     LeastSquaresSolverParam,
 )
-from nengo.utils.magic import DocstringInheritor
 
 logger = logging.getLogger(__name__)
 
 
-class Solver(FrozenObject, metaclass=DocstringInheritor):
+class Solver(FrozenObject):
     """Decoder or weight solver.
 
     A solver can be compositional or non-compositional. Non-compositional
@@ -64,6 +69,8 @@ class Solver(FrozenObject, metaclass=DocstringInheritor):
 
 
 class SolverParam(Parameter):
+    """A parameter in which the value is a `.Solver` instance."""
+
     def coerce(self, instance, solver):
         self.check_type(instance, solver, Solver)
         return super().coerce(instance, solver)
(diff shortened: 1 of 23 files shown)


nengo/solvers.py
```diff
@@ -116,7 +116,8 @@ class _LstsqNoiseSolver(Solver):
     solver = LeastSquaresSolverParam("solver")
 
     def __init__(self, weights=False, noise=0.1, solver=lstsq.Cholesky()):
-        """
+        """Create new solver.
```
tbekolay: I think this (and the other changes to `__init__` methods) will result in some pretty awkward documentation in the end ... Sphinx concatenates together the class docstring and the `__init__` docstring, so this sentence will appear in after the class docstring because we set `autoclass_content = "both"` in `conf.py`.

At the same time, I agree that if anyone happened to look at the `__init__` docstring in isolation, it would be awkward to just see the parameter lists and whatnot.

The reason that we don't just put the parameters etc in the class docstring is because there's a base class implementation that many (but not all) solvers use. So some solvers override the basic description (class docstring) but not the parameter / attribute lists (`__init__` docstring).

So what we have right now basically results in good Sphinx docs, but is not great for reading the docstrings in source or when using `help`.

One possible way forward that still lets us reduce duplication of parameter lists would be to use a simple decorator, as we did in pytest-allclose (see https://github.com/nengo/pytest-allclose/blob/master/pytest_allclose/plugin.py#L9). That way everything will be in the class docstring, and we simply leave `__init__` undocumented (if I wanted to know about a class, I definitely wouldn't think to do `help(ClassName.__init__)`.

Any other ideas?

nengo/builder/connection.py
```diff
@@ -225,7 +225,6 @@ def build_connection(model, conn):
     -----
     Sets ``model.params[conn]`` to a `.BuiltConnection` instance.
     """
-
```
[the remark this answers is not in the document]
hunse: It's a function, so `pydocstyle` requests that it not be there, which is why I did it. But in the new nengo-bones pydocstyle PR, I think you disabled that check, so I'm fine if you want to change them back.

nengo/solvers.py
hunse: The simple decorator seems straightforward, even though it's a bit of work to set up.

The only other thing I can think of is a fancy decorator that goes through base classes, finds the "Parameters" (or other sections) in the docstring, and adds the things in there to the parameter list. Seems like something Sphinx or some other package ought to have. But I don't think it's worth us putting in that work at this point.

hunse:
I'm fine for you to go ahead if you want @tbekolay. Otherwise, I could do them tomorrow.

tbekolay:
Cool, shouldn't take long so I'll do it now.

nengo/solvers.py
tbekolay: Implemented the somewhat simple decorators; the Nnls one was a bit less simple, but reduced any duplication, which is better than what is currently done.

nengo/solvers.py
```diff
@@ -302,6 +314,28 @@ class LstsqDrop(Solver):
...
+    Attributes
+    ----------
+    drop : float
+        Fraction of decoders or weights to set to zero.
+    solver1 : Solver
+        Solver for finding the initial decoders.
+    solver2 : Solver
+        Used for re-solving for the decoders after dropout.
+    weights : bool
+        If False, solve for decoders. If True, solve for weights.
```
hunse: This is a general comment, but there's often a lot of redundancy between our "Parameters" and "Attributes" sections.

hunse:
Everything LGTM!

nengo/solvers.py
tbekolay: Yeah for sure, it would be nice to reduce that duplication in a way that isn't too hard to read. Might be a thing to do when also moving the parameter stuff to a separate package? If we want to do that?

drasmuss: I find the Attributes sections tend to get out of date a lot as well (since it's harder to line them up with the code, versus just looking at the function definition).

tbekolay: Made #1558 to track that.

Issue #1534, opened 2019-05-29
Gains grow without bound as the neuron count is increased


arvoelke:
### Related Issues

 - https://github.com/nengo/nengo/issues/1212
 - https://github.com/nengo/nengo/pull/1243#issuecomment-267530076

### Background Analysis

As the intercept of a neuron, `c`, approaches `1` from the left-hand side, the gain on the response curve `r` grows without bound. To make this precise, the curve must transition from `r(c) = 0` to `r(1) = m`, where `m` is the `max_rate`, which is by default sampled from `U[200, 400)`. This results in a tangent of `(m - 0) / (1 - c)` between these two points. As `c -> 1` this tangent goes off to infinity.

Let `n` be the number of neurons. The probability, `p`, of generating at least one uniform random variable from `[-1, 1)` that falls within the interval `[c, 1)` is:

![CodeCogsEqn (1)](https://user-images.githubusercontent.com/5420057/58518360-eb974d00-817c-11e9-8c92-efab2ca9e46b.gif)

Plugging this into the equation for the tangent gives a slope of:

![CodeCogsEqn (2)](https://user-images.githubusercontent.com/5420057/58518510-83953680-817d-11e9-8357-dcaf7aeda67c.gif)

### Example Manifestation

For example, given `20000` neurons there is a **99% chance** that at least one of the ReLU's gains is on the order of 10^6 (see this by plugging in `m = 300`, `p = 0.99`, `n = 20000` into Mathematica). This is validated by the following figure+code, which also demonstrates that this happens regardless of neuron model (although the analysis for the precise gain differs by a constant factor for models other than ReLU):

![intercept_vs_gain](https://user-images.githubusercontent.com/5420057/58518627-e25ab000-817d-11e9-91c2-6a01040241bf.png)

```python
import nengo

import numpy as np
import seaborn as sns
import matplotlib.pyplot as plt

def sample(neuron_type, n_batches=20, batch_size=1000, seed=0):
    intercepts = np.empty(n_batches * batch_size)
    gain = np.empty(n_batches * batch_size)
    for i in range(n_batches):
        with nengo.Network(seed=seed+i) as model:
            ens = nengo.Ensemble(n_neurons=batch_size, dimensions=1,
                                 neuron_type=neuron_type)
        with nengo.Simulator(model, progress_bar=False) as sim:
            pass
        intercepts[i*batch_size:(i+1)*batch_size] = sim.data[ens].intercepts
[...]

PR #1561, opened 2019-08-07
Change point at which neurons achieve max_rates

hunse:
The new `NeuronType._max_x` controls the point at which neurons
achieve their max rates. By moving this point above 1 (which it
was before), ensembles that use intercepts in (-1, 1) will no
longer have arbitrarily large gains as intercepts approach 1.

**Motivation and context:**

Addresses #1534.

**How has this been tested?**

Will post plots below.

**Types of changes:**

**Checklist:**

**TODO**
- [ ] Fix `Ensemble.max_rates` documentation.


arvoelke:
As a quick check, this is equivalent to `LIFRateSafe` from #1534 and not `LIFRateUnbiased`?

gain_distributions 
(Note: log-scaled y-axis)

Also might this interact with the need for Loihi intercept adjustments, e.g., in https://github.com/nengo/nengo-loihi/pull/126?

hunse:
Here's a plot of the old gains vs intercepts and the new gains (error bars show 1 standard deviation).

neuron_gains1 5

jgosmann:
I don't have the time to think about his a lot right now. But setting `_max_x` seems rather arbitrary and more like a ad hoc workaround than a proper solution. I think it also would be quite surprising if one tries to get a specific max firing rate at |x|=1. At the very least I think this should be properly documented (in some appropriate place in the documentation, not just the changelog).

hunse:
BTW, I've changed this to use `max_x = 1.2` instead of 1.5. The more that edge neurons are not allowed to have high gains, the more error we have around the edges of our representation. This is because large weights would be needed to get these edge neurons to really contribute, and the regularization on the solvers keeps them from generating large weights. There's an argument that we should have less regularization on the solvers (I played around with it a bit), but less regularization means more spike noise (among other things). I think we'd really need to do a deeper study before we make changes. `LstsqL2nz` might be the first place to look, because it bases the amount of regularization for each neuron on the number of non-zero values in the neuron's tuning curve, and is meant to help edge neurons not be over-regularized (as they currently are). The reason we didn't switch over to that is that it changed the saturation behaviour of the ensemble for values outside (-1, 1), since edge neurons are now contributing much more. But maybe with the change to `max_x`, the effects there will be a little different.

Here are what the new gains look like with `max_x == 1.2`.

neuron_gains1 2

> setting `_max_x` seems rather arbitrary and more like a ad hoc workaround than a proper solution. I think it also would be quite surprising if one tries to get a specific max firing rate at |x|=1. At the very least I think this should be properly documented (in some appropriate place in the documentation, not just the changelog).

The point at which we evaluate the max rates has always been arbitrary. Previously it was 1, now it's 1.2. I made a note above to re-write the Ensemble documentation to reflect this (I think that's the main place we describe max_rates).

jgosmann:
> The point at which we evaluate the max rates has always been arbitrary.

I think 1 is less arbitrary because it corresponds to the default radius.

I might actually find it more intuitive if the default intercept distribution would be something like `Uniform(-0.8, 0.8)` instead of `Uniform(-1, 1)` which should have a similar effect. Still arbitrary, so I am also not super happy with it, but seems more transparent to the user than a hidden parameter to me.

tbekolay:
Since this change affects almost all models in ways we might not be able to predict well, we decided to merge in #1563 first so we could look at the relative RMSE values across all tests to ensure that the change in this PR was not detrimental. Additionally, we discussed the possibility of just changing the default intercepts, which we preferred over the `_max_x` change, assuming the RMSE values are not worse.

Here are the results of my testing. I ran the whole test suite for

- what's in master now
- five different `_max_x` values
- changing default intercepts to `Uniform(-1, 1-x)` for x in 0.01, 0.05, 0.1, 0.2 (shrinking upper bound)
- changing default intercepts to `Uniform(-1+x, 1-x)` for x in 0.01, 0.05, 0.1, 0.2 (shrinking upper and lower bounds)
- changing default intercepts to `Uniform(-1-x, 1-x)` for x in 0.01, 0.05, 0.1, 0.2 (shifting bounds negatively)

| Condition                 | Mean relative RMSE     | Tests              |
|---------------------------|------------------------|--------------------|
| master                    | _**0.06122 +/- 0.3240**_     | :heavy_check_mark: |
| max\_x=1.01               | 0.06006 +/- 0.3183     | 2 :x:              |
| max\_x=1.05               | 0.05907 +/- 0.3130     | :heavy_check_mark: |
| max\_x=1.1                | 0.05873 +/- 0.3116     | :heavy_check_mark: |
| max\_x=1.2                | **0.05840 +/- 0.3103** | :heavy_check_mark: |
| max\_x=1.3                | *0.05832 +/- 0.3098*   | 3 :x:              |
| intercepts=U(-1, 0.99)    | 0.06010 +/- 0.3180     | :heavy_check_mark: |
| intercepts=U(-1, 0.95)    | **0.05898 +/- 0.3130** | :heavy_check_mark: |
| intercepts=U(-1, 0.9)     | 0.05862 +/- 0.3118     | 1 :x:              |
| intercepts=U(-1, 0.8)     | *0.05830 +/- 0.3105*   | 1 :x:              |
| intercepts=U(-0.99, 0.99) | 0.06009 +/- 0.3180     | :heavy_check_mark: |
| intercepts=U(-0.95, 0.95) | 0.05910 +/- 0.3131     | 1 :x:              |
| intercepts=U(-0.9, 0.9)   | **0.05875 +/- 0.3116** | :heavy_check_mark: |
| intercepts=U(-0.8, 0.8)   | *0.05841 +/- 0.3103*   | 1 :x:              |
| intercepts=U(-1.01, 0.99) | **0.06011 +/- 0.3180** | :heavy_check_mark: |
| intercepts=U(-1.05, 0.95) | 0.05901 +/- 0.3134     | 2 :x:              |
| intercepts=U(-1.1, 0.9)   | 0.05870 +/- 0.3119     | 1 :x:              |
| intercepts=U(-1.2, 0.8)   | *0.05753 +/- 0.3077*   | 2 :x:              |

Notes:

- bolded values are the best for that test set with all tests passing
- italicized values are the best for that test set regardless of tests passing
- the `_max_x` changes were reverted before doing the default intercept tests
- some tests were changed in that commit, which were not changed in the intercept tests
- I didn't modify tests in the intercepts tests, nor did I look at the test failures to see how easy it would be to fix those tests

Conclusion:

From this table, my thought is that we should change the default intercepts to `Uniform(-1, 0.95)`, because

- it's the smallest delta from what we currently have now
- it avoids the most troubling situation where the intercept is near 1
- it doesn't change anything for people who have specified intercepts manually, which they presumably did for a reason and we shouldn't change their model without a really good reason

----

I'm going to push that change to this branch now, but we can always change it before merging if anyone has good reasons to go with a different value.

hunse:
What are the tests that aren't passing for the intercept ranges (-1, 0.9) and (-1, 0.8)? My inclination is to make the max intercept slightly smaller still, just to limit the range of gains a bit more. I think having a more limited range of gains is advantageous on fixed-point hardware, so if we can have people designing networks like that from the start, it might make it easier to transfer models to that hardware.

For anyone interested, here are the distributions of gains for the original LIF, the new LIF with `max_x=1.2`, and the original LIF with max intercepts of 0.95 and 0.9.

Figure_1

tbekolay:
> What are the tests that aren't passing for the intercept ranges (-1, 0.9) and (-1, 0.8)?

I didn't write them down. I'm sure they're fixable if we have some reasons to limit the range more, the fixed point hardware note makes sense.

jgosmann:
>     * it doesn't change anything for people who have specified intercepts manually, which they presumably did for a reason and we shouldn't change their model without a really good reason

I like that especially for Nengo SPA which uses the CosineSimilarity distribution that already makes high gains less likely.

One interesting point that ocured to me: for a 1-dimensional ensemble the CosineSimilarity distribution reduces to the Uniform(-1, 1) distribution. This is a nice property, because it makes the 1-dimensional case a special case of a more general case. This is sort of lost by limiting the intercept distribution to a smaller range.

Regarding the RMSE values ... I'm not sure if the data is sufficient to deduce a difference in the different conditions. The standard deviation (I assume that's the number behind the +/-) is much larger then the mean. I think, the standard error (considering the sample size) might more useful or if you want to get fancy do a statistical test to compare the means (though I assume to properly do the statistics might get complicated). But I suppose the data is enough to see that the change won't have a huge impact on the RMSE, which is probably good enough.

tbekolay:
> I'm not sure if the data is sufficient to deduce a difference in the different conditions.

We're not looking to publish a paper here, it's just a sanity check to make sure we're not significantly worsening performance.

> the standard error (considering the sample size) might more useful or if you want to get fancy do a statistical test to compare the means

You're right, standard error would be better here. Made https://github.com/nengo/pytest-allclose/issues/4 to track that.

tbekolay:
I went down to `Uniform(-1, 0.9)`; I think 0.8 is too big a range to cut off. I also fixed both of the tests that were failing for 0.9 and 0.8; it was kind of difficult to find ranges for `test_weights` that passed for both 0.9 and 0.8 so there might be some more serious issues for that test. Didn't want to spend the time to think about it at the moment.

I also looked at all the examples rendered in our docs to make sure nothing looks off, and they all look essentially the same. I did notice a minor plotting bug in the `two_neurons` example, which I fixed, and I don't think needs a separate commit.

I'll leave this open for the rest of the day in case anyone has any more concerns. Tomorrow morning I'll fix up the history and merge.

Issue #1543, opened 2019-06-13
Error in optimizer when signal contains nans


drasmuss:
**Describe the bug**

1. All-zero encoders will become `nan`, because we normalize them (and they have length 0, so dividing by 0).

2. The optimizer has a line that asserts `np.all(sig0 == sig1)` (paraphrasing).  But in numpy, `np.nan != np.nan`, so if the signals contain any nan values that assertion will fail.

**To reproduce**

```python
import nengo
import numpy as np

with nengo.Network() as net:
    a = nengo.Ensemble(10, 1, encoders=np.zeros((10, 1)))
    b = nengo.Ensemble(10, 1, encoders=np.zeros((10, 1)))

with nengo.Simulator(net) as sim:
    pass
```

**Expected behavior**

The optimizer shouldn't produce any errors, regardless of the values in the signals.  

We should probably also give a warning for length zero encoders.

We should probably also give a warning when signals contain nans (outside the optimizer), as it is likely unintentional and would make it easier to track down these kinds of bugs.

**Error message and logs**

```
Traceback (most recent call last):
  File "...\test.py", line 8, in <module>
    with nengo.Simulator(net) as sim:
  File "...\nengo\nengo\simulator.py", line 157, in __init__
    opmerge_optimize(self.model, self.dg)
  File "...\nengo\nengo\builder\optimizer.py", line 87, in optimize
    single_pass(only_merge_ops_with_view)
  File "...\nengo\nengo\builder\optimizer.py", line 161, in __call__
    self.perform_merges()
  File "...\nengo\nengo\builder\optimizer.py", line 187, in perform_merges
    self.perform_merges_for_subset(by_type[optype])
  File "...\nengo\nengo\builder\optimizer.py", line 214, in perform_merges_for_subset
    self.perform_merges_for_view_subset(view_subset)
  File "...\nengo\nengo\builder\optimizer.py", line 275, in perform_merges_for_view_subset
    self.merge(tomerge)
  File "...\nengo\nengo\builder\optimizer.py", line 287, in merge
    merged_op, merged_sig = OpMerger.merge(tomerge.ops)
  File "...\nengo\nengo\builder\optimizer.py", line 475, in merge
    return cls.mergers[type(ops[0])].merge(ops)
  File "...\nengo\nengo\builder\optimizer.py", line 682, in merge
    assert np.all(s.initial_value == A_sigr[s].initial_value)
AssertionError
```

**Versions**

 - OS: Windows 10
 - Python: 3.6
[...]

PR #1571, opened 2019-10-14
Guard against zero-length encoders

tbekolay:
**Motivation and context:**
Having NaN scaled encoders does not give a great error message. See #1543 for more details.

**How has this been tested?**
Added a few tests to ensure that the right error messages are being raised. The only thing not explicitly tested is that the check in the `DotIncMerger` succeeds if there are NaNs, but that situation should not be possible after the changes in this PR, so testing it seemed like overkill.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/ensemble.py
+++ b/nengo/builder/ensemble.py
@@ -189,6 +189,12 @@ def build_ensemble(model, ens):
         encoders = npext.array(ens.encoders, min_dims=2, dtype=rc.float_dtype)
     if ens.normalize_encoders:
         encoders /= npext.norm(encoders, axis=1, keepdims=True)
+    if np.any(np.isnan(encoders)):
+        raise BuildError(
+            "NaNs detected in %r encoders. This usually means that you had zero-length "
+            "encoders that were normalized, resulting in NaNs. Ensure all encoders "
+            "have non-zero length, or set `normalize_encoders=False`." % ens
+        )
 
     # Build the neurons
     gain, bias, max_rates, intercepts = get_gain_bias(ens, rng, dtype=rc.float_dtype)
--- a/nengo/builder/signal.py
+++ b/nengo/builder/signal.py
@@ -61,7 +61,7 @@ class Signal:
         offset=0,
     ):
         if self.assert_named_signals:
-            assert name
+            assert name is not None
         self._name = name
 
         if initial_value is None:
--- a/nengo/builder/signal.py
+++ b/nengo/builder/signal.py
@@ -71,6 +71,11 @@ class Signal:
             assert initial_value.shape == shape
 
         self._initial_value = initial_value
+        if np.any(
+            np.isnan(self._initial_value.data if self.sparse else self._initial_value)
+        ):
+            raise SignalError("%r contains NaNs." % self)
+
         if self.sparse:
             assert initial_value.ndim == 2
             assert offset == 0
--- a/nengo/spa/utils.py
+++ b/nengo/spa/utils.py
@@ -5,7 +5,7 @@ import numpy as np
 import nengo
 import nengo.utils.numpy as npext
 from nengo.exceptions import ValidationError
-from nengo.spa.vocab import VocabularyParam
+from nengo.spa.vocab import Vocabulary, VocabularyParam
 
 
 def enable_spa_params(model):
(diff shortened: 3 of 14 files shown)


nengo/builder/ensemble.py
```diff
@@ -189,6 +189,13 @@ def build_ensemble(model, ens):
         encoders = npext.array(ens.encoders, min_dims=2, dtype=rc.float_dtype)
     if ens.normalize_encoders:
         encoders /= npext.norm(encoders, axis=1, keepdims=True)
+    n_nans = np.sum(np.isnan(encoders))
+    if np.any(np.isnan(encoders)):
```
jgosmann: The calculation of `n_nans` could be move into the if clause or the if condition could be based on `n_nans`.

nengo/builder/signal.py
```diff
@@ -264,6 +272,18 @@ def sparse(self):
         """(bool) Whether the signal is sparse."""
         return is_sparse(self.initial_value)
 
+    def check_for_nans(self):
+        if self.allow_nan:
+            return
```
jgosmann: This if clause is redundant. The function will only be invoked if `allow_nan` is `False` except if someone calls it externally (in which case it is debatable which behaviour would be expected).

nengo/builder/ensemble.py
tbekolay: Good catch!

nengo/builder/signal.py
tbekolay: It's only invoked in one place now, but it could be used elsewhere later. But the expected behavior is a little ambiguous, so I'm going to change how it works.

nengo/builder/signal.py
```diff
@@ -42,6 +42,9 @@ class Signal:
         is a view already (in which case the signal base offset will be 0
         because it starts where the view starts. That NumPy view can have
         an offset of itself).
+    allow_nan : bool, optional
```
drasmuss: Would there ever be a case where someone would want to set this to True?

drasmuss: Just went ahead and removed it to keep things simple, but can add it back in if you had a use case in mind!

nengo/spa/utils.py
```diff
@@ -5,7 +5,7 @@
 import nengo
 import nengo.utils.numpy as npext
 from nengo.exceptions import ValidationError
-from nengo.spa.vocab import VocabularyParam
+from nengo.spa.vocab import VocabularyParam, Vocabulary
```
tbekolay: Vocab should go before VocabParam (alphabetically)

nengo/builder/ensemble.py
```diff
@@ -189,12 +189,11 @@ def build_ensemble(model, ens):
         encoders = npext.array(ens.encoders, min_dims=2, dtype=rc.float_dtype)
     if ens.normalize_encoders:
         encoders /= npext.norm(encoders, axis=1, keepdims=True)
-    n_nans = np.sum(np.isnan(encoders))
-    if n_nans > 0:
+    if np.any(np.isnan(encoders)):
         raise BuildError(
-            "%d encoders in %r are NaN. This usually means that you had zero-length "
+            "NaNs detected in %r encoders. This usually means that you had zero-length "
```
tbekolay: Why remove the number of nans? Seems like useful information. I guess the downside is that the warning could show up multiple times for the same ensemble?

nengo/builder/signal.py
tbekolay: The only time we intentionally use NaNs that I know of is in Nengo GUI for sliders to indicate which dimensions are overridden by the GUI and which are to continue as normal. We could always do that another way though, and that situation probably still wouldn't trip the NaN error because it happens _during_ the simulation, not in the `initial_value`. So I'm fine removing it until we get a real use case.

nengo/builder/ensemble.py
drasmuss: Just didn't seem like useful information to me (I frequently want to know whether there are nans, but I have never cared whether there are 1 or 10, it's just a binary error). So removing that just allows us to keep the code simpler (and slightly faster, since `np.any` will be quicker than `np.sum`).  Relatively minor advantages, but slightly outweighs the even more minor benefit of knowing how many nans there are. Those were my thoughts anyway!

tbekolay: My main reason for including it is that I can basically see two situations where you end up with zero length encoders. One is that you're doing random sampling poorly, like you're sampling a quantized range and you include zero in that range so you end up with 1 or 2 encoders that are zero. The other is that you are just setting all encoders to zero for some reason. But yeah, I guess in both of those situations, the underlying issue is one or more zeros, so it probably isn't important to differentiate :+1:

PR #1576, opened 2019-10-30
Update two_neurons.ipynb

pedrombmachado:
Uniform is characterized by low and high values and in this case the low=high. Corrected this typo.

**Motivation and context:**

**How has this been tested?**

**Types of changes:**

**Checklist:**

**Still to do:**

### DIFF
--- a/docs/examples/basic/two_neurons.ipynb
+++ b/docs/examples/basic/two_neurons.ipynb
@@ -51,7 +51,7 @@
     "with model:\n",
     "    neurons = nengo.Ensemble(\n",
     "        2, dimensions=1,  # Representing a scalar\n",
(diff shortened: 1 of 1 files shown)


drasmuss:
In this case that is actually intentional, rather than a typo. The goal is to set the intercepts to exactly `-0.5`, so sampling from a distribution with low=high will achieve that. Now admittedly that is kind of a weird way of achieving that goal. I believe that example was written a long time ago, before we had more options for setting the intercepts (e.g. directly passing an array of values), and has never been updated.  We have plans to do a big reworking of all the Nengo examples in the near future, which should hopefully remove some of these confusing aspects!

pedrombmachado:
Hi @drasmuss,
  Thanks for the clarification.
Bw
Pedro

PR #1577, opened 2019-11-01
Remove obsolete/deprecated code

drasmuss:
**Motivation and context:**

Cleaning up old code before the 3.0 release.

Also increased the minimum numpy version to 1.13.

**How has this been tested?**

Ran all the tests with `--slow --spa` to verify that nothing was relying on this deprecated code.

**Types of changes:**

**Checklist:**


tbekolay:
Aside from the docs failures from timeouts, which seem to be rampant today, there's an odd failure of the pickling code on Windows (https://ci.appveyor.com/project/nengo/nengo/builds/28553332/job/27aq8m02o56q21q1#L1573). I'll see if I can reproduce it.

drasmuss:
With the numpy builds removed, I think we might as well get rid of the Basic/Advanced stage distinction as well. It's only two jobs we're capturing under "Advanced", and by removing the distinction we'll be able to take better advantage of all the available TravisCI jobs.

drasmuss:
I've seen that `test_pickle_sim` randomly failing a few times (not just on Windows, in the regular travis-ci builds as well). So I think there's just some non-determinism in that test (which we should look into, but it's unrelated to this PR).

Issue #1582, opened 2019-11-11
Cannot do advanced indexing into ensemble.neurons


drasmuss:
**Describe the bug**

Trying to connect to an `Ensemble.neurons` object using advanced indexing produces an error.

**To reproduce**

```python
import nengo

with nengo.Network() as net:
    node = nengo.Node([0, 0])
    ens = nengo.Ensemble(10, 10)
    nengo.Connection(node, ens.neurons[[0, 2]])

with nengo.Simulator(net) as sim:
    pass
```

**Expected behavior**

This should work without error (forming connections to the neurons indicated by the indices).

**Error message and logs**

```
...
  File "...\nengo\nengo\builder\connection.py", line 301, in build_connection
    model.sig[conn]["out"][post_slice],
  File "...\nengo\nengo\builder\signal.py", line 149, in __getitem__
    raise SignalError("Can only index or slice into signals")
nengo.exceptions.SignalError: Can only index or slice into signals
```

**Versions**

 - OS: Windows 10
 - Python: 3.6
 - Nengo: master

PR #1583, opened 2019-11-11
Fix error when advanced indexing Ensemble.neurons

drasmuss:
**Motivation and context:**

Fixes https://github.com/nengo/nengo/issues/1582

We do not allow advanced indexing on Signals (since we want any views on Signals to share memory, whereas advanced indexing forces a copy).  Previously when building a Connection to a Neurons object we directly incremented the Connection output signal when applying the gains. This works fine, unless we're using advanced indexing on that output signal.

So this PR adds special case handling for that edge case, by creating an intermediate signal that is incremented by the gains op, and then performing the advanced indexing in a separate Copy op.

**How has this been tested?**

Added a new test based on https://github.com/nengo/nengo/issues/1582.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.templates/LICENSE.rst.template
+++ b/.templates/LICENSE.rst.template
@@ -5,8 +5,8 @@ Licensed code
 
 Nengo imports or vendorizes several open source libraries.
 
-* `NumPy <http://www.numpy.org/>`_ - Used under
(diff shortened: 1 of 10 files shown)


drasmuss:
Fixups look good to me.

PR #1588, opened 2019-11-28
Fix __eq__ functions for disparate types

hunse:
**Motivation and context:**

Bug in our `__eq__` functions for `Neurons` and `LearningRule` meant that if you tried to compare it with another type, it would result in an error (because it wouldn't have the attributes we use to do the check). So just check the type first.

**How has this been tested?**

Tested that this fixed the problem in the situation that it came up.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,7 +103,7 @@ before_install:
   - if [[ "$TRAVIS_PYTHON_VERSION" < "3.6" ]]; then
         echo "Skipping bones-check because Python $TRAVIS_PYTHON_VERSION < 3.6";
     else
-        bones-check;
+        bones-check --verbose;
     fi
   # display environment info
   - pip freeze
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,12 @@ Release history
 - Added a new example notebook for Legendre Memory Units.
   (`#1589 <https://github.com/nengo/nengo/pull/1589>`__)
 
+**Fixed**
+
+- Fixed a bug when comparing equality with ``Ensemble.neurons`` or
+  ``Connection.learning_rule`` objects.
+  (`#1588 <https://github.com/nengo/nengo/pull/1588>`__)
+
 3.0.0 (November 18, 2019)
 =========================
 
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -653,7 +653,8 @@ class LearningRule:
 
     def __eq__(self, other):
         return (
-            self._connection is other._connection
+            type(self) == type(other)
+            and self._connection is other._connection
             and self.learning_rule_type == other.learning_rule_type
         )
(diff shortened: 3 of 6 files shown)


drasmuss:
Added tests, which revealed two kind of odd cases:

1. 
``` python
        conn = nengo.Connection(
            ens, ens, learning_rule_type=[
                nengo.PES(learning_rate=0), nengo.PES(learning_rate=10)]
        )
        assert conn.learning_rule[0] == conn.learning_rule[1]
```
Right now this check would pass, but should we consider two learning rules with different parameterizations to be equal?

2.
``` python
        ens = nengo.Ensemble(10, 1, neuron_type=nengo.LIF())
        old_neurons = ens1.neurons
        ens1.neuron_type = nengo.RectifiedLinear()
        assert ens1.neurons == old_neurons
```
Again this assertion passes at the moment, but should we consider two `neurons` objects with not only different parameterizations, but completely different neuron types to be equal?

hunse:
1. Appears to be a bug to me. In `conn.learning_rule.__eq__` we check if the learning rule types are equal, which should be false if they have different learning rates. So seems like a bug in `LearningRuleType.__eq__`
2. All that `Neurons` references is the ensemble. So when you change the neuron type on the ensemble, it will automatically change for every instance of `ens.neurons` that has ever been made.

drasmuss:
Yeah 1. was a mistake, I mixed up the cases I was testing.  For 2, we could change the behaviour (if we wanted) by storing the `neuron_type` in the `Neurons` object (analogously to how we store `learning_rule_type` in `LearningRules`). Just depends on what we want the semantics of the equality to be (i.e. is it checking that these two neurons objects belong to the same ensemble, or is it checking that they belong to the same ensemble _and_ have the same type).

hunse:
`ens.neurons` is just supposed to be a way to refer to the ensemble's neurons, so the less state it can have the better in my opinion. i.e., if I do something like:
```
        ens = nengo.Ensemble(10, 1, neuron_type=nengo.LIF())
        nengo.Connection(ens.neurons, ens.neurons)
        ens.neuron_type = nengo.RectifiedLinear()
```
I wouldn't want that connection to break because the `Neurons` object is storing the old neuron type. So I like the way that it's currently set up.

If I could, I would change `LearningRule` to act the same (i.e. just reference its connection, so if the associated learning rule type changes on the connection, it will change automatically on the `LearningRule`). Wouldn't be too hard; `LearningRule` would just have to store an index or key of which learning rule type it's associated with in the connection.

drasmuss:
I remember us talking about removing the multiple-learning-rules-on-one-connection functionality (since it isn't really used), which would make that even easier. I'll just leave all the current behaviour as-is for now, and we can worry about it if/when the time comes.

Issue #1580, opened 2019-11-07
Allow transform=None on Connections to indicate no weights


drasmuss:
**Is your feature request related to a problem? Please describe.**

Often in models we want to transfer information from one component to another without applying any transformation (e.g., connecting an input `Node` up to something else, or using passthrough nodes on Networks, or when using a connection Probe). Right now those connections all get built with `transform=1`, which adds an unnecessary weight multiplication to all those connections. Even though the multiplication itself is trivial, depending on the backend there can be a certain amount of overhead associated with any operation like that, which can add up.

This also makes a difference in NengoDL, where all those `transform=1` connections are actually trainable parameters. This is often unexpected for users, since they haven't explicitly opted in to creating a transform on the Connection.

**Describe the solution you'd like**

Allow Connections to be created with `transform=None`. This would be functionally the same as `transform=1`, but it wouldn't create the weights or add the weight multiplication operations to the graph.

A further consideration would be whether we want to change the default from `transform=1` to `transform=None`. I think that'd make sense (it seems like a straightforward improvement, with no functional change). The only case where it would make a difference is in something like NengoDL, where we'd be removing trainable parameters as mentioned above. But in my experience, we don't want those trainable parameters most of the time (in fact I often have to work around them by setting those parameters to be non-trainable), so I think that would be an improvement.

**Describe alternatives you've considered**

We could also have a special `transform=nengo.Null()` transform type (in the style of `Dense` or `Convolution`). But since `transform=None` isn't being used for anything else, it seems like a natural syntax for what we want here.

tbekolay:
This sounds great to me, and I like changing the default to `None` as well. :+1:

tbekolay:
Done in #1591.

PR #1591, opened 2020-01-06
Change default transform to None

drasmuss:
**Motivation and context:**

This implements the idea described in https://github.com/nengo/nengo/issues/1580.  Basically, rather than always having an implicit `transform=1` operation on Connections, we just don't apply any transform if none is specified.  In most cases this won't make any difference in the model's behaviour (since `transform=1` wasn't doing anything anyway), but it can save time and memory.

The only situation I can see it being noticeable is if someone was using the default transform and then later on accessing `my_conn.transform`. Previously that would return a `Dense` transform object, now it will return a `NoTransform` object. That seems relatively unlikely (and easy to resolve), but I updated the next version to 3.1.0 for that reason.

**How has this been tested?**

Added a new test in `test_learning_rules` for the new validation checks, and since this is the default transform it is being tested pretty thoroughly in most of the existing tests.

The main implementation change is in `builder/transforms.py`, which uses the new `Null` transform type defined in `transforms.py`. And then there is new validation added for learning rule parameters in `connection.py` to make sure that there are weights for learning to be applied to. Other than that it's relatively minor changes.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,13 +19,22 @@ Release history
    - Removed
    - Fixed
 
-3.0.1 (unreleased)
+3.1.0 (unreleased)
 ==================
 
 **Added**
 
 - Added a new example notebook for Legendre Memory Units.
   (`#1589 <https://github.com/nengo/nengo/pull/1589>`__)
+
+**Changed**
+
+- The default Connection transform is now ``None``, meaning that there will be
+  no transform applied. This only changes behavior when learning on a
+  neuron-neuron connection with the default scalar transform. In that situation
+  there are now no weights to apply learning to, so this will result in an
+  error. The old behaviour can be obtained by setting ``transform=1``.
+  (`#1591 <https://github.com/nengo/nengo/pull/1591>`__)
 
 **Fixed**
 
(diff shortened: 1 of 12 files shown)


tbekolay:
What do you think of the name `NoTransform` to match with the `NoSolver`?

drasmuss:
Did some quick tests with this benchmark script:

``` python
import tracemalloc
import timeit

import nengo
import numpy as np

n = 100
d = 1000

with nengo.Network() as net:
    node = nengo.Node(np.zeros(d))

    for i in range(n):
        new_node = nengo.Node(size_in=d)
        nengo.Connection(node, new_node, synapse=None, transform=None)
        node = new_node

tracemalloc.start()

    print("memory", tracemalloc.get_traced_memory())

    start = timeit.default_timer()
    sim.run(1.0)
    print("time", timeit.default_timer() - start)
```

With `transform=1` this gives

```
memory (3960376, 7072549)
Simulation finished in 0:00:01.                                                 
time 0.8064769
```

With `transform=None` we get
```
memory (2088612, 3018165)                                             
time 0.40419190000000005
```

So about a 2x improvement in memory usage and run time. Admittedly this is kind of a worst case scenario for `transform=1`, it's just a bunch of big empty `transform=1` connections. So we wouldn't expect that kind of improvement in all models. But an improvement nonetheless 🤷‍♂.

PR #1598, opened 2020-02-03
Fix errors when unpickling Ensemble

drasmuss:
**Motivation and context:**

Number of neurons/dimensions need to be set first, as they are used in shape checks of other parameters.

Also changed the max_rates/intercepts check to use equality rather than object identity (as after unpickling the objects may not match).

**How has this been tested?**

Added a pickle test in `test_ensemble.py`.

**Types of changes:**

**Checklist:**


tbekolay:
See #1584, not sure how much overlap there is?

tbekolay:
Looks like there's some overlap but not 100% so we should merge these two PRs together.

drasmuss:
Test looks good to me!

PR #1606, opened 2020-04-16
Specify pytest markers in conftest.py

### LINKED ISSUE #1570: Switch to using config.addinivalue_line for pytest markers
I've always thought it was kind of awkward that you had to specify pytest markers in `setup.cfg`. Apparently there is another way, which is used in Pytest's [introductory docs](https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option), which is to call `config.addinivalue_line` in `pytest_configure`. We should start doing this.

hunse:
**Motivation and context:**

Fixes #1570 

**How has this been tested?**

- Ensured they show up in `pytest --markers`
- Ensured they still function (e.g. `slow` stops slow tests from being run, can be overridden with `--slow` at command line)
- [x] Ensure that they function on TravisCI

**Types of changes:**

**Checklist:**

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -84,11 +84,6 @@ setup_cfg:
       - "ignore:Object <BasalGanglia \"bg\"> has already been built"
       - "ignore:Object <Thalamus \"thalamus\"> has already been built"
       - "ignore:IPython.core.inputsplitter is deprecated"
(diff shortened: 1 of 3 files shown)


hunse:
LGTM!

PR #1609, opened 2020-04-21
Neuron refactor

hunse:
**Motivation and context:**

- Add Tanh neuron type
- Add `RegularSpiking` and `PoissonSpiking` neuron types (updated code from #1225). Completes #579, and includes suggestions from #1487.

Replaces #1225.

**How has this been tested?**

Added a test for `RegularSpiking` and `PoissonSpiking`, and added a `RegularSpiking(Tanh())` neuron to the neurons we regularly run in tests, which should test both `RegularSpiking` and `Tanh` (though maybe we need a bit more testing for `Tanh`?). Unfortunately, the `PoissonSpiking` versions of models appear to be a bit too noisy to meet some of our current tolerances, so I didn't include a `PoissonSpiking` neuron in the neurons we regularly use on tests.

Look at the `Tanh` addition first (the two commits after it go with it and will probably be squashed in in the end). Then look at the other stuff.

Discussion points:
- I've added the `NeuronType.negative` attribute to distinguish which neuron types can emit negative outputs or not. This is mostly needed for the tests, since a number of the tests had assumed that neuron outputs are non-negative. I also use it once in `PoissonSpiking`, but that's just to save multiplying by `np.sign(rates)` if we know all the rates are non-negative. Do we need this attribute? If so, are we happy with the default being `True` (i.e. assuming negative rates are possible unless explicitly specified otherwise) or is `False` better (since most neuron types do not allow negative rates)? I would argue for `True`, because even non-negative neuron types will still work with `negative=True`, but not vice-versa. Also, is there an automatic way to figure out `negative` that we could just use for tests (e.g. by looking at the `.rates` for `[-np.inf, -1, 0, 1, np.inf]` or some other set of test values), and just always multiply by the sign in `PoissonSpiking`?
- Currently, I have both `Tanh` and `RegularSpiking(Tanh)` added to `nl`; they're also tested as part of `test_neurons.test_spiking_types`. We probably don't *need* to test them both places. Options are a) remove one of the two from `nl`, probably `Tanh` since `RegularSpiking(Tanh)` is testing `Tanh` as well, b) remove `Tanh` from `test_neurons.test_spiking_types`, and move back to the earlier version of the test that compares with `LIF` and `SpikingRectifiedLinear`, c) replace `RegularSpiking(Tanh)` in `nl` with `RegularSpiking(RectifiedLinear)` or `RegularSpiking(LIFRate)`, d) keep things as-is. I'd probably lean towards (c), since I think the effect on test time of the extra type isn't huge (so we don't need (a)), but I'm also fine with (d).

**Types of changes:**

**Checklist:**


nengo/neurons.py
```diff
@@ -230,10 +233,27 @@ def step_math(self, dt, J, output):
             Input currents associated with each neuron.
         output : (n_neurons,) array_like
             Output activities associated with each neuron.
+        rng : `numpy.random.mtrand.RandomState`
+            Random number generator for stochastic neuron types.
+
+        Notes
+        -----
+
```
arvoelke, line 241: Will this empty line affect sphinx? I usually see no empty line here but maybe it doesn't matter.

nengo/neurons.py
```diff
@@ -727,7 +814,106 @@ def step_math(self, dt, J, spiked, voltage, recovery):
...
+        return self.base_type.max_rates_intercepts(gain, bias)
+
+    def rates(self, x, gain, bias):
+        return self.base_type.rates(x, gain, bias)
+
+
+class RegularSpiking(_Spiking):
+    """Turn a rate neuron model into a spiking one with regular inter-spike intervals.
+
+    Spikes at regular intervals based on the rates of the base neuron type.
```
arvoelke, line 854: Could also add a reference to https://arxiv.org/abs/2002.03553 assuming `np.trunc` is changed to `np.floor`?

nengo/neurons.py
hunse: I don't think it makes a difference. It looks fine in the rendered docs. But we could remove it for consistency I guess.

nengo/neurons.py
hunse: Sure, go for it.

nengo/neurons.py
```diff
@@ -727,7 +814,155 @@ def step_math(self, dt, J, spiked, voltage, recovery):
...
+            rates = np.abs(rates)
+
+        n_spikes = rng.poisson(rates * dt, output.size)
+        output[...] = (self.amplitude / dt) * (
+            n_spikes if signs is None else n_spikes * signs
+        )
+
+
+class StochasticSpiking(_Spiking):
+    """Turn a rate neuron model into a spiking one with semi-Poisson spiking statistics.
```
arvoelke, line 937: Looks like semi-Poisson already has a meaning other than what we mean here (e.g., https://arxiv.org/abs/cond-mat/0511171).

hunse: Hmm, maybe "Poisson-like"? I wanted to get across the fact that spikes are independently generated each timestep. Maybe we just say that right out: "one with independent spiking statistics"? This seems to me to be the main way to do that other than true Poisson. We could even rename the whole class to `IndependentSpiking`, since it might be a bit more descriptive than `StochasticSpiking`.

hunse: Or what about `DitheredSpiking`, since this is actually dithering?

hunse: Decided to keep `StochasticSpiking` and make a specific reference to "stochastic rounding" in the docstring.

hunse:
I just noticed that `StochasticSpiking` doesn't actually have a builder. The reason it works is that the base neuron type builder should work for any neuron type that doesn't have states. So things will only break if you have a `base_type` with states, e.g. `AdaptiveLIFRate`. I'll add a fix and a test.

nengo/neurons.py
```diff
@@ -215,7 +219,24 @@ def rates(self, x, gain, bias):
...
+            result["voltage"] = phases
+
+        return result
+
+        # # distribute phases equally across states (each neuron will be at a random
+        # # position between 0 and 1 in one of the states, and 0 in all other states)
+        # phase_frac, phase_int = np.modf(phases * len(states))
+        # return OrderedDict(
+        #     (state, (phase_int == i) * phase_frac) for i, state in enumerate(states)
+        # )
```
hunse, line 237: I realized why this code to also set the initial refractory period wasn't working: the refractory period has range [0, tau_ref], not [0, 1]. So to initialize it, we'd have to know `tau_ref`.

nengo/neurons.py
```diff
@@ -727,7 +832,147 @@ def step_math(self, dt, J, spiked, voltage, recovery):
...
+            )
+
+    @property
+    def probeable(self):
+        base = list(self.base_type.probeable)
+        base[base.index("rates")] = "spikes"
+        base.insert(1 if base[0] == "spikes" else 0, "voltage")
+        return tuple(base)
+
+    def step_math(self, dt, J, output, voltage, *base_states):
```
hunse, line 906: Should we use a double asterisk here, because in `SimNeurons.make_step` we always pass states by name (i.e. with a double asterisk)?

hunse:
The commit to fix the `RandomState` sphinx references is here: ff3a8706.

arvoelke:
Randomly considered a weird corner case: is this tested for when the activity matrix is all negative (conceivable if someone used `Tanh` with weird eval points / biases)? I think the default `LstsqL2` solver will then use negative regularization, because `A.max()` will be negative. Maybe should be `np.abs(A).max()`?

nengo/builder/processes.py
```diff
@@ -98,7 +102,7 @@ def make_step(self, signals, dt, rng):
         shape_in = input.shape if input is not None else (0,)
         shape_out = output.shape
         rng = self.process.get_rng(rng)
-        state = {name: signals[sig] for name, sig in self.state.items()}
+        state = {name: signals[self.updates[idx]] for name, idx in self.state.items()}
```
hunse, line 105: Do we need this for the optimizer to work on processes? If so, what was the optimizer doing before? Should we add a test that the optimizer works properly with processes?

nengo/neurons.py
```diff
@@ -697,13 +909,19 @@ def __init__(
     def rates(self, x, gain, bias):
         """Estimates steady-state firing rate given gain and bias."""
         J = self.current(x, gain, bias)
-        voltage = np.zeros_like(J)
-        recovery = np.zeros_like(J)
         return settled_firingrate(
-            self.step_math, J, [voltage, recovery], settle_time=0.001, sim_time=1.0
+            self.step,
+            J,
+            {
```
hunse, line 915: `state=`

nengo/params.py
```diff
@@ -49,10 +49,18 @@ def iter_params(obj):
 
 def equal(a, b):
     """Check if two (possibly array-like) objects are equal."""
+    if isinstance(a, dict) and isinstance(b, dict):
+        if set(a) != set(b):
```
hunse, line 53: `return set(a) == set(b) and all(equal(a[key], b[key]) for key in a)`

nengo/tests/test_ensemble.py
```diff
@@ -388,13 +389,16 @@ def test_noise_copies_ok(Simulator, nl_nodirect, seed, plt, allclose):
 
     process = FilteredNoise(synapse=nengo.Alpha(1.0), dist=Choice([0.5]))
     with nengo.Network(seed=seed) as model:
-        inp, gain, bias = 1, 5, 2
-        model.config[nengo.Ensemble].neuron_type = nl_nodirect()
+        if "spikes" in nl_nodirect.state or RegularSpiking in nl_nodirect.__bases__:
```
hunse, line 392: Could we just check if "voltage" is in `nl_nodirect.state`?

nengo/builder/neurons.py
```diff
@@ -35,54 +29,61 @@ class SimNeurons(Operator):
...
     def make_step(self, signals, dt, rng):
         J = signals[self.J]
-        output = signals[self.output]
-        states = [signals[state] for state in self.states]
+        state = {name: signals[self.sets[idx]] for name, idx in self.state_sigs.items()}
+        state.update(self.state_extra)
 
         def step_simneurons():
-            self.neurons.step_math(dt, J, output, *states)
+            self.neurons.step_math(dt, J, **state)
```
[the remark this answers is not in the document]
tbekolay: No, this is a rebase issue. I had changed it to `step` an earlier commit but then forgot to change it later when I was changing how `rng`s are passed through to the state. I'll switch it to `step`.

nengo/builder/optimizer.py
```diff
@@ -716,20 +716,17 @@ def is_mergeable(op1, op2):
...
-            states.append(st)
-            states_sigr.update(st_sigr)
+        J, J_sigr = SigMerger.merge([op.J for op in ops])
+        state = {}
+        state_sigr = {}
+        for key in ops[0].state_sigs:
+            st, st_sigr = SigMerger.merge([op.sets[op.state_sigs[key]] for op in ops])
+            state[key] = st
+            state_sigr.update(st_sigr)
+        state.update(ops[0].state_extra)
```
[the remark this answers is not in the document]
tbekolay: Yeah that's a good point. I'll add a warning message!

nengo/builder/processes.py
tbekolay: I believe right now the optimizer does not work on processes, you can't merge two processes into one. However, it might be possible for a signal involved in the process to be merged as a result of merging some other op. My guess is that it doesn't actually occur for process state variables, even though it did occur for neuron type state variables (which is where I encountered the issue, and spurred me to make the same changes for process just in case it comes up in the future). I'll see if I can think of a way to test it, but probably it'll take more effort than it's worth.

nengo/params.py
tbekolay: I think I did this for efficiency reasons and possible typeerrors if `a` or `b` aren't dicts, but I'll look again.

nengo/tests/test_ensemble.py
tbekolay: There is no class-level `state` in `RegularSpiking`, it's an instance variable added during `__init__`, so we have to check the type of the class.

nengo/params.py
hunse: It should be just as efficient and safe the way I have it, because both `and` and `all` are lazy (as long as you use a generator in `all` as above).

tbekolay: It's the conversion to `set` that I'm aiming to avoid, can be costly for big dicts. But it's probably a premature optimization.

nengo/builder/processes.py
hunse: Yeah, not worth putting time into. Makes sense to have this for the future.

nengo/params.py
hunse: I don't understand how your code avoids a conversion to `set`. It seems to me that both `a` and `b ` get converted to set in this first line.

hunse: Could do `all(k in b for k in a) and all(k in a for k in b)` instead of the `set(a) == set(b)` to avoid converting to `set`.

tbekolay: > It seems to me that both `a` and `b ` get converted to set in this first line.

Not if either a or b is not a dict. E.g. if one is a tuple or list or really long numpy array (the most likley scenario in which it could be bad performance-wise). Anyway, I'll look at this again, probably it can be simplified to one statement.

arvoelke: The `if isinstance(a, dict) and isinstance(b, dict)` above this line means both `a` and `b` will be dicts though? But if you were to get rid of that guard, converting to `set` may be generally cheaper than `k in a for k in b` or vice-versa as the latter is `O(n^2)` worst case for two arrays whereas `set(a) == set(b)` is `O(n)`.

hunse: Yes, I assumed we were keeping the guard (i.e. we know they're both dicts). I don't see how this would work without it. My original comment was just a way to condense the same logic as Trevor originally had into one line (two if you count the guard) so that it's more readable. There shouldn't be any change to the logic.

hunse: i.e. instead of
```
    if isinstance(a, dict) and isinstance(b, dict):
        if set(a) != set(b):
            return False
        for key in a:
            if not equal(a[key], b[key]):
                return False
        return True
```
we have
```
    if isinstance(a, dict) and isinstance(b, dict):
        return set(a) == set(b) and all(equal(a[key], b[key]) for key in a)
```

tbekolay:
> Randomly considered a weird corner case: is this tested for when the activity matrix is all negative (conceivable if someone used `Tanh` with weird eval points / biases)? I think the default `LstsqL2` solver will then use negative regularization, because `A.max()` will be negative. Maybe should be `np.abs(A).max()`?

Good call; I added a test for this and both `LstsqNoise` and `LstsqL2nz` raised errors that we fixed with some `np.abs`ing. `LstsqL2` and all the others had no issues and the decoded activity passed normal tolerances :shrug:

tbekolay:
I addressed all the review comments, but since we were touching the `nl` / `nl_nodirect` stuff anyway I wanted to rename them to `nt`. Originally we envisioned a more general notion of a "nonlinearity", but we call them `NeuronType`s regardless so `nt` is a better short form. After that change, I'll fix up the history and make sure all the changelog entries are there and in the right commits.

hunse:
I do use `nt` sometimes to refer to the number of timesteps. If we're switching `nl` to `nt`, probably better to switch those instances to something else. Here are the places I've found that in tests:
```
nengo/tests/test_synapses.py:    nt = len(t)
nengo/tests/test_synapses.py:    nt = len(t)
nengo/tests/test_synapses.py:    nt = len(t)
nengo/tests/test_synapses.py:    nt = 3000
nengo/tests/test_synapses.py:    nt = 50
nengo/tests/test_processes.py:    nt = len(sim.trange())
nengo/utils/tests/test_testing.py:    nt = 100  # signal length
```

tbekolay:
> If we're switching `nl` to `nt`, probably better to switch those instances to something else.

Good to know, I think normally we call it `n_steps` so I'll make the switch.

PR #1610, opened 2020-05-06
Correct docstring for oscillator's frequency

arvoelke:
The units are in radians per second. To specify the frequency in Hz, this value must be multiplied by ``2*np.pi``. FYI, can also make it more accurate using equation 5.9 from [my thesis](https://uwspace.uwaterloo.ca/bitstream/handle/10012/14625/Voelker_Aaron.pdf) but this requires knowing `dt`.

**Motivation and context:**
Corrects documentation.

**How has this been tested?**
N/A

**Types of changes:**

**Checklist:**
- [N/A] I have updated the documentation accordingly.
- [N/A] I have run the test suite locally and all tests passed.

### DIFF
--- a/docs/examples/dynamics/controlled-oscillator.ipynb
+++ b/docs/examples/dynamics/controlled-oscillator.ipynb
@@ -108,7 +108,7 @@
    "outputs": [],
    "source": [
     "tau = 0.1  # Post-synaptic time constant for feedback\n",
-    "w_max = 10  # Maximum frequency is w_max/(2*pi)\n",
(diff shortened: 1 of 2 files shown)


arvoelke:
Currently failing with:

> `nengo/tests/test_solvers.py:420:25: E741 ambiguous variable name 'l'`

Not sure why that is happening here.

tbekolay:
> Not sure why that is happening here.

It was due to a new check in a more recent flake8 version. It's been fixed in master so it'll go away on a rebase.

PR #1611, opened 2020-05-07
Nengolib

hunse:
**Motivation and context:**

Migrate some key features from https://github.com/arvoelke/nengolib:
- Allow synapses to be specified in state-space, and add a number of new synapse models (including `LegendreDelay`)
- Allow neurons' initial voltages to be set, and default to uniformly distributing voltages (addresses #316)
- Allow connections' initial value(s) to be set (addresses #316)
- Add the `ScatteredHypersphere` distribution for more evenly distributed spacing of points on the hypersphere, and use this to generate encoders and eval points.
- Add the RLS learning rule

Based on #1609.

**How has this been tested?**

- Added many unit tests
- For `ScatteredHypersphere`, the large new test produces plots that give a good idea how the point distribution compares with `UniformHypersphere`

- The new features can be reviewed relatively independently.
- The new `LinearSystem` is best understood in the context of `LinearFilter`, and how it's used there.
- Setting the ensemble voltages has been done in two commits, but should be understood all together (I left the second part separate because the breaking change to identifying states by name may be controversial).
- For `ScatteredHypersphere`, the math is a bit complicated. @arvoelke or I can send you the referenced book pages if you want to look at them. Probably best to start by looking at the test plots to understand the result, then looking more at the code.

**Types of changes:**

  - We now store neurons' state as (key, value) pairs. This allows the user to not worry about order if they're using `make_neuron_state`; they just have to make sure the state names are consistent.

**Checklist:**


nengo/connection.py
```diff
@@ -319,6 +319,28 @@ def repeated_inds(x):
...
+    def coerce(self, conn, value):
+        value = super().coerce(conn, value)
+
+        if isinstance(value, np.ndarray):
+            if isinstance(conn.post_obj, Ensemble) and conn.solver.weights:
+                n_synapses = conn.post_obj.n_neurons
+            else:
+                n_synapses = conn.size_out
+
+            if value.ndim > 0 and value.shape != (n_synapses,):
```
arvoelke, line 334: I think `value.ndim` has to be `> 0`? Since there is `sample_shape=("*",)` the parent class seems to require that the array is at least 1D. I tried `initial_value=1` and got `nengo.exceptions.ValidationError: Connection.initial_value: ndarray must be 1D (got 0D)`. Unsure if this is intentional (would we want to support something like `initial_synapse=1` or would we require `initial_synapse=Choice([1])` or `initial_synapse=[1]`?

nengo/connection.py
```diff
@@ -319,6 +319,28 @@ def repeated_inds(x):
...
+    """Connection-specific validation for initial values."""
+
+    def coerce(self, conn, value):
+        value = super().coerce(conn, value)
+
+        if isinstance(value, np.ndarray):
+            if isinstance(conn.post_obj, Ensemble) and conn.solver.weights:
+                n_synapses = conn.post_obj.n_neurons
+            else:
+                n_synapses = conn.size_out
```
arvoelke, line 332: Does this code path have good testing coverage? I wonder if this logic handles all cases, given how complex the build connection function is when it comes to all the case logic with slicing and direct neuron connections. An easy way to enforce testing coverage here would be to change the default to `initial_value=0` but that raises the error I mention below.

nengo/builder/processes.py
```diff
@@ -176,3 +181,54 @@ def build_process(model, process, sig_in=None, sig_out=None, mode="set"):
...
+    """
+    if sig_out is None:
+        assert sig_in is not None, "Both `sig_in` and `sig_out` cannot be None"
+        output_shape = sig_in.shape
+
+        y0 = np.asarray(y0, dtype=rc.float_dtype)
+        if y0.ndim > 0:
+            assert y0.shape == output_shape
+        else:
+            y0 = y0 * np.ones(output_shape, dtype=rc.float_dtype)
```
arvoelke, line 222: Why ones?

nengo/builder/learning_rules.py
```diff
@@ -461,6 +461,94 @@ def step_simvoja():
...
+    r"""Calculate connection weight change according to the RLS rule.
+
+    Implements the Recursive Least Squares (RLS) learning rule of the form
+
+    .. math::
+
+        g_i &= \sum_j P_{ij}(n-1) r_i
+        P_{ij}(n) &= P_{ij}(n-1)
+                   - g_i g_j / \left( 1 + \sum_{ij} r_i P_{ij}(n-1) r_j \right)
+        \Delta \omega_{ji} &= \frac{\kappa \delta_t}{n} e_j \sum_{k} P_{ik}(n) r_k
```
arvoelke, line 474: I built the documentation locally and this is what this math looked like:

Screenshot from 2020-08-04 23-19-08

Not sure why. Maybe something to do with the newline break on 472-473?

nengo/utils/numpy.py
```diff
@@ -255,3 +255,86 @@ def rmse(x, y, axis=None, keepdims=False):  # pragma: no cover
 
     def rfftfreq(n, d=1.0):
         return np.abs(np.fft.fftfreq(n=n, d=d)[: n // 2 + 1])
+
+
+_betaincinv22_file = os.path.join(
+    os.path.abspath(os.path.dirname(__file__)), "betaincinv22_table.npz"
```
arvoelke, line 261: Is there a reason to prefer the location `nengo/utils` over `nengo-data` or vice-versa?

nengo/utils/tests/test_numpy.py
```diff
@@ -87,3 +95,87 @@ def test_expm(rng, allclose):
...
+        assert allclose(y, yref, atol=atol), "dims=%d" % dims[i]
+
+
+def test_make_betaincinv22_table(monkeypatch, tmpdir, plt, allclose):
+    pytest.importorskip("scipy.special")
+    monkeypatch.setattr(nengo.utils.numpy, "_betaincinv22_table", None)
+
+    filename = os.path.join(str(tmpdir), "betaincinv22_test_table.npz")
+
+    _make_betaincinv22_table(filename=filename, n_interp=200, n_dims=50)
```
arvoelke, line 162: Is there a test that the contents of the file in `nengo/utils` matches the one generated by `_make_betaincinv22_table` with whatever set of arguments?

nengo/builder/processes.py
hunse: If `y0` is a scalar, broadcast it across all dimensions.

arvoelke: Ah right. Thanks

nengo/connection.py
hunse: So I flip-flopped on whether to allow a scalar to initialize all synapses to the same value. My final consensus was to not allow it, because you can accomplish the same thing with `Choice([0])` (it is a bit more verbose, but it fits more with how we typically use `DistOrArrayParam` and makes it a bit clearer what the user is asking for). That said, it's easy to allow it (all we need to do is remove the `sample_shape` when we instantiate this param, and I think it will already have the validation it needs due to this line here).

So thoughts on whether to allow it @arvoelke @tbekolay @drasmuss?

If we don't allow it, then we should remove the vestiges of when it was allowed that you've noted here, @arvoelke (namely this line, the docstring, and there's a bit of stuff in `builder/processes.py` I think).

nengo/connection.py
hunse: I think it's actually the other part of the path (using a weight solver) that isn't well tested. This is tested in `test_connection.py:test_bad_initial_values`, and that case isn't covered there. You're right that we don't test all possible types of connections, but having `initial_value=0` as the default wouldn't check this either because then `value.ndim > 0` would be false and `n_synapses` wouldn't even get checked. If we wanted to test this, we'd have to explicitly test all those connection types, either by adding to `test_initial_value` or by setting initial values as vectors in some of the tests that test those other connection types. Personally, I'm not too worried. This choice is based of an understanding of how the connection builder works, and the number of synapses always is equal to the connection `size_out`.

nengo/utils/tests/test_numpy.py
hunse: Not explicitly. There's some implicit testing, in that we make sure that the `nengo/utils` table yields results close to the actual `betaincinv` function. We could test it explicitly, though.

nengo/utils/numpy.py
hunse: :shrug: @tbekolay?

nengo/builder/learning_rules.py
hunse: Fixed. It was missing the `\\` at the end of each line. Not sure why that caused it to not render at all (I would have expected it to all render on one line).

nengo/synapses.py
```diff
@@ -450,8 +455,329 @@ class Alpha(LinearFilter):
...
 
     def __init__(self, tau, **kwargs):
-        super().__init__([1], [tau ** 2, 2 * tau, 1], **kwargs)
         self.tau = tau
+        super().__init__(([1], [tau ** 2, 2 * tau, 1]), analog=True, **kwargs)
+
+    @property
+    def cutoff(self):
+        """Cutoff frequency in Hz; frequencies above this are attenuated."""
+        return np.sqrt(np.sqrt(2) - 1) / (2 * np.pi * self.tau)
```
arvoelke, line 464: Checked this via https://www.wolframalpha.com/input/?i=%7C1+%2F+%28tau*i*sqrt%28sqrt%282%29+-+1%29+%2F+%28tau%29+%2B+1%29%5E2%7C :+1:

nengo/linear_system.py
```diff
@@ -0,0 +1,369 @@
...
+
+        # create state memory variable X
+        assert shape_in[-1] == self.input_size
+        assert shape_out[-1] == self.output_size
+        assert shape_in[:-1] == shape_out[:-1]
+        X = np.zeros(shape_out[:-1] + (self.state_size,), dtype=dtype)
+        X[:] = self._expand_x0(x0, shape=shape_out[:-1], broadcast=True)
+        return {"X": X}
+
+    def make_step(self, shape_in, shape_out, dt, rng, state):  # noqa: C901
```
arvoelke, line 311: Is there a way to have only one `make_step` for the linear system? It seems right now there's two (this one, and the old one in `synapses.py`). Can they be combined, e.g., by making the synapse call out to this one? Or not for some reason?

nengo/linear_system.py
```diff
@@ -0,0 +1,369 @@
...
+    @classmethod
+    def _find_size(cls, matrices, index):
+        for M in matrices:
+            if M is not None and M.size > 0:
+                return M.shape[index]
+
+        return 0
+
+    def __init__(self, sys, analog=True, method="zoh", x0=0, default_dt=0.001):
+        self._CombineClass = LinearSystem
```
arvoelke, line 85: What's the reason for this being a variable? I don't see any place where this is modified for instance.

arvoelke, line 85: Since this is used in both combine and discretize, calling it `_CombineClass` is a bit confusing. Not sure of a better name though.

nengo/linear_system.py
```diff
@@ -0,0 +1,369 @@
...
+from nengo._vendor.scipy.signal import (
+    cont2discrete,
+    ss2tf,
+    tf2ss,
+    tf2zpk,
+    zpk2ss,
+)
+
+
+def _as2d(M, last=True, dtype=np.float64):
```
arvoelke, line 16: Is there anywhere this is used with `last=True`? I only see the one use below (with `last=False`).

nengo/linear_system.py
```diff
@@ -0,0 +1,369 @@
...
+            A, B, C, D = sys
+        else:
+            raise ValidationError(
+                "Must be a tuple in (num, den), (z, p, k), or (A, B, C, D) form "
+                "(the received length %d is not 2, 3, or 4)" % len(sys),
+                "sys",
+                obj=self,
+            )
+
+        A, B, C, D = [_as2d(M, last=False) for M in (A, B, C, D)]
```
arvoelke, line 106: Should `dtype` be passed here?

nengo/networks/tests/test_linear_system.py
```diff
@@ -0,0 +1,138 @@
...
+    omega = 4 * np.pi
+    A = [[0, -omega], [omega, 0]]
+    C = np.eye(2)
+    x0 = [1, 0]
+    sys = LinearSystem((A, None, C, None), x0=x0)
+
+    t, ref, u, x, y = _test_linear_system_network(
+        sys, synapse, NetworkType, Simulator, dt=dt, seed=seed, plt=plt, n_neurons=200,
+    )
+    e = npext.rms(y - ref, axis=0) / npext.rms(ref, axis=0)
```
arvoelke, line 91: Can use the same `nrmse` function from `nengo/tests/test_learning_rules.py`?

nengo/networks/tests/test_linear_system.py
```diff
@@ -0,0 +1,138 @@
...
+    dt=None,
+    simtime=1.0,
+    input_f=None,
+    **kwargs
+):
+    probe_synapse = nengo.Alpha(0.005)
+
+    with nengo.Network(seed=seed) as net:
+        ref = nengo.Node(sys, label="ref")
+        ref_p = nengo.Probe(ref, synapse=probe_synapse)
```
arvoelke, line 39: There's also a way to do this without probing `ref` with a synapse. You can take `subnet.state_input` which is `x` and then do `Cx + Du` directly to get the network's output. The `subnet.state_input` is already filtered and so no additional filtering is required on either the reference or the network. This is a feature that I've wanted a number of times in nengolib but was just doing it manually when needed. If you think it's a good idea I think this would be a nice add-on to `LinearNetwork`, possibly as an alternative output node...? The main drawback of this approach is if `u` is spiking then it won't look great.

nengo/connection.py
hunse: Based on a discussion we had on the Nengo dev team, I have now moved the `initial_value` parameter to `Synapse`, and called it `initial_output` instead. Scalars are now accepted (as is any array that is broadcastable to the required shape). This is in line with how things work for the new `NeuronType.initial_state`.

nengo/connection.py
hunse: This check is now gone. Since we've moved `initial_output` to `Synapse`, we can only validate the shape once we actually build.

nengo/linear_system.py
hunse: This one is fully general, in that it supports systems with any input/output size. The step functions on `LinearFilter` have been optimized for filtering (where the input and output size are both equal to 1). There's also some even more specific functions for cases like only one state, and no state. `LinearFilter` could use this `make_step`, but I think it would be much less efficient (particularly for synapses with one state like `Lowpass`).

nengo/linear_system.py
hunse: No, it's not. I've changed it so that we use `last=True` on `C` and `D`, and `last=False` on `A` and `B`. Essentially this amounts to assuming the system is SISO if vectors are provided for `B` and `C`, which seems like a fine assumption to me (I doubt this will be used much).

nengo/linear_system.py
hunse: Good call

nengo/networks/tests/test_linear_system.py
hunse: So it would be an alternative output node that uses `state_input` instead of `state_output` as the `pre` for the "C" connection? That seems like a reasonable thing to want to do sometimes, but the fact that the connection doesn't get represented using neurons is somewhat concerning. I would leave this as something users can do themselves if they wish (it seems like more of a power-user move, and only requires a line or two of code).

nengo/utils/tests/test_numpy.py
hunse: I've added an explicit test

nengo/utils/numpy.py
hunse: Trevor said `nengo/utils` is good, because we're going to retire `nengo-data` at some point. I've added it to the manifest and tested that it gets installed with `python setup.py install` (this required `include_package_data=True` in `setup.py`).

nengo/linear_system.py
hunse: I can't think of a better one ATM either. If you do, let me know, otherwise I'll keep it as-is for now.

nengo/networks/tests/test_linear_system.py
hunse: We could, but this is already a one-liner so I'm happy just to leave it.

CHANGES.rst
```diff
@@ -46,6 +46,28 @@ Release history
...
+  (`#1611`_)
+- Added the ``RLS`` (recursive least-squares) learning rule. This is an online version
+  of the least-squares method typically used for offline decoder-solving in Nengo.
+  It is demonstrated in the new ``learning/force-learning.ipynb`` notebook. (`#1611`_)
+- Added ``LinearSystemNetwork``. This network takes a ``LinearSystem`` process and
+  implements it in neurons, properly accounting for the synapse model used on the
+  recurrent synapse. For ``Lowpass`` synapses, this is equivalent to Principle 3 of the
+  Neural Engineering Framework (NEF), but it generalizes to other synapse types, as well
+  as discrete synapses. (`#1611`_)
+- Added ``rng`` argument to ``Process.make_state``, for processes that use randomness
```
arvoelke, line 66: Should this be moved to changed, or should there otherwise be a note under changed, that processes written for v3.0.0 that override `make_state` need to be updated to work with v3.1.0? Or is this not a breaking change?

nengo/linear_system.py
arvoelke: I think `_factory` might be a bit more telling of its purpose, even if that might not be the most accurate use of the 'factory method pattern' terminology (since it's not really a method that's creates instances of a class). But it functions in that sort of way. I'll add a commit for this and if you or anyone else doesn't like it feel free to revert. :)

nengo/networks/ensemblearray.py
```diff
@@ -245,6 +245,14 @@ def add_output(self, name, function, synapse=None, **conn_kwargs):
             should stay as None, and synaptic filtering should be performed in
             the connection from the output node.
         """
+        if hasattr(self, name):
+            raise ValidationError(
+                "Cannot add output named %r; there is already an attribute by this name"
```
arvoelke, line 250: Should this be added to the changelog either under fixed or changed? Since before it would "work" by overwriting it, and this could break someone's example if they were relying on this behaviour (or doing it accidentally)?

nengo/networks/tests/test_linear_system.py
arvoelke: Yeah it is a bit of a power move. And can also mean some interesting things sometimes, like the `Cx` connection that doesn't get represented in neurons could be multiplied together with the connection that it's eventually projecting to. And if there is a lowpass synapse on that connection too, then it becomes an `DoubleExp` and can give you better signal to noise. It would be nice if these nuances were discoverable, as it's one of those things that's not readily obvious, but can become clear and understandable with an example, IMO. 

It can be done by the user in a line or two of code, but requires the user to look into the code and pull out `C, D = subnet.mapped_system.ss[2:]` to build the two connections, as well as realize that each connection needs to go from `subnet.state.input`. To avoid having to know these internals, there could be a method like `add_output` (akin to the `EnsembleArray`) method. But I am wary of wanting to add more to this already large PR, so we could also move this discussion to a later feature request issue.

arvoelke:
Still reviewing, but I cherry-picked the automatic nengo-bones update from master to fix the build. I did this rather than rebasing so that I wouldn't be force-pushing. Just FYI.

nengo/networks/linear_system.py
```diff
@@ -0,0 +1,216 @@
...
+    system = LinearSystemParam("system")
+    synapse = LinearFilterParam("synapse")
+
+    def __init__(
+        self,
+        system,
+        synapse,
+        n_neurons=100,
+        dt=None,
+        output_synapse=None,
```
arvoelke, line 165: `output_synapse` missing from docstring parameters.

nengo/networks/linear_system.py
```diff
@@ -0,0 +1,216 @@
...
+    an accurate implementation of the system using the provided synapse.
+
+    Parameters
+    ----------
+    system : `.LinearSystem`
+        The LinearSystem to be implemented.
+    synapse : `.LinearFilter`
+        The synapse to use on the recurrent connection and input connection. Must be
+        an instance or subclass of `.LinearFilter`.
+    n_neurons : int
```
arvoelke, line 124: optional

nengo/networks/linear_system.py
```diff
@@ -0,0 +1,216 @@
...
+        The number of neurons per ensemble in the state `.EnsembleArray`.
+    dt : float (optional)
```
arvoelke, line 126: Think `, optional` is used everywhere instead of `(optional)`. Unsure if it matters.

nengo/networks/linear_system.py
```diff
@@ -0,0 +1,216 @@
...
+        The time constant used to discretize the synapses. If provided, it should be
+        the same as the simulator time constant. If not provided (default), the system
```
arvoelke, line 128: I realize this isn't strictly required to be true. In general it serves as controlling the time-scale of integration. So for example if you have an oscillator, dropping this `dt` by 10x is the same as increasing the frequency of the oscillator by 10x (because the simulator is going 10x faster than it thought it would be going, or more simply you can just look at the dimensions of the units `s` and `1/s`). This can also be useful if you are for instance feeding it the linear system for the LMU and want to control the size of the window without redefining the system. So this wording could be relaxed?

nengo/networks/linear_system.py
```diff
@@ -0,0 +1,216 @@
...
+                n_neurons, x_dims, **ens_kwargs
+            )
+
+            # --- make connections
+            self.ss_connections = {}
+            self.ss_connections["A"] = nengo.Connection(
+                self.state_output,
+                self.state_input,
+                transform=A,
+                synapse=self.synapse.copy(initial_output=self.mapped_system.x0),
```
arvoelke, line 193: What would happen if the synapse they pass in already has an `initial_output` specified (in relation to what it would do here and below on the "B" connection)?

nengo/networks/tests/test_linear_system.py
arvoelke: I also realized you can do this same thing but from the output of the `EnsembleArray`, which is equivalent to what I was saying above but is perhaps more conventional since we're no longer connecting from the input to the array. So I added a commit which shows how to do this in the unit test. It was actually pretty nontrivial to do this (took me longer than I thought because there are some nuances to initializing the synapse). Hoping this can serve as a reference if in the future someone wants to apply the same trick or if there is some desired to turn this into a helper method.

arvoelke:
Still have a few more things to look at, but I added a commit which shows (in a unit test) how to apply a more advanced Principle 3 trick with the `LinearSystemNetwork` that lets you decode the _unfiltered_ output (even though it being decoded from filtered spike trains). This is useful if you want to represent the output in some other ensemble without filtering it (note the synapse can also be "moved" by linearity). This is something I've wanted to support in `nengolib` for a while now but have just done manually myself as needed. The details are a bit nuanced.

While I'm thinking about it there are two things that I think would be great additions (e.g., as future PRs):
 - Rewrite https://arvoelke.github.io/nengolib-docs/nengolib.networks.RollingWindow.html with `legendre=True` and `realized=Identity()` by building off `LegendreDelay` and `LinearSystemNetwork` and making the readout more transparent. This can also use the same "trick" mentioned above to avoid needing to filter the state.
 - Create another interface like `LinearSystemNetwork` that supports nonlinear dynamical systems. I have an example up in our GitLab at `arvoelke/scratchpad/blob/master/benchmarks/oscillator.py` for a nonlinear controlled oscillator with discrete Principle 3. The more general case is in my thesis (Theorems 5.1.3 and 5.1.4).

nengo/networks/linear_system.py
hunse: How would you relax it? We could say "If provided, it is typically the same as the simulator time constant." I think in general, users would be surprised if they're creating something (e.g. an oscillator, a filter) with well-defined frequencies/time constants, and effectively end up with something different due to a different time constant here. So I'd definitely like to still promote this as typically being the same as the simulator time constant, and then anybody who knows what they're doing can go from there.

nengo/networks/linear_system.py
hunse: Ahh, yeah, good point. Right now, it would be a back-door undocumented way to set the initial input to the network, except it's setting that initial input *after* the `B` transform, which is weird. Probably best to make that synapse a copy, too, with an initial output of 0.

arvoelke: Maybe could throw an error if they pass in a synapse with an initial output, telling the user to set it on the system instead? That way it's clear that the case is not currently supported (and not really well-defined either). I think what it would end up doing right now is (indirectly) add the extra initial value to the same initial state `x0`. As you say I don't think this would be used much, so this would just be a safeguard for a user doing something weird (and probably not what they want to be doing).

nengo/networks/linear_system.py
arvoelke: Sure makes sense. And I just realized what I said wasn't strictly true either because of the way `ss2sim` is written. If you want to do this speed change 'trick' you still want to use the real simulator dt when discretizing the synapse, but a different dt when discretizing the system. Currently `ss2sim` assumes the same for both, so yeah probably worth just changing the word "should" for now.

nengo/networks/linear_system.py
hunse: Yeah, I think an error is a good idea. I've added one. It's possible that it might be annoying occasionally (e.g. if the user's already got a synapse with an `initial_output=Choice([0])`, they'd get the error, even though that's identical to `initial_output=None`), but I think with the new ability to copy a FrozenObject and change a value at the same time, it should be easy to work around (i.e. passing in `synapse.copy(initial_output=None)` would always get around the error).

hunse:
OK, I think I've made all of the requested changes. Also rebased to master and squashed the previous changes. (I do have a backup of the old unsquashed branch if we need it, but I looked through all of @arvoelke's changes and they all looked good to me, so it made sense to squash them in.)

nengo/linear_system.py
```diff
@@ -0,0 +1,382 @@
...
+        Whether ``sys`` is in analog (continuous-time) or discrete form.
+    method : string, optional
+        The method to use for discretization (if ``analog`` is True). See
+        `scipy.signal.cont2discrete` for information about the options.
+    x0 : array_like, optional
+        Initial values for the system state. The last dimension must equal the
+        ``state_size``.
+    default_dt : float, optional
+        If ``dt`` is not specified in `~.Process.run`, `~.Process.run_steps`,
+        `~.Process.ntrange`, or `~.Process.trange`, this will be used.
```
arvoelke, line 55: I initially tried to make this reference `LinearSystem` rather than `Process`, but the docs build failed because it couldn't find a reference to those methods.

nengo/synapses.py
```diff
@@ -119,24 +161,53 @@ def filtfilt(self, x, **kwargs):
...
+    sys : tuple
+        A tuple describing the linear system used for filtering, in
+        ``(numerator, denominator)``, ``(zeros, poles, gain)``, or ``(A, B, C, D)``
+        (state space) form. See `.LinearSystem` for more details on these forms.
+        The system must have both an input size and output size of 1.
+    den : array_like, optional (DEPRECATED)
+        Denominator coefficients of the transfer function. If used, the first (``sys``)
+        argument is treated as the numerator coefficients of the transfer function.
+        This way of specifying transfer functions is deprecated; the preferred method
+        is to pass a ``(num, den)`` tuple to the ``sys`` argument.
```
arvoelke, line 205: Do these need to have any sort of `.. versionchanged` or `.. versiondeprecated` sphinx annotation or similar?

nengo/synapses.py
```diff
@@ -385,12 +468,12 @@ class NoD(Step):
         """
 
         def __call__(self, t, signal):
-            self.X[:] = np.dot(self.A, self.X) + self.B * signal
-            return np.dot(self.C, self.X)[0]
+            self.X[:] = np.dot(self.X, self.AT) + signal[..., None] * self.BT
+            return np.dot(self.X, self.CT)[..., 0]
```
arvoelke, line 472: Would `.squeeze(axis=-1)` be more failsafe or robust?

nengo/synapses.py
```diff
@@ -466,8 +560,332 @@ class Alpha(LinearFilter):
...
+
+class DoubleExp(LinearFilter):
+    r"""A second-order (two-pole) lowpass filter synapse with no zeros.
+
+    The transfer function is:
+
+    .. math::
+
+        h(t) &= \frac{\exp(-t / \tau_1) - \exp(-t / \tau_2)}{\tau_1 - \tau_2} \\
+        H(s) &= \frac{1}{(\tau_1 s + 1)(\tau_2 s + 1)}
```
arvoelke, line 580: I added the `h(t)` equation ([see validation](https://www.wolframalpha.com/input/?i=InverseLaplace%5B%5Cfrac%7B1%7D%7B%28%5Ctau_1+s+%2B+1%29%28%5Ctau_2+s+%2B+1%29%7D%5D)) to mirror how they other synapses are presented. Note that even though there is a divide-by-zero when `tau_1 == tau_2` you can take the limit from either side to show that it equals the alpha synapse.

nengo/synapses.py
```diff
@@ -466,8 +560,332 @@ class Alpha(LinearFilter):
...
+        plt.legend()
+    """
+
+    tau = NumberParam("tau", low=0)
+    order = IntParam("order", low=1)
+
+    def __init__(self, tau, order=1, **kwargs):
+        self.tau = tau
+        self.order = order
+        num = [tau] + [0] * order
```
arvoelke, line 755: In nengolib this is `tau ** order` instead of `tau`. I think this is necessary to keep them unit gain. https://arvoelke.github.io/nengolib-docs/nengolib.synapses.Highpass.html 

nengolib-synapses-Highpass-1

nengo/synapses.py
```diff
@@ -161,41 +244,50 @@ class LinearFilter(Synapse):
...
+        **kwargs
+    ):
+        if den is not None:
+            warnings.warn(
+                DeprecationWarning(
+                    "`den` is deprecated. Pass systems in transfer function form as a "
+                    "`(numerator, denominator)` 2-tuple instead."
+                )
+            )
+            sys = (sys, den)
```
arvoelke, line 266: `num` and `den` are still listed as attributes of this class, but I don't think they are anymore. Maybe they should be attributes of `LinearSystem`?

nengo/synapses.py
```diff
@@ -466,8 +560,332 @@ class Alpha(LinearFilter):
...
+        Q = np.arange(order, dtype=np.float64)
+        R = (2 * Q + 1)[:, None] / theta
+        j, i = np.meshgrid(Q, Q)
+
+        A = np.where(i < j, -1, (-1.0) ** (i - j + 1)) * R
+        B = (-1.0) ** Q[:, None] * R
+        C = np.ones((1, order))
+        D = np.zeros((1,))
+
+        super().__init__((A, B, C, D), analog=True, **kwargs)
```
arvoelke, line 888: Note to selves: following this PR, we can update examples such as https://www.nengo.ai/nengo-dl/examples/lmu.html to obtain the (A, B) matrices by instantiating this class. :)

nengo/synapses.py
arvoelke: This change would also make it consistent with the equation in the class-level docstring (`.. math:: H(s) = \left( \frac{\tau s}{\tau s + 1} \right)^{order}`).

nengo/tests/test_params.py
```diff
@@ -349,26 +351,33 @@ class Test:
...
+                ),
+            ),
+            (params.BoolParam, lambda attr: not attr.default),
+            (
+                params.NumberParam,
+                lambda attr: (1 if attr.default is None else attr.default + 1),
+            ),
+            (params.StringParam, lambda attr: "abc"),
+            (params.NdarrayParam, lambda attr: np.zeros([1] * len(attr.shape))),
+        ]
```
arvoelke, line 379: Missing `DistOrArrayParam` and `IntParam`?

nengo/tests/test_synapses.py
```diff
@@ -54,29 +68,82 @@ def test_lowpass(Simulator, plt, seed, allclose):
...
+def test_double_exp(Simulator, plt, seed, allclose):
+    dt = 1e-3
+    synapse = DoubleExp(0.01, 0.002)
+    t, x, yhat = run_synapse(Simulator, seed, synapse, dt=dt)
+    y = synapse.filt(x, dt=dt)
+    assert signals_allclose(t, y, yhat, delay=dt, atol=5e-6, plt=plt, allclose=allclose)
+
+    check_synapse = Lowpass(0.002).combine(Lowpass(0.01))
+    assert np.allclose(check_synapse.tf[0], synapse.tf[0])
+    assert np.allclose(check_synapse.tf[1], synapse.tf[1])
```
arvoelke, line 105: Added this test. Was expecting there to be a `.num` and `.den` attribute when I first tried this.

nengo/tests/test_synapses.py
```diff
@@ -54,29 +68,86 @@ def test_lowpass(Simulator, plt, seed, allclose):
...
+
+    freqs = [0, synapse.cutoff_low, synapse.freq, synapse.cutoff_high, 1e16]
+    gains = np.abs(synapse.evaluate(freqs))
+    assert (np.diff(freqs) > 0.1).all()
+    assert allclose(gains, [0, 1 / np.sqrt(2), 1, 1 / np.sqrt(2), 0])
+
+
+def test_highpass(Simulator, plt, seed, allclose):
+    dt = 1e-3
+    synapse = Highpass(0.1)
```
arvoelke, line 133: Missing test for `order != 1`. Also see comment there about how I think the numerator should be `tau ** order` instead of just `tau`.

nengo/tests/test_synapses.py
```diff
@@ -137,38 +208,90 @@ def test_linearfilter_evaluate(plt):
...
+
+    assert synapse == nengo.LinearFilter(sys)
+
+
+@pytest.mark.parametrize(
+    "synapse, rmse_range",
+    [
+        (DiscreteDelay(20), [0, 1e-8]),
+        (LegendreDelay(0.09, 9), [0.001, 0.03]),
+        (LegendreDelay(0.09, 21), [0.001, 0.03]),
```
arvoelke, line 259: I think something might be off by +/- 1 time-steps as this tolerance should be able to be dropped significantly for larger orders. For example, if I bump the order up to 512 then the unit test generates thie plot:

Screenshot from 2020-08-30 15-35-23

which looks like it might be off by one step (or maybe [half a step](https://github.com/arvoelke/nengolib/issues/169)). The RMSE should be able to approach 0 as this is scaled up.

I spent some time fiddling with a few things in the unit test and couldn't figure it out though. So unless you can think of something I think it's fine as is (they plots look pretty accurate).

nengo/tests/test_reprs.py
```diff
@@ -378,15 +378,14 @@ def test_distributions():
     check_repr(ScatteredHypersphere(surface=True))
     check_repr(ScatteredHypersphere(min_magnitude=0.3))
     check_repr(ScatteredHypersphere(base=Uniform(0, 1)))
-    check_repr(ScatteredHypersphere(method="sct"))
```
arvoelke: This failed due to importing scipy, so removed (can just check "tfww" instead).

nengo/utils/tests/test_numpy.py
arvoelke: I couldn't find the explicit test so I added one. Feel free to drop commit `9e51fcf6e7cd55ad9cc83fcec9532418034a70b4` if this was an oversight on my part.

nengo/synapses.py
```diff
@@ -212,6 +212,18 @@ class LinearFilter(LinearSystem, Synapse):
...
+        ``state_size``.
+
+        .. versionadded:: 3.1.0
+    initial_output : Distribution or float or (n_synapses,) array_like, optional
+        Initial output value(s) represented by the synapses. ``n_synapses`` is typically
+        equal to the connection's ``size_out``, except when ``post_obj`` is an
+        ``Ensemble`` and ``solver.weights`` is set, in which case ``n_synapses`` equals
+        the number of neurons in the post ``Ensemble``.
+
+        .. versionadded:: 3.1.0
```
hunse, line 226: So these are both parameters from base classes, which is why I didn't document them here. I can understand the desire to document them here, particularly because they're passed explicitly rather than as part of a ``**kwargs``. This is especially helpful because we've got diamond inheritance here, so it's extra confusing for a user where to look if they want documentation for these arguments. The downside, though, is then we have duplication of docstrings. I'm wondering if it's possible to make some sort of note (ideally both here and in the original docstrings) so that if one of them changes, we know to update the other location. I'm not sure how best to do that, though.

arvoelke: Could something like https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html be used to define and reuse the same description in multiple places? There could also be some sort of 'inherit params' magic annotations. Seems like a fairly general problem that we wouldn't be the first to encounter.

nengo/utils/tests/test_numpy.py
hunse: Yeah, I found it in my stashes. Guess I forgot to commit it! This looks fine, though.

nengo/synapses.py
hunse: Sure!

nengo/synapses.py
hunse: Yes, probably a good idea.

hunse: I left `OneXScalar` untouched, just in case `squeeze` has a bit of extra overhead. I changed the rest, though.

nengo/synapses.py
hunse: Fixed. Also added a test that catches this (we were never testing `Highpass` with higher order before).

nengo/synapses.py
hunse: I removed them. They're not attributes of `LinearSystem` either, since it stores things in state-space form. It does have the attributes `A`, `B`, `C`, and `D`, but I haven't listed them there because the same information can be accessed through the `ss` property, which *is* documented.

nengo/tests/test_params.py
hunse: `IntParam` is a child of `NumberParam`, so it would just fall back on that `conv_func`. Similarly, `DistOrArrayParam` just falls back on `NdArrayParam`. I think this is fine in both cases.

nengo/tests/test_synapses.py
hunse: Added this.

nengo/tests/test_synapses.py
hunse: It does look slightly off. I'm not sure about the reason either, though.

nengo/tests/test_processes.py
```diff
@@ -538,3 +539,310 @@ def func(t):
...
+    def test_discretize(self, Simulator, plt):
+        dt = 0.003
+        sys = LinearSystem(([1], [0.02, 1]), method="euler")
+        f_in = lambda t: np.sin(20 * t)
+        t, y0 = self.run_sim(Simulator, sys, dt=dt, plt=plt, f_in=f_in)
+        t, y1 = self.run_sim(Simulator, sys.discretize(dt), dt=dt, plt=plt, f_in=f_in)
+        t, y2 = self.run_sim(
+            Simulator, sys.discrete_ss(dt), analog=False, dt=dt, plt=plt, f_in=f_in
+        )
+        assert np.allclose(y1, y0)
```
tbekolay, line 681: Are these and other `np.allclose` calls intentionally not using the `allclose` fixture?

nengo/dists.py
```diff
@@ -306,6 +344,273 @@ def sample(self, n, d=None, rng=np.random):
...
+    References
+    ----------
+    .. [#] Martin Roberts. "The Unreasonable Effectiveness of Quasirandom Sequences."
+       http://extremelearning.com.au/unreasonable-effectiveness-of-quasirandom-sequences/
+
+    Examples
+    --------
+    .. testcode::
+
+        from nengo.dists import Rd
```
tbekolay, line 368: Note to self: indentation

nengo/builder/learning_rules.py
```diff
@@ -461,6 +461,94 @@ def step_simvoja():
...
 
 
+class SimRLS(Operator):
+    r"""Calculate connection weight change according to the RLS rule.
+
+    Implements the Recursive Least Squares (RLS) learning rule of the form
+
+    .. math::
+
+        g_i &= \sum_j P_{ij}(n-1) r_i \\
```
tbekolay, line 471: NTS: indentation

nengo/tests/test_processes.py
hunse: No, just an oversight I think. I'd definitely change the ones in this test to `allclose`, because it uses `Simulator`. The others don't matter so much because they don't use `Simulator`, so they aren't for backends. But there's no harm in using `allclose`, I don't think.

tbekolay (commented):
As we discussed offline, I split this up into the more straightforward improvements (this PR) and the LinearSystem stuff (#1650). I still have a few things left to look into in this PR, but had a few questions that I wanted to talk about at tomorrow's scrum, so I pushed what I have so far.

nengo/dists.py
```diff
@@ -306,6 +346,362 @@ def sample(self, n, d=None, rng=np.random):
...
+    for sampling ``eval_points`` and ``encoders`` in Nengo.
+
+    Parameters
+    ----------
+    surface : ``boolean``
+        Set to ``True`` to restrict the points to the surface of the hypersphere.
+        Set to ``False`` to sample from the ball (i.e. the volume of the hypersphere).
+    base : `nengo.dists.Distribution`, optional
+        The base distribution from which to draw "quasi Monte Carlo" samples.
+        Defaults to :class:`.Rd` and should not be changed unless
```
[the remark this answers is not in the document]
arvoelke: In nengolib this parameter could take on two different values, with the second being: https://arvoelke.github.io/nengolib-docs/nengolib.stats.Sobol.html#nengolib.stats.Sobol. That one wasn't ported over but there are dozens of these number theoretic methods out there that one could plug in.

nengo/dists.py
```diff
@@ -306,6 +346,362 @@ def sample(self, n, d=None, rng=np.random):
...
+                ratio0 = (dims1 - dims) / (dims1 - dims0)
+                ratio1 = 1 - ratio0
+                xp = (ratio0 * xp0 + ratio1 * xp1) if len(xp0) == len(xp1) else xp0
+                yp = ratio0 * np.interp(xp, xp0, yp0) + ratio1 * np.interp(xp, xp1, yp1)
+
+        return np.interp(x, xp, yp)
+
+
+class ScatteredHypersphere(UniformHypersphere):
+    r"""Number--theoretic distribution over the hypersphere or hyperball.
```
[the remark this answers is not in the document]
arvoelke: These terms are often used interchangeably in the literature. See the first line of this abstract for instance: https://ui.adsabs.harvard.edu/abs/2003SPIE.5091..364H/abstract but we can be more consistent sure. This is why though.

nengo/dists.py
```diff
@@ -306,6 +346,362 @@ def sample(self, n, d=None, rng=np.random):
...
+    @staticmethod
+    def random_orthogonal(d, rng=np.random):
+        """Returns a random orthogonal matrix."""
+        m = rng.standard_normal((d, d))
+        u, _, v = np.linalg.svd(m)
+        return np.dot(u, v)
+
+    def sample(self, n, d=1, rng=np.random):
+        if d == 1:
+            return super().sample(n, d, rng=rng)
```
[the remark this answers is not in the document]
arvoelke: The super here shouldn't be needed but I haven't looked at that code path in a long time.

nengo/dists.py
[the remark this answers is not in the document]
arvoelke: Another way to respond is there don't exist any methods of generating quasirandom sequences that don't use number theory to ensure that they are low-discrepancy. None that I'm aware of..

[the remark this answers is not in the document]
tbekolay: Thanks for the context! Googling the number-theory based terms (e.g. [number-theoretic net](https://www.google.com/search?q=number-theoretic+net)) gives unhelpful results, while [quasirandom sequence](https://www.google.com/search?q=quasirandom+sequence) gives spot on results, so I'm going to use that terminology throughout.

nengo/dists.py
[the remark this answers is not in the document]
tbekolay: I see that the code for `Sobol` is [not very Pythonic](https://github.com/arvoelke/nengolib/blob/v0.5.1/nengolib/stats/_sobol_seq.py), is that the main reason it wasn't ported over? In any case, if there are other things that can be plugged in here then leaving it in makes sense. I'll reword the docs here though.

nengo/dists.py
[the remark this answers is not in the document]
tbekolay: I'll test it out :+1:

nengo/dists.py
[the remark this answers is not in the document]
arvoelke: Yeah it was also slow and didn't scale past 40 dimensions. There are very likely good Python implementations nowadays for Sobol and Halton as they are considered the classics, but I haven't checked in a long times. :+1: to rewording.

nengo/dists.py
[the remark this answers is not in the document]
tbekolay: Alright I did some testing. I think it's a larger issue so I'm going to take this discussion offline.

nengo/dists.py
```diff
@@ -306,6 +346,362 @@ def sample(self, n, d=None, rng=np.random):
...
+            radius **= 1.0 / d
+
+        if self.method == "sct":
+            mapped = self.spherical_transform_sct(samples, approx=False)
+        elif self.method == "sct-approx":
+            mapped = self.spherical_transform_sct(samples, approx=True)
+        elif self.method == "tfww":
+            mapped = self.spherical_transform_tfww(samples)
+        else:
+            raise NotImplementedError(self.method)
```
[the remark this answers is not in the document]
hunse: Don't we not care about coverage for `NotImplementedErrors`?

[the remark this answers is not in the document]
tbekolay: Ah yeah, that's a good point. I guess it feels weird though because it's not like we have any intention of implementing more methods later?

nengo/dists.py
```diff
@@ -306,6 +346,358 @@ def sample(self, n, d=None, rng=np.random):
...
+        """Returns a random orthogonal matrix."""
+        m = rng.standard_normal((d, d))
+        u, _, v = np.linalg.svd(m)
+        return np.dot(u, v)
+
+    def sample(self, n, d=1, rng=np.random):
+        if d == 1:
+            return super().sample(n, d, rng=rng)
+
+        base = QuasirandomSequence()
```
hunse, line 671: I might put this in the constructor, partly for performance and partly so that if someone (probably one of us) wants to play around and try a different base, we can. I'm fine keeping it hidden (e.g. in `self._base`).

nengo/dists.py
```diff
@@ -306,6 +346,358 @@ def sample(self, n, d=None, rng=np.random):
...
+            z = (seed + (1 / n) * np.arange(n, dtype=np.float64)) % 1
+            return z[:, np.newaxis]
+        else:
+            phi_inv = 1.0 / self._phi(d)
+            alpha = phi_inv ** np.arange(1, d + 1)
+            z = (seed + np.outer(np.arange(1, n + 1), alpha)) % 1
+            return z
+
+
+class _betaincinv22:
```
hunse, line 407: I might move this to the bottom so it's not distracting from the actual distributions.

nengo/dists.py
tbekolay: Ended up reverting this part of things since there are other sequences that could be put in eventually.

nengo/dists.py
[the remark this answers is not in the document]
tbekolay: The resolution to that offline discussion is to use the `base` sequence for `d == 1`. It is only slightly random (we randomly offset the first sample) but eval_points and encoders aren't biological things anyway so we're not changing how biologically plausible anything is.

I am also making a separate backlog item to look into using the quasirandom `d == 1` behavior as an alternative to the `Uniform` distribution for things like max rates and intercepts. That'll be a bigger change (in terms of model impact) so definitely out of the scope of this PR.

tbekolay (approved):
Alright I encountered a few minor things but I think I've got it all worked out now and this all looks good to me. I made some changes though that would be good for @hunse and/or @arvoelke to sign off on.

The big thing is changing the behavior of `ScatteredHypersphere` when `d == 1`. I _believe_ the references that are linked in various docstrings don't give any guidance as to what we should do for `d == 1` so I did what felt right after chatting about it with Eric (but if I'm mistaken and those references do have things to say, please point me to that).

So here's what happens:

- For `d == 1 and surface is True`, we alternate between 1 and -1. The only randomness is whether we start with 1 or -1; so, if you're sampling 4 numbers it'll be `[-1, 1, -1, 1]` or `[1, -1, 1, -1]`. I went with this for simplicity and because Eric didn't want it to clump such that you have long runs of 1s or -1s.

  One downside of this is that two ensembles with the same `n_neurons` and `dimensions` will have the same encoders half of the time. We have a test that ensures that changing seeds changes all attributes of an ensemble, and this fails 50% of seeds because encoders will be same when changing the seed. This is reflective of there being little diversity in encoders for 1D populations by default. I'm not sure if this is an issue or not, and if it is an issue, how it should be resolved.

- For `d == 1 and surface is False`, we draw `n` samples from `self.base` and then map from the `[0, 1)` range to `(-1, min_mag] U [min_mag, 1)` by scaling to the `[min_mag, 1)` range and then changing the sign of alternating samples in the same manner as we change the sign of encoders. When you inspect the generated samples, you get what you would expect in that it optimally tiles the `(-1, 1)` space with the `(-min_mag, min_mag)` cutout with a random offset.

Does this seem like an appropriate thing to do for `d == 1`? Keep in mind that right now we're only doing this for `eval_points` and `encoders`. We could, in the future, do something similar for `Uniform` and use this for more things (like `intercepts`, `max_rates`, etc) but that's a bigger change so it's out of the scope of this PR.

----

The only other potentially controversial change that I made was to add `import matplotlib.pyplot as plt` to our `doctest_setup`. I'm not sure if this has a significant performance impact, but I don't think it should. The bigger change would be to enable [`numpydoc_use_plots`](https://github.com/arvoelke/nengolib/blob/master/docs/conf.py#L57), but that is also outside of the scope of this PR so I'll track that separately.

Oh, and I also removed the force-learning example from this branch because it depended on `Bandpass`. It's still in #1650, but for now I instead added `nengo.RLS` to the `learn-product` example.

tbekolay:
After some discussion with @hunse, I made the following changes:

1. For the `d == 1 and surface is False` case, we sample and scale twice, and on the second sampling we make all those values negative.
2. For both `d == 1 and surface is False` and all `QuasirandomSequence` cases, we shuffle the samples before returning them to break any unwanted correlations. This means that in the `d == 1 and surface is True` case I could simplify to just using `np.sign` as the samples we get from `self.base` are in shuffled order, which also means that we have far more diversity in encoders.

The lack of encoder diversity was the major issue from the previous implementation IMO, so I think with those things this is good to go. There are some lines in `synapses.py` that are now uncovered for some reason, so I'm going to look into that and make sure they're covered before merging. Hopefully that won't take too long though so if anyone else wants to give some thoughts before merging, better act fast!

tbekolay:
The uncovered lines were because vendorizing the filter design stuff (3bb2b39509d14ed596558005a9ae7d7413450508) made it such that it was impossible to make a `LinearFilter.NoX` step function; not sure if that's because of something that changed in the newer functions, or if we did something custom there that we missed (or intentionally changed) when vendorizing. Whatever the case may be, I just cut that commit out for now and we'll deal with it #1650. Merging once CI finishes!

PR #1615, opened 2020-05-15
Added Repr Tests

AlexEliasmith:
Added tests for processes, solvers, transforms, and learning rules. The reason this is a new PR is because the old one had bad names and descriptions that were throwing errors, and I couldn't seem to fix it.


nengo/learning_rules.py
```diff
@@ -138,12 +126,12 @@ def __init__(self, learning_rate=Default, pre_synapse=Default):
...
 
-    @property
-    def _argdefaults(self):
-        return (
-            ("learning_rate", PES.learning_rate.default),
-            ("pre_synapse", PES.pre_synapse.default),
-        )
+
+def _remove_default_post_synapse(argreprs, pre_synapse):
+    default_post_synapse = "post_synapse=%r" % (pre_synapse,)
```
hunse, line 131: I would maybe call it `default` instead of `pre_synapse`. I thought it was a typo (supposed to be `post_synapse`) until I looked below and realized that the default value is always `pre_synapse`, that's where the naming comes from.

tbekolay: Good call, done!

PR #1619, opened 2020-05-21
Code completion

AlexEliasmith:
Code Completion

- Medium (Plenty of code, but changes are often straightforward tests)

**Still to do:**
Will add more tests, this is mostly an intermediate step so travis-ci tells me which lines I have visited


nengo/utils/ipython.py
```diff
@@ -9,18 +9,18 @@
     from IPython import get_ipython
     from IPython.display import HTML
 
-    if IPython.version_info[0] <= 3:
+    if IPython.version_info[0] <= 3:  # pragma: no cover
         from IPython.nbconvert import PythonExporter
-    else:
+    else:  # pragma: no cover
```
hunse, line 14: We should only have to "no cover" one of these two branches (I'm guessing the one for older IPython).

nengo/utils/matplotlib.py
```diff
@@ -200,23 +198,3 @@ def rasterplot(time, spikes, ax=None, use_eventplot=False, **kwargs):  # noqa
     ax.yaxis.set_ticks_position("none")
 
     return ax
-
-
-def plot_tuning_curves(ensemble, sim, connection=None, ax=None):
```
hunse, line 205: Are we okay to just get rid of this, or do we want to deprecate and remove in 4.0 just in case somebody is using it?

nengo/utils/numpy.py
```diff
@@ -251,7 +251,7 @@ def rmse(x, y, axis=None, keepdims=False):  # pragma: no cover
 
 if hasattr(np.fft, "rfftfreq"):
     rfftfreq = np.fft.rfftfreq
-else:
+else:  # pragma: no cover
```
hunse, line 254: We might want to monkeypatch into here just so that we can test that our `rfftfreq` does the same as Numpy's.

nengo/utils/paths.py
```diff
@@ -2,10 +2,10 @@
 import sys
 
 
-if sys.platform.startswith("win"):
+if sys.platform.startswith("win"):  # pragma: no cover
     config_dir = os.path.expanduser(os.path.join("~", ".nengo"))
     cache_dir = os.path.join(config_dir, "cache")
-else:
+else:  # pragma: no cover
```
hunse, line 8: Shouldn't need this

nengo/utils/tests/test_ensemble.py
```diff
@@ -4,7 +4,20 @@
...
+    """Ensures all cases are run"""
+    encoders = [1, 2, 3]
+    index = 1
+    rows = 1
+    cols = 2
+
+    assert _similarity(encoders, index, rows, cols) == 2.0
+    index = 0
+    cols = 2
+    assert _similarity(encoders, index, rows, cols) == 2.0
```
hunse, line 20: It would be good to test a case that doesn't return `2.0`

nengo/utils/tests/test_filter_design.py
```diff
@@ -1,11 +1,133 @@
...
+
+def test_ss2zpk():
+    """test the function ss2zpk"""
+    predicted = (np.array([0.0]), np.array([1.0]), 1.0)
+    assert repr(ss2zpk([1], [1], [1], [1])) == repr(predicted)
 
 
 @pytest.mark.parametrize("dt", [1e-3, 1e-2, 1e-1])
 def test_cont2discrete_zoh(dt, allclose):
+    """test the function ss2zpk"""
```
hunse, line 130: Should say `cont2discrete`

nengo/utils/tests/test_least_squares_solvers.py
```diff
@@ -0,0 +1,66 @@
...
+def test_randsvd():
+    """tests RandomizedSVD"""
+    pytest.importorskip("sklearn")
+    mySVD = RandomizedSVD()
+    A = np.array([[1], [2], [3]])
+    Y = np.array([1, 2, 3])
+    sigma = 0
+    assert (
+        repr(mySVD(A, Y, sigma)) == "(array([[1.]]), {'rmses': "
+        "array([6.69210662e-16])})"
```
hunse, line 63: Too fragile to check an exact RMSE.

nengo/utils/tests/test_magic.py
```diff
@@ -1,10 +1,135 @@
...
+        " '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__',"
+        " '__weakref__', 'test']"
+        or str(dir(my_proxy))
+        == "['__class__', '__delattr__', '__dict__', '__dir__', '__doc__',"
+        " '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__',"
+        " '__hash__', '__init__', '__le__', '__lt__',"
+        " '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__',"
+        " '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__',"
+        " '__weakref__', 'test']"  # Travis-ci has this one
+    )
```
hunse, line 69: This seems really fragile. Why do we care about the exact `dir`?

nengo/utils/tests/test_progress.py
```diff
@@ -1,9 +1,50 @@
...
+        elapsed_seconds = myFunction
+
+        def eta(self):
+            return 0
+
+    data = Test(False)
+    prog.update(data)
+
+    data = Test(True)
+    prog.update(data)
```
hunse, line 47: What does this test assert? We should probably open the written file and make sure it contains what we expect.

nengo/utils/tests/test_progress_file
```diff
@@ -0,0 +1 @@
+0%, ETA: 0:00:00
```
hunse, line 1: Also committed accidentally I assume.

nengo/utils/tests/test_stdlib.py
```diff
@@ -208,6 +211,44 @@ def test_weakkeydict_update():
...
+    assert repr(d.iterkeys()).startswith(repr(d._keyrefs.values())[0:43])
+
+    results = list(d.iteritems())
+    start = "(<nengo.utils.tests.test_stdlib.C object "
+    end = [
+        "1)",
+        "2)",
+        "3)",
+    ]
+    end2 = [  # ending for travis-ci
```
hunse, line 233: So the reason there's a difference in TravisCI is likely because in Python <= 3.5, dictionary items do not necessarily appear/iterate in the order that they were added. We should modify this test so that it's not dependent on dictionary order (e.g. just asserting that the expected items appear in the dictionary, not that they appear in a particular order).

nengo/builder/connection.py
```diff
@@ -94,12 +94,7 @@ def build_linear_system(model, conn, rng):
     eval_points = get_eval_points(model, conn, rng)
     ens = conn.pre_obj
     activities = get_activities(model.params[ens], ens, eval_points)
-    if np.count_nonzero(activities) == 0:
-        raise BuildError(
```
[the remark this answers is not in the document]
tbekolay: It was non-trivial to hit this `BuildError` -- you basically had to be calling it with unrealistic inputs, since the builder will never get to this point if there were all zero inputs. So we opted to make it an `assert` so that we'd still know if we got here, but didn't have to make an extremely artificial test case.

nengo/utils/matplotlib.py
tbekolay: Up to you, I suggested removing it because I didn't realize it existed and we don't use it anywhere in our examples (though we do plot tuning curves, just not with this function).

nengo/utils/tests/test_magic.py
tbekolay: `ObjectProxy` explicitly defines the `__dir__` magic method, so to cover that we would need to call `dir`. That said, we should check the lists directly rather than comparing the string representation of those lists.

nengo/utils/tests/test_progress_file
tbekolay: If these get generated as part of testing the progress bar, we should add these to `.gitignore` also.

nengo/tests/test_rc.py
```diff
@@ -0,0 +1,52 @@
...
+
+    file_dir = os.path.abspath(
+        os.path.join(os.path.dirname(__file__), os.pardir, os.pardir)
+    )
+
+    fp = open("%s/nengo/tests/test_rc_file1.txt" % file_dir, "r")
+
+    fp2 = open("%s/nengo/tests/test_rc_file1.txt" % file_dir, "r")
+
+    class TestFile:
```
hunse, line 16: What's the point in testing with this mock file class? This function is just testing `test_read_file`, which is a pretty simple function on our end. I don't think we need this to cover the lines.

nengo/builder/connection.py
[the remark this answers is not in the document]
hunse: I've been able to hit it by having neurons with high intercepts, then using eval points that never cross those intercepts.

nengo/utils/tests/test_progress_file
hunse: I opted to create the file on-the-fly in the test itself.

nengo/utils/tests/test_magic.py
```diff
@@ -228,3 +232,149 @@ def __init__(self, a, b):
...
+    assert str(proxy) == str(obj)
+
+    proxy.newattr = "newval"  # pylint: disable=assigning-non-slot
+    assert obj.newattr == "newval"
+    assert proxy.newattr == "newval"
+
+    assert repr(proxy).startswith("<ObjectProxy at ")
+
+
+def test_boundfunctionwrapper():
```
hunse, line 337: It would be nice to have a more functional test of BoundFunctionWrapper, but I can't figure out how to use a lot of its functionality. I can't trigger the `__get__` function at all, and I also can't seem to get `self.instance is None`. Any ideas @tbekolay?

tbekolay: I really don't think it's that important; everything in the `magic.py` file is from [wrapt](https://pypi.org/project/wrapt/), which has its own test suite. We could vendorize it.

PR #1625, opened 2020-07-15
Removed items from pylint disable list

WAEliasmith:
Items were removed from disable list.
Code was updated to remove pylint errors.

**Motivation and context:**
Following more of pylint's guidelines means cleaner code

**Checklist:**


nengo/base.py
```diff
@@ -287,7 +287,7 @@ def __init__(
         self.default_dt = default_dt
         self.seed = seed
 
-    def apply(self, x, d=None, dt=None, rng=np.random, copy=True, **kwargs):
+    def apply(self, x, d=None, dt=None, rng=np.random, make_copy=True, **kwargs):
```
hunse, line 290: I would prefer to keep this as `copy`. `np.array` uses `copy`, so there's similarity there.

nengo/builder/builder.py
```diff
@@ -131,8 +132,15 @@ def build(self, obj, *args, **kwargs):
...
+
+        def NoneFunction(other):
+            return
+
+        function_to_call = NoneFunction
         if self.build_callback is not None:
-            self.build_callback(obj)
+            function_to_call = self.build_callback
+        if function_to_call is not NoneFunction:
+            function_to_call(obj)
```
hunse, line 143: This seems much more confusing than the original code. What was `pylint`'s problem?

Also, as a general principle, I don't mind adding a `pylint: disable=...` to a particular line if it means keeping the code cleaner (sometimes, `pylint` is just out to lunch IMO). But I think having the disables on specific lines is still better than having them in the disable list, since then we're not ignoring them everywhere.

drasmuss (approved):
Added various fixups, LGTM

docs/examples/advanced/functions-and-tuning-curves.ipynb
```diff
@@ -43,7 +43,7 @@
    "outputs": [],
    "source": [
     "n_neurons = 150\n",
-    "duration = 2."
```
hunse, line 46: With Python 3 we could just make these integers.

nengo/processes.py
```diff
@@ -299,7 +299,7 @@ def coerce(self, instance, data):
             # make sure this is the same size as previous items
             if size != size_out and size_out is not None:
                 raise ValidationError(
-                    "time %g has size %d instead of %d" % (time, size, size_out),
+                    "time %g has size %d instead of %s" % (time, size, size_out),
```
hunse, line 302: Why did this change?

nengo/tests/test_config.py
```diff
@@ -49,18 +49,18 @@ def test_config_basic():
...
         model.config[a].something_else = 1
     with pytest.raises(AttributeError):
         model.config[a2b].something = 1
 
     with pytest.raises(ConfigError, match="'str' is not set up for configuration"):
-        model.config["a"].something
+        assert model.config["a"].something
     with pytest.raises(ConfigError, match="'NoneType' is not set up for configuration"):
-        model.config[None].something
+        assert model.config[None].something
```
hunse, line 63: I know why `pylint` wants this, but in this case, I feel like it's misleading because it looks like the statement is what we're asserting, rather than the error.

Could we do `_ = model.config[None].something`? Or would `pylint` complain about that, too.

nengo/tests/test_transforms.py
```diff
@@ -175,6 +182,7 @@ def run_sim():
 @pytest.mark.parametrize("encoders", (True, False))
 @pytest.mark.parametrize("decoders", (True, False))
 def test_sparse_nef(encoders, decoders, Simulator):
+    """tests errors when using nef for sparse"""
     # Sparse transforms currently don't work with NEF connections,
     # so just check the errors
```
hunse, line 187: Make this comment the docstring

nengo/utils/progress.py
```diff
@@ -782,11 +781,7 @@ def get_default_progressbar():
     if pbar.lower() == "none":
         return NoProgressBar()
 
-    try:
-        return _load_class(pbar)()
-    except Exception as e:
-        warnings.warn("Could not load progress bar: %s" % e)
-        return NoProgressBar()
```
hunse, line 789: Do we want to fail if we can't load it? A warning seems better to me.

nengo/tests/test_simulator.py
```diff
@@ -35,7 +37,7 @@ def test_steps(Simulator, allclose):
 
 @pytest.mark.parametrize("bits", ["16", "32", "64"])
 def test_dtype(Simulator, request, seed, bits):
-    # Ensure dtype is set back to default after the test, even if it fails
```
hunse, line 38: This is not the test description, it's a comment on the finalizer.

nengo/tests/test_learning_rules.py
```diff
@@ -615,19 +633,23 @@ def test_custom_type(Simulator, allclose):
...
     class TestRule(nengo.learning_rules.LearningRuleType):
+        """Test class"""
+
         modifies = "decoders"
 
         def __init__(self):
             super().__init__(1.0, size_in=3)
 
-    @Builder.register(TestRule)
+    # @Builder.register(TestRule)
```
hunse, line 643: Remove this comment

hunse (changes requested):
I would break the test docstring additions out into a separate commit, particularly since we're not actually requiring them yet. I didn't go through them all, so I'll come back and do that later.

nengo/builder/builder.py
```diff
@@ -132,15 +132,8 @@ def build(self, obj, *args, **kwargs):
...
-
-        def NoneFunction(other):
-            return
-
-        function_to_call = NoneFunction
         if self.build_callback is not None:
-            function_to_call = self.build_callback
-        if function_to_call is not NoneFunction:
-            function_to_call(obj)
+            self.build_callback(obj)  # pylint: disable=not-callable
```
tbekolay, line 136: Does adding `and callable(self.build_callback)` silence the pylint warning? Would be more explicit.

nengo/builder/optimizer.py
```diff
@@ -94,8 +94,9 @@ def optimize(model, dg):
         scaled_rate = threshold * mean_reduction_rate
         if 0.0 < last_reduction_rate < scaled_rate:  # pragma: no cover
             logger.debug(
-                "Operator reduction rate fell below {} mean reduction rate. "
-                "Stopping optimizer.".format(threshold)
+                "Operator reduction rate fell below %d mean reduction rate. "
```
tbekolay, line 97: This can't be `%d` as `threshold` is a flaot

nengo/networks/assoc_mem.py
```diff
@@ -14,7 +18,7 @@ def filtered_step(t, shift=0.5, scale=50, step_val=1):
     return np.maximum(-1 / np.exp((t - shift) * scale) + 1, 0) * step_val
 
 
-class AssociativeMemory(nengo.Network):
+class AssociativeMemory(Network):
```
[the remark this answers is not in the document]
tbekolay: I kind of like this style for network subclasses personally; they tend to be larger / more complicated than the networks we put in examples, so saving on typing seems more useful here. And if people happen to look at these files as they're getting into larger / more complicated stuff they get to see both styles and can decide which they like for their own models.

nengo/processes.py
tbekolay: The warning you get is `Argument 'builtins.NoneType' does not match format type 'd' (bad-string-format-type)` because pylint doesn't look at the if block context, at least in this case. I'd stick with `%s` though because it's going to give you the same thing as `%d`.

nengo/tests/test_config.py
tbekolay: I've been changing these to `print`s instead of asserts. It silences pylint, which is good, and if the statement is raising an error (as it should) it won't print anything. It has the side benefit that if it doesn't raise an error then it prints out the thing that you're not supposed to be able to do.

nengo/utils/progress.py
tbekolay: Eh, if a user gets to this point they're doing something pretty specific, so I think they would not be upset to be told about it (a warning is easy to ignore / miss). I think 0% of users have seen this warning.

nengo/utils/stdlib.py
```diff
@@ -239,7 +239,7 @@ def get_terminal_size(fallback=(80, 24)):
 
     try:
         return shutil.get_terminal_size(fallback)
-    except Exception:  # pragma: no cover
+    except Exception:  # pragma: no cover pylint: disable = broad-except
```
tbekolay, line 242: I think we actually did this for backwards compat reasons as `shutil` didn't have `get_terminal_size` until Python 3.3. Will just remove it.

tbekolay (approved):
I made several changes, I think it all looks good to me now, but will let @hunse do a last pass of it. I removed most of the test docstrings as they were mostly rephrasings of the test names; possibly down the line if we choose to reenable `missing-docstring` we could omit that check for functions in the test folders?

hunse (approved):
LGTM. I made a minor change to just import all the operators rather than having `operator.` in front of all of them in `nengo/builder/optimizer.py`.

PR #1626, opened 2020-07-17
Weight solver warning

WAEliasmith:
Changed the error to a warning for nodes.
Added tests to ensure this is working.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -68,6 +68,10 @@ Release history
   values. (`#1609`_)
 - The ``nl`` and ``nl_nodirect`` test arguments have been renamed to ``AnyNeuronType``
   and ``NonDirectNeuronType``. (`#1609`_)
+- Weight solvers (i.e. those with ``weights=True``) are now allowed on all connections.
+  For connections that are not between ``Ensembles``, though, weight solvers have the
+  same effects as solvers with ``weights=False``, and a warning will be raised.
+  (`#1626 <https://github.com/nengo/nengo/pull/1626>`__)
 
 **Deprecated**
 
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy as np
 
 from nengo.base import NengoObject, NengoObjectParam, ObjView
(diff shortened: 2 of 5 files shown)


nengo/connection.py
```diff
@@ -122,20 +123,28 @@ class ConnectionSolverParam(SolverParam):
...
                 raise ValidationError(
                     "weight solvers only work for connections from ensembles "
                     "(got %r)" % type(conn.pre_obj).__name__,
                     attr=self.name,
                     obj=conn,
                 )
-            if solver.weights and not isinstance(conn.post_obj, Ensemble):
+            elif solver.weights and isinstance(conn.pre_obj, Node):
+                warnings.warn(
+                    "weight solvers only work for connections to ensembles, not nodes"
```
hunse, line 135: Should say "from ensembles", not "to ensembles".

I'm actually wondering whether we want a warning at all in this case. For Node->Ensemble connections, using a solver with `weights=True` will result in weights that go from the node directly to the neurons (i.e. the encoders are incorporated into the weights). This is different from one with `weights=False` (where weights and encoders are separate), so there is a reason that someone might want to set `weights=True` on a Node->Ensemble connection.

nengo/connection.py
```diff
@@ -122,20 +123,28 @@ class ConnectionSolverParam(SolverParam):
...
+            if solver.weights and not isinstance(conn.post_obj, (Ensemble, Node)):
                 raise ValidationError(
                     "weight solvers only work for connections to ensembles "
                     "(got %r)" % type(conn.post_obj).__name__,
                     attr=self.name,
                     obj=conn,
                 )
+            elif solver.weights and isinstance(conn.post_obj, Node):
+                warnings.warn(
+                    "weight solvers only work for connections to ensembles, not nodes"
```
hunse, line 146: I wonder if we want to say a bit more about what's happening in this case. In this case, there's no difference between a solver with `weights=True` and `weights=False`, so maybe we want to make that clear (saying something like "setting `weights=True` on a solver has no effect when the post object is a Node"). Furthermore, the current statement is a bit misleading, because weight solvers do work, they just act the same as non-weight solvers.

CHANGES.rst
```diff
@@ -68,6 +68,10 @@ Release history
   values. (`#1609`_)
 - The ``nl`` and ``nl_nodirect`` test arguments have been renamed to ``AnyNeuronType``
   and ``NonDirectNeuronType``. (`#1609`_)
+- Weight solvers (i.e. those with ``weights=True``) now work on connections from
```
tbekolay, line 71: Probably "are now allowed" rather than "work on", which feels ambiguous as to what it means to "work". If you're consider their previous disallowance as a bug, then this should be in the "fixed" section.

CHANGES.rst
```diff
@@ -68,6 +68,10 @@ Release history
...
+  ``Node`` to ``Ensemble``. They also work on connections from ``Ensemble`` to ``Node``,
+  but issue a warning, since here they are identical those with ``weights=False``.
```
tbekolay, line 73: should be "identical _to_ those with"

CHANGES.rst
```diff
@@ -68,6 +68,10 @@ Release history
   values. (`#1609`_)
 - The ``nl`` and ``nl_nodirect`` test arguments have been renamed to ``AnyNeuronType``
   and ``NonDirectNeuronType``. (`#1609`_)
+- Weight solvers (i.e. those with ``weights=True``) now work on connections from
+  ``Node`` to ``Ensemble``. They also work on connections from ``Ensemble`` to ``Node``,
```
tbekolay, line 72: Also here, "work on" -> "allowed on"

nengo/connection.py
```diff
@@ -122,20 +124,27 @@ class ConnectionSolverParam(SolverParam):
     def coerce(self, conn, solver):
         solver = super().coerce(conn, solver)
         if solver is not None:
-            if solver.weights and not isinstance(conn.pre_obj, Ensemble):
+            if solver.weights and not isinstance(conn.pre_obj, (Ensemble, Node)):
                 raise ValidationError(
-                    "weight solvers only work for connections from ensembles "
+                    "weight solvers only work for connections from Ensembles and Nodes "
```
tbekolay, line 129: Could also clarify "work" here too

CHANGES.rst
hunse: And I was so happy with my line being under 88 characters.

tbekolay: Haha... for changelogs I still like to keep it something more like 80 characters, though not that strictly. :shrug:

hunse:
@tbekolay I think I've addressed all your suggestions. Does the new version look good to you?

PR #1627, opened 2020-07-24
Fix Signal.may_share_memory comparison with self

hunse:
**Motivation and context:**

A signal always shares memory with itself. This bug caused problems in NengoOCL (it's unclear whether it made things less efficient in Nengo core or NengoDL).

**How has this been tested?**

The idea has been tested with NengoOCL. NengoOCL doesn't yet support Nengo `master`, so this exact code hasn't been tested there yet.

Also, should add a unit test.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Add unit test

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -93,8 +93,11 @@ Release history
   (`#1541 <https://github.com/nengo/nengo/issues/1541>`__,
   `#1607 <https://github.com/nengo/nengo/pull/1607>`__)
 - Fixed a bug in the validation of ``Choice`` distributions. (`#1630`_)
+- Fixed a bug where a ``Signal`` did not register as sharing memory with itself.
+  (`#1627`_)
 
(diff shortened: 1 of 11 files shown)


hunse:
I modified the PES builder to do the encoder multiply outside of the `SimPES` operator. This is something that both NengoDL and NengoOCL were already doing (they were overriding the PES builder here), so we might as well make it part of this. It also makes it possible for the optimizer or backend to combine that multiply with other ones, for efficiency. The `SimPES` operator itself is also simplified significantly.

Unfortunately, NengoDL still will have to override the builder here, because it has a special requirement that the encoders not be sliced along the second axis, which requires an additional copy. NengoOCL will no longer need to override this builder.

drasmuss (approved):
Two minor fixups, LGTM

Issue #1473, opened 2018-09-15
nengo.dists.Choice([1]) behaviour can be improved


arvoelke:
If we create an ensemble such as:

```python
    nengo.Ensemble(100, 1, encoders=np.ones(100))
```

then we get a good error message:

```
ValidationError: Ensemble.encoders: ndarray must be 2D (got 1D)
```

However, as a `Choice` distribution:

```python
with nengo.Network() as model:
    nengo.Ensemble(100, 1, encoders=nengo.dists.Choice([1]))
with nengo.Simulator(model) as sim:
    pass
```

```
~/CTN/nengo/nengo/simulator.py in __init__(self, network, dt, seed, model, progress_bar, optimize)
    158                 # Build the network into the model
    159                 self.model.build(network,
--> 160                                  progress=pt.next_stage("Building", "Build"))
    161 
    162             # Order the steps (they are made in `Simulator.reset`)

~/CTN/nengo/nengo/builder/builder.py in build(self, obj, *args, **kwargs)
    121             The object to build into this model.
    122         """
--> 123         built = self.builder.build(self, obj, *args, **kwargs)
    124         if self.build_callback is not None:
    125             self.build_callback(obj)

~/CTN/nengo/nengo/builder/builder.py in build(cls, model, obj, *args, **kwargs)
    216                 "Cannot build object of type %r" % type(obj).__name__)
    217 
--> 218         return cls.builders[obj_cls](model, obj, *args, **kwargs)
    219 
    220     @classmethod

~/CTN/nengo/nengo/builder/network.py in build_network(model, network, progress)
     95         logger.debug("Network step 1: Building ensembles and nodes")
     96         for obj in network.ensembles + network.nodes:
---> 97             model.build(obj)
     98 
     99         logger.debug("Network step 2: Building subnetworks")

~/CTN/nengo/nengo/builder/builder.py in build(self, obj, *args, **kwargs)
    121             The object to build into this model.
    122         """
--> 123         built = self.builder.build(self, obj, *args, **kwargs)
    124         if self.build_callback is not None:
    125             self.build_callback(obj)

~/CTN/nengo/nengo/builder/builder.py in build(cls, model, obj, *args, **kwargs)
    216                 "Cannot build object of type %r" % type(obj).__name__)
    217 
--> 218         return cls.builders[obj_cls](model, obj, *args, **kwargs)
    219 
    220     @classmethod

~/CTN/nengo/nengo/builder/ensemble.py in build_ensemble(model, ens)
[...]

PR #1630, opened 2020-08-06
Fix validation of Choice dist with scalar samples

drasmuss:
**Motivation and context:**

`Choice([0, 1, 2, 3])` should have `dimensionality==0` (because we're sampling scalar values from that list of options). But it was being treated as `dimensions==1`, because `options.shape[1:] == []` and `np.prod([]) == 1`. This caused it to miss the validation for, e.g., `nengo.Connection(a.neurons, b.neurons, transform=Choice([0])` (this would correctly raise an error if `a.n_neurons > 1`, but the validation would be missed for `a.neurons == 1`, and we'd end up with a weird error downstream).

This supercedes https://github.com/nengo/nengo-dl/pull/166

Fixes https://github.com/nengo/nengo-dl/issues/124

**How has this been tested?**

Added a test for the 0-dimensionality validation check.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -91,9 +91,11 @@ Release history
 - Fixed a bug where the ``LstsqDrop`` solver errored when solving for zero weights.
   (`#1541 <https://github.com/nengo/nengo/issues/1541>`__,
   `#1607 <https://github.com/nengo/nengo/pull/1607>`__)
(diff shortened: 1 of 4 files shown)


arvoelke:
Also see: https://github.com/nengo/nengo/issues/1473

PR #1640, opened 2020-09-21
Fix bug with sliced neuron-to-neuron PES learning

drasmuss:
**Motivation and context:**

Applying PES learning to a neuron-to-neuron connection with a slice on the post-synaptic neurons did not work correctly (we were incorrectly applying the slicing to the state dimension instead of the neuron dimension).

**How has this been tested?**

Added a new test, see `test_pes_neuron_neuron_slice`.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/connection.py
+++ b/nengo/connection.py
@@ -60,7 +60,14 @@ class ConnectionLearningRuleTypeParam(LearningRuleTypeParam):
             if not isinstance(conn.post_obj, Ensemble):
                 raise ValidationError(
                     "'post' must be of type 'Ensemble' (got %r) "
-                    "for learning rule '%s'" % (type(conn.pre_obj).__name__, rule),
+                    "for learning rule '%s'" % (type(conn.post_obj).__name__, rule),
+                    attr=self.name,
+                    obj=conn,
+                )
+            if conn._to_neurons:
+                raise ValidationError(
+                    "Cannot use an encoder learning rule with a direct neural "
+                    "connection (since encoders are not used in that case).",
                     attr=self.name,
                     obj=conn,
                 )
@@ -91,10 +98,10 @@ class ConnectionLearningRuleTypeParam(LearningRuleTypeParam):
 
         if rule.modifies == "weights":
             # If the rule modifies 'weights', then it must have full weights
-            if conn.is_decoded:
+            if not conn._to_neurons:
                 raise ValidationError(
-                    "Learning rule '%s' can not be applied to decoded "
-                    "connections. Try setting solver.weights to True or "
+                    "Learning rule '%s' can only be applied on connections to neurons. "
+                    "Try setting `solver.weights` to True or "
                     "connecting between two Neurons objects." % rule,
                     attr=self.name,
                     obj=conn,
@@ -411,10 +418,6 @@ class Connection(NengoObject):
 
     Attributes
     ----------
-    is_decoded : bool
-        True if and only if the connection is decoded. This will not occur
-        when ``solver.weights`` is True or both pre and post are
-        `~nengo.ensemble.Neurons`.
     function : callable
         The given function.
     function_size : int
@@ -562,11 +565,24 @@ class Connection(NengoObject):
 
     @property
     def is_decoded(self):
+        warnings.warn(
+            "is_decoded is deprecated; directly check the pre/post objects for the "
+            "properties of interest instead",
+            DeprecationWarning,
+        )
         return not (
             self.solver.weights
             or (
                 isinstance(self.pre_obj, Neurons) and isinstance(self.post_obj, Neurons)
             )
+        )
+
+    @property
+    def _to_neurons(self):
+        return isinstance(self.post_obj, Neurons) or (
+            isinstance(self.pre_obj, Ensemble)
+            and isinstance(self.post_obj, Ensemble)
+            and self.solver.weights
         )
 
     @property
--- a/nengo/tests/test_connection.py
+++ b/nengo/tests/test_connection.py
@@ -1015,10 +1015,12 @@ def test_connectionlearningruletypeparam():
         a = nengo.Ensemble(10, 1)
         b = nengo.Ensemble(11, 1)
 
-        with pytest.raises(ValueError):  # need a 2D transform for BCM
+        with pytest.raises(
+            ValidationError, match="can only be applied on connections to neurons"
+        ):
             nengo.Connection(a, b, learning_rule_type=nengo.BCM())
 
-        with pytest.raises(ValueError):  # transform must be correct shape
+        with pytest.raises(ValidationError, match="does not match expected shape"):
             nengo.Connection(
                 a, b, transform=np.ones((10, 11)), learning_rule_type=nengo.BCM()
             )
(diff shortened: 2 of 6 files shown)


drasmuss:
Added a commit that changes `Connection.is_decoded` to `Connection.to_neurons`. I think the notion of whether a connection is "decoded" is fairly ambiguous. E.g., is `nengo.Connection(nengo.Node(...), nengo.Ensemble(...).neurons)` decoded? It is under the previous definition of `is_decoded`, but I think intuitively most people would say no, it's not decoded. Likely due to this ambiguity, we were actually using `is_decoded` incorrectly in several places in the codebase. We were using `not is_decoded` to infer that we were connecting directly to a neuron object, but in many cases that isn't true (e.g. `nengo.Connection(nengo.Ensemble(...), nengo.Ensemble(...).neurons)` has `is_decoded=True`, but `to_neurons` is also True). So I removed the ambiguous `is_decoded`, and added the `to_neurons` attribute (which is the information we actually wanted all the places we were using `is_decoded`).

A couple open questions though
1) Should we deprecate `is_decoded`? It seemed obscure enough that I just removed it, but it couldddd be a breaking change if someone had a model using that attribute for some reason (which I've never seen).
2) Should we even expose `to_neurons` in the front-end Connection API? As with `is_decoded`, I can't see people really using this in their code. We could just make it a builder function instead.

tbekolay:
> * Should we deprecate `is_decoded`? It seemed obscure enough that I just removed it, but it couldddd be a breaking change if someone had a model using that attribute for some reason (which I've never seen).

I would just in case; we're (supposed to) make a release soon so we can remove that deprecation quickly.

> * Should we even expose `to_neurons` in the front-end Connection API? As with `is_decoded`, I can't see people really using this in their code. We could just make it a builder function instead.

I think I would move it to the builder, yeah... it feels like a temporary thing since it'll change with the NEF transform API stuff and so I wouldn't want anyone to rely on it.

nengo/tests/test_learning_rules.py
```diff
@@ -138,8 +173,26 @@ def test_pes_neuron_ens(Simulator, plt, seed, rng, allclose):
         allclose,
         pre_neurons=True,
         post_neurons=False,
+        weight_solver=weight_solver,
```
hunse, line 176: If `pre` is neurons, the solver shouldn't be used at all. Why test this?

nengo/tests/test_learning_rules.py
```diff
@@ -138,8 +173,26 @@ def test_pes_neuron_ens(Simulator, plt, seed, rng, allclose):
...
+        Simulator,
+        nengo.LIF,
+        plt,
+        seed,
+        allclose,
+        pre_neurons=False,
+        post_neurons=True,
         n=n,
         transform=initial_weights,
+        rate=1e-4,
```
hunse, line 195: It seems like we should be testing weight solvers here.

nengo/tests/test_learning_rules.py
drasmuss: Because this test failed before these changes :) (because we were assuming, incorrectly, that if a weight solver is used then `to_neurons=True`)

nengo/tests/test_connection.py
```diff
@@ -1015,10 +1015,12 @@ def test_connectionlearningruletypeparam():
         a = nengo.Ensemble(10, 1)
         b = nengo.Ensemble(11, 1)
 
-        with pytest.raises(ValueError):  # need a 2D transform for BCM
+        with pytest.raises(
+            ValidationError, match="can not be applied to decoded connections"
```
hunse, line 1019: "cannot"

nengo/tests/test_learning_rules.py
hunse: Got it. We should add a comment to that effect.

nengo/tests/test_learning_rules.py
drasmuss: These pes tests are kind of expensive, so I ended up adding a new test (see `test_build_pes`) that doesn't run any actual learning, just checks that these build with the correct shapes we expect (since that's where these different cases were failing).

nengo/tests/test_learning_rules.py
drasmuss: Moved this case into the new `test_build_pes` (see below) instead.

nengo/builder/tests/test_learning_rules.py
```diff
@@ -10,6 +16,50 @@ def test_simpes():
     assert str(sim) == "SimPES{pre=pre_filtered, error=error -> delta}"
 
 
+@pytest.mark.parametrize("pre_neurons", (True, False))
+@pytest.mark.parametrize("post_neurons", (True, False))
+@pytest.mark.parametrize("weight_solver", (True, False))
+def test_build_pes(pre_neurons, post_neurons, weight_solver):
```
hunse, line 22: Since this test is fast, maybe it would be good to test slices on pre/post too? The one downside to a builder test like this is it won't apply to backends.

drasmuss: Good call on checking slices, because it turns out the PES rule didn't work at all with a slice on `pre` 😂. I also modified it to use a Simulator (it should still be pretty fast, because we're only running it for one step), since NengoOCL and NengoDL both have their own PES builders which we'd want to check for these cases as well.

drasmuss:
> > * Should we deprecate `is_decoded`? It seemed obscure enough that I just removed it, but it couldddd be a breaking change if someone had a model using that attribute for some reason (which I've never seen).
> 
> I would just in case; we're (supposed to) make a release soon so we can remove that deprecation quickly.

Done

> > * Should we even expose `to_neurons` in the front-end Connection API? As with `is_decoded`, I can't see people really using this in their code. We could just make it a builder function instead.
> 
> I think I would move it to the builder, yeah... it feels like a temporary thing since it'll change with the NEF transform API stuff and so I wouldn't want anyone to rely on it.

I left this on the Connection, as it ended up being used in the frontend and the backend (and we don't want to be using backend functions in the frontend). But I made it private.

nengo/connection.py
```diff
@@ -91,9 +98,9 @@ def check_transform(self, conn, rule):
 
         if rule.modifies == "weights":
             # If the rule modifies 'weights', then it must have full weights
-            if conn.is_decoded:
+            if not conn._to_neurons:
```
hunse, line 101: I'm not convinced that all "weights" learning rules will work for all `_to_neurons` connections. We should test this more thoroughly, but it probably makes sense to do that with the NEFTransform changes.

nengo/connection.py
```diff
@@ -91,9 +98,9 @@ def check_transform(self, conn, rule):
...
                 raise ValidationError(
-                    "Learning rule '%s' can not be applied to decoded "
+                    "Learning rule '%s' cannot be applied to decoded "
```
hunse, line 103: I'll get rid of the phrase "decoded connections" here because it's no longer accurate.

nengo/connection.py
drasmuss: Yeah I suspect that lots of other learning rules will have similar issues to the bugs fixed here, I just didn't want to get into checking/debugging them all right now (and figured it wasn't as urgent since they're less commonly used).

hunse (approved):

PR #1644, opened 2020-09-30
Fix cache bug if index version is too new

hunse:
**Motivation and context:**

If the cache version is too new, we should just ignore the cache. Currently, there's an error because it tries to use the index when it's `None`.

**How has this been tested?**

I added a test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/tests/test_cache.py
+++ b/nengo/tests/test_cache.py
@@ -1,6 +1,7 @@
 import errno
 import multiprocessing
 import os
+import pickle
 import sys
 from subprocess import CalledProcessError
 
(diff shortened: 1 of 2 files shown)


hunse:
This was caused by 4caba86b9c4f12d2be17394ee3eff1af8d4e0e31, because we changed a number of `except Exception` statements to more specific exceptions, but obviously missed catching the exception that causes this problem. We might want to consider switching a few of those back and just disabling the pylint warning (in `cache.py` specifically).

nengo/tests/test_cache.py
```diff
@@ -180,6 +181,19 @@ def test_corrupted_decoder_cache_index(tmpdir):
     assert len(os.listdir(cache_dir)) == 2  # index, index.lock
 
 
+def test_too_new_decoder_cache_index(tmpdir):
+    cache_dir = str(tmpdir)
+
+    # Write index with super large version numbers
+    with open(os.path.join(cache_dir, CacheIndex._INDEX), "wb") as f:
+        pickle.dump((1000, 1000), f)
```
jgosmann, line 189: Could use `(CacheIndex.VERSION + 1, pickle.HIGHEST_PROTOCOL + 1)` to make it perfectly clear that these number are supposed to be higher than the current version numbers and avoid the "magic" numbers (but this is not a blocker to me).

jgosmann (commented):
I think this will print a ["Decoder cache could not acquire lock and was deactivated."](https://github.com/nengo/nengo/blob/494fec80fda535d2d268bfdfd1c11c675932fff5/nengo/cache.py#L611) warning even though it might be a version mismatch. While this broke prior to this PR, it might make sense to fix that as part of this PR?

PR #1649, opened 2020-10-20
Drop support for Python 3.5

drasmuss:
**Motivation and context:**

Python 3.5 reached its end of life in September, so time to upgrade. In addition to dropping any backwards compatibility for 3.5, this also allows us to depend on some of the nice features of 3.6+ (specifically pathlib, f-strings, and naturally ordered dicts).

Depends on https://github.com/nengo/nengo-bones/pull/123

**How has this been tested?**

Covered by existing tests.

I've separated it into a couple different steps for the different upgrades being adopted, I'd look at them individually just to make the diff a bit easier to digest.

**Types of changes:**

**Checklist:**


nengo/transforms.py
```diff
@@ -150,7 +150,7 @@ def coerce(self, instance, value):
         ):
             raise ValidationError(
                 "Must be `nengo.transforms.SparseMatrix` or "
-                "`scipy.sparse.spmatrix`, got %s" % type(value),
+                f"`scipy.sparse.spmatrix`, got {type(value)}",
```
hunse, line 153: This should have quotes

README.rst
```diff
@@ -62,8 +62,9 @@ If you'd like to install Nengo from source,
 please read the `developer installation instructions
 <https://www.nengo.ai/nengo/contributing.html#developer-installation>`_.
 
-Nengo is tested to work on Python 3.5 and above.
+Nengo is tested to work on Python 3.6 and above.
 Python 2.7 and Python 3.4 were supported up to and including Nengo 2.8.0.
+Python 3.5 was supported up to and including Nengo 3.1.
```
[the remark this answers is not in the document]
hunse: Never mind, I guess if we're merging this, we're upping the version to 3.2, so it's fine.

hunse:
I'll go and draft the changes.

Out of curiosity, did you use a tool to do all the changes to f-strings, or did you do them by hand @drasmuss?

nengo/cache.py
```diff
@@ -434,7 +435,9 @@ class WriteableCacheIndex(CacheIndex):
 
     def __init__(self, cache_dir):
         super().__init__(cache_dir)
-        self._lock = FileLock(self.index_path + ".lock")
+        self._lock = FileLock(
+            self.index_path.with_suffix(self.index_path.suffix + ".lock")
```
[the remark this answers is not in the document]
drasmuss: I assume that `with_suffix` is designed to handle more edge cases, but it probably doesn't matter here 🤷 .

nengo/tests/test_cache.py
```diff
@@ -524,12 +525,15 @@ def test_writeablecacheindex_removes(tmpdir):
 
 
 def test_writeablecacheindex_warning(monkeypatch, tmpdir):
-    def raise_error(*args, **kwargs):
-        raise CalledProcessError(-1, "move")
 
-    monkeypatch.setattr(os, "replace", raise_error)
     with pytest.warns(CacheIOWarning):
-        with WriteableCacheIndex(cache_dir=str(tmpdir)):
+        with WriteableCacheIndex(cache_dir=str(tmpdir)) as idx:
```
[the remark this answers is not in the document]
drasmuss: Yeah I didn't bother updating some of the non-core test files (e.g. in utils and here) if it would have involved a lot of lines of code, just didn't seem worth the effort.

drasmuss:
I used [flynt](https://github.com/ikamensh/flynt) to do the f-string conversion (although you still have to go and fix some things up by hand after that). The rest was all just by hand.

hunse:
Ok, I've made the changes I suggested above, and all the same types of changes (f-strings, pathlib, no OrderedDict) to the nengolib commits that weren't yet covered here. You can review that at your leisure, @drasmuss.

.nengobones.yml
```diff
@@ -239,6 +239,9 @@ travis_yml:
 
 ci_scripts:
   - template: static
+    codespell_ignore_words:
+      - nin
+      - nIn
```
drasmuss, line 244: I think we don't need to ignore these words any more right, after the changes here (or in the deterministic-plan PR)?

drasmuss:
Just the one question above, other than that fixups all look good.

.nengobones.yml
hunse: Correct, I don't think we need these ignores anymore.

hunse:
Ok, I think I've fixed the Windows cache problem here (see #1655 for details). Turns out we were trying to compare a `Path` to a string, which works well on Ubuntu but not on Windows.

I've kept in the "Fix cache bug if index version is too new" since it makes sense that if the cache index does fail, we should still be able to ignore the cache and solve for the decoders. But that commit shouldn't be necessary for getting this PR working.

Let me know if this all looks good @drasmuss.

drasmuss:
Looks good

hunse:
@drasmuss I've added a number of commits:
- I found some old formatting that wasn't caught by `flynt` so I fixed that.
- I got `setup.py` working with Python 3.5 (by removing f-strings in `version.py`), and added a test (we can squash that commit with the rest of the Python 3.5 stuff). (We could add a test for `version.py` too, but I did put a comment there, and when we move the version info to `.nengobones.yml` then it shouldn't be an issue anymore, so I wouldn't bother.) 
- I added a bunch of tests to get 100% patch coverage.

nengo/tests/test_network.py
```diff
@@ -67,28 +67,52 @@ def test_nested_context():
...
+            model.add(object())
+
     # Okay if add_to_container=False
     nengo.Ensemble(1, dimensions=1, add_to_container=False)
     nengo.Node(output=[0], add_to_container=False)
 
+    # Errors if context is in a bad state (one in `Network.__exit__`)
+    with pytest.raises(NetworkContextError, match="bad state; was expecting current"):
+        with nengo.Network():
+            nengo.Network.context.append("bad")
```
drasmuss, line 102: I would have thought that messing with the global contexts like this would have bad effects in other tests (since we're never restoring it to a good state). But none of the other tests are failing. Is the context actually being restored in some way, or is Nengo in general just insensitive to these contexts being in a bad state?

nengo/tests/test_config.py
```diff
@@ -298,7 +299,11 @@ def test_classparams_str_repr():
 def test_config_repr():
     """tests the repr function in Config class"""
     model = nengo.Network()
-    assert repr(model.config) == "<Config(Connection, Ensemble, Node, Probe)>"
+    print(repr(model.config))
```
drasmuss, line 302: Leftover print statement, or is this intentional?

nengo/tests/test_processes.py
```diff
@@ -557,9 +557,13 @@ def func(t):
...
+
         # cubic interpolation with 0 not in times
         process = Piecewise({0.001: 0, 0.1: 0.1}, interpolation="cubic")
         with pytest.warns(UserWarning, match="'cubic' interpolation.*for t=0.0"):
-            # this ValueError comes from inside Scipy
-            with pytest.raises(ValueError, match="number of derivatives at boundaries"):
+            try:
                 process.run(0.001)
+            except ValueError:
+                pass  # scipy may raise a ValueError
```
drasmuss, line 569: Is scipy stochastically raising an error here? That seems really weird.

nengo/tests/test_config.py
hunse: Just leftover

nengo/tests/test_network.py
hunse: That's a good point. My guess is we're just insensitive to it. For example, when you enter a network `with` block, both that network and its config get added to their respective lists. Then when we're using them, we're always looking at the last one in each list, so it doesn't really matter if one is longer than the other. Then when we exit, we remove from both.

We should probably use `monkeypatch` for this, though, just to be safe.

nengo/tests/test_processes.py
hunse: Good question. When I first wrote this test, I didn't make it Scipy-specific. Then I realized we'll only get the warning if Scipy is installed, so I added the `importorskip`. Originally, I had a `pytest.raises` for this ValueError instead of the `try` block, and I'm not sure if I changed it because I wanted the test to run with or without Scipy, or if it was because TravisCI's Scipy did something different than mine. Either way, we don't really care about whether Scipy raises an error or not, so I think it's fine to leave it.

drasmuss:
Fixups LGTM

PR #1652, opened 2020-11-11
Reset op resets to initial signal value

tbekolay:
**Motivation and context:**
When I was preparing the 3.1 release, I found that a slow test was failing. To fix it, I added a `Reset` op, which required resetting the value to a two-dimensional value instead of a scalar. Rather than deal with that, I though it made more sense to have `Reset` reset signals to their initial values, whatever they may be.

In the end, I went with another approach to fix the slow test failure, but this PR still seems like a useful one to have IMO.

**How has this been tested?**
Existing tests cover it.

**Types of changes:**

**Checklist:**

**Still to do:**
- [ ] Discuss if this change is worthwhile
- [ ] Add changelog entry

### DIFF
--- a/nengo/builder/learning_rules.py
+++ b/nengo/builder/learning_rules.py
@@ -722,9 +722,9 @@ def build_voja(model, voja, rule):
 
     # Learning signal, defaults to 1 in case no connection is made
     # and multiplied by the learning_rate * dt
-    learning = Signal(shape=rule.size_in, name="Voja:learning")
+    learning = Signal(initial_value=np.ones(rule.size_in), name="Voja:learning")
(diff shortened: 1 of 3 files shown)


hunse:
One downside to this for a backend like NengoOCL is it means we'll have to either store all the arrays of initial values on the GPU, or copy them over each timestep. Neither seems ideal.

I'd probably suggest a hybrid approach where for `Reset` ops that are resetting the whole signal to the same value, we specify that value (e.g. `Reset(sig, value=0)`), but we also allow for a way to reset to initial value in the case where we need that (e.g. `Reset(sig, value=None)` or something like that).

PR #1654, opened 2020-11-24
Ensure op graph order is deterministic

drasmuss:
**Motivation and context:**

Rebuilding the same network multiple times could produce a different op order each time (as long as all those orders respected the constraints of the topological sort). While this doesn't have any functional impact on the model, it can be problematic for backends that are more sensitive to op order (e.g., NengoDL, where the order of operators can impact the layout of signals in memory).

This changes all the usages of `set`/`frozenset` in the builder to `OrderedSet`/`FrozenOrderedSet` (analogous to `OrderedDict`, these are sets that preserve their insertion order).

https://github.com/nengo/nengo-dl/pull/187 depends on this PR

**How has this been tested?**

See new tests in `test_builder.py` and `test_graphs.py`.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/tests/test_builder.py
+++ b/nengo/builder/tests/test_builder.py
@@ -1,3 +1,8 @@
+import os
+import subprocess
+import sys
+import textwrap
+
 import pytest
 
 import nengo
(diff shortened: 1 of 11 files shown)


nengo/utils/stdlib.py
```diff
@@ -138,6 +138,50 @@ def discard(self, key):
             del self._data[key]
 
 
+class FrozenOrderedSet(Set):
+    """A set that preserves insertion order and is hashable."""
+
+    def __init__(self, data=None):
+        if data is None:
+            data = []
+        self.data = dict((d, None) for d in data)
```
hunse, line 147: I would use `OrderedDict` here for now, so that we don't have differences between Python 3.5 and 3.6+.

nengo/builder/tests/test_builder.py
```diff
@@ -51,3 +56,54 @@ class Test:
     my_builder(1)
     with pytest.warns(Warning, match="Type .* already has a builder. Overwriting"):
         my_builder(1)  # repeat setup warning
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6, 0), reason="requires ordered dicts")
```
hunse, line 61: If we use `OrderedDict` in `OrderedSet` then we shouldn't need this.

nengo/utils/stdlib.py
drasmuss: I was thinking we'd merge this after the PR dropping Python 3.5 support, but yeah if not then we should do that.

nengo/builder/tests/test_builder.py
drasmuss: Unfortunately not, there's other places that we use regular `dicts` within the build process that introduce non-determinism, besides `sets`. And fixing those didn't seem worth it, since we'll be dropping Python 3.5 support very soon.

hunse: So I looked into making this deterministic in Python 3.5, and it's actually quite difficult. There are a lot of places where dict order can randomize things, including in e.g. `NeuronType.state`, other parts of making the graph, etc. I haven't been able to track them all down, and I don't think it's worth putting more time into. Looking at the NengoDL PR, things might actually be deterministic enough that the previously broken functionality there is now fixed for Python 3.5 as well, even though things aren't completely deterministic.

So we'll just leave this full determinism as something we don't support in Python 3.5.

hunse:
There are a couple `set`s in some of the other functions in `graphs.py` still. Should we make them `OrderedSet`s, too? I think they're mostly used in the optimizer. I know that we're not going for full determinism in the optimizer at the moment, but if performance is similar between `OrderedSet` and `set`, then I don't see a strong reason not to. And it could avoid some bugs in the future (in general, non-determinism is annoying to debug). Or should we be doing more profiling before making that change, to ensure there's not a significant performance hit?

drasmuss:
Yeah that's intentional. Because `OrderedSet` is slower than `set`, I tried to only use `OrderedSet` in the minimum number of places necessary to ensure determinism.

hunse (approved):

PR #1659, opened 2020-12-16
Fix test_pytest_nengo.py to work with new pytest

hunse:
**Motivation and context:**

`test_pytest_nengo.test_unsupported` is breaking with new pytest (6.2.1), since we can't overwrite `tmpdir` anymore.

Also increase minimum pytest to 3.9.1 (which has `tmp_path`).

**How has this been tested?**

- Ran modified test with pytest 6.2.1, and it works. Also double-checked that file is moved properly.
- Ran all tests with pytest 3.9.1, and they all pass.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -47,7 +47,7 @@ setup_py:
   tests_req:
     - jupyter
     - matplotlib>=1.4
-    - pytest>=3.6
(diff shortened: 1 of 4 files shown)


drasmuss (approved):
Looks good

PR #1671, opened 2021-03-17
Fix conjgrad test to be more robust

hunse:
It would fail for some seeds on the `cond=50` case. Using extra
iterations ensures it converges (it was sometimes stopping on
iterations, not tolerance), and comparing `y` instead of `x`
ensures the tested error is in the same space as the stopping
tolerance criterion.

**Interactions with other PRs**

Based on #1670.

**How has this been tested?**

Ran the test over 100 times with random seeds, it now passes every time.

**Types of changes:**

**Checklist:**


hunse:
Merged as part of #1676.

PR #1676, opened 2021-09-24
Fix some signals not correctly marked readonly

tbekolay:
**Motivation and context:**
A user reported issues in resetting a simulator with recent versions of Nengo. The issue was that some signals were not marked readonly, but their underlying NumPy arrays were readonly. This PR fixes the case that was causing the issue for the user.

**How has this been tested?**
Existing tests still run, the user's model now resets correctly.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/utils/tests/test_least_squares_solvers.py
+++ b/nengo/utils/tests/test_least_squares_solvers.py
@@ -29,14 +29,14 @@ def run_solver(solver, m=30, n=20, d=1, cond=10, sys_rng=np.random, **kwargs):
     y = np.dot(A, x)
     x2, _ = solver(A, y, **kwargs)
 
-    return x, x2
+    return x, x2, A
 
 
 @pytest.mark.parametrize("cond, sigma", [(5, 1e-4), (500, 1e-6), (1e5, 1e-8)])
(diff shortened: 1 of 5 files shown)


nengo/utils/tests/test_least_squares_solvers.py
```diff
@@ -13,7 +13,9 @@
 )
 
 
-def run_solver(solver, m=30, n=20, d=1, cond=10, sys_rng=np.random, **kwargs):
+def run_solver(
+    solver, m=30, n=20, d=1, cond=10, sys_rng=np.random, measure="x", **kwargs
```
tbekolay, line 17: New arguments that modify what the function does (especially what it returns) is an anti-pattern IMO... I think in this case I would just `return x, x2, A` and calculate `y` and `y2` in `test_conjgrad`.

hunse: Hah, I had something more like that before (I just returned `x, x2, y, y2`), but then it seemed silly to modify all the other tests to `x, x2, _, _ = run_solver` when nothing had really changed about them. I can change it back, though...

hunse (approved):

PR #1677, opened 2021-11-19
Fix pylint errors

hunse:
**Motivation and context:**

This fixes errors from the new pylint version, as well as one breaking test and some links for the NEF book. 

**Interactions with other PRs:**

All other PRs will have to be rebased on to this to get them running again.

**How has this been tested?**

Ran the unit tests.

**How long should this take to review?**

- Quick (less than 40 lines changed or changes are straightforward)

**Types of changes:**

- Bug fix (non-breaking change which fixes an issue)
- Non-code change (touches things like tests, documentation, build scripts)

**Checklist:**

- [x] I have read the **CONTRIBUTING.rst** document.
- [n/a] I have updated the documentation accordingly.
- [n/a] I have included a changelog entry.
- [n/a] I have added tests to cover my changes.
- [x] I have run the test suite locally and all tests passed.

PR #1678, opened 2021-12-06
Allow NumPy scalars for eval points

tbekolay:
**Motivation and context:**
In #1677, we fixed a NumPy warning in which an array was being set with list, tuple, or other iterable. We fixed it by checking if it was a number or ndarray. We didn't consider NumPy scalar types, which NengoDL uses, and which fail when `np.hstack` is called on them.

This fixes that issue. Since the issue exists only in master and not in a release version, we don't need a changelog entry.

**How has this been tested?**
NengoDL tests pass with this.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/builder/connection.py
+++ b/nengo/builder/connection.py
@@ -85,7 +85,7 @@ def get_targets(conn, eval_points, dtype=None):
                     f"Building {conn}: Connection function returned "
                     "None. Cannot solve for decoders."
                 )
-            if not isinstance(out, (int, float, np.ndarray)):
+            if isinstance(out, (tuple, list)):
                 out = np.hstack(out)
             targets[i] = out
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 
 repos:
 -   repo: https://github.com/psf/black
-    rev: 20.8b0
+    rev: 21.12b0
     hooks:
     - id: black
       files: \.py$
--- a/nengo/utils/numpy.py
+++ b/nengo/utils/numpy.py
@@ -18,7 +18,6 @@ try:
         """Check if ``obj`` is a sparse matrix."""
         return isinstance(obj, scipy_sparse.spmatrix)
 
-
 except ImportError as e:
     logger.info("Could not import scipy.sparse:\n%s", str(e))
     scipy_sparse = None


drasmuss (approved):

nengo/builder/connection.py
```diff
@@ -85,7 +85,7 @@ def get_targets(conn, eval_points, dtype=None):
                     f"Building {conn}: Connection function returned "
                     "None. Cannot solve for decoders."
                 )
-            if not isinstance(out, (int, float, np.ndarray)):
+            if not isinstance(out, (int, float, np.ndarray, np.number)):
```
hunse, line 88: It seems strange to me to check for the things that it can't be. Why not just `isinstance(out, (tuple, list))`? I think those are what we're looking for. Is there anything else?

Also, why `hstack`? Why not just have an `asarray` call on the output no matter what it is?

tbekolay: There are lots of iterables, checking for tuple and list isn't sufficient. Checking if it's an iterable isn't ideal because ndarrays are iterables.

I don't think it's as simple as just making it into an array, but it might be. You can try that if you want.

hunse: But what other iterables are supported by `hstack` and won't return an array with `dtype=object`? Numpy functions don't allow generators. If you look at our implementation of `is_array_like`, we basically assume list/tuple/ndarray. If we want to generally support more, I think we'd need to modify that function and possibly make some changes throughout the codebase.

tbekolay: Yeah, I mean checking only tuple and list would be simpler. I'll switch.

hunse: I looked into just using `asarray`, and there's at least one test where it breaks. We currently allow users to do something like `function=lambda x: [x, -x]` which, since `x` is a vector (even if it's a single-element vector), returns a 2D array when passed through `asarray`. So `hstack` is what we want.

As for what we allow, in general I'm in favour of being a bit more limited and forcing users to be a bit more explicit. You're right that somebody can implement a custom `Sequence` class and `hstack` appears to work with that. There are some situations where we get what I believe is unintended/unexpected behaviour though. For example, `np.hstack({1:2, 3:4}) == [1, 3]`, but a user returning a dictionary was probably trying to do something else, and to silently do something different than expected seems like an easy avenue for bugs. Note that this has just come up with using `hstack`; before the recent changes, returning something like a `dict` would fail.

tbekolay: I'm not too concerned with cases like the `dict` where it should be obvious to anyone that what they're doing shouldn't work. Whitelisting tuple and list is probably fine; like you said, there are other places where we assume some types and the model probably wouldn't get to this point anyway. In the future we'll add explicit types to NengoCore and these things will be more clear.

tbekolay:
I amended in the whitelist (instead of blacklist) check, so this should be good to go.

hunse (approved):

tbekolay:
Had a random Appveyor failure, but this is all passing now. If one of you could merge, I'll rerun the failing tests in https://github.com/nengo/nengo-dl/pull/183 and merge that in.

hunse:
I'll merge.

PR #1679, opened 2021-12-15
Copy inputs for process step functions

tbekolay:
**Motivation and context:**
While working on https://github.com/nengo/nengo-loihi/pull/312 we discovered that processes don't copy their input `x` value, even though they are copied for non-process node step functions. This PR copies the `x` value for processes as well, making them behave the same.

**How has this been tested?**
Tested that this PR makes the `np.copy` in [this line](https://github.com/nengo/nengo-loihi/pull/312/files#diff-d9c021912f7de6b7afade53e8289d8c593282c0e60e76eb8a514df6b4b3f70f0R16) not necessary. However, we should probably write a test for this within NengoCore.

Should probably start by writing a test for the behavior. We might have a test that normal node output functions are copies which we can add a process to.

**Types of changes:**

**Checklist:**

**Still to do:**
- [x] Add test

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,7 +1,7 @@
 environment:
   matrix:
     - PYTHON: "C:\\Miniconda37"
-      CONDA_PYTHON: "python=3.6"
+      CONDA_PYTHON: "python=3.7"
     - PYTHON: "C:\\Miniconda37-x64"
       CONDA_PYTHON: "python=3.9"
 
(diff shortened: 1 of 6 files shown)


hunse:
OK, added a test. I didn't make it quite the same as our test for nodes, since there we check `x is not last_x`, which could miss a problem if we return a view on `x` without the underlying data being copied. (Maybe we should fix that?)

tbekolay:
Test looks good. I guess we could also `assert x.base is None` to ensure it's not a view?

hunse:
Yeah, could do. That's probably the easiest fix for now.

hunse (approved):
OK, I've made that change. I also added another test to cover all the branches of `SimProcess.make_step`, since one was uncovered.

PR #1681, opened 2021-12-23
Remove duplicate output probeable in Neurons

tbekolay:
**Motivation and context:**
When you look at `nengo.Ensemble(10, 1, add_to_container=False).neurons.probeable`, you get:

```
('output', 'input', 'output', 'voltage', 'refractory_time')
```

Why's that duplicate `'output'` in there? Seems to have been an oversight from this change: 036a0ce07b24ec247423ff6651c4222d5e2aba38

This PR fixes it:

```
>>> nengo.Ensemble(10, 1, add_to_container=False).neurons.probeable
('output', 'voltage', 'refractory_time', 'input')
```

Note that the order changes such that `'input'` is now at the end. That's so that `'output'` still appears first.

**How has this been tested?**
Verified in the console, but we still need to write tests for this.

Implement a test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo/ensemble.py
+++ b/nengo/ensemble.py
@@ -244,7 +244,7 @@ class Neurons:
     @property
     def probeable(self):
         """(tuple) Signals that can be probed in the neuron population."""
-        return ("output", "input") + self.ensemble.neuron_type.probeable
+        return self.ensemble.neuron_type.probeable + ("input",)
 
     @property
     def size_in(self):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,12 +32,6 @@ Release history
 - Added ``Conv`` and ``ConvTranspose`` aliases for ``Convolution`` and
   ``ConvolutionTranspose``. (`#1648`_)
 
-**Fixed**
-
-- Operator graph step order will now be deterministic. (`#1654`_)
-- Fixed an issue in which some simulators could not be reset due to signals
-  not being marked as readonly. (`#1676`_)
-
 **Removed**
 
 - Removed support for Python 3.5 (which reached its end of life in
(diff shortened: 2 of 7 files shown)


tbekolay:
Added tests and changelog entry so this is ready for review.

nengo/ensemble.py
```diff
@@ -244,7 +244,7 @@ def ensemble(self):
     @property
     def probeable(self):
         """(tuple) Signals that can be probed in the neuron population."""
-        return ("output", "input") + self.ensemble.neuron_type.probeable
+        return self.ensemble.neuron_type.probeable + ("input",)
```
hunse, line 247: Interesting effect of this is that the first element of the neuron type's `probeable` will be the default. Everything looks good for our neuron types, but I've added a comment to this docstring, just to help users (particularly those potentially implementing their own neuron types) to be aware. It'd be nice if we had a better place to note this for custom neuron types (e.g. if we had a notebook about custom neuron types), but I don't see a better place than here unfortunately.

tbekolay: > the first element of the neuron type's `probeable` will be the default

Yes, but the first element of every neuron type's probeable is `"output"` so the behavior is the same and predictable: https://github.com/nengo/nengo/blob/master/nengo/neurons.py#L91

I don't believe it's possible to make a custom neuron type that doesn't have an `"output"` state, as it's automatically managed in the base class.

hunse: Yeah, it would only be if somebody overrode that base class implementation and messed it up. For example, in `RatesToSpikesNeuronType`, we do such an override. On the one hand, it's really not our problem if somebody messes that up, but on the other hand, there's nowhere in the documentation that we say that the first probeable is the default (I didn't even realize that until looking at this PR), so to me it makes sense to mention it.

hunse:
It seemed strange that the learning rules all had different values for the first probeable, so I made "delta" first for all of them. Really, we probably want to encourage users to be explicit for learning rules anyway, but there's no way to force that (i.e. have no default) without significant changes.

nengo/ensemble.py
tbekolay: I mentioned elsewhere, but this is documented in the `nengo.Probe` docstring, so I would document something specific to `NeuronType`s here, not something generic to probes.

hunse:
You're right that it's technically a breaking change. I don't think anybody is depending on it, but if you feel strongly we could make a separate PR and leave it to 4.0 or something. Frankly, I think it's so uncommon that I would do it now since that'll be easier, but I don't feel strongly.

hunse (approved):

tbekolay:
> I don't think anybody is depending on it, but if you feel strongly we could make a separate PR and leave it to 4.0 or something.

Yeah ... I'm conflicted as I agree it's super uncommon but I think I'd leave it as I don't want people's models to break, and the plan for 4.0 will require migration details so it's easier to document it there, I think.

nengo/ensemble.py
hunse: Having a comment in the Probe docstring is probably as good as having it here, so let's just leave it at that. I hadn't realized there was a comment there.

hunse:
OK, learning rule changes are moved to #1682. The only changes I've made here are to combine the two "Fixed" sections in the changelog.

Issue #1675, opened 2021-08-05
Add groups parameter to nengo.Convolution


Michaeljurado24:
nengo.Convolution supports many of the same arguments and options as its keras counterpart. However there is no groups parameter in nengo as in the keras docs for[ Conv2D](https://keras.io/api/layers/convolution_layers/convolution2d/).

A groups paramter would be useful to allow for more robust conversions between keras networks and nengo networks via the dl converter. This would also allow people to implement depthwise convolutions with ease also in nengo.

Michaeljurado24:
I discovered a potential fix for this.

Michaeljurado24:
Sorry I did not mean to close this issue

PR #1684, opened 2022-02-17
Add groups parameter to Convolutions

Michaeljurado24:
**Motivation and context:**

Currently nengo does not support group convolution. Grouped convolutions are important because when groups == n_channels == n_filters, they are functionally equivalent to Depthwise Convolutions. (These in turn can be used to construct efficient separable convolutions which [greatly reduce computational overhead](https://towardsdatascience.com/a-basic-introduction-to-separable-convolutions-b99ec3102728) compared to regular convolutions. 

Furthermore adding grouped convolutions in nengo is a stepping stone to adding grouped convolutions in nengo-loihi. 

Issue 
https://github.com/nengo/nengo/issues/1675

Related Forum Posts
https://forum.nengo.ai/t/adding-a-groups-parameter-for-convolutions-in-nengo-loihi/2035/2
https://forum.nengo.ai/t/trying-to-understand-nengo-transforms-convolution/1844/4?u=mjurado3

None

**How has this been tested?**
Primarily Through [this function](https://github.com/Michaeljurado24/nengo/blob/Convolution_groups/nengo/builder/tests/test_transforms_conv.py#L18). I also manually compared the new convolutions against a tensorflow implementation of a convolution. 

Quick since the only major change comes from [this function](https://github.com/Michaeljurado24/nengo/blob/Convolution_groups/nengo/_vendor/npconv2d/conv2d.py#L184)

1. The reviewer should probably first ask himself if it is worth accepting this pull request since TransposeConvolutions still do not currently accept a groups argument. 
2. If they deem that this is an acceptable, they should then move onto reviewing nengo/_vendor/npconv2d.py. 
3. Then should move onto the modified [Convolution Transform](https://github.com/Michaeljurado24/nengo/blob/Convolution_groups/nengo/transforms.py#L351)

**Types of changes:**

- New-Feature: Low level backend convolution changes
- New-Feature: High Level addition of groups parameter to nengo.Convolution

**Checklist:**

**Still to do:**
- [ ] Add groups param to nengo.ConvolutionTranspose
- [ ] Add groups param to this unit [test](https://github.com/Michaeljurado24/nengo/blob/Convolution_groups/nengo/tests/test_transforms_conv.py#L53). (This involves completing the prior task first)

### DIFF
--- a/nengo/_vendor/npconv2d/conv2d.py
+++ b/nengo/_vendor/npconv2d/conv2d.py
@@ -181,7 +181,7 @@ def extract_sliding_windows_gradx(x,
     return y
 
 
-def conv2d(x, w, pad='SAME', stride=(1, 1)):
+def conv2d(x, w, pad="SAME", stride=(1, 1)):
     """2D convolution (technically speaking, correlation).
     Args:
         x: [N, H, W, C]
@@ -191,6 +191,7 @@ def conv2d(x, w, pad='SAME', stride=(1, 1)):
     Returns:
         y: [N, H', W', K]
     """
+    assert x.ndim == 4 and w.ndim == 4
     ksize = w.shape[:2]
     x = extract_sliding_windows(x, ksize, pad, stride)
     ws = w.shape
@@ -199,6 +200,43 @@ def conv2d(x, w, pad='SAME', stride=(1, 1)):
     x = x.reshape([xs[0] * xs[1] * xs[2], -1])
     y = x.dot(w)
     y = y.reshape([xs[0], xs[1], xs[2], -1])
+    return y
+
+
+def conv2d_groups(x, w, pad="SAME", stride=(1, 1)):
+    """2D convolution (technically speaking, correlation). Compatible with groups > 1
+    Args:
+        x: [N, H, W, C]
+        w: [I, J, C/G, K]
+        pad: [PH, PW]
+        stride: [SH, SW]
+    Returns:
+        y: [N, H', W', K]
+    """
+    assert x.ndim == 4 and w.ndim == 4
+    c = x.shape[-1]  # input channels
+    ksize = w.shape[:2]
+    cg, k = w.shape[2:]  # channels-per-group and output channels
+
+    # infer number of groups
+    assert (
+        c % cg == 0
+    ), f"Number of channels ({c}) must be divisible by channels-per-group ({cg})"
+    groups = c // cg
+
+    x = extract_sliding_windows(x, ksize, pad, stride)
+    x = x.reshape(x.shape[:-1] + (groups, c // groups))  # split windows into groups
+    x = np.moveaxis(x, -2, 0)  # move groups to axis 0
+    xs = x.shape
+    x = x.reshape([groups, xs[1] * xs[2] * xs[3], xs[4] * xs[5] * xs[6]])
+
+    w = w.reshape(w.shape[:-1] + (groups, k // groups))  # split weights into groups
+    w = np.moveaxis(w, -2, 0)  # move groups to axis 0
+    ws = w.shape
+    w = w.reshape([groups, ws[1] * ws[2] * ws[3], ws[4]])
+
+    y = np.einsum("ikj,ijm->kim", x, w)
+    y = y.reshape([xs[1], xs[2], xs[3], k])
     return y
 
 
--- a/nengo/builder/tests/test_transforms_conv.py
+++ b/nengo/builder/tests/test_transforms_conv.py
@@ -8,28 +8,44 @@ from nengo.builder.transforms import ConvInc, ConvTransposeInc
 from nengo.transforms import ChannelShape, Convolution, ConvolutionTranspose
 
 
-@pytest.mark.parametrize("channels_last", (True, False))
-@pytest.mark.parametrize("stride0", (1, 2))
-@pytest.mark.parametrize("stride1", (1, 2))
-@pytest.mark.parametrize("kernel0", (4, 5))
-@pytest.mark.parametrize("kernel1", (4, 5))
-@pytest.mark.parametrize("padding", ("same", "valid"))
-def test_convinc_2d(
-    channels_last, stride0, stride1, kernel0, kernel1, padding, rng, allclose
-):
+def scipy_conv2d(x, w, in_channels, out_channels, groups=1):
     correlate2d = pytest.importorskip("scipy.signal").correlate2d
 
-    shape0 = 16
-    shape1 = 17
+    out_c_per_group = out_channels // groups
+    in_c_per_group = in_channels // groups
+    result = []
+    for i in range(out_channels):
+        j_offset = (i // out_c_per_group) * in_c_per_group
+        result_i = [
+            correlate2d(x[..., j + j_offset], w[..., j, i], mode="valid")
+            for j in range(in_c_per_group)
+        ]
+        result.append(np.sum(result_i, axis=0))
+
+    return np.stack(result, axis=-1)
+
+
+def _test_convinc_2d(
+    channels_last,
+    stride0,
+    stride1,
+    kernel0,
+    kernel1,
+    padding,
+    groups,
+    out_channels,
+    rng,
+    allclose,
+):
+    shape0, shape1 = 16, 17
     in_channels = 32
-    out_channels = 64
     x_shape = (
         (shape0, shape1, in_channels)
         if channels_last
         else (in_channels, shape0, shape1)
     )
     x = Signal(rng.randn(*x_shape))
-    w = Signal(rng.randn(kernel0, kernel1, in_channels, out_channels))
+    w = Signal(rng.randn(kernel0, kernel1, in_channels // groups, out_channels))
 
     conv = Convolution(
         out_channels,
@@ -38,6 +54,7 @@ def test_convinc_2d(
         strides=(stride0, stride1),
         padding=padding,
         channels_last=channels_last,
+        groups=groups,
     )
 
     y = Signal(np.zeros(conv.output_shape.shape))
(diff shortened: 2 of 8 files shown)


nengo/_vendor/npconv2d/conv2d.py
```diff
@@ -185,21 +185,30 @@ def conv2d(x, w, pad='SAME', stride=(1, 1)):
...
     ws = w.shape
-    w = w.reshape([ws[0] * ws[1] * ws[2], ws[3]])
+    w = w.reshape([groups, ws[1] * ws[2] * ws[3], ws[4]])
     xs = x.shape
-    x = x.reshape([xs[0] * xs[1] * xs[2], -1])
-    y = x.dot(w)
-    y = y.reshape([xs[0], xs[1], xs[2], -1])
-    return y
+    x = x.reshape([groups, xs[1] * xs[2]* xs[3], -1])
+    y = np.einsum('ikj,ijm->kim', x, w)
```
hunse, line 209: I think it might be better to avoid the use of `einsum` in the original case (i.e. where `groups=1`), both for speed and clarity. I think the best approach might be to have a separate function that deals with the grouped case. Preserving the original function will mean we can both test against it, and have it for reference.

nengo/_vendor/npconv2d/conv2d.py
```diff
@@ -185,21 +185,30 @@ def conv2d(x, w, pad='SAME', stride=(1, 1)):
...
-        w: [I, J, C, K]
+        w: [I, J, C/G, K]
         pad: [PH, PW]
         stride: [SH, SW]
     Returns:
         y: [N, H', W', K]
     """
     ksize = w.shape[:2]
+    c = x.shape[-1]
+    groups = c // w.shape[2] # number of groups inferred
```
hunse, line 196: It might be good to assert that `c % w.shape[2] == 0`. If that isn't the case, my guess is that at best, the output will not be what the user expects (as they probably made a mistake in specifying the weight shape).

nengo/builder/tests/test_transforms_conv.py
```diff
@@ -14,8 +14,9 @@
 @pytest.mark.parametrize("kernel0", (4, 5))
 @pytest.mark.parametrize("kernel1", (4, 5))
 @pytest.mark.parametrize("padding", ("same", "valid"))
+@pytest.mark.parametrize("groups",  (1, 8, 32))
```
hunse, line 17: There are already a lot of parametrizations for this test, and adding "groups" makes it go from 64 to 192. I know they're quite fast, but it adds clutter to the log and I'm not convinced it's necessary. Also, it might be good to test with different numbers of in and out channels, since that's directly related to groups.

The way I would do this is to take the body of this function, and make a `_test_convinc_2d` helper function. Then you can have two functions, `test_convinc_2d` and `test_grouped_convinc_2d`, that test the ungrouped and grouped cases. Since we're splitting up `conv2d` into grouped and ungrouped versions, this will also allow testing each one separately.

hunse (changes requested):

hunse:
I don't think it's necessary to also add groups for `ConvolutionTranspose` at this time, particularly if it's not something that you need.

nengo/tests/test_reprs.py
```diff
@@ -576,13 +576,15 @@ def test_transforms():
...
             "init",
+            "groups"
         ],
     )
     check_repr(Convolution(n_filters=3, input_shape=(3, 4, 5)))
     check_repr(Convolution(n_filters=3, input_shape=(3, 2, 5), kernel_size=(3, 2)))
     check_repr(Convolution(n_filters=3, input_shape=(3, 4, 5), channels_last=False))
     check_repr(Convolution(n_filters=3, input_shape=(3, 4, 5), strides=(2, 2)))
     check_repr(Convolution(n_filters=3, input_shape=(3, 4, 5), padding="same"))
+    check_repr(Convolution(n_filters=3, input_shape=(3, 4, 5), groups = 3))
```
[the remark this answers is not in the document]
Michaeljurado24: One thing I notice is that when I run `pre-commit run --all-files` all of the changes I made get reformatted. However, a ton of other files get changed as well as in the example below. 
image

Is this okay?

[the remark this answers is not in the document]
tbekolay: This is due to the `pre-commit` hook being out of date the latest version of Black; if you don't mind, can you remove the hook with `pre-commit uninstall`, then in your usual environment do `pip install black` and then `black .`? That should remove those erroneous changes.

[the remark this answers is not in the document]
Michaeljurado24: thank you

Michaeljurado24:
In the new commit I addressed these changes and added a few unit tests to specifically test for grouped convolutions. Here they are

1. https://github.com/Michaeljurado24/nengo/blob/Convolution_groups/nengo/builder/tests/test_transforms_conv.py#L95
2. https://github.com/Michaeljurado24/nengo/blob/Convolution_groups/nengo/tests/test_transforms_conv.py#L52
3. https://github.com/Michaeljurado24/nengo/blob/Convolution_groups/nengo/tests/test_transforms_conv.py#L199

nengo/_vendor/npconv2d/conv2d.py
```diff
@@ -202,6 +202,39 @@ def conv2d(x, w, pad='SAME', stride=(1, 1)):
...
+        pad: [PH, PW]
+        stride: [SH, SW]
+    Returns:
+        y: [N, H', W', K]
+    """
+    ksize = w.shape[:2]
+    c = x.shape[-1]
+
+    # infer number of groups
+    assert c % w.shape[2] == 0
```
hunse, line 219: Add a message like `f"Number of channels ({c}) must be divisible by number of groups ({w.shape[2]})"`

hunse: Actually, I guess `w.shape[2]` isn't the number of groups, it's the channels divided by groups. I'm not quite sure what to call it...

hunse: Maybe "channels per group"? Thus e.g. `f"Number of channels ({c}) must be divisible by channels-per-group ({w.shape[2]})"`

nengo/_vendor/npconv2d/conv2d.py
```diff
@@ -202,6 +202,39 @@ def conv2d(x, w, pad='SAME', stride=(1, 1)):
...
+    """2D convolution (technically speaking, correlation). Compatible with groups > 1
+    Args:
+        x: [N, H, W, C]
+        w: [I, J, C/G, K]
+        pad: [PH, PW]
+        stride: [SH, SW]
+    Returns:
+        y: [N, H', W', K]
+    """
+    ksize = w.shape[:2]
```
hunse, line 215: While I'm thinking of it, we should probably assert that both `x` and `w` have `ndim == 4`, both here and in `conv2d`.

nengo/builder/tests/test_transforms_conv.py
```diff
@@ -88,6 +88,93 @@ def test_convinc_2d(
...
+@pytest.mark.parametrize("channels_last", (True, False))
+@pytest.mark.parametrize("padding", ("same", "valid"))
+@pytest.mark.parametrize("groups", (1, 8, 32))
+@pytest.mark.parametrize("out_channels", (32, 64))
+def test_grouped_convinc_2d(
+    channels_last, padding, groups, out_channels, rng, allclose
+):
+    correlate2d = pytest.importorskip("scipy.signal").correlate2d
+
+    shape0, shape1 = 16, 17
```
hunse, line 100: What I was thinking for the test is that we could have one `_test_convinc_2d` helper function, that is called by both `test_convinc_2d` and `test_grouped_convinc_2d`. That way, there's not all the duplicated code.

nengo/builder/tests/test_transforms_conv.py
```diff
@@ -96,11 +183,9 @@ def test_convtransposeinc_2d(
     channels_last, strides, kernel_size, padding, rng, allclose, plt
 ):
     """Test ConvTransposeInc by ensuring it is the transpose of ConvInc.
-
```
hunse, line 99: I think these blank lines were mistakenly removed.

nengo/tests/test_transforms_conv.py
```diff
@@ -44,6 +44,46 @@ def test_convolution_shape(padding, stride, k_size, x_mul, rng, allclose):
         assert allclose(y_np, y_tf)
 
 
+@pytest.mark.parametrize("groups", (1, 2))
+@pytest.mark.parametrize("x_mul", (1, 2))
+@pytest.mark.parametrize("k_size", (1, 2))
+@pytest.mark.parametrize("stride", (1, 2))
+@pytest.mark.parametrize("padding", ("same", "valid"))
+def test_convolution_group_shape(padding, stride, k_size, x_mul, groups, rng, allclose):
```
hunse, line 52: Again, I would have a common helper function for both this and the test above, so that we're not duplicating code.

nengo/tests/test_transforms_conv.py
```diff
@@ -151,6 +191,93 @@ def test_convolution(
...
+    channels_last,
+    fixed_kernel,
+    groups,
+    output_channels,
+    Simulator,
+    allclose,
+    rng,
+    seed,
+):
+    input_d = 4
```
hunse, line 210: Again, common helper function with the test above.

Michaeljurado24:
In the latest commit I added a few helper functions as suggested and also added a check for ndim == 4. 

I should also mention that I added an extra helper method called [scipy_conv2d](https://github.com/Michaeljurado24/nengo/blob/Convolution_groups/nengo/builder/tests/test_transforms_conv.py#L11) and scipy_conv2d_grouped to reduce tab nesting.

hunse:
Thanks for the changes! I'll take a look now.

Can you please sign the CAA [here](https://www.nengo.ai/caa/)? Just add your name to the People page as it describes at the top to sign. You'll have to do that before I can merge this PR.

hunse:
I've made some fixups. Let me know if they look good, @Michaeljurado24. I'll squash everything together before merge.

I'll take one last look once the CAA is signed before I merge, but I think that this is looking good!

I've made the `convolution-groups-ci` branch, that's just a copy of this for running CI. The tests passed for me locally, but I'll need to check the CI tests and coverage before merge.

Michaeljurado24:
Everything looks good! I made a pull request with my name added to the CAA

hunse:
I've squashed everything down, which I think should get the last CI test to pass (it was a problem with the commit message being too long).

Note that it'll still be at least another week before this gets merged, because there are others on the team who want to take a quick look first. That said, I think this is quite close to what will be the final version, so feel free to build off this branch for e.g. a NengoLoihi implementation.

Michaeljurado24:
So for adding grouped convolutions to nengo-loihi, I need to first add them to nengo-dl. 
For adding grouped convolutions to nengo-dl, I am thinking that I only need to modify the  [ConvertConv function](https://github.com/nengo/nengo-dl/blob/4abc90fa7a020cb30a58edfd749815884661605e/nengo_dl/converter.py#L1320) and [a unit test](https://github.com/nengo/nengo-dl/blob/master/nengo_dl/tests/test_converter.py#L37). Are there any other areas you know of which would need to be edited for a pull request to be accepted?

hunse:
The most important thing to update is actually this class, [`ConvIncBuilder`](https://github.com/nengo/nengo-dl/blob/26b7511d7a9d4d273c70eae5f01fcd1e794387f3/nengo_dl/transform_builders.py#L53). This is the class that takes a Nengo `ConvInc` operator and produces the appropriate TensorFlow code. So implementing the `groups` parameter there and adding a unit test would be the first step.

The `ConvertConv` class is part of the Keras to Nengo converter, which is used to import Keras models into Nengo. It should also be updated, but hopefully that's not too much more involved than just passing the groups parameter from the Keras layer to the Nengo `Convolution` transform.

Michaeljurado24:
Okay, I just made a pull request for adding grouped convolutions to nengo-dl. Surprisingly I did not need to edit ConvIncBuilder. We can move further discussion about this to the [other pull request.](https://github.com/nengo/nengo-dl/pull/223)

tbekolay:
Hi @Michaeljurado24, thanks for this contribution! It's been merged now as b6002e4035f4d7641d0c740c3a9c8fab9e4d4889.

PR #1687, opened 2022-10-18
Fix LinearFilter for tensors

hunse:
**Motivation and context:**

LinearFilter wasn't working properly for tensors with `ndim >= 3`.

Also upgraded black and isort since pre-commit hooks were failing due to changes in click (https://github.com/psf/black/issues/2964). 

**How has this been tested?**

Added a unit test that reproduces the failure in the absence of the fix.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,8 +26,14 @@ Release history
 
 - Added ``groups`` parameter to ``nengo.Convolution``. (`#1675`_, `#1684`_)
 
+**Fixed**
+
+- Fixed an issue where ``nengo.LinearFilter`` and subclasses (e.g. ``Lowpass``,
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@
 
 repos:
 -   repo: https://github.com/psf/black
-    rev: 21.12b0
+    rev: 22.10.0
```
tbekolay, line 5: This file's generated by Bones, so we should just have to run `bones-generate`.

tbekolay (approved):
Pushed a fixup to fix a small issue I noticed, the rest LGTM. Currently CI isn't running due to having to TravisCI credits, but I'm in the process of switching to GH Actions; once it's switched over I'll fix up the history and merge.

